PR #4, opened 2019-09-16
Setup E2E test suite in Golang

mgencur:
Hey, this is for https://jira.coreos.com/browse/SRVCOM-550

The strategy is now the following:
- deploy ServiceMesh using the script
- deploy via script the OperatorSource generated by ./hack/catalog.sh 
- deploy and undeploy the remaining resources from Go (including Subscription, KnativeServing, Knative service)

Regarding the single test that currently exists, it just checks that the Knative service installed successfully. It does not check that the service returns a valid response. I could add that if needed. At the end of the test it is checked that the openshift-ingress and knative-serving operators were removed successfully.

The e2e-tests.sh script is close to what we use in other repos. And when I setup CI it will be used  there. Right now it can be run against a runing OpenShift cluster after setting the KUBECONFIG env property


mgencur:
The PR is huge due to dependencies. I'm using the "prune" option for dep but suggestions for pruning the deps further are welcome.

PR #5, opened 2019-09-17
CI setup

mgencur:
Please let me know who else should be on the OWNERS list or whether I should remove someone. I'll need to put the same people in the other PR before it can be merged.
I'll send a PR for openshift/release shortly, link: https://github.com/openshift/release/pull/5046

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -1,0 +1,15 @@
+aliases:
+  serverless-approvers:
+  - alanfx
+  - mgencur
+  - bbrowning
+  - jcrossley3
+  - markusthoemmes
+  - vdemeester
+  serverless-reviewers:
+  - alanfx
+  - mgencur
+  - bbrowning
+  - jcrossley3
+  - markusthoemmes
+  - vdemeester
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -6,6 +6,7 @@ aliases:
   - jcrossley3
   - markusthoemmes
   - vdemeester
+  - warrenvw
   serverless-reviewers:
   - alanfx
   - mgencur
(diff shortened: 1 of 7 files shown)


openshift/ci-operator/generate-ci-config.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+branch=${1-'openshift-serverless-v1.0.0'}
```
mgencur: I'm not sure about this name. Please let me know if this should be changed. For core knative image we use "knative-v0.7" and

OWNERS_ALIASES
```diff
@@ -0,0 +1,15 @@
+aliases:
```
jcrossley3: Maybe add @warrenvw in here, too?

openshift/ci-operator/generate-ci-config.sh
jcrossley3: The CSV is using `serverless-operator.v1.0.0` but I'm not sure what the correct convention is.

OWNERS_ALIASES
mgencur: Will do.

openshift/ci-operator/generate-ci-config.sh
mgencur: This name determines where the image will be available for tests in other openshift repositories (in case we ever need it). The image is promoted there after a PR is merged. 
The current location would then be: `registry.svc.ci.openshift.org/openshift/openshift-serverless-v1.0.0:serverless-operator` (this is the same scheme that we use for e.g. knative serving: `registry.svc.ci.openshift.org/openshift/knative-v0.7.1:knative-serving-activator`)
I think this scheme is alright.
If we used `serverless-operator.v1.0.0` there it would look like this: `registry.svc.ci.openshift.org/openshift/serverless-operator.v1.0.0:serverless-operator`

mgencur:
Thanks.

PR #8, opened 2019-09-30
Facilitate local testing

jcrossley3:


test/e2e/deployment_test.go
```diff
@@ -31,12 +32,9 @@ func TestDeployment(t *testing.T) {
 	image := "gcr.io/knative-samples/helloworld-go"
 	_, err = test.WithServiceReady(ctx, "helloworld-go", testNamespace, image)
 	if err != nil {
-		t.Fatal("Knative Service not ready")
+		t.Fatal("Knative Service not ready", err)
```
mgencur: Yeah. Good point. I forgot to add the original error here and there.

test/e2e/deployment_test.go
```diff
@@ -31,12 +32,9 @@ func TestDeployment(t *testing.T) {
...
 	}
 
-	// delete manually so that we can verify removing dependent operators
-	ctx.Cleanup()
```
mgencur: The cleanup was here on purpose so that the following function `WaitForOperatorDepsDeleted` can verify that all the dependencies that are installed automatically will be removed in the end. 
I suppose you're moving this to have a proper cleanup in the event of error? Maybe the Cleanup function should be idempotent so that we can call it multiple times - first in this place and second in the defer statement

test/e2e-tests.sh
```diff
@@ -224,18 +224,23 @@ EOF
...
 
   header "CatalogSource installed successfully"
 }
 
 function tag_serverless_operator_image(){
-  oc policy add-role-to-group system:image-puller system:serviceaccounts:${OPERATORS_NAMESPACE} --namespace=${OPENSHIFT_BUILD_NAMESPACE}
-  oc tag --insecure=false -n ${OPERATORS_NAMESPACE} ${OPENSHIFT_REGISTRY}/${OPENSHIFT_BUILD_NAMESPACE}/stable:${SERVERLESS_OPERATOR} ${SERVERLESS_OPERATOR}:latest
+  if [[ -n "${OPENSHIFT_BUILD_NAMESPACE}" ]]; then
+    oc policy add-role-to-group system:image-puller system:serviceaccounts:${OPERATORS_NAMESPACE} --namespace=${OPENSHIFT_BUILD_NAMESPACE}
+    oc tag --insecure=false -n ${OPERATORS_NAMESPACE} ${OPENSHIFT_REGISTRY}/${OPENSHIFT_BUILD_NAMESPACE}/stable:${SERVERLESS_OPERATOR} ${SERVERLESS_OPERATOR}:latest
```
mgencur: Looks like the output of these lines needs to be sent to /dev/null so that we don't get the error in logs:
`+ sed -e 's+\(.* containerImage:\)\(.*\)+\1 Tag serverless-operator:latest set to registry.svc.ci.openshift.org/ci-op-bqg1pvq8/stable:serverless-operator.
image-registry.openshift-image-registry.svc:5000/openshift-operators/serverless-operator+g'`

jcrossley3: fixed

test/e2e/deployment_test.go
jcrossley3: Good call, fixed

mgencur:
/lgtm
@jcrossley3 let me know if some more logging should be done during the test run. I could try to improve it. But I believe it helped reveal some bugs already? Thanks

PR #11, opened 2019-10-03
Create test users with view and edit roles

mgencur:
This PR depends on https://github.com/openshift-knative/serverless-operator/pull/12

This is to fix https://jira.coreos.com/browse/SRVKS-275

The tests cover create/delete/get/list operations for knative.serving.dev API group. They don't test patch/update/watch but I think the current coverage would be quite sufficient. Please let me know if we really need to cover the rest as well.


mgencur:
@jcrossley3 it's now ready. Thanks

test/e2e/knative_serving_test.go
```diff
@@ -0,0 +1,163 @@
...
+		userContext: editCtx,
+		wantErr:     false,
+	}, {
+		name:          "user with edit role can delete",
+		userRole:      "edit",
+		operationName: "delete",
+		operation: func(c *test.Context) error {
+			return test.DeleteService(c, "useredit-service", testNamespace)
+		},
+		userContext: contexts[1],
```
markusthoemmes: nit: 
```suggestion
		userContext: editCtx,
```

test/e2e/knative_serving_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+func testUserPermissions(t *testing.T, contexts []*test.Context) {
+	editCtx := contexts[1]
+	viewCtx := contexts[2]
+	tests := []struct {
+		name          string
+		userRole      string
+		userContext   *test.Context
+		operationName string
+		operation     func(context *test.Context) error
```
markusthoemmes: `operationName` and `operation` seem entirely redundant. Maybe infer the operation from the string in the `Run` block below?

test/e2e/knative_serving_test.go
```diff
@@ -0,0 +1,163 @@
...
+			return err
+		},
+		userContext: editCtx,
+		wantErr:     false,
+	}, {
+		name:          "user with edit role can delete",
+		userRole:      "edit",
+		operationName: "delete",
+		operation: func(c *test.Context) error {
+			return test.DeleteService(c, "useredit-service", testNamespace)
```
markusthoemmes: What sets this service up to be deleted? Does this rely on the other test before it to pass as well?

test/e2e/knative_serving_test.go
```diff
@@ -0,0 +1,163 @@
...
+		name:          "user with view role cannot create",
+		userRole:      "view",
+		operationName: "create",
+		operation: func(c *test.Context) error {
+			_, err := test.CreateService(c, "userview-service", testNamespace, image)
+			return err
+		},
+		userContext: viewCtx,
+		wantErr:     true,
+		wantErrStr:  "is forbidden",
```
markusthoemmes: Should we infer `wantErr` from `wantErrStr`? Right now they seem redundant.

test/clients.go
```diff
@@ -38,26 +39,30 @@ type Clients struct {
...
-	clients, err := NewClients()
-	if err != nil {
-		t.Fatalf("Couldn't initialize clients: %v", err)
-	}
-
-	ctx := &Context{
-		T:       t,
-		Clients: clients,
+// Setup creates context objects for all kubeconfigs passed from the command line
+func Setup(t *testing.T) []*Context {
```
markusthoemmes: I wonder if this should always just return the Admin context and if we should create `SetupView` and `SetupEdit` or something that specifically craetes the other two contexts.

That way, we can also add a "name" string to the context to be used to print enough info in tests without providing that additionally.

test/e2e/knative_serving_test.go
mgencur: I can just remove the operationName, it's always clear what the operation is - either from the test name or from the error being thrown.

test/e2e/knative_serving_test.go
mgencur: Yes.

test/e2e/knative_serving_test.go
mgencur: Possibly, it will make it a little less clear but can save some code/text.

test/clients.go
mgencur: OK. Probably makes sense because we already depend on certain order of parameters to `--kubeconfig`, so we will always know that the second kubeconfig is for `edit` role and third for `view` role

PR #22, opened 2019-10-25
Add Service Mesh as an OLM dependency in Serverless Operator CSV

wtam2018:
Fix https://jira.coreos.com/browse/SRVKS-284

Add Service Mesh as an OLM dependency in Serverless Operator CSV

### DIFF
--- a/olm-catalog/serverless-operator/serverless-operator.v1.2.0.clusterserviceversion.yaml
+++ b/olm-catalog/serverless-operator/serverless-operator.v1.2.0.clusterserviceversion.yaml
@@ -89,7 +89,12 @@ spec:
         displayName: Istio Service Mesh Member Roll
         kind: ServiceMeshMemberRoll
         name: servicemeshmemberrolls.maistra.io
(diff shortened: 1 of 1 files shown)


olm-catalog/serverless-operator/serverless-operator.v1.2.0.clusterserviceversion.yaml
```diff
@@ -84,6 +84,12 @@ spec:
         displayName: Version
         path: version
       version: v1alpha1
+    required:
+      - description: A list of namespaces in Service Mesh
+        displayName: Istio Service Mesh Member Roll
+        kind: ServiceMeshMemberRoll
```
markusthoemmes: We should also depend on ServiceMeshControlPlane here. It won't make any difference in effect, but we really depend on both of these CRDs.

wtam2018: Sure.

PR #23, opened 2019-10-25
Using OLM namespace for CatalogSource does not work

wtam2018:
Using openshift-operator-lifecycle-manager for internal CatalogSource does not work.   Subscription will not be able to find the it.  We should use openshift-operators instead.


jcrossley3:
Closing in lieu of #21

PR #24, opened 2019-10-29
[SRVKS-255] Add the keyword 'Knative' into the short description.

markusthoemmes:
As the title says. Per SRVKS-255 it seems like the search functionality is limited to the title and the description annotation.


bbrowning:
Has anyone actually spun up a cluster with this fix to verify it now shows up when searching for knative?

PR #27, opened 2019-11-06
Use the newest releases of k-s-o and k-o-i for auto-smcp goodness.

markusthoemmes:
As the title says. This contains all the goodness around auto-installing an SMCP and managing an SMMR.


hack/lib/servicemesh.bash
```diff
@@ -1,24 +1,6 @@
...
-      return 0
-    fi
-  fi
-  install_service_mesh
-}
-
-function install_service_mesh {
-  logger.info 'Installing ServiceMesh'
-
+function install_service_mesh_operator {
```
jcrossley3: I would expect only the `CatalogSourceConfig` is needed in this function. All of the Subscriptions can be deleted if the CRD dependencies in the CSV's are correct, I think.

markusthoemmes: Openshift 4.1 needs the manual subscription stuff and this repo is currently only setup against 4.1 CI.

markusthoemmes: On 4.2 this entire function is no longer needed. See https://github.com/openshift/knative-serving/pull/298 for how that looks like.

jcrossley3:
We can conditionally add the subs based on ocp version in a different PR
/lgtm

cardil:
I was using that scripts in upstream `eventing-operator` tests. This removal broke my tests :cry:

jcrossley3:
> I was using that scripts in upstream `eventing-operator` tests. This removal broke my tests

Sorry about that. Can we re-organize to accommodate both?

cardil:
Yes. We are meeting with Matthias tomorrow morning to groom on how to do shared libraries. We should have a proposition for the team

PR #29, opened 2019-11-15
Update k-s-o and k-o-i to contain the newest bugfixes.

markusthoemmes:
Title has it all.


test/knativeserving.go
```diff
@@ -34,11 +34,15 @@ func CreateKnativeServing(ctx *Context, name, namespace string) (*servingoperato
 		return nil, err
 	}
 	ctx.AddToCleanup(func() error {
-		return ctx.Clients.ServingOperator.KnativeServings(namespace).Delete(serving.Name, &metav1.DeleteOptions{})
+		return DeleteKnativeServing(ctx, name, namespace)
```
mgencur: In the other PR I'm removing the one-line helpers. So perhaps we could remove it here as well and call the .Delete operation directly.

mgencur:
LGTM

PR #32, opened 2019-11-19
[WIP] Feature: run upstream serving operator tests

cardil:
This PR will let us run upstream serving operator tests, a compliance suite, on a serverless operator.

PR #35, opened 2019-11-25
Run knative-serving tests.

markusthoemmes:
This is for a PoC. Let's see how hard it is to run actual Knative Serving tests here.

### DIFF
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -21,7 +21,7 @@ failed=0
 
 # Setup serverless and run upstream e2e and conformance tests.
 (( !failed )) && ensure_serverless_installed || failed=5
-(( !failed )) && run_upstream_tests "v0.9.0" || failed=6
+(( !failed )) && run_knative_serving_tests "v0.9.0" || failed=6
 (( !failed )) && teardown_serverless || failed=7
 
 (( failed )) && dump_state
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -14,7 +14,9 @@ failed=0
 
 (( !failed )) && install_service_mesh_operator || failed=2
 (( !failed )) && install_catalogsource || failed=3
-(( !failed )) && logger.success 'Cluster prepared for testing.' && run_e2e_tests || failed=4
+(( !failed )) && ensure_serverless_installed || failed=4
+(( !failed )) && run_conformance_tests || failed=5
+#(( !failed )) && logger.success 'Cluster prepared for testing.' && run_e2e_tests || failed=4
 (( failed )) && dump_state
 (( failed )) && exit $failed
 
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -14,9 +14,16 @@ failed=0
 
 (( !failed )) && install_service_mesh_operator || failed=2
 (( !failed )) && install_catalogsource || failed=3
-(( !failed )) && ensure_serverless_installed || failed=4
-(( !failed )) && run_conformance_tests || failed=5
-#(( !failed )) && logger.success 'Cluster prepared for testing.' && run_e2e_tests || failed=4
+(( !failed )) && logger.success 'Cluster prepared for testing.'
+
+# Run serverless-operator specific tests.
+(( !failed )) && run_e2e_tests || failed=4
+
+# Setup serverless and run upstream e2e and conformance tests.
+(( !failed )) && ensure_serverless_installed || failed=5
+(( !failed )) && run_upstream_tests "release-v0.9.0" || failed=6
+(( !failed )) && teardown_serverless || failed=7
+
 (( failed )) && dump_state
 (( failed )) && exit $failed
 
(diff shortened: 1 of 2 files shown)


test/e2e-tests.sh
```diff
@@ -14,7 +14,16 @@ failed=0
...
 (( !failed )) && install_catalogsource || failed=3
-(( !failed )) && logger.success 'Cluster prepared for testing.' && run_e2e_tests || failed=4
+(( !failed )) && logger.success 'Cluster prepared for testing.'
+
+# Run serverless-operator specific tests.
+(( !failed )) && run_e2e_tests || failed=4
+
+# Setup serverless and run upstream e2e and conformance tests.
+(( !failed )) && ensure_serverless_installed || failed=5
+(( !failed )) && run_upstream_tests "v0.9.0" || failed=6
```
mgencur: Could we rename this to run_knative_serving_tests ?

test/lib.bash
```diff
@@ -45,7 +45,49 @@ function run_e2e_tests {
     && return 1
 }
 
+function run_upstream_tests {
+  (
+  # Checkout the relevant code to run
+  mkdir -p "$GOPATH/src/knative.dev"
+  cd "$GOPATH/src/knative.dev" || exit
+  #git clone -b "release-v$1" --single-branch https://github.com/openshift/knative-serving.git serving
+  git clone -b "ocp-flag-0.9.0" --single-branch https://github.com/markusthoemmes/knative-serving.git serving
```
mgencur: We should probably unify how we consume other repositories. For example, in this other PR we use Go dependencies: https://github.com/openshift-knative/serverless-operator/pull/32/files#diff-836546cc53507f6b2d581088903b1785R41

test/lib.bash
```diff
@@ -45,7 +45,49 @@ function run_e2e_tests {
...
+  go test -v -tags=e2e -count=1 -timeout=30m -parallel=3 ./test/conformance/api/... --resolvabledomain --kubeconfig "$KUBECONFIG" \
+    --imagetemplate "$image_template" \
+    || failed=1
+  
+  return $failed
+  )
+}
+
 function teardown {
+  if [[ "${OPENSHIFT_BUILD_NAMESPACE}" != "" ]]; then
```
cardil: Small change here, that I didn't pushed yet. It should be:

```bash
if [[ -v OPENSHIFT_BUILD_NAMESPACE ]]; then
```

Previous check wasn't working in case you don't have `OPENSHIFT_BUILD_NAMESPACE` variable defined.

test/lib.bash
```diff
@@ -47,6 +47,12 @@ function run_e2e_tests {
 
 function run_knative_serving_tests {
   (
+  # Setup a temporary GOPATH to safely check out the repository without breaking other things.
+  local tmp_gopath=/tmp/gotmp
```
cardil: Let's use uniq path each time:

```bash
local tmp_gopath
tmp_gopath="$(mktemp -d -t gopath-XXXXXXXXXX)"
```

test/lib.bash
markusthoemmes: Discussed this with @cardil and landed on the current code as a way forward.

test/lib.bash
```diff
@@ -45,7 +45,55 @@ function run_e2e_tests {
...
+  (
+  # Setup a temporary GOPATH to safely check out the repository without breaking other things.
+  local tmp_gopath
+  tmp_gopath="$(mktemp -d -t gopath-XXXXXXXXXX)"
+  cp -r "$GOPATH/bin" "$tmp_gopath"
+  export GOPATH="$tmp_gopath"
+
+  # Checkout the relevant code to run
+  mkdir -p "$GOPATH/src/knative.dev"
+  cd "$GOPATH/src/knative.dev" || exit
```
cardil: This should be `|| return 42` or `|| exit 42` or `|| return $?` So it will exit with non zero exitcode

test/lib.bash
```diff
@@ -45,7 +45,55 @@ function run_e2e_tests {
     && return 1
 }
 
+function run_knative_serving_tests {
+  (
+  # Setup a temporary GOPATH to safely check out the repository without breaking other things.
+  local tmp_gopath
```
cardil: `local tmp_gopath failed` to register multiple variables as local

test/e2e-tests.sh
```diff
@@ -14,7 +14,16 @@ failed=0
 
 (( !failed )) && install_service_mesh_operator || failed=2
 (( !failed )) && install_catalogsource || failed=3
-(( !failed )) && logger.success 'Cluster prepared for testing.' && run_e2e_tests || failed=4
+(( !failed )) && logger.success 'Cluster prepared for testing.'
+
+# Run serverless-operator specific tests.
+(( !failed )) && run_e2e_tests || failed=4
+
+# Setup serverless and run upstream e2e and conformance tests.
```
cardil: Maybe we would like to have control, if that part should be run or not?

test/lib.bash
markusthoemmes: You lost me here. Didn't understand your comment.

test/e2e-tests.sh
markusthoemmes: Locally, maybe. The e2e-tests.sh script should be a uber script for CI imo though. We can always create smaller scripts that only run specific suites locally?

PR #36, opened 2019-11-27
Do not use serverless-operator image in tests

mgencur:
The image is not used even if it's part of the catalog source. Only the metadata is consumed from the catalog source.


markusthoemmes:
Obsolete.

PR #38, opened 2019-11-28
Rewrite openshift/knative-serving test suite execution to Golang

cardil:
That will be better in number of ways:
* It will be easier to understand and write properly as bash isn't well understood
* test reports will be the same as in other test
* It will be easier to reuse in different tests
* It's easier to select what tests should be run, with build tags
* In future when we rewrite cluster setup to go, we will be able to run and DEBUG single test directly from IDE

The last point is biggest for me. That's what I would like to see. To be able to run everything just from IDE. Just as you run Spring Test or Arquillian!

### DIFF
--- a/test/clients.go
+++ b/test/clients.go
@@ -173,7 +173,7 @@ func CleanupAll(contexts ...*Context) {
 	}
 }
 
-// Cleanup iterates through the list of registered CleanupFunc functions and calls them
+// Cleanup iterates through the list of registered CleanupFunc functions in reverse order and calls them
 func (ctx *Context) Cleanup() {
 	for _, f := range ctx.CleanupList {
 		f()
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 catalogsource-ci.yaml
 user*.kubeconfig
 users.htpasswd
+.idea
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -19,11 +19,6 @@ failed=0
 # Run serverless-operator specific tests.
 (( !failed )) && run_e2e_tests || failed=4
 
-# Setup serverless and run upstream e2e and conformance tests.
-(( !failed )) && ensure_serverless_installed || failed=5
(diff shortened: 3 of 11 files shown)


matzew:
Didn't we say that we want to keep the scripts as well ? 

I do still find them handy 

creation of a golang lib was discussed as a future (alternative?) option - not sure we really need that now

cardil:
> Didn't we say that we want to keep the scripts as well ?
> 
> I do still find them handy
> 
> creation of a golang lib was discussed as a future (alternative?) option - not sure we really need that now

I'm not removing any scripts. I just rewritten a test that Marcus added.

mgencur:
I'm not a big fan of rewriting this to Go. We removed 45 lines of bash and added 300 lines of Go. I also find the bash scripts handy.

markusthoemmes:
> I'm not a big fan of rewriting this to Go. We removed 45 lines of bash and added 300 lines of Go. I also find the bash scripts handy.

That was also my sentiment.

test/clients.go
```diff
@@ -173,9 +173,11 @@ func CleanupAll(contexts ...*Context) {
 	}
 }
 
-// Cleanup iterates through the list of registered CleanupFunc functions and calls them
+// Cleanup iterates through the list of registered CleanupFunc functions in reverse order and calls them
 func (ctx *Context) Cleanup() {
-	for _, f := range ctx.CleanupList {
+	last := len(ctx.CleanupList)-1
```
mgencur: The AddToCleanup function adds them to this list in reverse order so you're effectively changing it to be removing NOT in reverse order.

cardil: I didn't notice that. I'll restore it.

matzew:
can we keep the bash scripts ?

cardil:
> can we keep the bash scripts ?

@matzew It is only TEST rewritten. Not, scripts. They are left as they were.

test/clients.go
cardil: Restored.

PR #39, opened 2019-11-29
Remove quay.io image references.

markusthoemmes:
Replace quay.io references to the knative-serving images with references to the openshift CI registry. These are public as well and can be pulled via the internet (for example from your local machine). This removes the need for us to mirror images from the knative-serving repository.


markusthoemmes:
Obsolete.

PR #41, opened 2019-11-29
[DO NOT SUBMIT] Replace OwnerRefs on Routes with Finalizer.

markusthoemmes:
This is a PR that showcases how the monorepo + the newly setup CI works. 

This entire change is validated by the operator tests and then run through the integration tests of Knative Serving. If it passes, confidence is high that the entire system would still work after this change has been merged.

In addition, it also runs the unit tests that we currently have for extra confidence.


markusthoemmes:
Obsolete.

PR #46, opened 2019-12-03
Move top-level vendor directory into test.

markusthoemmes:
`dep ensure` tries to be smart and tries to vendor things for all of our subdirs, even if they have their own vendor. This properly scopes this vendor dir just to the tests for now.


mgencur:
Where would I run the `dep ensure` now? in the test directory? Does it make sense to run it in the top-level dir as well?

markusthoemmes:
@mgencur Running it top-level doesn't work really at all. You'll need to run it in the individual subdirectory.

mgencur:
/lgtm

It makes sense to have the `vendor` in the test dir because the dependencies are only related to tests.

PR #47, opened 2019-12-04
Tighten the security on Knative Serving pods from cluster local ingress accesses

wtam2018:
Fixes: https://jira.coreos.com/browse/SRVKS-227
The current Network Policies on knative serving pods grant ingress access from anywhere.  The proposed changes are to change the Network Polices on the following pods:
* Activator, autoscaler, controller - allow pods in the namespace of monitoring policy group member
* Networking-istio networking-certmanager - disallow pods from anyway.

Network prolicy on Webhook is unchanged.

Note that pods in namespace member of service mesh control plane and openshift ingress router group can still access to knative serving pods as allowed by other Network Polices.

ServiceMonitor as documented https://docs.openshift.com/container-platform/4.2/serverless/monitoring-serverless.html should continue to work.


mgencur:
Can you please add some tests? The test framework is there and already includes some tests for security aspects (normal user vs. admin).

wtam2018:
@mgencur Can you point me to where the security tests are?

mgencur:
@wtam2018 https://github.com/openshift-knative/serverless-operator/blob/release-1.3/test/e2e/knative_serving_test.go#L54 (security tests)

wtam2018:
> @wtam2018 https://github.com/openshift-knative/serverless-operator/blob/release-1.3/test/e2e/knative_serving_test.go#L54 (security tests)

Unit test has been added.

markusthoemmes:
/hold

We shouldn't merge against 1.3 anymore. Will cut a proper master soon.

bbrowning:
This also doesn't really do anything to fix the underlying issue. Every namespace in the cluster that uses Knative will be part of the same Service Mesh AND a user can automatically get any namespace added to that Service Mesh by just deploying a Knative Service. So, any user in any namespace on the cluster can still hit those metrics endpoints.

markusthoemmes:
Closing this as outdated. Please reopen against master if still relevant.

PR #48, opened 2019-12-04
Improve the developer experience.

markusthoemmes:
- Clearly spell out the requirements for the scripts.
- Remove minikube docs to not confuse people.
- Move `envsubst` call into the catalog.sh script so it produces a meaningful result as well.

PR #50, opened 2019-12-04
CI_PROMOTION_NAME env, to test with CI-promoted images

maschmid:
Add the ability to use the promoted CI images, add CI_PROMOTION_NAME env 

e.g. CI_PROMOTION_NAME=openshift-serverless-v1.3.0

### DIFF
--- a/hack/lib/catalogsource.bash
+++ b/hack/lib/catalogsource.bash
@@ -12,13 +12,18 @@ function ensure_catalogsource_installed {
 function install_catalogsource {
   logger.info "Installing CatalogSource"
 
-  # Determine if we're running locally or in CI.
-  if [ -z "$OPENSHIFT_BUILD_NAMESPACE" ]; then
-    export IMAGE_KNATIVE_SERVING_OPERATOR="${DOCKER_REPO_OVERRIDE}/knative-serving-operator"
-    export IMAGE_KNATIVE_OPENSHIFT_INGRESS="${DOCKER_REPO_OVERRIDE}/knative-openshift-ingress"
-  else
+  # Use current build images in CI.
+  if [ -n "$OPENSHIFT_BUILD_NAMESPACE" ]; then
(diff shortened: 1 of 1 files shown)


hack/lib/catalogsource.bash
```diff
@@ -12,13 +12,18 @@ function ensure_catalogsource_installed {
...
-  else
+  # Use current build images in CI.
+  if [ -n "$OPENSHIFT_BUILD_NAMESPACE" ]; then
     export IMAGE_KNATIVE_SERVING_OPERATOR="registry.svc.ci.openshift.org/${OPENSHIFT_BUILD_NAMESPACE}/stable:knative-serving-operator"
     export IMAGE_KNATIVE_OPENSHIFT_INGRESS="registry.svc.ci.openshift.org/${OPENSHIFT_BUILD_NAMESPACE}/stable:knative-openshift-ingress"
+  # Use DOCKER_REPO_OVERRIDE for developing locally
+  elif [ -n "$DOCKER_REPO_OVERRIDE" ]; then
+    export IMAGE_KNATIVE_SERVING_OPERATOR="${DOCKER_REPO_OVERRIDE}/knative-serving-operator"
+    export IMAGE_KNATIVE_OPENSHIFT_INGRESS="${DOCKER_REPO_OVERRIDE}/knative-openshift-ingress"
+  # Use CI_PROMOTION_NAME to use current promoted images from CI
```
markusthoemmes: Wondering if we should hardcode this to the actual version maybe? With the scripts in this branch, only the 1.3.0 promotion is supposed to work, right? Maybe we can even roll without the new variable then.

mgencur: I think the only problem is that you will need to update this script with every new branch.

markusthoemmes: We kinda need to do that anyway, we have references to v0.10 and stuff

maschmid:
Sure, I'll update/rebase it after #48 is merged

maschmid:
Obsolete by #48

PR #51, opened 2019-12-06
Fix OLM namespace

mgencur:
* also fix condition to not fail the whole script when install_plan is
empty (this is due to setting set -e in bash)


mgencur:
/test 4.3-e2e-aws-ocp-43

flake...

PR #52, opened 2019-12-06
Completely separate E2E and rolling upgrades

mgencur:
We were using flags to choose between E2E and rolling upgrades. This separates them. I think this is a slightly better approach but let me know what you think:)

### DIFF
--- a/test/lib.bash
+++ b/test/lib.bash
@@ -29,7 +29,7 @@ function register_teardown {
 function run_e2e_tests {
   declare -al kubeconfigs
   local kubeconfigs_str
-  
+
   logger.info "Running tests"
   kubeconfigs+=("${KUBECONFIG}")
   for cfg in user*.kubeconfig; do
@@ -37,7 +37,7 @@ function run_e2e_tests {
   done
   kubeconfigs_str="$(array.join , "${kubeconfigs[@]}")"
 
-  go test -v -tags=e2e -count=1 -timeout=30m -parallel=1 ./test/e2e \
+  go_test_e2e -tags=e2e -timeout=30m -parallel=1 ./test/e2e \
     --kubeconfig "${kubeconfigs[0]}" \
     --kubeconfigs "${kubeconfigs_str}" \
     && logger.success 'Tests has passed' && return 0 \
@@ -53,23 +53,29 @@ function make_temporary_gopath {
   cp -rv "$(go env GOPATH)/bin" "${tmp_gopath}"
   logger.info "Temporary GOPATH is: ${tmp_gopath}"
   export GOPATH="$tmp_gopath"
+  export PATH="$GOPATH/bin":$PATH
 }
 
-function run_knative_serving_tests {
-  (
+function remove_temporary_gopath {
+  if [[ "$GOPATH" =~ .*gopath-[0-9a-zA-Z]{10} ]]; then
+    logger.info "Removing GOPATH: ${GOPATH}"
+    rm -rf "${GOPATH}"
+  fi
+}
+
+function checkout_knative_serving {
   local knative_version=$1
+  # Setup a temporary GOPATH to safely check out the repository without breaking other things.
   make_temporary_gopath
 
-  # Save the rootdir before changing dir
-  rootdir="$(dirname "$(dirname "$(realpath "${BASH_SOURCE[0]}")")")"
-
   # Checkout the relevant code to run
-  mkdir -p "$GOPATH/src/knative.dev"
-  git clone --branch "release-${knative_version}" \
-    --depth=1 https://github.com/openshift/knative-serving.git \
-    "${GOPATH}/src/knative.dev/serving"
-  pushd "${GOPATH}/src/knative.dev/serving" || return $?
+  export KNATIVE_SERVING_HOME="$GOPATH/src/knative.dev/serving"
+  mkdir -p "$KNATIVE_SERVING_HOME"
+  git clone -b "release-${knative_version}" --depth 1 https://github.com/openshift/knative-serving.git "$KNATIVE_SERVING_HOME"
+  git describe --always --tags --dirty
+}
 
+function prepare_knative_serving_tests {
   # Remove unneeded manifest
   rm test/config/100-istio-default-domain.yaml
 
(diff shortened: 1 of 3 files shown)


test/lib.bash
```diff
@@ -45,24 +45,27 @@ function run_e2e_tests {
...
-function run_knative_serving_tests {
-  (
+function checkout_knative_serving {
   local knative_version=$1
   # Setup a temporary GOPATH to safely check out the repository without breaking other things.
   local tmp_gopath
   tmp_gopath="$(mktemp -d -t gopath-XXXXXXXXXX)"
-  cp -r "$GOPATH/bin" "$tmp_gopath"
+  if [[ -d "$GOPATH/bin" ]]; then
+    cp -r "$GOPATH/bin" "$tmp_gopath"
```
cardil: This cp command works improperly. Bins are directly on `$tmp_gopath` directory. They should be inside `bin/` directory. See, https://github.com/cardil/serverless-operator/blob/feature/rewrite-to-golang/test/e2e/temp_gopath.go#L28-L30

mgencur: Hmm the command at line 54 will copy the whole "bin" directory inside the $tmp_gopath directory. So the path will be `$tmp_gopath/bin/<any_binaries_in_that_dir>` . Your command does the same, no?

cardil: I think I was mistaken with this. Sorry. When I was doing this those binaries were copied to parent directory.

test/lib.bash
```diff
@@ -2,7 +2,7 @@
 
 # == Overrides & test releated
 
-# shellcheck disable=SC1091,SC1090
+# shellcheck disable=SC1091,SC1090,SC2120
```
cardil: Skipping SC2120 isn't good idea, as it might find nasty errors

test/lib.bash
```diff
@@ -45,24 +45,27 @@ function run_e2e_tests {
...
-  # Save the rootdir before changing dir
-  rootdir="$(dirname "$(dirname "$(realpath "${BASH_SOURCE[0]}")")")"
-
   # Checkout the relevant code to run
   mkdir -p "$GOPATH/src/knative.dev"
-  cd "$GOPATH/src/knative.dev" || return $?
-  git clone -b "release-${knative_version}" --single-branch https://github.com/openshift/knative-serving.git serving
-  cd serving || return $?
+  pushd "$GOPATH/src/knative.dev" || return $?
+  git clone -b "release-${knative_version}" --depth 1 https://github.com/openshift/knative-serving.git serving
```
cardil: You can clone directly. No need to have 2 pushed and popd. One is sufficient.

test/lib.bash
```diff
@@ -73,34 +76,52 @@ function run_knative_serving_tests {
...
+  image_template="registry.svc.ci.openshift.org/openshift/knative-${knative_version}:knative-serving-test-{{.Name}}"
 
-function run_knative_serving_e2e_and_conformance_tests {
-  logger.info "Running E2E and conformance tests"
   go test -v -tags=e2e -count=1 -timeout=30m -parallel=3 ./test/e2e ./test/conformance/... \
     --resolvabledomain --kubeconfig "$KUBECONFIG" \
-    --imagetemplate "$image_template" || return 1
+    --imagetemplate "$image_template" || failed=1
+
+  if [[ "$GOPATH" =~ .*gopath-[0-9a-zA-Z]{10} ]]; then
```
cardil: +1

test/lib.bash
mgencur: Fixed.

test/lib.bash
mgencur: Actually, in my latest commit I don't remove the folder at all. We might need it later for extracting JUnit XML files or something else. I think it's not necessary to remove it in CI. It will be clean anyway when CI ends.

cardil: Sure. But, for running locally it will be nice to remove that junk.

test/lib.bash
```diff
@@ -121,10 +151,12 @@ function run_knative_serving_rolling_upgrade_tests {
   PROBER_PID=$!
 
   if [[ $UPGRADE_SERVERLESS == true ]]; then
-    local serving_version=$(oc get knativeserving knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.version}")
+    local serving_version
```
cardil: Worth mention that you can stack multiple local variables. I do that in first line of function. Like:
```bash
function example_func {
  local version upgrade target_version

  version="$1"
  upgrade="${SCRIPT_UPGRADE:-true}"
  target_version="$(calculate_target_version "$version" "$upgrade")"
  # [..]
}
```

test/lib.bash
```diff
@@ -53,23 +53,29 @@ function make_temporary_gopath {
...
   # Checkout the relevant code to run
-  mkdir -p "$GOPATH/src/knative.dev"
-  git clone --branch "release-${knative_version}" \
-    --depth=1 https://github.com/openshift/knative-serving.git \
-    "${GOPATH}/src/knative.dev/serving"
-  pushd "${GOPATH}/src/knative.dev/serving" || return $?
+  export KNATIVE_SERVING_HOME="$GOPATH/src/knative.dev/serving"
+  mkdir -p "$KNATIVE_SERVING_HOME"
+  git clone -b "release-${knative_version}" --depth 1 https://github.com/openshift/knative-serving.git "$KNATIVE_SERVING_HOME"
+  git describe --always --tags --dirty
```
cardil: This is the single minor issue that I have with this code. Invoking it just after checkout will always result in clean checkout, `--dirty` flag will report clean, always. Then we modify repository in some way (the `rm` command). 

Better will be to put this directly before test execution.

cardil: Other than this it looks okay to me. This is minor.

test/lib.bash
```diff
@@ -208,7 +234,7 @@ function run_knative_serving_operator_tests {
 
   logger.info "Run tests of knative/serving-operator @ ${gitdesc}"
   env TEST_NAMESPACE='knative-serving' \
-    go test -v -tags=e2e -count=1 -timeout=30m -parallel=1 ./test/e2e \
+    go_test_e2e -tags=e2e -timeout=30m -parallel=1 ./test/e2e \
```
cardil: Tests failed here. I was using `env TEST_NAMESPACE='knative-serving'` to override the namespace where tests create knative serving CR. But, it could not work with `go_test_e2e` which is bash function, not a command.

I think you need to export that variable, and after test restore it to original value (other places rely on the same name `TEST_NAMESPACE`).

PR #53, opened 2019-12-06
Invocation of upstream knative/serving-operator e2e tests

cardil:
A bash version of test invocation of upstream `knative/serving-operator`. Previous proposition is #32.

There is a patch that skips `TestKnativeServingDeployment/configure` due to SRVKS-241.

There is a `0.9.0` version of `knative/serving-operator` used, as current serverless operator do not handle `knativeservings.operator.knative.dev` definitions yet. After supporting that, versions should be set to `support/v0.10.0`. 

For `0.11+` we can use knative instead of fork (cardil), as master has configurable test namespace knative/serving-operator#223.

### DIFF
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -19,13 +19,18 @@ failed=0
 # Run serverless-operator specific tests.
 (( !failed )) && run_e2e_tests || failed=4
 
-# Run upstream knative serving tests
-RUN_KNATIVE_SERVING_UPGRADE_TESTS=false
-RUN_KNATIVE_SERVING_E2E=true
+RUN_KNATIVE_SERVING_UPGRADE_TESTS=${RUN_KNATIVE_SERVING_UPGRADE_TESTS:-false}
+RUN_KNATIVE_SERVING_E2E=${RUN_KNATIVE_SERVING_E2E:-true}
+export RUN_KNATIVE_SERVING_UPGRADE_TESTS RUN_KNATIVE_SERVING_E2E
 
-(( !failed )) && ensure_serverless_installed || failed=5
-(( !failed )) && run_knative_serving_tests "v0.10.0" || failed=6
-(( !failed )) && teardown_serverless || failed=7
+# Run upstream knative serving operator tests
+(( !failed )) && deploy_serverless_operator_latest || failed=11
+(( !failed )) && run_knative_serving_operator_tests 'support/v0.9.0' 'cardil' || failed=12
+
+# Run upstream knative serving tests
+(( !failed )) && ensure_serverless_installed || failed=6
+(( !failed )) && run_knative_serving_tests "v0.10.0" || failed=7
+(( !failed )) && teardown_serverless || failed=8
 
 (( failed )) && dump_state
 (( failed )) && exit $failed
--- a/test/lib.bash
+++ b/test/lib.bash
@@ -45,23 +45,30 @@ function run_e2e_tests {
     && return 1
 }
 
+# Setup a temporary GOPATH to safely check out the repository without breaking other things.
+# CAUTION: function overrides GOPATH so use it in subshell or restore original value!
+function make_temporary_gopath {
+  local tmp_gopath
+  tmp_gopath="$(mktemp -d -t gopath-XXXXXXXXXX)"
+  cp -rv "$(go env GOPATH)/bin" "${tmp_gopath}"
+  logger.info "Temporary GOPATH is: ${tmp_gopath}"
+  export GOPATH="$tmp_gopath"
+}
+
 function run_knative_serving_tests {
   (
   local knative_version=$1
-  # Setup a temporary GOPATH to safely check out the repository without breaking other things.
-  local tmp_gopath
-  tmp_gopath="$(mktemp -d -t gopath-XXXXXXXXXX)"
-  cp -r "$GOPATH/bin" "$tmp_gopath"
-  export GOPATH="$tmp_gopath"
+  make_temporary_gopath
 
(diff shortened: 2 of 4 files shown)


test/lib.bash
```diff
@@ -170,6 +180,51 @@ function end_prober_test {
...
+function run_knative_serving_operator_tests {
+  (
+  local version target serverless_rootdir exitstatus patchfile fork gitdesc
+  version="$1"
+  fork="${2:-knative}"
+  serverless_rootdir="$(pwd)"
+  make_temporary_gopath
+
+  logger.info "Checkout the code ${fork}/serving-operator @ ${version}"
+  mkdir -p "${GOPATH}/src/knative.dev"
```
mgencur: What about this?:
```
target="${GOPATH}/src/knative.dev/serving-operator"
mkdir -p $target
git clone .... "$target"
```
This would remove a bit of redundancy.

test/lib.bash
```diff
@@ -170,6 +180,51 @@ function end_prober_test {
...
+
+  if (( !exitstatus )); then
+    logger.success 'Tests have passed'
+  else
+    logger.error 'Tests have failures!'
+  fi
+
+  logger.info "Removing GOPATH: ${GOPATH}"
+  rm -rf "${GOPATH}"
+  popd || return $?
```
mgencur: We're running in a subshell so this is probably not necessary?

test/lib.bash
```diff
@@ -170,6 +180,51 @@ function end_prober_test {
...
+  [ -f "${patchfile}" ] || return $?
+  patch --strip=0 < "${patchfile}" || return $?
+
+  gitdesc=$(git describe --always --tags --dirty)
+
+  exitstatus=0
+
+  logger.info "Run tests of knative/serving-operator @ ${gitdesc}"
+  env TEST_NAMESPACE='knative-serving' \
+    go test -v -tags=e2e -count=1 -timeout=30m -parallel=1 ./test/e2e \
```
mgencur: Let's replace this line with `go_test_e2e -tags=e2e -timeout=30m -parallel=1 ./test/e2e \`
That way we'll have JUnit reports generated.

test/lib.bash
```diff
@@ -45,23 +45,30 @@ function run_e2e_tests {
     && return 1
 }
 
-function run_knative_serving_tests {
-  (
-  local knative_version=$1
-  # Setup a temporary GOPATH to safely check out the repository without breaking other things.
+# Setup a temporary GOPATH to safely check out the repository without breaking other things.
+# CAUTION: function overrides GOPATH so use it in subshell or restore original value!
+function make_temporary_gopath {
```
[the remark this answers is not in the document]
cardil: Why the check if GOPATH exists is important? Okay for putting $GOPATH/bin on PATH

test/e2e-tests.sh
```diff
@@ -19,13 +19,18 @@ failed=0
 # Run serverless-operator specific tests.
 (( !failed )) && run_e2e_tests || failed=4
 
-# Run upstream knative serving tests
-RUN_KNATIVE_SERVING_UPGRADE_TESTS=false
-RUN_KNATIVE_SERVING_E2E=true
+RUN_KNATIVE_SERVING_UPGRADE_TESTS=${RUN_KNATIVE_SERVING_UPGRADE_TESTS:-false}
```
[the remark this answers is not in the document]
cardil: I see. So, it will be gone. This change of mine wont damage anything.

test/lib.bash
```diff
@@ -77,6 +84,8 @@ function run_knative_serving_tests {
   image_template="registry.svc.ci.openshift.org/openshift/knative-${knative_version}:knative-serving-test-{{.Name}}"
   export GATEWAY_NAMESPACE_OVERRIDE="knative-serving-ingress"
 
+  git describe --always --tags --dirty
+
```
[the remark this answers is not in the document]
cardil: I deliberately putted it lower. `--dirty` flag shows `-dirty` in output if repo has been modified, and it does in this case.

test/lib.bash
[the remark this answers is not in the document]
mgencur: Because when it doesn't exist the `cp` command will fail and since we're using `-e` the whole script will fail. We want to re-use the script in downstream (Jenkins) testing where GOPATH might not be available as a variable.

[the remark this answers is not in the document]
cardil: That's why I use `$(go env GOPATH)`, instead of `$GOPATH`

test/lib.bash
cardil: Sure.

test/lib.bash
cardil: Yes. Not necessary, but a good practice to restore.

test/lib.bash
cardil: Sure.

test/lib.bash
```diff
@@ -86,7 +95,8 @@ function run_knative_serving_tests {
     run_knative_serving_e2e_and_conformance_tests || failed=1
   fi
 
-  rm -rf "$tmp_gopath"
+  rm -rf "${GOPATH}"
```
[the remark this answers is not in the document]
mgencur: We will fix it in the other PR

test/lib.bash
mgencur: Let's fix it in the other PR. and get this in

test/lib.bash
mgencur: We can fix this in the other PR as well.

mgencur:
/lgtm

The things I mentioned are rather cosmetic and there's another PR open that touches the code so I'll fix them there.

PR #54, opened 2019-12-08
Import knative-serving-openshift platform-specific operator

jcrossley3:
Ported over from https://github.com/openshift-knative/knative-serving-openshift/tree/master

I'm not sure if all of the files make sense within the monorepo, e.g. `.gitignore`, but we'll start with everything and then prune.


jcrossley3:
/hold
I don't think I did this in a way that preserved my history :(

PR #56, opened 2019-12-09
Require at least ocp 4.2, k8s 1.14.0 [SRVKS-371]

jcrossley3:
I've confirmed that a virgin install of this will fail if the required min version isn't available, but it is not clear how this will work for upgrading existing operators because OLM doesn't seem to distinguish versions when updating operators, i.e. only an "uninstall" option is provided when you attempt to install the upgraded CatalogSource, even after deleting the older CatalogSource first.


markusthoemmes:
/test 4.2-e2e-aws-ocp-42

just trying something.

markusthoemmes:
@jcrossley3 this is obsolete now. Martin had to add the same line to his branch to test that stuff actually worked.

jcrossley3:
Closing since #57 includes these commits.

PR #60, opened 2019-12-17
WIP Adds kourier as a postInstall step for openshift

jmprusi:
Adds kourier as a postinstall step (requires Knative crd's to be created before)

To enable kourier adding the proper ingress.class to the KnativeServing object is enough: 

```yaml
kind: KnativeServing
metadata:
  name: knative-serving
  namespace: knative-serving
spec:
  config:
   (..)
    network:
      ingress.class: kourier.ingress.networking.knative.dev

```

Currently kourier will always be installed, but we could add an specific trigger to enable/disable it's installation.

### DIFF
--- a/serving/operator/deploy/resources/kourier/kourier.yaml
+++ b/serving/operator/deploy/resources/kourier/kourier.yaml
@@ -1,0 +1,273 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: kourier
+  namespace: knative-serving
+spec:
+  ports:
+    - name: http
+      port: 80
+      protocol: TCP
+      targetPort: 8080
+  selector:
+    app: 3scale-kourier-gateway
+  type: LoadBalancer
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: 3scale-kourier-gateway
+  namespace: knative-serving
+spec:
+  progressDeadlineSeconds: 600
+  replicas: 1
+  revisionHistoryLimit: 10
+  selector:
+    matchLabels:
+      app: 3scale-kourier-gateway
+  strategy:
+    rollingUpdate:
+      maxSurge: 25%
+      maxUnavailable: 25%
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        app: 3scale-kourier-gateway
+    spec:
+      containers:
+        - args:
+            - -c
+            - /tmp/config/envoy-bootstrap.yaml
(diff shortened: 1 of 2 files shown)


mgencur:
hey @jmprusi , what's the status of this PR please? I went ahead and wanted to check performance of Kourier through this PR but the installation fails for me because the 3scale-kourier-gateway logs an error: 
```
Unable to parse JSON as proto (INVALID_ARGUMENT:(route_config.virtual_hosts[0].routes[0].match) safe_regex: Cannot find field.): {"http_filters":[{"name":"envoy.router","config":{}}],"stat_prefix":"stats_server","route_config":{"virtual_hosts":[{"routes":[{"route":{"cluster":"service_stats"},"match":{"safe_regex":{"google_re2":{},"regex":"/(certs|stats(/prometheus)?|server_info|clusters|listeners|ready)?"},"headers":[{"exact_match":"GET","name":":method"}]}}],"domains":["*"],"name":"admin_interface"}]}}
```
Perhaps it's not ready for any testing yet... ?

A few observations so far:
* All Service Mesh pods are installed even if I choose the Kourier ingress. Wouldn't it be better to only install ServiceMesh control plan if I don't choose Kourier (to save a lot of resources)
* It installs 3scale-kourier-control and 3scale-kourier-gateway pods in the knative-serving namespace. I think I like it more than having it in a separate namespace but the fact is that so far we were installing the ingress in a separate one. Just curious what's your thought on that.

serving/operator/deploy/resources/kourier/kourier.yaml
```diff
@@ -0,0 +1,271 @@
...
+  namespace: knative-serving
+spec:
+  ports:
+    - name: http2
+      port: 80
+      protocol: TCP
+      targetPort: 8081
+  selector:
+    app: 3scale-kourier-gateway
+  type: ClusterIP
```
mgencur: Upstream has this as LoadBalancer. And all the other services in this file as well. Not sure if this is relevant.. Just making sure you're aware

davidor: We changed this after this PR was opened. We'll need to update this PR with the latest changes.

jmprusi:
@mgencur I've been told to update the PR against master instead of the release-1.3. So I will close this one and open an update one against master. Does it make sense? 

Thanks.

mgencur:
Yes. Makes sense.

jmprusi:
closing in favour of #70

PR #62, opened 2020-01-08
API separation in tests

mgencur:
This is related to "Backwards compatibility for installation" in https://issues.redhat.com/browse/SRVLS-199

The idea is that we explicitly choose version of KnativeServing CR for the tests. There are two two types of tests:
* serverless-operator tests in test/e2e/ folder

    These tests are deploying KnativeServing from test/v1alpha1 package using Go. In the future they could deploy a newer version to test new features (e.g. v1beta1, v1, etc.)

* knative-serving tests in which are pulled from github.com/openshift/knative-serving on the fly when running `e2e-tests.sh`

  Before these tests run, the KnativeServing CR is deployed from Bash. Now that it is versioned we could make sure this KnativeServing CR stays at version v1alpha1 even for future releases and we could test backwards compatibility through it. Basically, making sure that this old CR can still be used with newer releases and doesn't break installation.

### DIFF
--- a/hack/lib/serverless.bash
+++ b/hack/lib/serverless.bash
@@ -97,14 +97,11 @@ function deploy_knativeserving_v1alpha1 {
   # Wait for the CRD to appear
   timeout 900 "[[ \$(oc get crd | grep -c knativeservings) -eq 0 ]]" || return 6
 
+  local rootdir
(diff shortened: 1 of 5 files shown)


mgencur:
Replaced by https://github.com/openshift-knative/serverless-operator/pull/62

PR #63, opened 2020-01-09
Bring over openshift-specific knative operator

jcrossley3:
History available in https://github.com/openshift-knative/knative-serving-openshift


matzew:
@aliok eventually we want this too, for eventing - as a heads up

PR #67, opened 2020-01-15
Add CSV for 1.4.0 containing Serving 0.11.1, freeze 1.3.0 image references.

markusthoemmes:
Title says it all. This introduces the base CSV for the upcoming 1.4.0 version, which contains the image references for Serving 0.11.1.

In addition to that, this freezes the image references for the 1.3.0 version as this no longer needs to be built in CI and thus can be used for upgrade testing.


mgencur:
Could you add -x on this line to run with more logging? https://github.com/openshift-knative/serverless-operator/blob/master/test/upgrade-tests.sh#L6 
I've tried your branch locally and it timeouts on this condition: https://github.com/openshift-knative/serverless-operator/blob/master/test/lib.bash#L176
Which means the KnativeServing CR still reports version 0.10 and not 0.11. I've checked the pods and they still show this:
```
labels:
    app: activator
    pod-template-hash: 6c4c78c774
    role: activator
    serving.knative.dev/release: v0.10.0
```
But the image seems to have the right version: registry.svc.ci.openshift.org/openshift/knative-v0.11.1:knative-serving-activator

mgencur:
I think we should be running with -x all the time

olm-catalog/serverless-operator/1.4.0/serverless-operator.v1.4.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,385 @@
...
+                "sample-rate": "0.1"
+              }
+            }
+          }
+        }
+      ]
+    capabilities: Seamless Upgrades
+    categories: Networking,Integration & Delivery,Cloud Provider,Developer Tools
+    certified: "false"
+    containerImage: quay.io/openshift-knative/serverless-operator:v1.4.0
```
jcrossley3: I think @warrenvw and I decided we could get rid of this. I think it's purely metadata (for display on some console somewhere) and mostly just causes confusion.

serving/operator/deploy/resources/knative-serving-0.11.1.yaml
```diff
@@ -1053,21 +1183,18 @@ data:
...
     # panic-window-percentage of the stable-window.
+    # When computing the panic window it will be rounded to the closest
+    # whole second.
     panic-window-percentage: "10.0"
 
-    # Absolute panic window duration.
-    # Deprecated in favor of panic-window-percentage.
-    # Existing revisions will continue to scale based on panic-window
-    # but new revisions will default to panic-window-percentage.
-    panic-window: "6s"
```
jcrossley3: This is the kind of thing that the operator could deal with: checking for a `panic-window` key in the CR's config map-of-maps and either converting to some `panic-window-percentage` or logging a warning. Probably not worth doing for this, but good to keep in mind for future updates.

serving/operator/deploy/resources/knative-serving-0.11.1.yaml
```diff
@@ -1908,8 +2039,12 @@ data:
 
     # Default Knative Gateway after v0.3. It points to the Istio
     # standard istio-ingressgateway, instead of a custom one that we
-    # used pre-0.3.
-    gateway.knative-ingress-gateway: "istio-ingressgateway.istio-system.svc.cluster.local"
```
jcrossley3: ibid

serving/operator/deploy/resources/knative-serving-0.11.1.yaml
```diff
@@ -1919,10 +2054,16 @@ data:
     # sidecar injection (like Knative's istio-lean.yaml).  Since every pod
     # is outside of the service mesh in that case, a cluster-local  service
     # will need to be exposed to a cluster-local gateway to be accessible.
-    local-gateway.cluster-local-gateway: "cluster-local-gateway.istio-system.svc.cluster.local"
```
jcrossley3: ibid

serving/operator/pkg/controller/knativeserving/knativeserving_controller.go
```diff
@@ -153,7 +151,6 @@ func (r *ReconcileKnativeServing) Reconcile(request reconcile.Request) (reconcil
 	stages := []func(*servingv1alpha1.KnativeServing) error{
 		r.ensureFinalizers,
 		r.initStatus,
-		r.checkWebhooks,
```
jcrossley3: Good riddance!

PR #68, opened 2020-01-15
Create ConsoleCLIDownload CR for kn CLI

### LINKED ISSUE #37: Add the creation / update for a ConsoleCLIDownload resource
In order to make the download of `kn` accessible in the console, the operator needs to create or update (or delete & create would also work) a resource that looks like

```yaml
apiVersion: console.openshift.io/v1
kind: ConsoleCLIDownload
metadata:
  name: kn
spec:
  description: The OpenShift Serverless client `kn` is a CLI tool that allows you to fully manage OpenShift Serverless Serving and Eventing resources without writing a single line of YAML.
  displayName: kn - OpenShift Serverless CLI
  links:
  - href: https://mirror.openshift.com/pub/openshift-v4/clients/serverless/0.9.0/kn-linux-amd64-0.9.0.tar.gz
    text: Download kn (Linux)
  - href: https://mirror.openshift.com/pub/openshift-v4/clients/serverless/0.9.0/kn-macos-amd64-0.9.0.tar.gz
    text: Download kn (macOS)
  - href: https://mirror.openshift.com/pub/openshift-v4/clients/serverless/0.9.0/kn-windows-amd64-0.9.0.zip
    text: Download kn (Windows)
```

The version (0.9.0) in the example is the version of the kn release included in the serverless release.

This will lead automatically lead to a download page that looks like 

grafik

In addition, the base URL (`https://mirror.openshift.com/pub/openshift-v4/clients`) must be configurable via the main Serverless CRD (e.g. via `knBaseDownloadUrl:` as there is also the requirement to support 'air-gapped' downloads when the cluster has no access to the internet. (the interesting bit would be from where to serve the binaries, but that's the second part of the story).

Ideally, the version number (0.9.0) is parameterizable and filled in just before the CO is created by the operator.

navidshaikh:
Fixes #37 

- Create ConsoleCLIDownload instance part of openshift-specific knative operator
install to create ConsoleCLIDownload for kn.
- Add ConsoleCLIDownload resource YAML knative-operator/deploy/resources/console_cli_download_kn.yaml
- Create and Delete functions for ConsoleCLIDownload added as separate package under knative-operator/pkg/controller/knativeserving/consoleclidownload
- Creation of ConsoleCLIDownload treated as stage like others stages in operator
- Update link for kn v0.11.0 version download from mirror

image

### DIFF
--- a/knative-operator/deploy/resources/console_cli_download_kn.yaml
+++ b/knative-operator/deploy/resources/console_cli_download_kn.yaml
@@ -1,0 +1,14 @@
+apiVersion: console.openshift.io/v1
+kind: ConsoleCLIDownload
+metadata:
+  name: kn
+spec:
+  description: The OpenShift Serverless client `kn` is a CLI tool that allows you to fully manage OpenShift Serverless Serving and Eventing resources without writing a single line of YAML.
+  displayName: kn - OpenShift Serverless CLI
+  links:
+  - href: https://mirror.openshift.com/pub/openshift-v4/clients/serverless/0.11.0/kn-linux-amd64-0.11.0.tar.gz
+    text: Download kn for Linux
+  - href: https://mirror.openshift.com/pub/openshift-v4/clients/serverless/0.11.0/kn-macos-amd64-0.11.0.tar.gz
+    text: Download kn for macOS
+  - href: https://mirror.openshift.com/pub/openshift-v4/clients/serverless/0.11.0/kn-windows-amd64-0.11.0.zip
+    text: Download kn for Windows
--- a/knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
+++ b/knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
@@ -5,6 +5,7 @@ import (
 
 	mf "github.com/jcrossley3/manifestival"
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/common"
+	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/controller/knativeserving/consoleclidownload"
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/controller/knativeserving/servicemesh"
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 	corev1 "k8s.io/api/core/v1"
@@ -93,12 +94,14 @@ func (r *ReconcileKnativeServing) Reconcile(request reconcile.Request) (reconcil
 		r.ensureCustomCertsConfigMap,
 		r.installNetworkPolicies,
 		r.installServiceMesh,
+		r.createConsoleCLIDownload,
 	}
 	for _, stage := range stages {
 		if err := stage(instance); err != nil {
 			return reconcile.Result{}, err
 		}
 	}
+
 	return reconcile.Result{}, nil
 }
 
@@ -179,6 +182,11 @@ func (r *ReconcileKnativeServing) installServiceMesh(instance *servingv1alpha1.K
 	return servicemesh.ApplyServiceMesh(instance, r.client)
 }
 
+// createConsoleCLIDownload creates CR for kn CLI download link
+func (r *ReconcileKnativeServing) createConsoleCLIDownload(instance *servingv1alpha1.KnativeServing) error {
+	return consoleclidownload.Create(instance, r.client)
+}
+
 // general clean-up, mostly service mesh resources
 func (r *ReconcileKnativeServing) delete(instance *servingv1alpha1.KnativeServing) error {
 	if len(instance.GetFinalizers()) == 0 || instance.GetFinalizers()[0] != finalizerName() {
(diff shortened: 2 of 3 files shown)


knative-operator/deploy/console_cli_download.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: console.openshift.io/v1
```
jcrossley3: Let's move this to `deploy/resources/` which is the (arbitrary) conventional place for any yaml applied by the operator. Humans are expected to apply the stuff in `deploy/`

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
```diff
@@ -87,6 +87,10 @@ func (r *ReconcileKnativeServing) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, r.delete(instance)
 	}
 
+	if err := r.createConsoleCLIDownloadCR(); err != nil {
```
jcrossley3: I'd prefer to treat this as a "stage", even if you never use the passed instance and never return an error.

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
```diff
@@ -174,6 +179,40 @@ func (a *ReconcileKnativeServing) installNetworkPolicies(instance *servingv1alph
 	return nil
 }
 
+// createConsoleCLIDownloadCR creates CR for kn CLI download link
+func (a *ReconcileKnativeServing) createConsoleCLIDownloadCR() error {
```
jcrossley3: Add an instance parameter so it can be treated as a "stage" and just for the sake of clarity, let's move the guts of this function, along with the accompanying `deleteConsoleCLIDownloadCR` function to their own package, creating a `cli/` subdir similar to `servicemesh/`. Within their own package, the function names could then be `Create` and `Delete`. That would keep the controller cleaner, I think. Make sense?

navidshaikh: Yes, thanks for the review!

I've updated the PR, also now the errors for create/delete of ConsoleCLIDownload are not suppressed but returned.

jcrossley3:
Looks good! Much cleaner! Once I get #66 sorted out, we'll rebase this and merge. Thanks!

knative-operator/deploy/resources/console_cli_download_kn.yaml
```diff
@@ -0,0 +1,14 @@
...
+spec:
+  description: The OpenShift Serverless client `kn` is a CLI tool that allows you to fully manage OpenShift Serverless Serving and Eventing resources without writing a single line of YAML.
+  displayName: kn - OpenShift Serverless CLI
+  links:
+  - href: https://mirror.openshift.com/pub/openshift-v4/clients/serverless/0.11.0/kn-linux-amd64-0.11.0.tar.gz
+    text: Download kn for Linux
+  - href: https://mirror.openshift.com/pub/openshift-v4/clients/serverless/0.11.0/kn-macos-amd64-0.11.0.tar.gz
+    text: Download kn for macOS
+  - href: https://mirror.openshift.com/pub/openshift-v4/clients/serverless/0.11.0/kn-windows-amd64-0.11.0.zip
+    text: Download kn for Windows
```
navidshaikh: Added a commit which adds to-be-synced-at-mirror kn version for 1.4.0 release (as kn v0.11.0 tar ball at mirror will be synced after QA is done, which is past our code-freeze date)

navidshaikh: Another thing we could do here is keep the link of `latest` i.e. https://mirror.openshift.com/pub/openshift-v4/clients/serverless/latest/ , which is sym-linked to latest available kn version at mirror.

Cc: @rhuss

navidshaikh:
Rebased and squashed the commits, PTAL.

knative-operator/deploy/resources/console_cli_download_kn.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: console.openshift.io/v1
+kind: ConsoleCLIDownload
+metadata:
+  name: kn
+spec:
+  description: The OpenShift Serverless client `kn` is a CLI tool that allows you to fully manage OpenShift Serverless Serving and Eventing resources without writing a single line of YAML.
+  displayName: kn - OpenShift Serverless CLI
+  links:
+  - href: https://mirror.openshift.com/pub/openshift-v4/clients/serverless/0.11.0/kn-linux-amd64-0.11.0.tar.gz
```
rhuss: I wonder whether we should parameterize the version number at some point in time (and make it part of the build process to put that into the resource file). Maybe manifestival know how to process templates of that kind ? Then we could add a placeholder here and add the real version from within the operator itself.

jcrossley3: I'm open to env substitution in manifestival, but in this case it might be just as easy to munge the appropriate `manifest.Resources` to replace some token in the file with `instance.Status.Version` before invoking `manifest.ApplyAll`

rhuss: yeah, whatever :) Doing the replace ourselves is probably easier (only if manifestival would already have supported templating then we could have used that). But could do now that it's easy to get to the operator version (which actually also corresponds to the `kn` version).

@navidshaikh do you think this would be hard to implement ? we can do that post 1.4 though (as the current approach works)

navidshaikh: We can make it more parameterized, sure!
With air-gapped scenario, will have to re-work the ConsoleCLIDownload and associated artifacts management, will keep this in mind. Thanks for the review and suggestions!

PR #70, opened 2020-01-16
WIP Adds kourier as a postInstall step for openshift

jmprusi:
Adds kourier as a postinstall step (requires Knative crd's to be created before)

To enable kourier adding the proper ingress.class to the KnativeServing object is enough:

```yaml
kind: KnativeServing
metadata:
  name: knative-serving
  namespace: knative-serving
spec:
  config:
   (..)
    network:
      ingress.class: kourier.ingress.networking.knative.dev
```

Currently kourier will always be installed, but we could add an specific trigger to enable/disable it's installation.

### DIFF
--- a/knative-operator/deploy/resources/kourier/kourier.yaml
+++ b/knative-operator/deploy/resources/kourier/kourier.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: kourier-system
+  name: knative-serving-ingress
 ---
 apiVersion: v1
 kind: ConfigMap
--- a/knative-operator/deploy/resources/kourier/kourier.yaml
+++ b/knative-operator/deploy/resources/kourier/kourier.yaml
@@ -1,8 +1,8 @@
-#apiVersion: v1
-#kind: Namespace
-#metadata:
-#  name: kourier-system
-#---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: kourier-system
+---
 apiVersion: v1
 kind: ConfigMap
 metadata:
--- a/knative-operator/deploy/resources/kourier/kourier.yaml
+++ b/knative-operator/deploy/resources/kourier/kourier.yaml
@@ -1,0 +1,275 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: kourier-system
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config-logging
+  namespace: kourier-system
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config-observability
+  namespace: kourier-system
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: kourier
+  namespace: kourier-system
+spec:
+  ports:
+    - name: http2
+      port: 80
+      protocol: TCP
+      targetPort: 8080
+    - name: https
+      port: 443
+      protocol: TCP
+      targetPort: 8443
+  selector:
+    app: 3scale-kourier-gateway
+  type: LoadBalancer
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: 3scale-kourier-gateway
+  namespace: kourier-system
+spec:
+  progressDeadlineSeconds: 600
+  replicas: 1
+  revisionHistoryLimit: 10
+  selector:
+    matchLabels:
+      app: 3scale-kourier-gateway
+  strategy:
+    rollingUpdate:
+      maxSurge: 25%
+      maxUnavailable: 25%
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        app: 3scale-kourier-gateway
+    spec:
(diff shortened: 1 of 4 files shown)


jmprusi:
@mgencur created this new PR against Master, with the latest version and fixed to work with the maistra envoy version. (the regex part...)

mgencur:
@jmprusi great, thanks! I'll run some tests to compare performance with Service Mesh.

mgencur:
hi @jmprusi I've tried to install the stuff. At first my knative service was not reachable. Then I looked at the route saw it was using the target port `http2`:

```
spec:
  host: >-
    autoscale-go.tests.apps.ocf-perf-10-ocf-ocp4-aws-release-mgencur-tmp.openshift-aws.rhocf-dev.com
  subdomain: ''
  to:
    kind: Service
    name: kourier
    weight: 100
  port:
    targetPort: http2
  tls:
    termination: edge
    insecureEdgeTerminationPolicy: Allow
  wildcardPolicy: None
```
The routes created by the Ingress operator use the "http2" target port" [link](https://github.com/openshift-knative/serverless-operator/blob/master/serving/ingress/pkg/controller/resources/route.go#L131)
I just tried to change the route to use "http" because the Kourier service defines that [link](https://github.com/openshift-knative/serverless-operator/pull/70/files#diff-5eb757727d79626f93f674d1501a0f9eR25) and my knative service started to work. 
So I think we need to change the Kourier service to use http2

serving/operator/deploy/resources/kourier/kourier.yaml
```diff
@@ -0,0 +1,275 @@
...
+  namespace: kourier-system
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: kourier
+  namespace: kourier-system
+spec:
+  ports:
+    - name: http
```
mgencur: This should be `http2`. See me other comment on this PR.

jmprusi: @mgencur Ok, I will fix it. Thanks!

mgencur:
hey @jmprusi I did another round of testing on this PR and I've spotted a problem there: Even if I use knative-serving without `ingress.class: kourier.ingress.networking.knative.dev` and expecting Service Mesh to take care of the ingress, it's not true. In fact, the Route for my Knative Service is created in kourier-system namespace and not in knative-serving-ingress. Also, my namespace that I added to ServiceMeshMemberRoll will get removed from it. The communication with my application still goes through the Kourier gateway.

mgencur:
It would be good to explicitly enable Kourier in tests so that failures can be caught by this CI.

jmprusi:
@mgencur I'm "rebasing" on top of the master branch and pushing the latest version of kourier that fixes a couple of new knative integration tests!

knative-operator/deploy/resources/kourier/kourier.yaml
```diff
@@ -0,0 +1,275 @@
...
+kind: ConfigMap
+metadata:
+  name: config-observability
+  namespace: kourier-system
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: kourier
+  namespace: kourier-system
```
nak3: I think we should deploy all kourier components (or only kourier gateway) in `knative-serving-ingress` namespace.
Otherwise, k-o-i needs to create OpenShift's Route in the same namespace with 3scale-kourier-gateway.

/cc @bbrowning

jmprusi: I'm deploying the control plane and the gateway in the knative-serving-ingress namespace. Working fine in my cluster...

jmprusi: I'm deploying the control plane and the gateway in the knative-serving-ingress namespace. Working fine in my cluster...

nak3:
/retest

Hmm... why knative-serving-* pod did not come up. https://gcsweb-ci.svc.ci.openshift.org/gcs/origin-ci-test/pr-logs/pull/openshift-knative_serverless-operator/70/pull-ci-openshift-knative-serverless-operator-master-4.3-e2e-aws-ocp-43/78/artifacts/e2e-aws-ocp-43/pods/

nak3:
Alright, it looks like tests just failed on serving's e2e test. To solve it, we need to fix at least following/reconcile istio configs.

1. Update config-istio configmap or (remove it), Especially these settings:

```
  gateway.knative-ingress-gateway: istio-ingressgateway.knative-serving-ingress.svc.cluster.local
  local-gateway.cluster-local-gateway: cluster-local-gateway.knative-serving-ingress.svc.cluster.local
```

They are from here https://github.com/openshift-knative/serverless-operator/blob/3df6ebc78f3c6da5f817eff6c71229b8d23aefe2/knative-operator/pkg/controller/knativeserving/servicemesh/servicemesh.go#L121-L122

2. Re-create kingress  due to the issue https://github.com/knative/serving/pull/6569

jmprusi:
superseded by #85

PR #73, opened 2020-01-17
Add relatedImages block

warrenvw:
This is currently required for disconnected installs.

### DIFF
--- a/olm-catalog/serverless-operator/1.4.0/serverless-operator.v1.4.0.clusterserviceversion.yaml
+++ b/olm-catalog/serverless-operator/1.4.0/serverless-operator.v1.4.0.clusterserviceversion.yaml
@@ -381,5 +381,24 @@ spec:
   maturity: alpha
   provider:
     name: Red Hat, Inc.
+  relatedImages:
+    - name: IMAGE_QUEUE
+      image: registry.svc.ci.openshift.org/openshift/knative-v0.11.1:knative-serving-queue
+    - name: IMAGE_activator
+      image: registry.svc.ci.openshift.org/openshift/knative-v0.11.1:knative-serving-activator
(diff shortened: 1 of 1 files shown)


olm-catalog/serverless-operator/1.4.0/serverless-operator.v1.4.0.clusterserviceversion.yaml
```diff
@@ -381,5 +381,24 @@ spec:
...
+    - name: IMAGE_autoscaler-hpa
+      image: registry.svc.ci.openshift.org/openshift/knative-v0.11.1:knative-serving-autoscaler-hpa
+    - name: IMAGE_controller
+      image: registry.svc.ci.openshift.org/openshift/knative-v0.11.1:knative-serving-controller
+    - name: IMAGE_networking-istio
+      image: registry.svc.ci.openshift.org/openshift/knative-v0.11.1:knative-serving-istio
+    - name: IMAGE_webhook
+      image: registry.svc.ci.openshift.org/openshift/knative-v0.11.1:knative-serving-webhook
+    - name: knative-operator
+      image: $IMAGE_KNATIVE_OPERATOR
```
markusthoemmes: This makes it dependent on #66, is that intentional? 

I'd rather we wait until #66 merges and move this in as-is or add the knative-serving-operator as it's used today instead of this to keep the documents correct (and to keep us sane).

If you change this to be knative-serving-operator, @jcrossley3 can change it to this reference in his PR to keep it all in one spot.

warrenvw: Let's merge #73 after #66 is merged.

PR #75, opened 2020-01-21
Add network policy for cluster-local-gateway

savitaashture:
Jira: https://issues.redhat.com/browse/SRVKS-378

```
kubectl get NetworkPolicy -n knative-serving-ingress cluster-local-gateway -o yaml
apiVersion: extensions/v1beta1
kind: NetworkPolicy
metadata:
  annotations:
    manifestival: new
  creationTimestamp: "2020-01-16T21:23:19Z"
  generation: 1
  name: cluster-local-gateway
  namespace: knative-serving-ingress
  ownerReferences:
  - apiVersion: maistra.io/v1
    blockOwnerDeletion: true
    controller: true
    kind: ServiceMeshControlPlane
    name: basic-install
    uid: 2dce7bf1-38a6-11ea-8cf9-52fdfc072182
  resourceVersion: "153229"
  selfLink: /apis/extensions/v1beta1/namespaces/knative-serving-ingress/networkpolicies/cluster-local-gateway
  uid: 6af24170-38a6-11ea-8cf9-52fdfc072182
spec:
  ingress:
  - from:
    - namespaceSelector: {}
  podSelector:
    matchLabels:
      istio: cluster-local-gateway
  policyTypes:
  - Ingress
```


knative-operator/pkg/controller/knativeserving/servicemesh/servicemesh.go
```diff
@@ -239,6 +242,39 @@ func isServiceMeshMemberRollReady(servingNamespace string, api client.Client) (b
...
+	ready, err := isServiceMeshControlPlaneReady(instanceNamespace, api)
+	if err != nil {
+		return err
+	}
+	if ready {
+		if err := api.Get(context.TODO(), client.ObjectKey{Namespace: namespace, Name: smcpName}, smcp); err != nil {
+			return err
+		}
+	}
+	transforms := []mf.Transformer{mf.InjectOwner(smcp), mf.InjectNamespace(namespace)}
```
nak3: ownerreferences for smcp causes some minor issue. For example, this network-policy remains if we uninstall serverless-operator. _But_ I guess this should be align with Maistra's istio-ingressgateway. So, let's leave this as you implemented.

knative-operator/pkg/controller/knativeserving/servicemesh/servicemesh.go
```diff
@@ -239,6 +242,39 @@ func isServiceMeshMemberRollReady(servingNamespace string, api client.Client) (b
...
+	namespace := ingressNamespace(instanceNamespace)
+	const path = "deploy/resources/serviceMesh/networkpolicies.yaml"
+
+	manifest, err := mf.NewManifest(path, false, api)
+	if err != nil {
+		log.Error(err, "Unable to create Mesh Network Policy install manifest")
+		return err
+	}
+	smcp := &maistrav1.ServiceMeshControlPlane{}
+	ready, err := isServiceMeshControlPlaneReady(instanceNamespace, api)
```
markusthoemmes: Do we really need to wait for it becoming ready? I feel like we could just add the networkpolicy once the SMCP is there.

knative-operator/pkg/controller/knativeserving/servicemesh/servicemesh.go
```diff
@@ -239,6 +242,39 @@ func isServiceMeshMemberRollReady(servingNamespace string, api client.Client) (b
...
+	if err != nil {
+		return err
+	}
+	if ready {
```
nak3: I guess this `ready` condition is not necessary. But if you want, you need to take care when `isServiceMeshControlPlaneReady()` returns `false, nil` from [here](https://github.com/openshift-knative/serverless-operator/pull/75/files#diff-4850e596b590230168aecc3c95fbd473R160).

knative-operator/pkg/controller/knativeserving/servicemesh/servicemesh.go
```diff
@@ -161,7 +164,7 @@ func isServiceMeshControlPlaneReady(servingNamespace string, api client.Client)
 func installServiceMeshControlPlane(instance *servingv1alpha1.KnativeServing, api client.Client) error {
 	log.Info("Installing serviceMeshControlPlane")
 	const (
-		path = "deploy/resources/servicemesh.yaml"
+		path = "deploy/resources/serviceMesh/servicemesh.yaml"
```
nak3: I think this is not "official" rule, but capital letter is not used in directory or file name in general. So, the directory should be `servicemesh` instead of `serviceMesh`?

jcrossley3: Yes, please :)

knative-operator/pkg/controller/knativeserving/servicemesh/servicemesh.go
jcrossley3: The GC should delete the NP when the SMCP is deleted due to the ownerref. And the KS controller will delete the entire `knative-serving-ingress` namespace when the KS CR is deleted. So I think everything will be properly cleaned up. Of course, none of that happens if this operator isn't running (i.e. uninstall serverless-operator), but that's expected, right?

knative-operator/pkg/controller/knativeserving/servicemesh/servicemesh.go
jcrossley3: I agree. If we can `Get` the resource, we should install the NP, regardless of the former's status.

savitaashture: Added **isServiceMeshControlPlaneReady** in order to make sure that before creating networkpolicy SMCP is ready

But i can just do GET operation on SMCP in order to get the data for `InjectOwner`

knative-operator/pkg/controller/knativeserving/servicemesh/servicemesh.go
savitaashture: Sure thank you

savitaashture:
Thank you @nak3 @jcrossley3 @markusthoemmes 
Updated PR by addressing all the review comments

markusthoemmes:
A test in https://github.com/openshift-knative/serverless-operator/blob/master/test/e2e/knative_serving_test.go#L90 that creates a cluster-local service and then verifies it can be hit from outside the mesh (i.e. from a k8s service in a non-mesh namespace) would be awesome!

markusthoemmes:
FWIW such a test should fail today and would repro the issue as well.

PR #76, opened 2020-01-22
Remove the serving operator.

markusthoemmes:
The old serving/operator has been replaced by the knative-operator + the upstream serving operator so all this code is no longer needed.


knative-operator/pkg/apis/addtoscheme_obsolete_serving_v1alpha1.go
```diff
@@ -1,7 +1,7 @@
 package apis
 
 import (
-	"github.com/openshift-knative/serverless-operator/serving/operator/pkg/apis/serving/v1alpha1"
+	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/apis/serving/v1alpha1"
```
jcrossley3: This is a mistake, isn't it? We need to refer to the old CRD to support migration. That's the what the "obsolete controller" watches.

markusthoemmes: Absolutely, changed the approach.

PR #77, opened 2020-01-23
Separate serving stuff

aliok:
To allow eventing code later, I refactored and moved around some code.

Note: I wanted to send a separate PR to your PR branch to have a clear boundary for eventing code.
I didn't want to refactor serving related stuff in the same PR where I add eventing stuff.

Changes:
- Moved code in `knative-operator/.../common` to `knative-operator/.../common/serving` as they were only relevant for serving. I will add common things for eventing into `knative-operator/.../common/eventing` and we can keep common things for both serving and eventing in `knative-operator/.../common`
- Deleted the confusing and unnecessary file `knative-operator/build/Dockerfle` (see my comment in the PR)
- Rename `REQUIRED_NAMESPACE` to `REQUIRED_SERVING_NAMESPACE`
- Add `serving` prefix to serving resources


knative-operator/build/Dockerfile
```diff
@@ -1,18 +0,0 @@
...
-
-COPY build/bin /usr/local/bin
-RUN  /usr/local/bin/user_setup
-
-# install manifest[s]
-COPY deploy /deploy
-
-ENTRYPOINT ["/usr/local/bin/entrypoint"]
-
-USER ${USER_UID}
```
aliok: I don't think this file is necessary anymore.

We're not using `operator-sdk build` for building the image.
This file actually creates some confusion and if we keep it to support `operator-sdk build` command, we need to make sure we sync it with upper level `Dockerfile`.

knative-operator/pkg/common/serving/openshift.go
```diff
@@ -1,4 +1,4 @@
-package common
+package serving
```
jcrossley3: I'm struggling with this, honestly. I kinda don't want to reorganize the common dir until we think we have something eventing-related that should go in here.

aliok: @jcrossley3 fair enough. let's keep this PR as is until the eventing related code is free to go into this repo.

aliok:
Closing in favor of https://github.com/openshift-knative/serverless-operator/pull/126

PR #80, opened 2020-01-27
Backwards compatibility tests for KnativeServing CRD

mgencur:
* pin the version of KnativeServing to 1.3.0

Replaces https://github.com/openshift-knative/serverless-operator/pull/62

This is related to "Backwards compatibility for installation" in https://issues.redhat.com/browse/SRVLS-199

The idea is that we explicitly choose version of KnativeServing CR for the tests. There are two two types of tests:

serverless-operator tests in test/e2e/ folder

These tests are deploying KnativeServing from test/v1alpha1 package using Go. In the future they could deploy a newer version to test new features (e.g. v1beta1, v1, etc.)

knative-serving tests in which are pulled from github.com/openshift/knative-serving on the fly when running e2e-tests.sh

Before these tests run, the KnativeServing CR is deployed from Bash. Now that it is versioned we could make sure this KnativeServing CR stays at version v1alpha1 even for future releases and we could test backwards compatibility through it. Basically, making sure that this old CR can still be used with newer releases and doesn't break installation.

We can also decide if we ensure backwards compatibility with the KnativeServing from Serverless 1.3.0 release or if the oldest config would be 1.4.0


test/v1alpha1/resources/serving.knative.dev_v1alpha1_knativeserving_cr_1.3.0.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: serving.knative.dev/v1alpha1
```
mgencur: The content of this file is taken from the example we provide when creating KnativeServing CR via console: https://github.com/openshift-knative/serverless-operator/blob/master/olm-catalog/serverless-operator/1.3.0/serverless-operator.v1.3.0.clusterserviceversion.yaml#L5

mgencur:
@jcrossley3 the tests fail on this line: https://github.com/openshift-knative/serverless-operator/pull/80/files#diff-a9cb53ea4b16710cd3d15c2b9f865f35R109 , should we wait for a kantiveserving in "operator" group to be ready even if installing with "serving" group? If not then I think something is still missing in the code.

mgencur:
I've looked at CI while it was running and the KnativeServing doesn't get processed it seems:
```
oc get knativeserving.serving.knative.dev -n knative-serving -oyaml
apiVersion: v1
items:
- apiVersion: serving.knative.dev/v1alpha1
  kind: KnativeServing
  metadata:
    annotations:
      kubectl.kubernetes.io/last-applied-configuration: |
        {"apiVersion":"serving.knative.dev/v1alpha1","kind":"KnativeServing","metadata":{"annotations":{},"name":"knative-serving","namespace":"knative-serving"},"spec":{"config":{"autoscaler":{"container-concurrency-target-default":"100","container-concurrency-target-percentage":"1.0","enable-scale-to-zero":"true","max-scale-up-rate":"10","panic-threshold-percentage":"200.0","panic-window":"6s","panic-window-percentage":"10.0","scale-to-zero-grace-period":"30s","stable-window":"60s","tick-interval":"2s"},"defaults":{"revision-timeout-seconds":"300"},"deployment":{"registriesSkippingTagResolving":"ko.local,dev.local"},"gc":{"stale-revision-create-delay":"24h","stale-revision-lastpinned-debounce":"5h","stale-revision-minimum-generations":"1","stale-revision-timeout":"15h"},"logging":{"loglevel.activator":"info","loglevel.autoscaler":"info","loglevel.controller":"info","loglevel.queueproxy":"info","loglevel.webhook":"info"},"observability":{"logging.enable-var-log-collection":"false","metrics.backend-destination":"prometheus"},"tracing":{"backend":"none","sample-rate":"0.1"}}}}
    creationTimestamp: "2020-01-28T14:01:48Z"
    generation: 1
    name: knative-serving
    namespace: knative-serving
    resourceVersion: "35375"
    selfLink: /apis/serving.knative.dev/v1alpha1/namespaces/knative-serving/knativeservings/knative-serving
    uid: 8ea569cb-e987-464b-a4b6-d6bb858d45ef
  spec:
    config:
      autoscaler:
        container-concurrency-target-default: "100"
        container-concurrency-target-percentage: "1.0"
        enable-scale-to-zero: "true"
        max-scale-up-rate: "10"
        panic-threshold-percentage: "200.0"
        panic-window: 6s
        panic-window-percentage: "10.0"
        scale-to-zero-grace-period: 30s
        stable-window: 60s
        tick-interval: 2s
      defaults:
        revision-timeout-seconds: "300"
      deployment:
        registriesSkippingTagResolving: ko.local,dev.local
      gc:
        stale-revision-create-delay: 24h
        stale-revision-lastpinned-debounce: 5h
        stale-revision-minimum-generations: "1"
        stale-revision-timeout: 15h
      logging:
        loglevel.activator: info
        loglevel.autoscaler: info
        loglevel.controller: info
        loglevel.queueproxy: info
        loglevel.webhook: info
      observability:
        logging.enable-var-log-collection: "false"
        metrics.backend-destination: prometheus
      tracing:
        backend: none
        sample-rate: "0.1"
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```

markusthoemmes:
@mgencur we'll need to ignore if we're not actually upgrading but installing from an old CR.

```diff
diff --git a/knative-operator/pkg/controller/knativeservingobsolete/knativeservingobsolete_controller.go b/knative-operator/pkg/controller/knativeservingobsolete/knativeservingobsolete_controller.go
index 437928dd..cf9bc94f 100644
--- a/knative-operator/pkg/controller/knativeservingobsolete/knativeservingobsolete_controller.go
+++ b/knative-operator/pkg/controller/knativeservingobsolete/knativeservingobsolete_controller.go
@@ -121,6 +121,10 @@ func (r *ReconcileKnativeServingObsolete) removeOldCertsConfig(ns string) error
 	const name = "controller"
 	deployment := &appsv1.Deployment{}
 	if err := r.client.Get(context.TODO(), client.ObjectKey{Namespace: ns, Name: name}, deployment); err != nil {
+		// Ignore an IsNotFound error. Nothing to do for us.
+		if errors.IsNotFound(err) {
+			return nil
+		}
 		return err
 	}
 	volumes := deployment.Spec.Template.Spec.Volumes
```

Should fix that.

hack/lib/serverless.bash
```diff
@@ -92,23 +92,21 @@ function find_install_plan {
...
-apiVersion: ${group}.knative.dev/v1alpha1
-kind: KnativeServing
-metadata:
-  name: knative-serving
-  namespace: ${SERVING_NAMESPACE}
-EOF
+  oc apply -n "${SERVING_NAMESPACE}" -f "${rootdir}/test/v1alpha1/resources/serving.knative.dev_v1alpha1_knativeserving_cr_1.3.0.yaml" || return $?
 
-  timeout 900 '[[ $(oc get knativeserving.${group}.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.conditions[?(@.type==\"Ready\")].status}") != True ]]'  || return 7
+  timeout 900 '[[ $(oc get knativeserving.serving.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.conditions[?(@.type==\"Ready\")].status}") != True ]]'  || return 7
```
jcrossley3: Right now, the behavior of the operator is to immediately convert and then delete an old CR when it detects one, so the old CR's status will never become ready.

mgencur:
Alright, so now the tests fail with `Error from server (NotFound): knativeservings.serving.knative.dev "knative-serving" not found` which is in line with what Jim said. But I hesitate to change the test until we agree if we keep the original CR.

markusthoemmes:
This got merged via #91

PR #81, opened 2020-01-28
Have the CSV own both CRD's to fix SRVKS-396

jcrossley3:


olm-catalog/serverless-operator/1.4.0/serverless-operator.v1.4.0.clusterserviceversion.yaml
```diff
@@ -76,7 +76,7 @@ spec:
   customresourcedefinitions:
     owned:
     - description: Represents an installation of a particular version of Knative Serving
-      displayName: Knative Serving
+      displayName: Knative Serving 0.11.1
```
markusthoemmes: Should this just stay "Knative Serving"?

olm-catalog/serverless-operator/1.4.0/serverless-operator.v1.4.0.clusterserviceversion.yaml
```diff
@@ -89,6 +89,20 @@ spec:
         x-descriptors:
         - 'urn:alm:descriptor:io.kubernetes.conditions'
       version: v1alpha1
+    - description: Represents an installation of a particular version of Knative Serving
+      displayName: Knative Serving 0.10.0 (obsolete)
```
markusthoemmes:
```suggestion
      displayName: Knative Serving (obsolete group, see documentation)
```

maybe?

mgencur:
Can you retrigger https://github.com/openshift-knative/serverless-operator/pull/80 when this is merged? It would be even better to include those tests here in this PR.

mgencur:
Please see my comment on https://github.com/openshift-knative/serverless-operator/pull/80  where I included this commit.

olm-catalog/serverless-operator/1.4.0/serverless-operator.v1.4.0.clusterserviceversion.yaml
jcrossley3: I don't know. I was trying to include what I personally would expect to see if we have to show two of them. If I show the version of one, I know that would trigger in myself to know what the version of the other one is.

olm-catalog/serverless-operator/1.4.0/serverless-operator.v1.4.0.clusterserviceversion.yaml
jcrossley3: I think we need to sufficiently convey in their names why there need to be two. Ideally, we'd have some means of telling the console *NOT* to display one of our owned CRD's, which is something @bbrowning has asked for in the past.

olm-catalog/serverless-operator/1.4.0/serverless-operator.v1.4.0.clusterserviceversion.yaml
markusthoemmes: So as discussed in Slack yesterday, do we want to make this "Knative Serving" and the other "Knative Serving (deprecated)"? I feel like that sufficiently conveys that we want people to move to the "proper" one while the other is going away in the future.

mgencur: Having the version there is confusing, IMO. I suppose people will think they can choose between https://github.com/knative/serving/releases/tag/v0.11.1 and https://github.com/knative/serving/releases/tag/v0.10.0 but they won't get Knative Serving of the given version, they will only get KnativeServing CRD of that version. And the version of Knative Serving won't be affected by the choice.

PR #82, opened 2020-01-28
Simplify logging utitilities and enable -x in CI.

markusthoemmes:
As the title says, this reduces the complexity of the logging utitilities to not produce too much noise when being run with `-x`.

### DIFF
--- a/hack/lib/common.bash
+++ b/hack/lib/common.bash
@@ -31,15 +31,9 @@ function timeout {
   while eval $*; do
     seconds=$(( seconds + interval ))
     logger.debug "Execution failed: ${*}. Waiting ${interval} sec ($seconds/${timeout})..."
-    if [[ "${LOG_LEVEL}" != 'DEBUG' ]]; then
-      echo -n '.'
-    fi
     sleep $interval
     [[ $seconds -gt $timeout ]] && logger.error "Time out of ${timeout} exceeded" && return 1
   done
-  if [[ "${LOG_LEVEL}" != 'DEBUG' ]] && [[ "$seconds" != '0' ]]; then
-    echo ''
-  fi
   return 0
 }
 
--- a/hack/lib/ui.bash
+++ b/hack/lib/ui.bash
@@ -1,105 +1,21 @@
 #!/usr/bin/env bash
 
-if [ -t 1 ]; then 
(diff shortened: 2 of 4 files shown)


mgencur:
So the colors are gone now. I think it wouldn't introduce much more noise if we kept them? Otherwise this looks good to me.

hack/lib/ui.bash
```diff
@@ -1,21 +1,73 @@
...
+  level="$1"
+  color="$2"
+  message="$3"
+  now="$(date '+%H:%M:%S.%3N')"
+  
+  printf "${color}%7s ${COLOR_CYAN}%s ${color}%s${COLOR_NC}\n" "${level}" "${now}" "${message}" 1>&2
 }
+
+if [[ "${SHOULD_COLOR}" == "false" ]]; then
+  function logger.debug {
```
mgencur: Nice idea. I have never gone that far with Bash but if it works...why not:)

mgencur:
I think it's good enough for our purposes but I'll let somebody else comment.

hack/lib/common.bash
```diff
@@ -31,15 +31,9 @@ function timeout {
   while eval $*; do
     seconds=$(( seconds + interval ))
     logger.debug "Execution failed: ${*}. Waiting ${interval} sec ($seconds/${timeout})..."
-    if [[ "${LOG_LEVEL}" != 'DEBUG' ]]; then
-      echo -n '.'
-    fi
```
jcrossley3: I actually kinda like the dots. :smile: Are you just removing them due to the -x noise? Can we set +x in timeout instead maybe?

markusthoemmes: I can bring the loglevels back. I just removed the entire loglevel setting + only showing logs of a certain level. You'd now also see debug logs locally. Want me to roll that back? The `x` noise wouldn't be affected because I override the entire functions for CI anyway.

jcrossley3: I may have misunderstood then. As long as I see *something*, whether dot or log message, I'm good. I just want to know something is doing something.

markusthoemmes: Yeah the log messages will now be printed on each run, giving you the failure reason.

PR #85, opened 2020-01-29
Install Kourier without removing Service Mesh

nak3:
This patch contains following two changes:

Add `GATEWAY_OVERRIDE=kourier` to lib.bash.
- This is necessary to pass gRPC e2e tests.

~~Remove service mesh from deploy.~~
- ~~This is necessary to pass service_to_service e2e tests. Precisely the
culprit of test failure is a NetworkPolicy in knative-serving-ingress.~~

Add NetworkPolicy for Kourier Gateway in knative-serving-ingress namespace.
- This fixes NetworkPolicy issue to pass service_to_service e2e tests.


nak3:
/test 4.2-e2e-aws-ocp-42
/test 4.3-e2e-aws-ocp-43

Looks like unrelated failure.

knative-operator/deploy/resources/kourier/kourier.yaml
```diff
@@ -0,0 +1,289 @@
+#apiVersion: v1
+#kind: Namespace
+#metadata:
+#  name: kourier-system
+#---
+apiVersion: networking.k8s.io/v1
```
markusthoemmes: Shall we move this to a different file maybe to signify that this is "from us"?

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
```diff
@@ -154,6 +156,31 @@ func (r *ReconcileKnativeServing) ensureCustomCertsConfigMap(instance *servingv1
...
+
+	manifest, err := mf.NewManifest(path, false, a.client)
+	if err != nil {
+		log.Error(err, "Unable to create Kourier Ingress install manifest")
+		return err
+	}
+	transforms := []mf.Transformer{mf.InjectOwner(instance)}
+	// let's hardcode this for now.
+	transforms = append(transforms, mf.InjectNamespace("knative-serving-ingress"))
+
```
markusthoemmes: Can we filter the kourier-system namespace out here to avoid having to manipulate a release manifst?

knative-operator/deploy/resources/kourier/kourier.yaml
```diff
@@ -0,0 +1,289 @@
+#apiVersion: v1
```
markusthoemmes: Generally speaking: Can we make this file to be a plain Kourier release manifest taken straight from their upstream? And add the version to the filename?

Just to make it easy for us to upgrade that manifest if needed and so we don't have to think about what we need to change and stuff. The namespace for instance will be injected by the code, so we can leave this to what upstream defines as well.

knative-operator/pkg/common/openshift.go
```diff
@@ -33,6 +34,11 @@ func Mutate(ks *servingv1alpha1.KnativeServing, c client.Client) error {
 	return nil
 }
 
+func ingressClass(ks *servingv1alpha1.KnativeServing, c client.Client) error {
+	Configure(ks, "network", "ingress.class", "kourier.ingress.networking.knative.dev")
```
savitaashture: Is that possible to use constants from knative.dev/serving ? instead of writing again

jcrossley3: I didn't realize those constants existed. But since they do, it might be a good idea to submit a separate PR that converts all the keys we pass to `Configure` to their upstream constants.

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
```diff
@@ -154,6 +156,31 @@ func (r *ReconcileKnativeServing) ensureCustomCertsConfigMap(instance *servingv1
...
+	const path = "deploy/resources/kourier/kourier.yaml"
+
+	manifest, err := mf.NewManifest(path, false, a.client)
+	if err != nil {
+		log.Error(err, "Unable to create Kourier Ingress install manifest")
+		return err
+	}
+	transforms := []mf.Transformer{mf.InjectOwner(instance)}
+	// let's hardcode this for now.
+	transforms = append(transforms, mf.InjectNamespace("knative-serving-ingress"))
```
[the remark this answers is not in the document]
jcrossley3: It's currently in the `servicemesh` package, but might be worth refactoring any functions common to any ingress out of there, possibly in a separate PR?

knative-operator/pkg/common/openshift.go
savitaashture: `ingress.class` is here https://github.com/knative/serving/blob/master/pkg/network/network.go#L67

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
```diff
@@ -154,6 +156,31 @@ func (r *ReconcileKnativeServing) ensureCustomCertsConfigMap(instance *servingv1
 	return nil
 }
 
+// Install Kourier Ingress Gateway
+func (a *ReconcileKnativeServing) installKourier(instance *servingv1alpha1.KnativeServing) error {
```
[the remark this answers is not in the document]
nak3: Yes, we should. Done.

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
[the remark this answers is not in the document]
nak3: Yes, it is in the service mesh package. It should be refactored when we remove `servicemesh` dependency.

knative-operator/deploy/resources/kourier/kourier.yaml
nak3: Sure. Done.

knative-operator/deploy/resources/kourier/kourier.yaml
nak3: Yes. Done.

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
nak3: Yes, we can. And I realized that `namespace` is transformed to create namespace by injectNamespace, so we just don't need to comment it out.

https://github.com/openshift-knative/serverless-operator/blob/55d9f27c68286fb9be22fc03f3c3aa842db87054/knative-operator/vendor/github.com/jcrossley3/manifestival/transform.go#L42-L43

knative-operator/deploy/resources/kourier/kourier.yaml
nak3: Done

knative-operator/pkg/common/openshift.go
nak3: Yes, Savita is correct. But it seems knative-operator still uses old code and we cannot find the value. https://github.com/openshift-knative/serverless-operator/tree/55d9f27c68286fb9be22fc03f3c3aa842db87054/knative-operator/vendor/knative.dev/pkg/network :cry:

savitaashture: @jcrossley3 @markusthoemmes Do we have plan to use new code for serverless-operator ?

nak3:
> I'm still unclear on exactly how we expose the decision of which ingress to use (kourier or maistra). I'd prefer not to install both of them, if possible.

Yes, I have started testing it in https://github.com/openshift-knative/serverless-operator/pull/87. It needs a lot of changes including unit tests, so I think separae PR would be better.

knative-operator/pkg/common/openshift.go
markusthoemmes: Let's relay that for a separate PR.

nak3:
NOTE:
The upgrade test failure needs the fix of https://github.com/knative/serving/issues/6568.
As I have updated the Route's port name to http (it was necessary to use "raw" Kourier's yaml) by https://github.com/openshift-knative/serverless-operator/pull/85/commits/27a58f4c1800f952effe64d27feaa76e9ae3b2fb, Route needs to be re-created. But currently KIngress is not reconciled and so Route is not updated.

markusthoemmes:
@nak3 If the route thing is an issue, we could also use a transformer to change the kourier port name? Or even change it in Kourier upstream? @jmprusi Is that port actually serving http only or does it also serve http2?

knative-operator/pkg/controller/knativeserving/kourier/kourier.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+	return nil
+}
+
+func apply(instance *servingv1alpha1.KnativeServing, api client.Client, path string) error {
+	manifest, err := mf.NewManifest(path, false, api)
+	if err != nil {
+		return err
+	}
+	transforms := []mf.Transformer{mf.InjectOwner(instance)}
```
markusthoemmes: Does this give us a cross-namespace ownerRef? Do we need to use a finalizer instead?

jmprusi:
> @nak3 If the route thing is an issue, we could also use a transformer to change the kourier port name? Or even change it in Kourier upstream? @jmprusi Is that port actually serving http only or does it also serve http2?

That port is Serving both. We can change it Upstream if you want

markusthoemmes:
Let's do that @jmprusi!

markusthoemmes:
@jcrossley3 There won't be a decision. We run with Kourier period. This enables that while keeping ServiceMesh around for upgrade cases. I believe a followup PR will take care of a migration of existing ServiceMesh users into Kourier and then deleting ServiceMesh.

knative-operator/pkg/controller/knativeserving/kourier/kourier.go
nak3: Ah, yes... This is a owner for knative-serving in `knative-serving` ns. So we need to use a finalizer instead.

nak3:
Thank you @jmprusi 
@markusthoemmes upstream Kourier changed it but we need to wait for the update till v0.3.9, so I updated the yaml by https://github.com/openshift-knative/serverless-operator/pull/85/commits/4971a9d218da4bd39986570cc57f59d2c28161d5 manually to confirm the upgrade test. (And it worked fine now.) If I should not change it manually I will revert it and wait for Kourier v0.3.9.

jmprusi:
@nak3 we don't have that many changes to create a release, but if you really need it, we can do it.

Due to maistra using an older envoy image, we should patch kourier deployment with:

```patch
diff --git a/deploy/kourier-knative.yaml b/deploy/kourier-knative.yaml
index 2628f96..331009f 100644
--- a/deploy/kourier-knative.yaml
+++ b/deploy/kourier-knative.yaml
@@ -57,10 +57,11 @@ spec:
         app: 3scale-kourier-gateway
     spec:
       containers:
-        - args:
+        - command: ["/usr/local/bin/envoy"]
+          args:
             - -c
             - /tmp/config/envoy-bootstrap.yaml
-          image: quay.io/3scale/kourier-gateway:v0.1.3
+          image: docker.io/maistra/proxyv2-ubi8:1.0.4
           imagePullPolicy: Always
           name: kourier-gateway
           ports:
@@ -247,9 +248,7 @@ data:
                             - "*"
                           routes:
                             - match:
-                                safe_regex:
-                                  google_re2: {}
-                                  regex: '/(certs|stats(/prometheus)?|server_info|clusters|listeners|ready)?'
+                                regex: '/(certs|stats(/prometheus)?|server_info|clusters|listeners|ready)?'
                                 headers:
                                   - name: ':method'
                                     exact_match: GET
```

nak3:
(rebased)

nak3:
Thanks @jmprusi
With regard to using image (`docker.io/maistra/proxyv2-ubi8:1.0.4`), do you have a plan to change it on [kourier upstream](https://github.com/3scale/kourier/blob/master/deploy/kourier-knative.yaml#L63)? 
I think that we want to use the same manifest with upstream's as much as possible.

markusthoemmes:
Can/should we carry this patch until this is hashed out upstream? Not sure if we'd want a random maistra image upstream.

knative-operator/pkg/common/openshift.go
```diff
@@ -18,6 +18,7 @@ var log = Log
 
 func Mutate(ks *servingv1alpha1.KnativeServing, c client.Client) error {
 	stages := []func(*servingv1alpha1.KnativeServing, client.Client) error{
+		ingressClass,
 		egress,
 		ingress,
 		configureIstio,
```
markusthoemmes: I think we need to ditch this now, no? And replace it with `ingressClass`? Feels like this would override Kourier and not use it at all 🤔

nak3: Yes, you are correct. We can remove the `configureIstio` as it is not used at all. 
Updated.

PR #86, opened 2020-01-29
Another attempt at supporting the obsolete CRD

jcrossley3:
Attempting to orphan resources without deleting the old CR.

TODO: Sync status conditions in both CR's without engaging in a controller war.


knative-operator/pkg/controller/knativeservingobsolete/knativeservingobsolete_controller.go
```diff
@@ -87,31 +92,106 @@ func (r *ReconcileKnativeServingObsolete) Reconcile(request reconcile.Request) (
...
 	}
+
+	// if !equality.Semantic.DeepEqual(current.Status, new.Status) {
+	// 	current.Status.Version = new.Status.Version
+	// 	current.Status.Conditions = deepCopyConditions(new.Status.Conditions)
+	// 	if err := r.client.Status().Update(context.TODO(), current); err != nil {
+	// 		return reconcile.Result{}, err
+	// 	}
+	// 	return reconcile.Result{}, nil
+	// }
```
markusthoemmes: This logic needs to stay intact I believe. It surfaces the status on the old thing to make it look like it used to before. Cloud Paks definitely depends on this.

PR #87, opened 2020-01-30
[WIP] [SRVKS-405] [SRVKS-407] Remove Service Mesh dependencies

nak3:
/hold

This is to see the result of e2e tests. Needs to merge https://github.com/openshift-knative/serverless-operator/pull/85 first.


mgencur:
I don't see anything related to upgrades from the previous release that uses Service Mesh to the latest release that uses Kourier. IMO, it's really important to do it in this PR or let's closely collaborate on adding/changing those tests in a separate PR that includes changes from this PR.

nak3:
Hi @mgencur The main PR is https://github.com/openshift-knative/serverless-operator/pull/85 and it will change the ingress.class config from istio to kourier by https://github.com/openshift-knative/serverless-operator/pull/85/files#diff-73539d886b2dd189cbb0105ac83f20b0R39.

Once the ingress.class is updated, all kingress is updated to use kourier. (we still need the backport of https://github.com/knative/serving/issues/6568 though)

For the test coverage, I wonder that `test/upgrade-tests.sh` covers but we need to add more test if necessary.

mgencur:
@nak3 Thanks, so when a user performs an upgrade to the current version it will automatically start using Kourier.... this is good. In this case the existing upgrade tests should cover that. They're currently designed to assert that there is no downtime (no failed requests) through the "prober" test. If there's any downtime it will need to be updated.

nak3:
Moved to https://github.com/openshift-knative/serverless-operator/pull/116 and closing this PR.

PR #88, opened 2020-01-30
Mirror the new and the old CRD to keep them both functional.

markusthoemmes:
This adds watching and updating between the old and the new CRD resource:
- Status of the **new** CRD will be reflected in the old CRD.
- Spec of the **old** CRD will be reflected in the new CRD.
- Migration code to orphan the old deployment to make it possible for the new CRD to take over the ownership of that.

We're specifically **not** setting an ownerRef from the old to the new CRD to allow the user to just remove the old one without any repercussions onto their system.


knative-operator/pkg/controller/knativeservingobsolete/knativeservingobsolete_controller.go
```diff
@@ -120,9 +203,14 @@ func (r *ReconcileKnativeServingObsolete) Reconcile(request reconcile.Request) (
 func (r *ReconcileKnativeServingObsolete) removeOldCertsConfig(ns string) error {
 	const name = "controller"
 	deployment := &appsv1.Deployment{}
-	if err := r.client.Get(context.TODO(), client.ObjectKey{Namespace: ns, Name: name}, deployment); err != nil {
+	err := r.client.Get(context.TODO(), client.ObjectKey{Namespace: ns, Name: name}, deployment)
+	if errors.IsNotFound(err) {
```
markusthoemmes: This is important for non-upgrade cases, where the users (because they have old scripts or whatnot) still apply the old CR without having a system around. We can safely ignore this condition anyway.

knative-operator/pkg/controller/knativeservingobsolete/deploy
```diff
@@ -0,0 +1 @@
+../../../deploy
```
markusthoemmes: This is not actually a file but a symlink 😂 . Github could do a better job at showing that. It's needed so that the tests can access the directory correctly to not break the orphaning logic.

knative-operator/pkg/controller/knativeservingobsolete/knativeservingobsolete_controller.go
```diff
@@ -87,31 +93,108 @@ func (r *ReconcileKnativeServingObsolete) Reconcile(request reconcile.Request) (
...
+		if err := r.client.Create(context.TODO(), new); err != nil {
+			return nil, err
+		}
+		return new, nil
+	} else if err != nil {
+		return nil, err
+	} else {
+		want := new.DeepCopy()
+		want.Spec.Config = old.Spec.Config
+		if err := common.Mutate(want, r.client); err != nil {
```
markusthoemmes: Left this here for simplicity. I know it's not cheap but I also didn't really bother to make it smart and make a cheaper check 🤔.

markusthoemmes:
The failure was due to the teardown routine not being able to connect to the cluster anyway. Unrelated to our code here.

/test 4.2-e2e-aws-ocp-42

markusthoemmes:
/test 4.2-e2e-aws-ocp-42

Flaky test from 0.11, should be fixed in newer versions anyway.

markusthoemmes:
/hold

Superseded by #91. Keeping this only if I screwed up the rebase.

PR #91, opened 2020-01-31
Support both the old and the new group of KnativeServings.

markusthoemmes:
This adds all the changes necessary to support knativeservings.serving.knative.dev and knativeservings.operator.knative.dev in one version. The old (serving) based API is supported by creating an entity of the new resource as a "child" of the old one. The `Spec` field of the old resource will be mirrored "down" into the new one, any status changes on the new one will be reflected in the old resource. That should make the old resource behave exactly like before, enabling this move to be backwards compatible.
Before creating the new resource, all of the old resources need to be orphaned to enable the new KnativeServing to gain ownership over all of these resources.
Added testcases ensure that different permutations of KnativeServings work as expected.

The migration path now looks like this:
1. Install the new version of the operators.
2.a. An installed system keeps working as before. You will have both an old and a new resource.
2.b. You can install the system freshly with the old resource and it still works as before. You will have both an old and a new resource.
2.c. You can install the system with the new CR. You will have only that new resource.
3. If you had both old and a new resource, you now have time to adjust your scripts etc. Once you've done that, you can remove the old resource without repercussions and continue to use just the new resource.

**Note:** If you want to adjust the Spec via the new CR, you **must** first delete the old CR.


mgencur:
Looks like most of the tests passed and the KnativeServing was installed successfully. So maybe some parameters from https://github.com/openshift-knative/serverless-operator/pull/91/commits/2186479293bc255d67f828b98e89bedaba4215ec#diff-12388c38e22670f0fcc4a55e1499916fR1 are different from defaults or could affect the failing tests?

markusthoemmes:
@mgencur pretty sure I found the error, see added commits.

mgencur:
Yeah, I think you're right. Now I remember that some time ago I saw similar errors when testing Kourier but my gateway was still pointing to Service Mesh so it didn't work.

knative-operator/pkg/common/openshift.go
```diff
@@ -20,6 +20,7 @@ func Mutate(ks *servingv1alpha1.KnativeServing, c client.Client) error {
 	stages := []func(*servingv1alpha1.KnativeServing, client.Client) error{
 		egress,
 		ingress,
+		configureIstio,
```
jcrossley3: Why is this needed for old CR sync'ing?

knative-operator/pkg/controller/knativeservingobsolete/knativeservingobsolete_controller.go
```diff
@@ -87,31 +94,113 @@ func (r *ReconcileKnativeServingObsolete) Reconcile(request reconcile.Request) (
...
+		if err := r.client.Create(context.TODO(), new); err != nil {
+			return nil, err
+		}
+		return new, nil
+	} else if err != nil {
+		return nil, err
+	} else {
+		want := new.DeepCopy()
+		want.Spec.Config = old.Spec.Config
+		if err := common.Mutate(want, r.client); err != nil {
```
jcrossley3: I have issues with this. I'm concerned that manual edits to the new config will be overwritten by manual edits to the old config. To me, that violates the principle of least surprise. I think the conservative approach is to only update new from old with fields not already configured.

knative-operator/pkg/common/openshift.go
markusthoemmes: The commit message describes it:
If we reconcile old -> new we override new's configMaps (because new's CR doesn't contain the istio config).

The istio config is currently added *after the fact* via a "manual" update to the ConfigMap itself and because `ApplyServiceMesh` is guarded by a state check it won't be retriggered after service mesh is ready once, causing us to loose that setting in the configMap.

I think all of the config map adjustions shall be done via the CR itself, not to the configmap directly, hence this change.

knative-operator/pkg/controller/knativeservingobsolete/knativeservingobsolete_controller.go
markusthoemmes: I hear you, though I think that is a somewhat dangerous assumption to begin with. My plan was to document that we're doing a "mirror migration" and if you want to start using the new CRD you **must** delete the old one first to avoid exactly those kinda situations.

Does that work for you as well?

jcrossley3: How about we add a validation webhook for the old which prevents any spec updates with an error referencing the new CR? That would support old CR creation and prevent updates with an error message encouraging migration.

markusthoemmes: I think that'd potentially also breaks support of the old CR, though I'm not sure if any of the systems we have actually update the old CR. I could imagine the Console to do that 🤔  but I don't know for sure.

jcrossley3: Well, if you must, how do you feel about replacing the "want/Mutate" logic with this instead?
https://gist.github.com/4b867fb15da60adb08e68fe24bf5df49

jcrossley3: BTW, that logic ^^ might obviate your need to add the istio config to Mutate, which is probably harmless, but I'd prefer to have conditioned on *something*, I think.

markusthoemmes: That's super elegant, taken!

I think I'd like us to keep the serviceMesh config in there. I agree it somewhat breaks an abstraction, but we don't really need an abstraction there anyway. IMO, config shall be done completely via the CR. In any case.

test/lib.bash
```diff
@@ -173,7 +173,8 @@ function run_knative_serving_rolling_upgrade_tests {
       [[ $(oc get knativeserving.serving.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.version}") == "$serving_version" ]] || return 1
     else
       approve_csv "$upgrade_to" || return 1
-      # The knativeserving CR should be updated now
+      # Both knativeserving CRs should be updated now
+      timeout 900 '[[ ! ( $(oc get knativeserving.serving.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.version}") != $serving_version && $(oc get knativeserving.serving.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.conditions[?(@.type==\"Ready\")].status}") == True ) ]]' || return 1
```
mgencur: I'm wondering if we should also add this after line 109 in serverless.bash when we're installing with the old CR - both CRs should be updated then.

PR #100, opened 2020-02-06
Exclude gRPC tests on vSphere

mgencur:
I'd like to first merge this and then setup periodic CI job.

Tests were failing with parallel=3 on vSphere. Using parallel=2 made the tests passed.

### DIFF
--- a/test/lib.bash
+++ b/test/lib.bash
@@ -107,7 +107,15 @@ function run_knative_serving_e2e_and_conformance_tests {
   local failed=0
   image_template="registry.svc.ci.openshift.org/openshift/knative-${knative_version}:knative-serving-test-{{.Name}}"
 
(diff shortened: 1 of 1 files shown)


mgencur:
/hold

I'm checking if `scale_up_workers ` works on vSphere. Otherwise will have to exclude the call.

mgencur:
Parallel==2 was working for me without errors.

mgencur:
/hold cancel

Workers cannot be scaled up on vSphere but we can skip this by setting SCALE_UP=-1 directly in ci-operator config. This PR doesn't need that.

mgencur:
I don't think we need to re-run upgrade tests on OCP 43. All the relevant script paths were exercised by the runs that passed.

PR #102, opened 2020-02-06
See if we can trigger the deleted knativeservings after upgrade and GC

bbrowning:


bbrowning:
Checking to see how/if these upgrade tests fail again.

/retest

markusthoemmes:
Obsolete.

PR #104, opened 2020-02-07
Remove resources from 1.3.0 carrying a wrong ownerRef.

markusthoemmes:
These old resources carried a cross-namespace ownerRef that would cause our correct ownerRefs to be killed by the GC.


knative-operator/pkg/controller/knativeservingobsolete/knativeservingobsolete_controller.go
```diff
@@ -170,8 +173,24 @@ func (r *ReconcileKnativeServingObsolete) reconcileNewResource(old *obsolete.Kna
...
+func (r *ReconcileKnativeServingObsolete) deleteObsoleteMonitoringResources() error {
+	const path = "deploy/resources/obsolete/monitoring-1.3.0.yaml"
+
+	manifest, err := mf.NewManifest(path, false, r.client)
+	if err != nil {
+		return err
+	}
+
+	if err := manifest.DeleteAll(); err != nil {
+		log.Info("Ignoring failed deletion of obsolete monitoring resources")
```
bbrowning: Are we sure we want to ignore delete failures? If we do and this bad ownerRef doesn't get deleted, we have a problem.

markusthoemmes: Fair point, I did this to ignore NotFound errors and thelike. Manifestival seems to do the right thing for that already though, so gonna change this. Thanks!

markusthoemmes:
/retest

Damn you!

bbrowning:
Looks good to me, assuming the tests pass and we manually verify the output of the upgrade tests to ensure nothing got deleted. We'll want to run this through a couple of retests just because the issue itself doesn't happen 100% of the time.

I don't want to hold up getting this PR in for this, but I'd love to see us make the tests actually fail if we have any invalid owner references created. Either via parsing the output we're already manually reviewing to see if the CRs got deleted or just moving this into Go code with our other tests.

markusthoemmes:
I agree on the retests. I think we should come up with a validation test that checks all of our resources for validity of their respective owner references after install and after update.

As this is somewhat time critical for those that already upgrade to 1.4.0, I'd like to get this in potentially without such a check.

bbrowning:
Also, I'd like a review from @jcrossley3 before actually merging.

jcrossley3:
I spot-checked the old operator code and couldn't find any other manifests we apply that aren't also applied by the new operator, so...
/lgtm
/approve

knative-operator/pkg/controller/knativeservingobsolete/knativeservingobsolete_controller.go
jcrossley3: confirmed, MF will eat the NotFounds but return any errors that might prevent deletion

PR #107, opened 2020-02-10
Check in htpasswd file and alias kubectl to drop yum based installs completely.

markusthoemmes:
It'd be neat to not depend on yum installs in our CI builds. This has bitten us more than once already.


mgencur:
Just FYI: The changes to openshift/ci-operator/build-image/Dockerfile won't take effect until this file is merged. So these changes are not reflected in CI results on this PR.

markusthoemmes:
Indeed, thanks for the reminder. Hmm. Do we want to risk it? 😂

mgencur:
I hope the alias could work but not 100% sure. Would it be possible to test this locally? Worst case we can merge this and immediately send a dummy PR that will just make the tests run and we'll see.

markusthoemmes:
Not following through for now.

PR #108, opened 2020-02-10
Bootstrap 1.5.0, upgrade to Serving 0.12.

markusthoemmes:
As the title says. This upgrade our latest CSV to 1.5.0 and uses 0.12 images for upstream operator and upstream itself.

### DIFF
--- a/hack/lib/vars.bash
+++ b/hack/lib/vars.bash
@@ -5,7 +5,7 @@ readonly BUILD_NUMBER=${BUILD_NUMBER:-$(uuidgen)}
 # shellcheck disable=SC1091,SC1090
 source "$(dirname "${BASH_SOURCE[0]}")/../../test/vendor/github.com/knative/test-infra/scripts/e2e-tests.sh"
 
-readonly KNATIVE_VERSION="${KNATIVE_VERSION:-v0.11.1}"
+readonly KNATIVE_VERSION="${KNATIVE_VERSION:-v0.12.1}"
 
 readonly CATALOG_SOURCE_FILENAME="${CATALOG_SOURCE_FILENAME:-catalogsource-ci.yaml}"
 readonly DOCKER_REPO_OVERRIDE="${DOCKER_REPO_OVERRIDE:-}"
--- a/olm-catalog/serverless-operator/1.4.1/serverless-operator.v1.4.1.clusterserviceversion.yaml
+++ b/olm-catalog/serverless-operator/1.4.1/serverless-operator.v1.4.1.clusterserviceversion.yaml
@@ -289,7 +289,7 @@ spec:
               serviceAccountName: knative-serving-operator
               containers:
                 - name: knative-serving-openshift
-                  image: $IMAGE_KNATIVE_OPERATOR
+                  image: registry.svc.ci.openshift.org/openshift/openshift-serverless-v1.4.0:knative-operator
                   command:
                   - knative-serving-openshift
                   imagePullPolicy: Always
@@ -334,7 +334,7 @@ spec:
               serviceAccountName: knative-openshift-ingress
               containers:
                 - name: knative-openshift-ingress
-                  image: $IMAGE_KNATIVE_OPENSHIFT_INGRESS
+                  image: registry.svc.ci.openshift.org/openshift/openshift-serverless-v1.4.0:knative-openshift-ingress
                   command:
                   - knative-openshift-ingress
                   imagePullPolicy: Always
@@ -442,9 +442,9 @@ spec:
     - name: IMAGE_webhook
       image: registry.svc.ci.openshift.org/openshift/knative-v0.11.1:knative-serving-webhook
     - name: knative-operator
-      image: $IMAGE_KNATIVE_OPERATOR
+      image: registry.svc.ci.openshift.org/openshift/openshift-serverless-v1.4.0:knative-operator
     - name: knative-openshift-ingress
-      image: $IMAGE_KNATIVE_OPENSHIFT_INGRESS
+      image: registry.svc.ci.openshift.org/openshift/openshift-serverless-v1.4.0:knative-openshift-ingress
   replaces: serverless-operator.v1.3.0
   skips:
     - serverless-operator.v1.4.0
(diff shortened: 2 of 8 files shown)


olm-catalog/serverless-operator/1.4.1/serverless-operator.v1.4.1.clusterserviceversion.yaml
```diff
@@ -289,7 +289,7 @@ spec:
               serviceAccountName: knative-serving-operator
               containers:
                 - name: knative-serving-openshift
-                  image: $IMAGE_KNATIVE_OPERATOR
+                  image: registry.svc.ci.openshift.org/openshift/openshift-serverless-v1.4.0:knative-operator
```
markusthoemmes: Yes, this looks wrong but we currently override the CI built images everytime we upgrade the branch. As 1.4.0 should be skipped anyway, I don't think it's harmful for our CI purposes. These are actually 1.4.1 images (they always reflect the latest of the respective branch).

Gonna align these names with the actual branch name in the next release to cause less confusion in the future.

test/lib.bash
```diff
@@ -115,7 +115,7 @@ function run_knative_serving_e2e_and_conformance_tests {
     parallel=2
   fi
 
-  go_test_e2e -tags=e2e -timeout=30m -parallel=$parallel ./test/e2e ./test/conformance/... \
+  go_test_e2e -tags=e2e -timeout=30m -parallel=$parallel ./test/e2e ./test/conformance/api/... ./test/conformance/runtime/... \
```
markusthoemmes: @nak3 we should likely revisit this after we ship Kourier. The Ingress conformance might pass happily there.

hack/lib/vars.bash
```diff
@@ -5,7 +5,7 @@ readonly BUILD_NUMBER=${BUILD_NUMBER:-$(uuidgen)}
 # shellcheck disable=SC1091,SC1090
 source "$(dirname "${BASH_SOURCE[0]}")/../../test/vendor/github.com/knative/test-infra/scripts/e2e-tests.sh"
 
-readonly KNATIVE_VERSION="${KNATIVE_VERSION:-v0.11.1}"
+readonly KNATIVE_VERSION="${KNATIVE_VERSION:-v0.12.1}"
```
jcrossley3: you remembered! :smile:

olm-catalog/serverless-operator/1.4.1/serverless-operator.v1.4.1.clusterserviceversion.yaml
jcrossley3: confirmed: this looks wrong! (i stopped reading after that :smile: )

olm-catalog/serverless-operator/1.5.0/serving_v1alpha1_knativeserving_crd.yaml
```diff
@@ -0,0 +1,99 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: knativeservings.serving.knative.dev
```
jcrossley3: i thought the plan was to remove this starting with 1.5.0

olm-catalog/serverless-operator/1.5.0/serverless-operator.v1.5.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,449 @@
...
+        displayName: Version
+        path: version
+      - description: Conditions of Knative Serving installed
+        displayName: Conditions
+        path: conditions
+        x-descriptors:
+        - 'urn:alm:descriptor:io.kubernetes.conditions'
+      version: v1alpha1
+    - description: Represents an installation of a particular version of Knative Serving
+      displayName: Knative Serving (obsolete)
```
jcrossley3: seriously? how long will this live and why longer than 1.4.0. that's the only one that needs to have both, right? so that clients can migrate within a single version?

test/lib.bash
nak3: Sure, thank you for letting me know.

nak3:
LGTM (hold lgtm label for Jim's comments.)
Also I verified that this change worked fine on OCP 4.3.

olm-catalog/serverless-operator/1.5.0/serving_v1alpha1_knativeserving_crd.yaml
markusthoemmes: There is no accepted plan AFAIK and I didn't concern myself with this decision for this PR. If we're removing it, I'd rather consciously do so in a separate PR that states that we're removing it. I'm happy if we are but I don't want us to block testing on 0.12 by that decision as @nak3 needs it for the Kourier migration work. It's orthogonal.

olm-catalog/serverless-operator/1.5.0/serverless-operator.v1.5.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,449 @@
...
+      displayName: Istio Service Mesh Member Roll
+      kind: ServiceMeshMemberRoll
+      name: servicemeshmemberrolls.maistra.io
+      version: v1
+    - description: An Istio control plane installation
+      displayName: Istio Service Mesh Control Plane
+      kind: ServiceMeshControlPlane
+      name: servicemeshcontrolplanes.maistra.io
+      version: v1
+  minKubeVersion: 1.14.0
```
mgencur: I suppose this should be changed? Right now it's same as in the previous release but we'll only support OCP 4.3 ?

mgencur: I think this should be 1.15.0 as we're moving to Knative Serving 0.12

markusthoemmes: Yes, you are absolutely right! Thanks for keeping me honest!

PR #110, opened 2020-02-10
Fix replacement chain for skipping 1.4.0.

markusthoemmes:
The OLM team told us that we actually need to both replace and skip 1.4.0.


jcrossley3:
There seems to be some disagreement among the OLM team, so for now...
/hold

mgencur:
This doesn't work for me when upgrading from 1.3.0 to 1.4.1. I get the same error as before: Conflicting CRD owner in namespace.

markusthoemmes:
Yeah this is garbage.

PR #111, opened 2020-02-11
Remove skips directive as it has been found to not be reliable.

markusthoemmes:
As discussed in Slack.


markusthoemmes:
/retest

Infra issues.

PR #117, opened 2020-02-13
Fix rolling upgrade wait condition

mgencur:
* take into account both cases when we installed previous version via
old CR and also via new CR

### DIFF
--- a/test/lib.bash
+++ b/test/lib.bash
@@ -173,8 +173,10 @@ function run_knative_serving_rolling_upgrade_tests {
       [[ $(oc get knativeserving.serving.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.version}") == "$serving_version" ]] || return 1
     else
       approve_csv "$upgrade_to" || return 1
-      # Both knativeserving CRs should be updated now
-      timeout 900 '[[ ! ( $(oc get knativeserving.serving.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.conditions[?(@.type==\"Ready\")].status}") == True ) ]]' || return 1
+      # If we used old CR for installing Serverless both knativeserving CRs should be updated now
+      if oc get knativeserving.serving.knative.dev knative-serving -n "$SERVING_NAMESPACE" >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


test/lib.bash
```diff
@@ -173,9 +173,13 @@ function run_knative_serving_rolling_upgrade_tests {
...
     else
       approve_csv "$upgrade_to" || return 1
-      # Both knativeserving CRs should be updated now
-      timeout 900 '[[ ! ( $(oc get knativeserving.serving.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.conditions[?(@.type==\"Ready\")].status}") == True ) ]]' || return 1
-      timeout 900 '[[ ! ( $(oc get knativeserving.operator.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.conditions[?(@.type==\"Ready\")].status}") == True ) ]]' || return 1
+      # If we used old CR for installing Serverless both knativeserving CRs should be updated now
+      if oc get knativeserving.serving.knative.dev knative-serving -n "$SERVING_NAMESPACE" >/dev/null 2>&1; then
+        timeout 900 '[[ ! ( $(oc get knativeserving.serving.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.conditions[?(@.type==\"Ready\")].status}") == True ) ]]' || return 1
+        timeout 900 '[[ ! ( $(oc get knativeserving.operator.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.conditions[?(@.type==\"Ready\")].status}") == True ) ]]' || return 1
+      else
```
markusthoemmes: I think you can drop the else block and always run the check for the new group. A bit less duplication.

mgencur: What if we installed Serverless via the new CR? Then it wouldn't wait at all?

mgencur: Oh. Sorry. Got it.

mgencur: Pushed.

markusthoemmes:
Should this go against master as well?

mgencur:
Yes. I'll send one against master as well.

PR #120, opened 2020-02-14
Include workaround for LoadBalancer cleanup on Azure

mgencur:
* workaround for https://bugzilla.redhat.com/show_bug.cgi?id=1798282

### DIFF
--- a/test/lib.bash
+++ b/test/lib.bash
@@ -50,10 +50,19 @@ function run_e2e_tests {
     logger.error 'Tests have failures!'
   fi
 
-  # On some platforms (e.g. Azure) it takes longer to clean namespaces from previous tests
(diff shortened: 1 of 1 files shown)


nak3:
So, bz#1798282 is talking about `ingresscontroller`, but `ingresscontroller` is not the main point. 
Strictly speaking, svc with type.LoadBalancer gets stuck when we delete namespace on Azure?

If yes, does Kourier's svc need this workaround as well?

mgencur:
That's right, all services of type LoadBalancer are affected. I'll have to fix this for Kourier. Mmt.

mgencur:
Updated with the Kourier check.

mgencur:
Thanks:)

PR #122, opened 2020-02-16
Add changes to support egress proxies in Knative.

savitaashture:
**JIRA:** https://issues.redhat.com/browse/SRVKS-359

**Description:**
The component which talks to internet in Knative is controller so need to add support for corporate egress proxies in Knative in order to avoid direct contact to external proxy.

There is a [cluster-wide](https://docs.openshift.com/container-platform/4.2/networking/enable-cluster-wide-proxy.html) proxy in OpenShift which helps to support this scenario

Currently added changes for **HTTP_PROXY** and **NO_PROXY**


knative-operator/pkg/common/openshift.go
```diff
@@ -71,6 +75,72 @@ func ingress(ks *servingv1alpha1.KnativeServing, c client.Client) error {
...
+	var proxyEnv = map[string]string{
+		"HTTP_PROXY":  "HTTP_PROXY",
+		"HTTPS_PROXY": "HTTPS_PROXY",
+		"NO_PROXY":    "NO_PROXY",
+	}
+	operatorNS, err := k8sutil.GetOperatorNamespace()
+	if err != nil {
+		return err
+	}
+	podData, err := k8sutil.GetPod(context.TODO(), c, operatorNS)
```
markusthoemmes: I don't quite understand why we need to fetch the pod here? 🤔 Can we not just do `os.Getenv("HTTP_PROXY")` etc. to get the same info?

savitaashture: Yes we can use i realized it after your comment
Thank you

test/e2e/knative_serving_test.go
```diff
@@ -301,3 +303,45 @@ func waitForRouteServingText(t *testing.T, caCtx *test.Context, routeDomain, exp
...
+	if err := test.UpdateGlobalProxy(caCtx, "http://1.2.4.5:8999"); err != nil {
+		t.Fatal("Failed to update proxy", err)
+	}
+
+	t.Run("wait for it to be ready after update", func(t *testing.T) {
+		test.GetPod(caCtx, knativeServing)
+	})
+
+	t.Run("deploy knative service after proxy update", func(t *testing.T) {
+		if _, err := test.WithServiceReady(caCtx, proxyHelloworldService, testNamespace, proxyImage); err == nil {
```
nak3: Is this `err == nil` expected? If yes, I feel that `Knative Service not ready` message is not correct.

test/e2e/knative_serving_test.go
```diff
@@ -301,3 +303,45 @@ func waitForRouteServingText(t *testing.T, caCtx *test.Context, routeDomain, exp
...
+	if err != nil {
+		t.Fatal("Failed to update proxy", err)
+	}
+
+	// Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1751903#c11
+	// Currently when we update cluster proxy by removing httpProxy, noProxy etc... OLM will not update controller
+	// once issue is fixed we can add
+	/*
+		t.Run("deploy knative service after proxy update", func(t *testing.T) {
+			if _, err := test.WithServiceReady(caCtx, proxyHelloworldService, testNamespace, proxyImage); err == nil {
```
nak3: It would be better to use `t.Skip()` instead of comment out - https://golang.org/pkg/testing/#hdr-Skipping

test/service.go
```diff
@@ -65,9 +65,41 @@ func CreateService(ctx *Context, name, namespace, image string) (*servingv1beta1
 	return service, nil
 }
 
+func GetPod(ctx *Context, ns string) {
```
nak3: We should change the name of function `GetPod()` . The name sounds like similar function `GetConfiguration()` but it does not actually.

test/e2e/knative_serving_test.go
```diff
@@ -301,3 +303,45 @@ func waitForRouteServingText(t *testing.T, caCtx *test.Context, routeDomain, exp
...
+		test.GetPod(caCtx, knativeServing)
+	})
+
+	t.Run("deploy knative service after proxy update", func(t *testing.T) {
+		if _, err := test.WithServiceReady(caCtx, proxyHelloworldService, testNamespace, proxyImage); err == nil {
+			t.Fatal("Knative Service not ready", err)
+		}
+	})
+
+	cfg, err := test.GetConfiguration(caCtx, proxyHelloworldService, testNamespace)
```
nak3: The error check is necessary here. Otherwise, `cfg.Status.Conditions` will get null pointer exception.

test/e2e/knative_serving_test.go
```diff
@@ -301,3 +303,45 @@ func waitForRouteServingText(t *testing.T, caCtx *test.Context, routeDomain, exp
...
+	for _, cond := range cfg.Status.Conditions {
+		// After global proxy update every call goes through proxy server
+		// Here it give unable to pull image because it tries to connect to not running http server
+		if !strings.Contains(cond.Message, "dial tcp 1.2.4.5:8999: i/o timeout") {
```
mgencur: I'm curious if this message is different if the traffic doesn't go through the proxy. In other words, are we verifying here that we use the proxy?

test/e2e/knative_serving_test.go
mgencur: It's hard to understand the bug in Bugzilla. Do we see the bug only if we first set a real value for HTTP_PROXY and later update it with an `empty` value? Or does it also happen if we first use `empty` value and later update with a real one (the non-existent server). If it doesn't happen in the second case maybe you could first test the empty value and later the non-existent server - and don't skip any tests.

knative-operator/pkg/common/openshift.go
```diff
@@ -71,6 +74,61 @@ func ingress(ks *servingv1alpha1.KnativeServing, c client.Client) error {
 	return nil
 }
 
+// Update updates Knative controller env to use cluster wide proxy information
+func Update(ks *servingv1alpha1.KnativeServing, c client.Client) error {
```
markusthoemmes: I'm not sure this function should live here as this file for now only contains mutations for the KnativeServing resource itself. Should we create a new file that encapsulates mutations to the deployment as it runs? Do we even want to do that or do we want to enhance the KS API to allow us to set these proxy settings through them?

Aaaaand the function name is somewhat non-descriptive. `ApplyProxySettings` or something along the lines would be more self-explanatory.

knative-operator/pkg/common/openshift.go
```diff
@@ -71,6 +74,61 @@ func ingress(ks *servingv1alpha1.KnativeServing, c client.Client) error {
...
+			return append(env[:i], env[i+1:]...)
+		}
+	}
+	return env
+}
+
+func appendUnique(orgEnv []v1.EnvVar, key, value string) []v1.EnvVar {
+	for i := range orgEnv {
+		if orgEnv[i].Name == key {
+			if value == "" {
```
markusthoemmes: I don't quite parse this piece of code 🤔. When would this be empty and why'd remove things only to set their value afterwards?

I think this can be simplified to:

```golang
// Set the value if the key is already present.
for i := range orgEnv {
  if orgEnv[i].Name == key {
    orgEnv[i].Value = value
    return orgEnv
  }
}
// If not, append a key-value pair.
return append(orgEnv, v1.EnvVar{
  Name:  key,
	Value: value,
})
```

test/Gopkg.toml
```diff
@@ -51,8 +51,8 @@ required = [
 
 [[constraint]]
   name = "github.com/openshift/api"
-  version = "3.9.0"
+  branch = "release-4.1"
```
markusthoemmes: Why not `release-4.3` to correspond more closely to what we actually use?

test/clients.go
```diff
@@ -173,6 +180,14 @@ func newOpenShiftRoutes(cfg *rest.Config) (routev1.RouteV1Interface, error) {
 	return routeClient, nil
 }
 
+func newOpenShiftProxy(cfg *rest.Config) (configV1.ConfigV1Interface, error) {
```
markusthoemmes:
```suggestion
func newOpenShiftProxyClient(cfg *rest.Config) (configV1.ConfigV1Interface, error) {
```

nit.

test/service.go
markusthoemmes: Yeah let's make this `WaitForControllerEnvironment` and pass in the key/value you want to see. You're going to need the inverse once we can enable the disabling test.

test/service.go
```diff
@@ -65,9 +65,41 @@ func CreateService(ctx *Context, name, namespace, image string) (*servingv1beta1
...
+	// run infinite for loop because once after the Global proxy update operator pod will be recreated
+	// and which intern updates knative serving controller pod so we don't no exact time duration for this to happen
+	for {
```
markusthoemmes: I'd recommend using apimachinery's `wait` package here. `wait.PollImmediate` should be a good fit.

test/service.go
```diff
@@ -65,9 +65,41 @@ func CreateService(ctx *Context, name, namespace, image string) (*servingv1beta1
...
+		pods, _ := ctx.Clients.Kube.CoreV1().Pods(ns).List(metav1.ListOptions{
+			LabelSelector: "app=controller",
+		})
+
+		for i := range pods.Items {
+			for _, container := range pods.Items[i].Spec.Containers {
+				for _, e := range container.Env {
+					if e.Name == "HTTP_PROXY" && e.Value != "" {
+						if pods.Items[i].Status.Phase == corev1.PodRunning {
```
markusthoemmes: This doesn't suffice to make sure the pod is actually running I think. I had lots of fun doing this for a separate upstream bug, this is what I came up with:

```golang

func isPodReady(pod *corev1.Pod) bool {
	if pod.DeletionTimestamp != nil || pod.Status.PodIP == "" {
		return false
	}
	for _, cond := range pod.Status.Conditions {
		if cond.Type == corev1.PodReady && cond.Status == corev1.ConditionTrue {
			return true
		}
	}
	return false
}
```

markusthoemmes: I'd also check if the pod is ready first, before even looping through the containers and looking for the proxy values.

test/service.go
```diff
@@ -233,3 +266,15 @@ func RouteHasHost(r *routev1.Route, err error) (bool, error) {
 }
 
 func int32Ptr(i int32) *int32 { return &i }
+
+func UpdateGlobalProxy(ctx *Context, value string) error {
```
markusthoemmes: So this file is called `service.go`. Can we create a new file `proxy.go` that then gets the new helpers?

test/e2e/knative_serving_test.go
markusthoemmes: Why do we need to fetch the configuration? Is the status not bubbling up into the service?

test/e2e/knative_serving_test.go
```diff
@@ -301,3 +303,45 @@ func waitForRouteServingText(t *testing.T, caCtx *test.Context, routeDomain, exp
 		t.Fatalf("The Route at domain %s didn't serve the expected text \"%s\": %v", routeDomain, expectedText, err)
 	}
 }
+
+func testKnativeServingForGlobalProxy(t *testing.T, caCtx *test.Context) {
+	if err := test.UpdateGlobalProxy(caCtx, "http://1.2.4.5:8999"); err != nil {
+		t.Fatal("Failed to update proxy", err)
+	}
+
+	t.Run("wait for it to be ready after update", func(t *testing.T) {
```
markusthoemmes: I'm not sure all these subtests is how testing in golang is supposed to work. Can we remove all the subtests from this function?

test/e2e/knative_serving_test.go
savitaashture: Yes this is to verify call goes via proxy or not 

and if proxy server not able to connect this is the error we get 
```
'Revision "helloworld-go-2fbt6" failed with message: Unable to fetch
      image "gcr.io/knative-samples/helloworld-go": failed to resolve image to digest:
      failed to fetch image information: Get https://gcr.io/v2/: proxyconnect tcp:
      dial tcp 1.2.4.5:8999: i/o timeout.'
```

test/service.go
savitaashture: I can use `wait.PollImmediate` but i am not sure how much time it will take for operator pod to recreate, update env and then operator pod will update those env to Knative **controller** POD and once Knative controller POD is in Running and have Env need to return from that function.

So i  am not able to estimate exact time for this to happen
that's why added infinite for loop

test/service.go
savitaashture: even if we check pod is ready first we need to validate whether that pod have env or not
So will that make any impact if we add check for podReady at the first step?

test/e2e/knative_serving_test.go
savitaashture: Yes issue produce only when we first set a real Asvalue for HTTP_PROXY and later update it with an empty value 
and that empty value wont be updated by OLM to operator controller.

As you mentioned to test first by setting empty value and then with real HTTP_PROXY even after this test at the end we need to set it empty and make sure knative service deployed successfully but right now as there is issue with OLM so need to skip that particular test

test/service.go
markusthoemmes: It won't, it just reads strange to me to check the env first and then check for readiness.

test/service.go
markusthoemmes: Well eventually the tests will just time out then? I think we should set a reasonable upper boundary. Maybe start with 5 minutes? We could in addition log how long it took to dial that value down in the future.

test/service.go
savitaashture: Why i am checking env because some time pod will not restart and pod will be always running in that case assuming Pod Ready will miss lead i feel

test/e2e/knative_serving_test.go
savitaashture: do you mean we need to have separate function for each t.Run ?

markusthoemmes: not necessarily, why not just have all of the code flat?

test/Gopkg.toml
savitaashture: If i update to `release-4.3`
and do `dep ensure` the build fails with error
```
test/vendor/github.com/openshift/api/route/v1/generated.pb.go:448:28: m.ObjectMeta.MarshalToSizedBuffer undefined (type "github.com/openshift-knative/serverless-operator/test/vendor/k8s.io/apimachinery/pkg/apis/meta/v1".ObjectMeta has no field or method MarshalToSizedBuffer)
test/vendor/github.com/openshift/api/route/v1/generated.pb.go:539:37: m.LastTransitionTime.MarshalToSizedBuffer undefined (type *"github.com/openshift-knative/serverless-operator/test/vendor/k8s.io/apimachinery/pkg/apis/meta/v1".Time has no field or method MarshalToSizedBuffer)

Comment (markusthoemmes) [reply test/Gopkg.toml]: meh, I don't think we care enough.

File: test/service.go (hunk 11 above)
Comment (savitaashture) [reply test/service.go]: I used max 10 min as timeout

File: test/e2e/knative_serving_test.go [hunk 15]
```diff
@@ -301,3 +304,58 @@ func waitForRouteServingText(t *testing.T, caCtx *test.Context, routeDomain, exp
...
+	t.Log("update global proxy with empty value")
+	if err = test.UpdateGlobalProxy(caCtx, ""); err != nil {
+		t.Fatal("Failed to update proxy", err)
+	}
+
+	// Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1751903#c11
+	// Currently when we update cluster proxy by removing httpProxy, noProxy etc... OLM will not update controller
+	// once bugzilla issue https://bugzilla.redhat.com/show_bug.cgi?id=1751903#c11 fixes need to run below test case
+	// in order to verify proxy update success and knative service deployed successfully
+	t.Skip("deploy knative service after proxy update and service should deploy successfully")
```
markusthoemmes: I think we should drop the Skip and just leave the comment.

test/e2e/knative_serving_test.go
```diff
@@ -301,3 +304,58 @@ func waitForRouteServingText(t *testing.T, caCtx *test.Context, routeDomain, exp
...
+			proxyExist = true
+			break
+		}
+	}
+	if !proxyExist {
+		t.Fatal("Service not ready", err)
+	}
+
+	t.Log("update global proxy with empty value")
+	if err = test.UpdateGlobalProxy(caCtx, ""); err != nil {
```
markusthoemmes: We need to do this on failure too, right? I'd `defer` + `OnInterrupt` this before setting the proxy to make sure we clean up after us correctly in this test.

knative-operator/pkg/common/openshift_test.go
```diff
@@ -116,3 +116,9 @@ func verifyCerts(t *testing.T, ks *servingv1alpha1.KnativeServing) {
 		t.Error("Missing custom certs config")
 	}
 }
+
+func verifyTimestamp(t *testing.T, ks *servingv1alpha1.KnativeServing) {
```
markusthoemmes: Merge error? This shouldn't be needed anymore.

savitaashture: Yep removed

test/e2e/knative_serving_test.go
```diff
@@ -303,6 +306,12 @@ func waitForRouteServingText(t *testing.T, caCtx *test.Context, routeDomain, exp
 }
 
 func testKnativeServingForGlobalProxy(t *testing.T, caCtx *test.Context) {
+	defer test.CleanupOnInterrupt(t, func() {
```
markusthoemmes: This shouldn't be deferred. It should rather look like this:

```golang
func resetProxy() {
  if err := test.UpdateGlobalProxy(caCtx, ""); err != nil {
   panic("Failed to update proxy", err)
  }
}

func testKnativeServingForGlobalProxy(t *testing.T, caCtx *test.Context) {
  test.CleanupOnInterrupt(t, resetProxy)
  defer resetProxy

...
}
```

test/e2e/knative_serving_test.go
```diff
@@ -73,9 +73,12 @@ func TestKnativeServing(t *testing.T) {
 		testKnativeVersusKubeServicesInOneNamespace(t, caCtx)
 	})
 
-	t.Run("update global proxy and verify calls goes through proxy server", func(t *testing.T) {
-		testKnativeServingForGlobalProxy(t, caCtx)
-	})
+	// t.Run for proxy testing
+	t.Skip("update global proxy and verify calls goes through proxy server")/*
+		func(t *testing.T) {
```
markusthoemmes: You shouldn't need to comment this out. The Skip should do the right thin, I think? Same goes for the `t.Run` call.

savitaashture: `func (c *common) Skip(args ...interface{}) {}` just take args not func

markusthoemmes: Right, but you can skip an intact test just fine. In this case

```golang
t.Run("update global proxy and verify calls goes through proxy server", func(t *testing.T) {
  t.Skip("Skipped because the test takes too long to run for now")
  testKnativeServingForGlobalProxy(t, caCtx)
}
```

Should work just fine.

test/e2e/knative_serving_test.go
```diff
@@ -73,16 +75,19 @@ func TestKnativeServing(t *testing.T) {
...
+	})
 
 	t.Run("remove knativeserving cr", func(t *testing.T) {
-		if err := v1a1test.DeleteKnativeServing(caCtx, knativeServing, knativeServing); err != nil {
-			t.Fatal("Failed to remove Knative Serving", err)
+		err := v1a1test.DeleteKnativeServing(caCtx, knativeServing, knativeServing)
+		if err != nil {
+			// Sometimes because of proxy unset takes time to login so need to wait until pods are up again
+			if strings.Contains(err.Error(), "Unauthorized") {
+				test.WaitForControllerEnvironment(caCtx, knativeServing, proxyVerificationEnvName)
```
markusthoemmes: Let's put this in the test above. The test updating the proxy should make sure that once it finishes, the state is as before, i.e.

1. Proxy has been unset
2. Everything has updated accordingly

We shouldn't bleed that over into the next test as that's going to cause maintainability issues down the line.

test/e2e/knative_serving_test.go
```diff
@@ -291,14 +296,79 @@ func testUserPermissions(t *testing.T, paCtx *test.Context, editCtx *test.Contex
...
+	t.Log("deploy successfully knative service after proxy update")
+	if _, err := test.WithServiceReady(caCtx, proxyHelloworldServiceSuccess, testNamespace, image); err != nil {
+		t.Fatal("Knative Service not ready", err)
+	}
+
+	t.Log("update global proxy with proxy server")
+	resetProxy(t, caCtx, "http://1.2.4.5:8999", httpProxy)
+
+	t.Log("deploy knative service after proxy update")
+	test.WithServiceReady(caCtx, proxyHelloworldService, testNamespace, proxyImage)
```
markusthoemmes: Isn't this supposed to not become ready?

test/e2e/knative_serving_test.go
```diff
@@ -291,14 +296,79 @@ func testUserPermissions(t *testing.T, paCtx *test.Context, editCtx *test.Contex
...
+		if strings.Contains(cond.Message, "dial tcp 1.2.4.5:8999: i/o timeout") {
+			// use bool variable here because service status have more than one conditions and if none of the condition matches
+			// then assume knative service failed because of some other reason
+			proxyExist = true
+			break
+		}
+	}
+
+	if !proxyExist {
+		t.Fatal("Service not ready", err)
```
markusthoemmes: This `err` will never be something else than `nil` cause the last time you set err is in line 336 where you also check it immediately for nil-ness. You'd rather want to log `svc.Status` here, I think.

test/service.go
```diff
@@ -65,9 +66,52 @@ func CreateService(ctx *Context, name, namespace, image string) (*servingv1beta1
...
 
+func WaitForControllerEnvironment(ctx *Context, ns, envName string) error {
+	return wait.PollImmediate(Interval, 10*time.Minute, func() (bool, error) {
+		pods, _ := ctx.Clients.Kube.CoreV1().Pods(ns).List(metav1.ListOptions{
+			LabelSelector: "app=controller",
+		})
+		for i := range pods.Items {
+			for _, container := range pods.Items[i].Spec.Containers {
+				for _, e := range container.Env {
+					if e.Name == envName && e.Value != "" {
```
markusthoemmes: Can we check for the actual value here? I think we want to check that this is exactly the configured proxy when configuring it and that this is unset when taking it out again.

test/service.go
```diff
@@ -65,9 +66,52 @@ func CreateService(ctx *Context, name, namespace, image string) (*servingv1beta1
 	return service, nil
 }
 
+func WaitForControllerEnvironment(ctx *Context, ns, envName string) error {
+	return wait.PollImmediate(Interval, 10*time.Minute, func() (bool, error) {
+		pods, _ := ctx.Clients.Kube.CoreV1().Pods(ns).List(metav1.ListOptions{
```
markusthoemmes: Please handle the error here.

test/service.go
```diff
@@ -65,9 +66,52 @@ func CreateService(ctx *Context, name, namespace, image string) (*servingv1beta1
...
+	}
+	for _, cond := range pod.Status.Conditions {
+		if cond.Type == corev1.PodReady && cond.Status == corev1.ConditionTrue {
+			return true
+		}
+	}
+	return false
+}
+
+func GetService(ctx *Context, name, namespace string) (*servingv1beta1.Service, error) {
```
markusthoemmes: With the changes proposed above, this is no longer needed.

test/service.go
```diff
@@ -65,9 +67,53 @@ func CreateService(ctx *Context, name, namespace, image string) (*servingv1beta1
...
+			for _, container := range pod.Spec.Containers {
+				for _, e := range container.Env {
+					if e.Name == envName && e.Value != envValue {
+						return false, nil
+					}
+				}
+
+			}
+		}
+		// We only reach here if all controller pods are ready and have the
```
markusthoemmes: Ugh, this is of course wrong. You might want to tune this back to the implementation you had before. Sorry I screwed this one up.

savitaashture: No issues i will change
I m just verifying by testing locally now

PR #124, opened 2020-02-18
Add previous version install option to make install-previous 

nak3:
This patch adds the support of previous version installation by `make install-previous`.

For example, to Install previous version:

```
DOCKER_REPO_OVERRIDE=quay.io/nak3 make install-previous 
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ images:
 	./hack/images.sh $(DOCKER_REPO_OVERRIDE)
 
 install:
-	./hack/install.sh
+	./hack/install.sh $(PREVIOUS_SERVERLESS_VERSION)
 
 teardown:
 	./hack/teardown.sh
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,10 @@ images:
 	./hack/images.sh $(DOCKER_REPO_OVERRIDE)
 
 install:
-	./hack/install.sh $(PREVIOUS_SERVERLESS_VERSION)
+	./hack/install.sh
+
+install-previous:
+	INSTALL_PREVIOUS_VERSION="true" ./hack/install.sh
 
 teardown:
 	./hack/teardown.sh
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -7,7 +7,7 @@ images:
 	./hack/images.sh $(DOCKER_REPO_OVERRIDE)
 
 install:
-	./hack/install.sh
+	./hack/install.sh $(PREVIOUS_SERVERLESS_VERSION)
```
markusthoemmes: I think I'd prefer a `make install-previous` to not have to remember the incantation 😂

hack/install.sh
```diff
@@ -11,7 +11,9 @@
 # shellcheck disable=SC1091,SC1090
 source "$(dirname "${BASH_SOURCE[0]}")/lib/__sources__.bash"
 
-set -Eeuo pipefail
+set -Eeo pipefail
```
markusthoemmes: Why drop the `u`?

hack/lib/serverless.bash
```diff
@@ -3,11 +3,16 @@
 function ensure_serverless_installed {
   logger.info 'Check if Serverless is installed'
   local group=${1:-operator}
+  local prev=${2}
```
markusthoemmes: I'd default this to `false` for good measure

hack/lib/serverless.bash
```diff
@@ -3,11 +3,16 @@
...
   if oc get knativeserving.${group}.knative.dev knative-serving -n "${SERVING_NAMESPACE}" >/dev/null 2>&1; then
     logger.success 'Serverless is already installed.'
     return 0
   fi
-  install_serverless_latest
+  if [[ -n $prev ]]; then
```
markusthoemmes: Let's check for `$prev == "true"`

hack/install.sh
nak3: Added back the `u` option.

`u` option does not allow empty value, so I dropped it to accept `install_prev="$1"` with empty. But never mind, as I could use env value by using `make install-previous`.

PR #125, opened 2020-02-18
Remove CI-namespace references.

markusthoemmes:
See https://issues.redhat.com/browse/SRVKS-423 for more information. In theory, these image references should already be set, so we might get lucky. Printing `env` though just to be sure.

### DIFF
--- a/hack/catalog.sh
+++ b/hack/catalog.sh
@@ -24,8 +24,8 @@ find $OLM_DIR -name '*_crd.yaml' | sort -n | xargs -I{} cp {} $CRD_DIR/
 
 # Determine if we're running locally or in CI.
 if [ -n "$OPENSHIFT_BUILD_NAMESPACE" ]; then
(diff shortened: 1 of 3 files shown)


hack/catalog.sh
```diff
@@ -24,8 +24,8 @@ find $OLM_DIR -name '*_crd.yaml' | sort -n | xargs -I{} cp {} $CRD_DIR/
 
 # Determine if we're running locally or in CI.
 if [ -n "$OPENSHIFT_BUILD_NAMESPACE" ]; then
```
maschmid: we should probably not even depend on existence of OPENSHIFT_BUILD_NAMESPACE then? (not mentioned here https://github.com/openshift/ci-tools/blob/master/TEMPLATES.md#image_format )

markusthoemmes: Let's do that separately. A lot of the variables actually mentioned in this doc are not available to our jobs so I asked for guidance here.

maschmid: ok

PR #126, opened 2020-02-19
Add eventing

aliok:
TODO:
- [x] ~consoleclidownload: move the code that adds that CR to a common place in order to ensure that CR is created by eventing or serving controller.~ Ignore for now. see https://coreos.slack.com/archives/CF5ANN61F/p1582117283254900
- [x]  Write a note about service account name and roles in CSV
- [x]  ~Duck type for serving and eventing CR types to deduplicate functions~ WON'T DO. See [here](https://coreos.slack.com/archives/CD87JDUB0/p1582228283048300)
- [x] Image override env vars for eventing controller deployments are using upstream nightly images atm --> Created https://issues.redhat.com/browse/SRVKE-334
- [x] Check image override env vars for things like BROKER_FILTER_IMAGE, which is defined within eventing-controller deployment. Discussion: https://coreos.slack.com/archives/CEXRYS5QC/p1582286657045800 I don't think we need to do something at this stage.
- [x] `make dev`/`make install` should also install eventing CR and do eventing related tests
- [x] `admission webhook "validating.knativeserving.openshift.io" denied the request: Version constraint not fulfilled: minimum version: 4.3.0, current version: 4.3.0-0.nightly-2020-01-28-050934` --> discussion https://coreos.slack.com/archives/CD87JDUB0/p1582547895003100
- [x] Need to ship the obsolete eventing CRD as well because of https://github.com/knative/eventing-operator/issues/112
- [x] Constant log messages in `eventing-webhook`:
```
{"level":"info","ts":"2020-02-20T19:32:43.226Z","logger":"eventing-webhook","caller":"webhook/webhook.go:191","msg":"Webhook ServeHTTP request=&http.Request{Method:\"POST\", URL:(*url.URL)(0xc000902280), Proto:\"HTTP/1.1\", ProtoMajor:1, ProtoMinor:1, Header:http.Header{\"Accept\":[]string{\"application/json, */*\"}, \"Accept-Encoding\":[]string{\"gzip\"}, \"Content-Length\":[]string{\"5239\"}, \"Content-Type\":[]string{\"application/json\"}, \"User-Agent\":[]string{\"kube-apiserver-admission\"}}, Body:(*http.body)(0xc001427640), GetBody:(func() (io.ReadCloser, error))(nil), ContentLength:5239, TransferEncoding:[]string(nil), Close:false, Host:\"eventing-webhook.knative-eventing.svc:443\", Form:url.Values(nil), PostForm:url.Values(nil), MultipartForm:(*multipart.Form)(nil), Trailer:http.Header(nil), RemoteAddr:\"10.128.0.1:46014\", RequestURI:\"/legacysinkbindings?timeout=30s\", TLS:(*tls.ConnectionState)(0xc002a62fd0), Cancel:(<-chan struct {})(nil), Response:(*http.Response)(nil), ctx:(*context.cancelCtx)(0xc001427680)}","commit":"8a92a8c"}
{"level":"info","ts":"2020-02-20T19:32:43.227Z","logger":"eventing-webhook","caller":"webhook/webhook.go:224","msg":"AdmissionReview for v1.GroupVersionKind{Group:\"apps\", Version:\"v1\", Kind:\"DaemonSet\"}: openshift-sdn/sdn-controller response=&v1beta1.AdmissionResponse{UID:\"\", Allowed:true, Result:(*v1.Status)(nil), Patch:[]uint8(nil), PatchType:(*v1beta1.PatchType)(nil), AuditAnnotations:map[string]string(nil)}","commit":"8a92a8c","knative.dev/kind":"apps/v1, Kind=DaemonSet","knative.dev/namespace":"openshift-sdn","knative.dev/name":"sdn-controller","knative.dev/operation":"UPDATE","knative.dev/resource":"{apps v1 daemonsets}","knative.dev/subresource":"","knative.dev/userinfo":"{system:serviceaccount:openshift-network-operator:default 98104fbb-b740-4f60-934b-70eb1886b80a [system:serviceaccounts system:serviceaccounts:openshift-network-operator system:authenticated] map[]}"}
```
--> ignore for now. see https://issues.redhat.com/browse/SRVKE-335
- [x] Unit tests
- [x] e2e and upgrade tests and adaption of scripting for them --> created https://issues.redhat.com/browse/SRVKE-333
- [x] Note about `updateImagesFromEnviron((*servingv1alpha1.Registry)(&ke.Spec.Registry))`
- [x] Don't use 1.5 CSV anymore, use 1.6
- [x] Remove obsolete CRD from CSV as the upstream operator doesn't need that anymore
- [x] Hide CRDs in developer catalog, similar to https://github.com/openshift-knative/serverless-operator/pull/170
- [x] Smoke tests (stuff in https://github.com/openshift-knative/serverless-operator/pull/126)
- [x] Update image references in 1.6 CSV with upstream 1.3 images
- [x] Looks like image override doesn't work with upstream operator --> issued https://github.com/knative/eventing/issues/2794 UPDATE: we don't care. we do it as here: https://github.com/openshift-knative/eventing-operator/pull/5

## Verification instructions:
* `DOCKER_REPO_OVERRIDE=docker.io/yourusername`
* `make images dev`
* Create a `KnativeEventing` CR in `knative-eventing` namespace
* Do some smoke tests:
```
# Create KSVC
cat <<EOF | kubectl apply -f -
kind: Service
metadata:
  name: eventinghello
spec:
  template:
    metadata:
      name: eventinghello-v1
      annotations:
        autoscaling.knative.dev/target: "1"
    spec:
      containers:
      - image: quay.io/burrsutter/eventinghello:0.0.1
EOF

# Create CronJobSource
cat <<EOF | kubectl apply -f -
apiVersion: sources.eventing.knative.dev/v1alpha1
kind: CronJobSource
metadata:
  name: event-greeter-cronjob-source
spec:
  schedule: "*/2 * * * *"
  data: '{"message": "Thanks for doing Knative Tutorial"}'
  sink:
    kind: Service
    name: eventinghello
EOF

# Wait for the Knative Service to be ready
while output=$(kubectl get ksvc eventinghello); do
  echo "$output"
  echo $output | grep True >/dev/null && break
  sleep 2
done

# Check logs of the ksvc
stern eventinghello-v1-deployment*

# You should see 
eventinghello-v1-deployment-68557fb879-zzlp7 user-container POST:{"message":"Thanks for doing Knative Tutorial"}

```


knative-operator/vendor/knative.dev/eventing-operator/test/cleanup.go
```diff
@@ -0,0 +1,46 @@
+/*
```
jcrossley3: Do we need to vendor all these test files?

knative-operator/pkg/common/eventing.go
```diff
@@ -17,6 +20,33 @@ func MutateEventing(ke *eventingv1alpha1.KnativeEventing, c client.Client) error
 	return nil
 }
 
+// eventingImagesFromEnviron overrides registry images
+func eventingImagesFromEnviron(ke *eventingv1alpha1.KnativeEventing, _ client.Client) error {
```
jcrossley3: I concur with @markusthoemmes that there is opportunity to factor some common code out here. Adding `eventing.go` in here makes me want to consider renaming `openshift.go` to `serving.go`. There is already a `util.go` in there that the common code for this function could go.

knative-operator/vendor/knative.dev/eventing-operator/test/cleanup.go
aliok: Well, as it is also like that in the serving part of things, let us do it separately.
I don't want to do any of these 2:
- Touch serving related things in this PR (i.e. deleting vendored serving operator tests)
- Do some improvements for _only_ for eventing which doesn't exist for serving

knative-operator/pkg/common/eventing.go
aliok: Done. See
- https://github.com/openshift-knative/serverless-operator/pull/126/commits/cc97fa54e147a23eaf472bf9a20714289c82e945
- https://github.com/openshift-knative/serverless-operator/pull/126/commits/8f2eb99ed1caf09b1694d25c70a2d9eb3b1c87c1

knative-operator/pkg/common/eventing.go
```diff
@@ -0,0 +1,25 @@
...
+		if err := stage(ke, c); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// eventingImagesFromEnviron overrides registry images
+func eventingImagesFromEnviron(ke *eventingv1alpha1.KnativeEventing, _ client.Client) error {
+	updateImagesFromEnviron((*servingv1alpha1.Registry)(&ke.Spec.Registry))
```
aliok: There are 2 `Registry` types, one in eventing one in serving upstream operators.

As we will have a unified type when the operators are merged and since these 2 `Registry` types are exactly the same, I made the `updateImagesFromEnviron` accept the serving `Registry` and I simply cast the eventing one to that.

knative-operator/deploy/resources/eventing_networkpolicies.yaml
```diff
@@ -0,0 +1,48 @@
...
+spec:
+  egress:
+  - {}
+  ingress:
+  - {}
+  podSelector: {}
+  policyTypes:
+  - Ingress
+  - Egress
+---
```
maschmid: We shouldn't be needed these anymore.

olm-catalog/serverless-operator/1.5.0/serverless-operator.v1.5.0.clusterserviceversion.yaml
```diff
@@ -274,6 +316,41 @@ spec:
...
+                - name: METRICS_DOMAIN
+                  value: knative.dev/eventing-operator
+                # TODO
+                image: gcr.io/knative-releases/knative.dev/eventing-operator/cmd/manager@sha256:294a0fcfa6c206f49c56b299c4377d27a8ac420aafa18243874ef8b7e28859c0
+                imagePullPolicy: IfNotPresent
+                name: knative-eventing-operator
+                ports:
+                - containerPort: 9090
+                  name: metrics
+              serviceAccountName: knative-serving-operator
```
aliok: This serviceAccount is used by the upstream serving operator deployment and our operator deployment too.

I didn't want to create a separate SA just for the upstream eventing operator as we will need our Knative operator deployment to have permissions for both serving and eventing. ...and, we cannot use 2 service accounts in a deployment.

So, I just added eventing related permissions to this service account. It needs to be renamed though (later)

knative-operator/pkg/controller/knativeeventing/knativeeventing_controller.go
```diff
@@ -0,0 +1,126 @@
...
+		if errors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+
+	// TODO: I assumed we don't need this
+	//if instance.GetDeletionTimestamp() != nil {
+	//	return reconcile.Result{}, r.delete(instance)
+	//}
```
aliok: @jcrossley3 I don't think we need to do anything with the deletion timestamp here. This timestamp is there in serving to do some checks around finalizers. ...and we don't need any finalizers in eventing side since we don't do any external things like deleting SMMR etc.

what do you think?

jcrossley3: If you have no clean-up to do, then no, you don't need it, but the upstream k-e-o should be deleting its resources (and any other clean-up) in a finalizer. Bad things can happen if you tie up that reconciler for too long.

olm-catalog/serverless-operator/1.5.0/serverless-operator.v1.5.0.clusterserviceversion.yaml
jcrossley3: Agree, there's still a lot of renaming to be done. We can do that in a later PR.

knative-operator/pkg/common/eventing.go
jcrossley3: I don't love that solution, tbh. I was thinking the common function would just return a `map[string]string]` that the caller could assign to its registry. The `default` feature complicates that, but there's no harm in returning it in the map, and just let the caller set it if it finds it. Maybe?

olm-catalog/serverless-operator/1.5.0/serverless-operator.v1.5.0.clusterserviceversion.yaml
```diff
@@ -274,6 +316,41 @@ spec:
...
+                - name: OPERATOR_NAME
+                  value: knative-eventing-operator
+                - name: SYSTEM_NAMESPACE
+                  valueFrom:
+                    fieldRef:
+                      fieldPath: metadata.namespace
+                - name: METRICS_DOMAIN
+                  value: knative.dev/eventing-operator
+                # TODO
+                image: gcr.io/knative-releases/knative.dev/eventing-operator/cmd/manager@sha256:294a0fcfa6c206f49c56b299c4377d27a8ac420aafa18243874ef8b7e28859c0
```
aliok: I will need to change this image with our midstream eventing operator

hack/lib/serverless.bash
```diff
@@ -103,6 +106,7 @@ function deploy_knativeserving_v1alpha1_1.3.0 {
   logger.info 'Deploy Knative Serving'
 
   # Wait for the CRD to appear
+  # TODO: use full path name to ensure we're checking the right thing. aka knativeservings.operator.knative.dev VS knativeservings.serving.knative.dev
   timeout 900 "[[ \$(oc get crd | grep -c knativeservings) -eq 0 ]]" || return 6
```
aliok: I think we need to change this `grep -c knativeservings` to `knativeservings.operator.knative.dev`
Otherwise, I am not sure from what group we would get `knativeservings`.
There's probably a rule there, but I would rather use the long name here. ...unless there's a proper reason.

cc @jcrossley3

aliok:
@jcrossley3 @markusthoemmes I can say this PR can be reviewed now. I don't expect any big changes.
Feel free to to review

hack/lib/serverless.bash
jcrossley3: @aliok technically, it shouldn't matter. either one should work. but since the function has `_latest` in its name, i suppose it should use the fqn to be sure.

knative-operator/deploy/resources/eventing_networkpolicies.yaml
aliok: +1

knative-operator/pkg/common/eventing.go
aliok: Thanks for the suggestion. https://github.com/openshift-knative/serverless-operator/pull/126/commits/52ac732151382a5cf78d3d43db7336e13815941c

hack/lib/serverless.bash
aliok: Ok, following the minimal disruption principle, I won't do anything around this

aliok:
@jcrossley3 @markusthoemmes I removed the WIP status. Ready to merge.
There are some follow up tasks created in JIRA for things written in PR description.

olm-catalog/serverless-operator/1.5.0/serverless-operator.v1.5.0.clusterserviceversion.yaml
aliok: TBD later with https://issues.redhat.com/browse/SRVKE-334

matzew:
This will add EVENTING to the `serverless-operator`, in version 0.12 

But, my vote would be to hold on, with this PR, until 1.5.0 cuts, so we can than include eventing 0.13, which is what we aim for Tech Preview.

0.12 bits will be still DP, and I guess we want to NOT include those than ?


This work was started EARLY to see _EARLY_ any short comings or "road blockers" on our way to TP 

So, unless the 1.5.0 repeats the 1.4x we can assume that we are here on a good and early way to land Eventing TP for the 1.6.0 release (at least that's my understanding is the targeted version, looking at version matrix and recent PM calls)

/cc @bbrowning @markusthoemmes

aliok:
Image registry.svc.ci.openshift.org/openshift/knative-v0.12.1:knative-eventing-operator now exists after https://github.com/openshift/release/pull/7305 is merged

Need to use that one in this PR

matzew:
I've tested this PR with a few e2e scenarios:

* https://github.com/matzew/knative-eventing-samples/tree/master/01-ping_service
* https://github.com/matzew/knative-eventing-samples/tree/master/01-source_service
* https://github.com/matzew/knative-eventing-samples/tree/master/02-source_channel
* https://github.com/matzew/knative-eventing-samples/tree/master/03-source_broker

The all work! Mostly!

However, only for the broker and the "trigger with annotation", I did not have a success. I will investigate on this, but the labeling works, so that should not prevent this PR from being merged

aliok:
ok, removing the WIP label.
@jcrossley3 mind doing a review now?

note that https://github.com/knative/eventing/issues/2794 is still open but I don't think it is a blocker for this PR. We just need to use the new version of the upstream eventing operator once it is released later.

markusthoemmes:
@aliok The issue in both CI runs is

```
 --- FAIL: TestKnativeServing (0.09s)
    --- FAIL: TestKnativeServing/create_subscription_and_wait_for_CSV_to_succeed (0.09s)
        knative_serving_test.go:40: Failed subscriptions.operators.coreos.com "serverless-operator-subscription" already exists 
```

That's not a known flake currently. Does the new test-suite have an issue here?

knative-operator/pkg/common/common_test.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	servingv1alpha1 "knative.dev/serving-operator/pkg/apis/serving/v1alpha1"
+	"testing"
+)
+
+func verifyImageOverride(t *testing.T, registry *servingv1alpha1.Registry, imageName string, expected string) {
+	if registry.Override[imageName] != expected {
+		t.Errorf("Missing queue image. Expected a map with following override in it : %v=%v, actual: %v", imageName, expected, registry.Override)
+	}
```
markusthoemmes: nit: This feels like something we might as well just inline.

knative-operator/pkg/common/serving.go
```diff
@@ -93,24 +90,17 @@ func ensureCustomCerts(ks *servingv1alpha1.KnativeServing, _ client.Client) erro
...
-			}
-		}
+	ks.Spec.Registry.Override = buildImageOverrideMapFromEnviron()
+
+	if defaultVal, ok := ks.Spec.Registry.Override["default"]; ok {
+		ks.Spec.Registry.Default = defaultVal
 	}
+
+	// special case for queue-proxy
+	if qpVal, ok := ks.Spec.Registry.Override["queue-proxy"]; ok {
```
markusthoemmes: I like this a lot more than the older code 😂

knative-operator/pkg/webhook/knativeserving/webhook_validating_test.go
```diff
@@ -80,6 +80,102 @@ func TestLoneliness(t *testing.T) {
...
+	os.Clearenv()
+	os.Setenv("MIN_OPENSHIFT_VERSION", "4.3.0-0")
+	validator := KnativeServingValidator{}
+	validator.InjectDecoder(&mockDecoder{ks1})
+	validator.InjectClient(fake.NewFakeClient(mockClusterVersion("4.3.0-0.nightly-2020-01-28-050934")))
+	result := validator.Handle(context.TODO(), types.Request{})
+	if !result.Response.Allowed {
+		t.Error("The version is later, but the request is not allowed")
+	}
+}
```
markusthoemmes: nit: The valid tests seem redundant to `TestPreReleaseVersionConstraint` above. Maybe we should fold them into there and make a similar one for invalid tests for less duplicated code?

test/e2e/knative_eventing_test.go
```diff
@@ -0,0 +1,74 @@
...
+			if err != nil {
+				t.Fatalf("Deployment %s is not gone: %v", deploymentName, err)
+			}
+		}
+
+		err := test.WithDeploymentCount(caCtx, knativeEventing, 0)
+		if err != nil {
+			t.Fatalf("Some deployments were to be deleted but not in namespace %s. Err: %v", knativeEventing, err)
+		}
+	})
```
markusthoemmes: This lacks a test for `"undeploy serverless operator and check dependent operators removed"` like we have in the serving suite. That should also fix the breaking CI.

aliok: +1. perfect!

markusthoemmes: We either add this test here or we remove the CSV/Subscription installation from each of the two suites. Or we make them gracefully accept things already installed.

aliok:
> Do I read this correct in that you're not adding the upstream Knative Eventing test suite?

@markusthoemmes yes. just created a JIRA for that: https://issues.redhat.com/browse/SRVKE-353

knative-operator/pkg/common/common_test.go
aliok: It was like that and I felt externalizing is better as 

```
verifyIngress(t, ks, domain)
verifyImageOverride(t, &ks.Spec.Registry, "queue-proxy", image)
verifyCerts(t, ks)
```

looks better than
```
verifyIngress(t, ks, domain)
if ks.Spec.Registry.Override["foo"] != image {
	t.Errorf("Missing queue image. Expected a map with following override in it : %v=%v, actual: %v", "foo", image, ks.Spec.Registry.Override)
}
verifyCerts(t, ks)
```

knative-operator/pkg/webhook/knativeserving/webhook_validating_test.go
aliok: +1. I didn't see that structure after I rebased. Good point!

test/e2e/knative_eventing_test.go
aliok: Just added that test

markusthoemmes:
/test 4.4-e2e-aws-ocp-44

Setup issue.

hack/lib/serverless.bash
```diff
@@ -3,7 +3,9 @@
 function ensure_serverless_installed {
   logger.info 'Check if Serverless is installed'
   local prev=${1:-false}
-  if oc get knativeserving.operator.knative.dev knative-serving -n "${SERVING_NAMESPACE}" >/dev/null 2>&1; then
+  if oc get knativeserving.operator.knative.dev knative-serving -n "${SERVING_NAMESPACE}" >/dev/null 2>&1&& \
```
jcrossley3: this looks weird to me: `2>&1&&`

can we group the two && commands with parentheses and apply the redirects to both only once?

knative-operator/Gopkg.lock
```diff
@@ -363,6 +363,14 @@
   revision = "24b83195037b3bc61fcda2d28b7b0518bce293b6"
   version = "v1.0.4"
 
+[[projects]]
+  branch = "master"
+  digest = "1:0e9bfc47ab9941ecc3344e580baca5deb4091177e84dd9773b48b38ec26b93d5"
+  name = "github.com/mattbaird/jsonpatch"
```
jcrossley3: What is pulling this in? How many different jsonpatch libs do we need? :)

knative-operator/pkg/common/serving.go
jcrossley3: `s/the older/jim's/`

knative-operator/pkg/webhook/knativeeventing/webhook_mutating.go
```diff
@@ -0,0 +1,102 @@
...
+		return admission.ErrorResponse(http.StatusInternalServerError, err)
+	}
+	return types.Response{
+		Patches: patches,
+		Response: &admissionv1beta1.AdmissionResponse{
+			Allowed:   true,
+			PatchType: func() *admissionv1beta1.PatchType { pt := admissionv1beta1.PatchTypeJSONPatch; return &pt }(),
+		},
+	}
+}
```
jcrossley3: Let's factor this out to a common lib. It's a temporary hack, so I'd prefer we not have two copies of it.

knative-operator/pkg/webhook/knativeeventing/webhook_validating.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// KnativeEventingValidator checks for a minimum OpenShift version
+func (v *KnativeEventingValidator) validate(ctx context.Context, ke *eventingv1alpha1.KnativeEventing) (allowed bool, reason string, err error) {
+	log := common.Log.WithName("validate")
+	stages := []func(context.Context, *eventingv1alpha1.KnativeEventing) (bool, string, error){
+		v.validateNamespace,
+		v.validateVersion,
+		v.validateLoneliness,
+	}
```
jcrossley3: I'd really like to figure out how to factor these out into a common lib. This is one of those times when type-safety is stupid. ;)

knative-operator/pkg/webhook/knativeeventing/webhook_validating_test.go
```diff
@@ -0,0 +1,93 @@
+package knativeeventing_test
```
jcrossley3: There is a lot of redundant code in here. I'm ok leaving it for now if it's too much effort to refactor, but I'd like it gone in a subsequent PR quickly.

knative-operator/pkg/webhook/knativeserving/webhook_validating.go
```diff
@@ -104,39 +103,12 @@ func (v *KnativeServingValidator) InjectDecoder(d types.Decoder) error {
...
-	if current.Major == 0 && current.Minor == 0 {
-		return true, "CI build detected, bypassing version check", nil
-	}
-
-	if current.LessThan(*minVersion) {
-		msg := fmt.Sprintf("Version constraint not fulfilled: minimum version: %s, current version: %s", minVersion.String(), current.String())
-		return false, msg, nil
-	}
-	return true, "", nil
+	return webhookutil.ValidateOpenShiftVersion(ctx, v.client)
```
jcrossley3: This is good -- can we not do more of this?

knative-operator/pkg/webhook/knativeeventing/webhook_validating_test.go
matzew: :+1:

hack/lib/serverless.bash
aliok: I simply lack bash skills to give you a good answer but I can give it a try.

knative-operator/Gopkg.lock
aliok: Well, honestly, I don't really check what `dep` does in `.lock` file. I just added the upstream eventing operator as a dependency and jsonpatch lib came. 

I think this is something maybe we can address in the upstream operator but not here.

knative-operator/pkg/webhook/knativeeventing/webhook_mutating.go
aliok: +1

hack/lib/serverless.bash
aliok: I just saw the space is somehow deleted.

I think following should be readable enough:
```
if oc get knativeserving.operator.knative.dev knative-serving -n "${SERVING_NAMESPACE}" >/dev/null 2>&1 && \
     oc get knativeeventing.operator.knative.dev knative-eventing -n "${EVENTING_NAMESPACE}" >/dev/null 2>&1
```

I will add that space back.

markusthoemmes:
@aliok the last commit doesn't work for a productized operator. We need all references to all images to be set in the CSV. Productization replaces those references with a productized image.

PR #127, opened 2020-02-20
Set default value to INSTALL_PREVIOUS_VERSION

nak3:
As `set -u` option does not allow empty value, when
`INSTALL_PREVIOUS_VERSION` has an empty value, the script fails due
to:

```
./hack/install.sh: line 25: INSTALL_PREVIOUS_VERSION: unbound variable
```

This patch fixes it.


nak3:
Sorry, I misunderstood https://github.com/openshift-knative/serverless-operator/pull/124#discussion_r381039030 and missed to test `make install`. It fails due to `INSTALL_PREVIOUS_VERSION: unbound variable`. `-u` does not allow any empty value.

markusthoemmes:
Can we instead set a global variable to `false` to make this available? Should go into `vars.bash`.

nak3:
@markusthoemmes Yes, I should have changed so. Now updated.

PR #129, opened 2020-02-20
[WIP] Try parallel=10 for Knative Serving E2E tests

mgencur:
Do not merge!

Let's just try out and see how many test failures we get and how much time we gain.


mgencur:
So, the execution time of the test container was 32m 17s. I can see in other PRs that parallel=3 was taking about 37 minutes.

mgencur:
/retest

Trying to see if we get some random failures and if the execution time is similar.

mgencur:
OK. The CI is extremely unstable but from the several runs that I've seen it is NOT apparent that parallel=10 would make the tests faster or less stable when running against openshift CI infrastructure. Closing this as it didn't help.

PR #138, opened 2020-02-25
[SRVKS-427] Override all configuration we set as we have authority. Apply mutations on update.

markusthoemmes:
Currently, when we update the operator, we will try to override the image setting in the respective config maps. Our `Configure` helper is too conservative though, and only writes config that was not set before.

All configs that we actively set are subject to our authority though. For now, we should always override all of these settings as they are integral to our integration.

### DIFF
--- a/knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
+++ b/knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
@@ -12,6 +12,7 @@ import (
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 	"github.com/operator-framework/operator-sdk/pkg/predicate"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -111,13 +112,15 @@ func (r *ReconcileKnativeServing) Reconcile(request reconcile.Request) (reconcil
 
 // configure default settings for OpenShift
 func (r *ReconcileKnativeServing) configure(instance *servingv1alpha1.KnativeServing) error {
-	if _, ok := instance.GetAnnotations()[common.MutationTimestampKey]; ok {
-		return nil
-	}
+	before := instance.DeepCopy()
 	log.Info("Configuring KnativeServing for OpenShift")
 	if err := common.Mutate(instance, r.client); err != nil {
 		return err
 	}
+	if equality.Semantic.DeepEqual(before, instance) {
+		return nil
+	}
+	// Only apply the update if something changed.
 	return r.client.Update(context.TODO(), instance)
 }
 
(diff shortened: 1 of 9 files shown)


nak3:
This looks like we need to backport to release-1.4 branch?

markusthoemmes:
@nak3 I don't think so. 1.4 introduced this mechanism and thus doesn't fail due to its effects.

jcrossley3:
i need to think this through, so...
/hold

markusthoemmes:
FWIW: The upgrade test failures are expected here. We probably shouldn't merge this until we've concluded on how we're going to approach them. We can alternaatively back out Kourier for now and move forward with this.

mgencur:
@markusthoemmes do you expect that with the next upgrade (1.5 to 1.6) the upgrade tests will start passing again? Are these failures in TestProbe only temporary? In this case we could do that same as with one of previous releases - Stop the TestProbe before the upgrade and document that a temporary outage is expected. And for the next release enable it again.

markusthoemmes:
@mgencur Yes and no. I want to have some more evidence that the regression I see isn't actually a bug we can fix. I have not reached that conclusion yet.

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
```diff
@@ -111,13 +111,14 @@ func (r *ReconcileKnativeServing) Reconcile(request reconcile.Request) (reconcil
 
 // configure default settings for OpenShift
 func (r *ReconcileKnativeServing) configure(instance *servingv1alpha1.KnativeServing) error {
-	if _, ok := instance.GetAnnotations()[common.MutationTimestampKey]; ok {
-		return nil
-	}
 	log.Info("Configuring KnativeServing for OpenShift")
-	if err := common.Mutate(instance, r.client); err != nil {
+	changed, err := common.Mutate(instance, r.client)
```
jcrossley3: I understand why we need to overwrite the image overrides, i.e. make `Configure` less conservative, but I don't understand why we need to know we changed something here.

This has also made me realize that we're not setting `version/version.go` properly. The operator should know definitively what version it's installing, so it should be able to tell exactly which images it should be configuring.

markusthoemmes: I want to know if something changed here to not unnecessarily make the api call to update the KnativeServing.

The second part I'm not quite parsing. We're configuring the correct images via env I think?

markusthoemmes:
Setup issue

markusthoemmes:
Weeee upgrade tests passed without dropping requests :)

The unrelated refactoring caused a nil pointer exception though.

nak3:
/retest

> The unrelated refactoring caused a nil pointer exception though.

Was it in upgrade test or e2e test?

nak3:

CI setup issue.

markusthoemmes:
/test 4.3-e2e-aws-ocp-43

Unrelated deletion issues. We should prolly doublecheck our finalizer logic 🙈

markusthoemmes:
/test 4.3-e2e-aws-ocp-43

Kourier instability that‘s also on master. We need to dig into this!

PR #139, opened 2020-02-26
Fix unsupported OCP version

mgencur:
In upgrade tests we install one version of Serverless back and that version is supposed to work on OCP 4.2, not 4.1. Then we upgrade to the latest and there the upgrade should fail.


mgencur:
Obsolete. Closing.

PR #140, opened 2020-02-26
Another whack at SRVKS-427

jcrossley3:
See commit messages


knative-operator/pkg/common/openshift.go
```diff
@@ -17,14 +17,17 @@ import (
 var log = Log
 
 func Mutate(ks *servingv1alpha1.KnativeServing, c client.Client) error {
+	if ks.GetAnnotations()[MutationKey] == version.Version {
```
markusthoemmes: We cannot unconditionally skip here based on the version. As mentioned in Slack, not all of these things just change on the boundary of a version. If the user reconfigures their openshift ingress settings for example, we should eventually pick that up once the KS is reconciled. With this check in place, we don't.

PR #141, opened 2020-02-27
Remove shortname ks from knativeservings.serving.knative.dev CRD

nak3:
When we run `oc get ks`, it hits `knative.serving.knative.dev` instead
of `knative.operator.knative.dev`.

Since we expect `knative.operator.knative.dev` now, this patch removes
ks from `knative.serving.knative.dev` CRD.

BEFORE
```
$ oc get ks -n knative-serving -o yaml  |grep apiVersion | head -2
apiVersion: v1
- apiVersion: serving.knative.dev/v1alpha1
```

AFTER
```
$ oc get ks -n knative-serving -o yaml  |grep apiVersion | head -2
apiVersion: v1
- apiVersion: operator.knative.dev/v1alpha1
```

### DIFF
--- a/olm-catalog/serverless-operator/1.5.0/serving_v1alpha1_knativeserving_crd.yaml
+++ b/olm-catalog/serverless-operator/1.5.0/serving_v1alpha1_knativeserving_crd.yaml
@@ -19,8 +19,6 @@ spec:
     listKind: KnativeServingList
     plural: knativeservings
     singular: knativeserving
-    shortNames:
-    - ks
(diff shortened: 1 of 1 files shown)


nak3:
/retest

CI infra issue.

jcrossley3:
I don't have a strong opinion. I've just gotten used to fully-qualifying the names ever since we've supported two. A case has been made for removing them upstream, too: https://github.com/knative/eventing-operator/issues/18#issuecomment-547698821

PR #142, opened 2020-02-27
Enable debug log for Kourier

nak3:
This patch enables debug log on Kourier.

Reconciler reconciles the settings, so this enable
on manifest, but it should be improved in the future.
It is tracked in https://issues.redhat.com/browse/SRVKS-428.


markusthoemmes:
This is outdated I believe.

PR #145, opened 2020-02-28
Bump to upstream v0.12.2 and set our version to be the same

jcrossley3:


knative-operator/Gopkg.toml
```diff
@@ -45,7 +45,7 @@ required = [
 
 [[constraint]]
   name = "knative.dev/serving-operator"
-  branch = "master"
+  version = "v0.12.2"
```
markusthoemmes: Should we vendor from our fork at `openshift-knative/serving-operator` for good measure? We could vendor the `release-0.12` branch to keep everything in sync.

jcrossley3: I'd rather not until we have a compelling reason to. I'm irrationally nervous about depending on a branch instead of a version.

knative-operator/vendor/knative.dev/serving-operator/version/version.go
```diff
@@ -2,5 +2,5 @@ package version
 
 var (
 	// Version is the version that will be applied to the KnativeServing resource.
-	Version = "0.11.0"
+	Version = "0.12.1"
```
warrenvw: Should this be changed to `0.12.2`?

knative-operator/Gopkg.toml
markusthoemmes: We should still vendor from our fork though.

markusthoemmes:
I believe we have went with 0.13.2 instead.

PR #146, opened 2020-03-02
Use official 0.12.1 operator image

nak3:
This patch uses official operator image which removed istio related resources


nak3:
/test 4.3-e2e-aws-ocp-43

Again, Kourier's probe gets 404 error after `rejected: route: unknown weighted cluster 'helloworld-go-5f282/'`. Then, ksvc became IngressNotConfigured.

PR #150, opened 2020-03-03
Use kourier image produced by fork and add related images.

markusthoemmes:
This ties some ends in that it uses the "correct" Kourier image that will be built through productization too.

Also adds the missing images to `relatedImages`.


markusthoemmes:
/test 4.3-e2e-aws-ocp-43

Kourier race we're hunting down.

markusthoemmes:
Kourier bug being hunted already.

/test 4.3-e2e-aws-ocp-43

markusthoemmes:
The failure is actually a test script failure. We try to delete both KS but the second is already gone because we deleted the other one first. Gonna fix this and merge this regardless.

PR #154, opened 2020-03-05
[POC] Support secure route for Knative Service by upstream's autoTLS mechanism

nak3:
Since we started using Kourier, we can support upstream's autoTLS
(nscert, cert-manager) on OpenShift. This patch tries to test it.

Note, we cannot fix following by code change:

- We must use [CertManager v0.12+](https://cert-manager.io/docs/) from Jetstack. (Let's encrypt is not mandatory.)
- OpenShift's Route uses `Passthrough` so HTTP access must be `Redirect` or `Disable`.
- KnativeServing must be v0.13+. (Between 0.12 and 0.13 has many changes for certs.)


nak3:
/hold

I just created this to discuss on Sprint 181 planning. We have to wait for Serving 0.13 on downstream and a few problems described in the first comments.

nak3:
/close

This is not necessary anymore as we use OpenShift's cluster wildcard cert.

PR #157, opened 2020-03-05
[SRVKS-436] remove support for obsolete CRD

jcrossley3:
This includes prep for 1.6.0, since the CSV shouldn't need to own the obsolete CRD any longer.


markusthoemmes:
Thanks! Going to hold this until we cut the 1.5.0 branch.

/hold

olm-catalog/serverless-operator/1.6.0/serverless-operator.v1.6.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,422 @@
...
+                    - name: IMAGE_autoscaler
+                      value: registry.svc.ci.openshift.org/openshift/knative-v0.12.1:knative-serving-autoscaler
+                    - name: IMAGE_autoscaler-hpa
+                      value: registry.svc.ci.openshift.org/openshift/knative-v0.12.1:knative-serving-autoscaler-hpa
+                    - name: IMAGE_controller
+                      value: registry.svc.ci.openshift.org/openshift/knative-v0.12.1:knative-serving-controller
+                    - name: IMAGE_webhook
+                      value: registry.svc.ci.openshift.org/openshift/knative-v0.12.1:knative-serving-webhook
+                    - name: IMAGE_3scale-kourier-gateway
+                      value: docker.io/maistra/proxyv2-ubi8:1.0.4
```
markusthoemmes:
```suggestion
                      value: docker.io/maistra/proxyv2-ubi8:1.0.8
```

olm-catalog/serverless-operator/1.6.0/serverless-operator.v1.6.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,422 @@
...
+                    - name: POD_NAME
+                      valueFrom:
+                        fieldRef:
+                          fieldPath: metadata.name
+                    - name: OPERATOR_NAME
+                      value: "knative-serving-openshift"
+                    - name: MIN_OPENSHIFT_VERSION
+                      value: "4.3.0"
+                    - name: REQUIRED_NAMESPACE
+                      value: "knative-serving"
```
markusthoemmes: This lacks the KOURIER_MANIFEST_PATH option that was recently added. CI fails because it can't install kourier because of that missing variable.

jcrossley3: Done

markusthoemmes:

The images have been missing but should be published now.

markusthoemmes:
sigh.. setup issue

markusthoemmes:

Interesting failure, seems like the upgrade test now takes too long with the GC run in between. We gotta watch this and possibly reduce the time/increase the timeouts.

markusthoemmes:
Heh, this is actually the upgrade test relying on a version change that never happens again 🤕

hack/lib/serverless.bash
```diff
@@ -100,35 +100,32 @@ function find_install_plan {
...
   logger.info 'Deploy Knative Serving'
 
   # Wait for the CRD to appear
   timeout 900 "[[ \$(oc get crd | grep -c knativeservings) -eq 0 ]]" || return 6
 
-  local rootdir
-  rootdir="$(dirname "$(dirname "$(dirname "$(realpath "${BASH_SOURCE[0]}")")")")"
-
-  # Install Knative Serving
-  oc apply -n "${SERVING_NAMESPACE}" -f "${rootdir}/test/v1alpha1/resources/serving.knative.dev_v1alpha1_knativeserving_cr_1.3.0.yaml" || return $?
```
mgencur: We still provide an example here: https://github.com/openshift-knative/serverless-operator/pull/157/files#diff-b466a37bd25835bf7c89df6eef60bb62R5 and we allow the user to specify various attributes. It would be good to create a file named `operator.knative.dev_v1alpha1_knativeserving_cr_1.6.0.yaml` with those various attributes (maybe same as the example?) and make sure we use it with the future releases. This way we will be able to verify that the previous config (with all the attributes) can still be used to install the latest Serverless.
What I see below is a minimal version of the config but it doesn't help check backwards compatibility. And this minimal config is already used in the Golang tests [here](https://github.com/openshift-knative/serverless-operator/blob/master/test/v1alpha1/knativeserving.go#L13) so no need to duplicate here, IMO.

mgencur:
@markusthoemmes I'm confused, do we report the installed version somewhere else?

markusthoemmes:
@mgencur this doesn't update Knative Serving to 0.13 yet, so the reported version stays at 0.12.1. We had the same issue when cutting the minor release for 1.4.

mgencur:
@markusthoemmes alright, thanks. Then I think this failure is simply expected - the new version 1.6 is not yet ready for an upgrade from 1.5 so this check fails as expected, IMO.

The next check at https://github.com/openshift-knative/serverless-operator/pull/157/files#diff-8b58d89183039c8064bef1680efdf24dR211 will fail as well because we rely on the version there too.

test/lib.bash
```diff
@@ -202,14 +202,10 @@ function run_knative_serving_rolling_upgrade_tests {
...
-        timeout 900 '[[ ! ( $(oc get knativeserving.serving.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.version}") != $serving_version && $(oc get knativeserving.serving.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.conditions[?(@.type==\"Ready\")].status}") == True ) ]]' || return 1
-      fi
-      timeout 900 '[[ ! ( $(oc get knativeserving.operator.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.version}") != $serving_version && $(oc get knativeserving.operator.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.conditions[?(@.type==\"Ready\")].status}") == True ) ]]' || return 1
+      timeout 900 '[[ ! ( $(oc get knativeserving knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.version}") != $serving_version && $(oc get knativeserving knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.conditions[?(@.type==\"Ready\")].status}") == True ) ]]' || return 1
 
       # Assert that the old image references eventually fade away
       timeout 900 "oc get pod -n $SERVING_NAMESPACE -o yaml | grep image: | uniq | grep $serving_version" || return 1
```
markusthoemmes:
```suggestion
      # TODO: reenable once we do a Knative version upgrade
      # timeout 900 '[[ ! ( $(oc get knativeserving knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.version}") != $serving_version && $(oc get knativeserving knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.conditions[?(@.type==\"Ready\")].status}") == True ) ]]' || return 1

      # Assert that the old image references eventually fade away
      # timeout 900 "oc get pod -n $SERVING_NAMESPACE -o yaml | grep image: | uniq | grep $serving_version" || return 1
```

I think we can comment these checks out until we actually do a Knative upgrade.

mgencur: Is it useful? You'll need to remember to re-enable this. We could keep it as it is and merge pull requests manually until the upgrade properly works (hopefully very soon). Ideally, we'd really define the new CSV together with the new images.

markusthoemmes: We can track reenablement via a JIRA. We actuially need smarter code here that detects if we're even installing a new version after all. A release might not entail a new Knative version (see patch releases for example).

mgencur: What about introducing an annotation on the CSV, something like `knative-update: "true"` , the test could check if knative-update is expected and change its behaviour?
I guess patch releases might and might not update Knative (both options are possible) and for changes like those in this PR we might as well have both options. So I think we need to set explicitly whether knative is expected to be updated. We can't detect that from e.g. the micro version in `version: 1.4.1`.

Anyway, I'm not sure disabling this code (and having JIRA to re-enable it) makes our life much easier.

markusthoemmes: Ooookay, let's just merge this one then to unblock CSV dependent work?

PR #159, opened 2020-03-06
Add KOURIER_MANIFEST_PATH env var to specify Kourier manifest.

nak3:
This PR supports `KOURIER_MANIFEST_PATH` to allow to switch Kourier's
manifest file. If the value is empty, default path is used.

This is implemented for switching the manifest between CI and production.

### DIFF
--- a/knative-operator/pkg/controller/knativeserving/kourier/kourier.go
+++ b/knative-operator/pkg/controller/knativeserving/kourier/kourier.go
@@ -17,13 +17,13 @@ import (
 )
 
 var (
-	log             = common.Log.WithName("kourier")
-	defaultManifest = "deploy/resources/kourier/kourier-latest.yaml"
+	log  = common.Log.WithName("kourier")
+	path = os.Getenv("KOURIER_MANIFEST_PATH")
 )
 
 // Apply applies Kourier resources.
 func Apply(instance *servingv1alpha1.KnativeServing, api client.Client, scheme *runtime.Scheme) error {
-	manifest, err := mf.NewManifest(manifestPath(), false, api)
+	manifest, err := mf.NewManifest(path, false, api)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


olm-catalog/serverless-operator/1.5.0/serverless-operator.v1.5.0.clusterserviceversion.yaml
```diff
@@ -305,6 +305,8 @@ spec:
                       value: docker.io/maistra/proxyv2-ubi8:1.0.4
                     - name: IMAGE_3scale-kourier-control
                       value: registry.svc.ci.openshift.org/openshift/knative-v0.12.1:kourier
+                    - name: KOURIER_MANIFEST_PATH
+                      value: $KOURIER_MANIFEST_PATH
```
markusthoemmes: I dunno if this works for us correctly. We'd need to reset this in productization I think? Can we instead set this to the "correct" path statically and replace it with `sed` for the CI case?

knative-operator/pkg/controller/knativeserving/kourier/kourier.go
```diff
@@ -17,13 +17,13 @@ import (
 )
 
 var (
-	log  = common.Log.WithName("kourier")
-	path = "deploy/resources/kourier/kourier-latest.yaml"
+	log             = common.Log.WithName("kourier")
+	defaultManifest = "deploy/resources/kourier/kourier-latest.yaml"
```
markusthoemmes: If you go by the suggestion below, we can just use `os.Getenv("KOURIER_MANIFEST_PATH")` here, I think?

olm-catalog/serverless-operator/1.5.0/serverless-operator.v1.5.0.clusterserviceversion.yaml
nak3: Updated
Sure, we can use the path statically and `sed`. It would be more simple.

PR #167, opened 2020-03-11
Add 4.3 channel for v1.6.0

warrenvw:
Looks to me like this change is all that's required in `package.yaml` for the 1.6.0 release.

### DIFF
--- a/olm-catalog/serverless-operator/serverless-operator.package.yaml
+++ b/olm-catalog/serverless-operator/serverless-operator.package.yaml
@@ -1,4 +1,6 @@
 packageName: serverless-operator
 channels:
 - name: techpreview
+  currentCSV: serverless-operator.v1.5.0
+- name: 4.3
   currentCSV: serverless-operator.v1.6.0
--- a/olm-catalog/serverless-operator/serverless-operator.package.yaml
+++ b/olm-catalog/serverless-operator/serverless-operator.package.yaml
@@ -1,7 +1,7 @@
 packageName: serverless-operator
 channels:
 - name: techpreview
-  currentCSV: serverless-operator.v1.5.0
+  currentCSV: serverless-operator.v1.6.0
 - name: 4.3
   currentCSV: serverless-operator.v1.6.0
 - name: 4.4
--- a/olm-catalog/serverless-operator/serverless-operator.package.yaml
+++ b/olm-catalog/serverless-operator/serverless-operator.package.yaml
@@ -4,3 +4,5 @@ channels:
   currentCSV: serverless-operator.v1.5.0
 - name: 4.3
   currentCSV: serverless-operator.v1.6.0
+- name: 4.4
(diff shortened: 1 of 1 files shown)


markusthoemmes:
I guess we need to wait with this until we actually merge a 1.6.0 CSV. Coming in #157.

olm-catalog/serverless-operator/serverless-operator.package.yaml
```diff
@@ -2,3 +2,5 @@ packageName: serverless-operator
 channels:
 - name: techpreview
   currentCSV: serverless-operator.v1.5.0
+- name: 4.3
```
markusthoemmes: Do we need a 4.4 channel too?

warrenvw: Yes. Based on the Openshift Serverless Release Component Matrix, it looks like we should have a 4.4. channel as well. I'll go ahead and add it.

olm-catalog/serverless-operator/serverless-operator.package.yaml
```diff
@@ -1,4 +1,8 @@
 packageName: serverless-operator
 channels:
 - name: techpreview
+  currentCSV: serverless-operator.v1.5.0
```
markusthoemmes: This should be 1.6.0 now as well.

warrenvw: Does it make sense for our GA release to be in the `techpreview` channel? At some point, I imagine we'll need to change our test scripts to use the `4.3`, `4.4`... or even a new `dev`/`nightly` channel. Question is when that should happen. Thoughts?

markusthoemmes: Maybe. Seems like our scripts need adjustments though wrt. these changes.

warrenvw: I'll update 1.5.0 -> 1.6.0 in techpreview channel, since 1.6.0 is still TP.

olm-catalog/serverless-operator/serverless-operator.package.yaml
```diff
@@ -2,3 +2,7 @@ packageName: serverless-operator
 channels:
```
mgencur: We need to place this above this line: `defaultChannel: techpreview`

olm-catalog/serverless-operator/serverless-operator.package.yaml
```diff
@@ -2,3 +2,7 @@ packageName: serverless-operator
...
 - name: techpreview
   currentCSV: serverless-operator.v1.6.0
+- name: 4.3
```
mgencur: We need to quote it: `"4.3"` and `"4.4"`

mgencur:
Replaced by https://github.com/openshift-knative/serverless-operator/pull/190

PR #168, opened 2020-03-11
Make sure to use the proper group for getting knativeserving version

mgencur:
It might happen that the old group is not available and then the serving_version remains empty. As a consequence, the upgrade might be marked as "finished" very quickly, before any pods are actually updated to the new Knative Serving.

### DIFF
--- a/test/lib.bash
+++ b/test/lib.bash
@@ -186,7 +186,10 @@ function run_knative_serving_rolling_upgrade_tests {
 
   if [[ $UPGRADE_SERVERLESS == true ]]; then
     local serving_version
-    serving_version=$(oc get knativeserving.serving.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.version}")
+    if oc get knativeserving.serving.knative.dev knative-serving -n "$SERVING_NAMESPACE" >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


test/lib.bash
```diff
@@ -186,7 +186,10 @@ function run_knative_serving_rolling_upgrade_tests {
 
   if [[ $UPGRADE_SERVERLESS == true ]]; then
     local serving_version
-    serving_version=$(oc get knativeserving.serving.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.version}")
+    if oc get knativeserving.serving.knative.dev knative-serving -n "$SERVING_NAMESPACE" >/dev/null 2>&1; then
+      serving_version=$(oc get knativeserving.serving.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.version}")
+    fi
+    serving_version=$(oc get knativeserving.operator.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.version}")
```
markusthoemmes: Can we not just always use the new version?

mgencur: No. We want to be able to install the previous version using the old CR (like we do in CI) and also through the new CR (like we do in Jenkins). Both options should work.

PR #172, opened 2020-03-12
extend support for HTTPS_PROXY to support egress proxies in Knative

savitaashture:
JIRA: https://issues.redhat.com/browse/SRVKS-421

### DIFF
--- a/knative-operator/pkg/common/deployment_test.go
+++ b/knative-operator/pkg/common/deployment_test.go
@@ -18,10 +18,10 @@ import (
 
 const (
 	deploymentName = "controller"
-	httpProxy      = "http://192.168.130.11:30001"
-	noProxy        = "index.docker.io"
+	proxyValue     = "http://192.168.130.11:30001"
 	namespace      = "default"
 	servingName    = "knative-serving"
+	noProxy        = "index.docker.io"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


knative-operator/pkg/common/deployment_test.go
```diff
@@ -51,20 +51,44 @@ func TestUpdateWithInvalidController(t *testing.T) {
...
+				},
+			},
+		}, deploymentName))
+	ks := &servingv1alpha1.KnativeServing{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      servingName,
+			Namespace: namespace,
+		},
+	}
+	if err := common.ApplyProxySettings(ks, client); err != nil && apierrors.IsNotFound(err) {
```
markusthoemmes: Why are we only looking at IsNotFound errors? I feel like all errors should cause a failure, no? THis applies to all tests below too.

savitaashture: Yep that was a mistake i don't know why i have added condition like that as i have handled [here](https://github.com/openshift-knative/serverless-operator/blob/master/knative-operator/pkg/common/deployment.go#L22-L24) my intention was if at all particular deployment not found i should not continue and return nil but other cases return error

PR #173, opened 2020-03-12
Update Knative Serving + operator to 0.13.

markusthoemmes:
Let's see what happens. Still waiting for the actual operator to release but this **should** work.

This reverts us not using the fully qualified api group for cli commands regarding knativeserving because the upgrade test still needs to deal with the "old" version.


markusthoemmes:
/retest

Proxy flake and shutdown flake.

PR #175, opened 2020-03-13
Always apply Kourier manifest in the reconcile loop

nak3:
This patch changes to apply Kourier manifest always in the reconcile loop.

Part of SRVKS-454.

### DIFF
--- a/knative-operator/pkg/controller/knativeserving/kourier/kourier.go
+++ b/knative-operator/pkg/controller/knativeserving/kourier/kourier.go
@@ -39,14 +39,22 @@ func Apply(instance *servingv1alpha1.KnativeServing, api client.Client, scheme *
 	}
 	if err := checkDeployments(&manifest, instance, api); err != nil {
 		log.Error(err, "")
+		prev := instance.Status.GetCondition(servingv1alpha1.DependenciesInstalled)
 		instance.Status.MarkDependencyInstalling("Kourier")
+		if prev == instance.Status.GetCondition(servingv1alpha1.DependenciesInstalled) {
+			return err
+		}
 		if apiErr := api.Status().Update(context.TODO(), instance); apiErr != nil {
 			return fmt.Errorf("failed to update KnativeServing status: %w", err)
 		}
 		return err
 	}
 	log.Info("Kourier is ready")
+	prev := instance.Status.GetCondition(servingv1alpha1.DependenciesInstalled)
 	instance.Status.MarkDependenciesInstalled()
+	if prev.Status == instance.Status.GetCondition(servingv1alpha1.DependenciesInstalled).Status {
+		return nil
+	}
 	return api.Status().Update(context.TODO(), instance)
 }
 
--- a/knative-operator/pkg/controller/knativeserving/kourier/kourier.go
+++ b/knative-operator/pkg/controller/knativeserving/kourier/kourier.go
@@ -39,11 +39,6 @@ func Apply(instance *servingv1alpha1.KnativeServing, api client.Client, scheme *
 	}
 	if err := checkDeployments(&manifest, instance, api); err != nil {
 		log.Error(err, "")
-		if !instance.Status.IsFullySupported() {
-			// If the instance is already DependencyInstalling status, do not update it.
-			// This avoids frequent LastTransitionTimestamp update.
-			return err
-		}
 		instance.Status.MarkDependencyInstalling("Kourier")
 		if apiErr := api.Status().Update(context.TODO(), instance); apiErr != nil {
 			return fmt.Errorf("failed to update KnativeServing status: %w", err)
--- a/knative-operator/pkg/controller/knativeserving/kourier/kourier.go
+++ b/knative-operator/pkg/controller/knativeserving/kourier/kourier.go
@@ -39,6 +39,11 @@ func Apply(instance *servingv1alpha1.KnativeServing, api client.Client, scheme *
 	}
 	if err := checkDeployments(&manifest, instance, api); err != nil {
 		log.Error(err, "")
+		if !instance.Status.IsFullySupported() {
+			// If the instance is already DependencyInstalling status, do not update it.
+			// This avoids frequent LastTransitionTimestamp update.
+			return err
+		}
 		instance.Status.MarkDependencyInstalling("Kourier")
 		if apiErr := api.Status().Update(context.TODO(), instance); apiErr != nil {
 			return fmt.Errorf("failed to update KnativeServing status: %w", err)
--- a/knative-operator/pkg/controller/knativeserving/kourier/kourier.go
+++ b/knative-operator/pkg/controller/knativeserving/kourier/kourier.go
@@ -33,28 +33,19 @@ func Apply(instance *servingv1alpha1.KnativeServing, api client.Client, scheme *
 	if err != nil {
 		return fmt.Errorf("failed to transform kourier manifest: %w", err)
 	}
-	if instance.Status.IsFullySupported() {
-		// TODO: verify deployed kourier is not different from kourier-latest.yaml accurately.
-		if err := checkDeployments(&manifest, instance, api); err == nil {
-			return nil
-		}
-	}
-
-	// Us reaching here means we need to do something and/or wait longer.
-	instance.Status.MarkDependencyInstalling("Kourier")
-	if err := api.Status().Update(context.TODO(), instance); err != nil {
-		return fmt.Errorf("failed to update KnativeServing status: %w", err)
-	}
-
 	log.Info("Installing Kourier Ingress")
 	if err := manifest.Apply(); err != nil {
 		return fmt.Errorf("failed to apply kourier manifest: %w", err)
 	}
 	if err := checkDeployments(&manifest, instance, api); err != nil {
-		return fmt.Errorf("failed to check kourier deployments: %w", err)
+		log.Error(err, "")
+		instance.Status.MarkDependencyInstalling("Kourier")
+		if apiErr := api.Status().Update(context.TODO(), instance); apiErr != nil {
+			return fmt.Errorf("failed to update KnativeServing status: %w", err)
+		}
+		return err
 	}
 	log.Info("Kourier is ready")
-
 	instance.Status.MarkDependenciesInstalled()
 	return api.Status().Update(context.TODO(), instance)
 }


nak3:
/retest

For some reason, OpenShift controller failed to be connected.

```
    --- FAIL: TestKnativeServing/update_global_proxy_and_verify_calls_goes_through_proxy_server (121.48s)
        knative_serving_test.go:332: update global proxy with empty proxy value
        knative_serving_test.go:337: deploy successfully knative service after proxy update
        knative_serving_test.go:342: update global proxy with proxy server
        knative_serving_test.go:347: wait for controller to be ready after update
        knative_serving_test.go:349: Get https://api.ci-op-sy06rz5i-de449.origin-ci-int-aws.dev.rhcloud.com:6443/api/v1/namespaces/knative-serving/pods?labelSelector=app%3Dcontroller: dial tcp 54.218.133.75:6443: connect: connection refused
```

knative-operator/pkg/controller/knativeserving/kourier/kourier.go
```diff
@@ -32,14 +32,6 @@ func Apply(instance *servingv1alpha1.KnativeServing, api client.Client, scheme *
...
 
-	if instance.Status.IsFullySupported() {
-		// TODO: verify deployed kourier is not different from kourier-latest.yaml accurately.
-		if err := checkDeployments(&manifest, instance, api); err == nil {
-			return nil
-		}
-	}
-
-	// Us reaching here means we need to do something and/or wait longer.
 	instance.Status.MarkDependencyInstalling("Kourier")
```
markusthoemmes: We mustn't do this without a condition I think. I think manifestival needs a `Check` function or something that allows us to check if we even need to apply an update or not. We only want to move into `Installing` state if we know we actually need to apply things.

nak3: I see. In that case, we also should fix Serving manifest's `ApplyAll()`, I think.

https://github.com/openshift-knative/serving-operator/blob/331df80ba3d1aba7f74065b02bc982d4ae99f905/pkg/reconciler/knativeserving/knativeserving_controller.go#L197

nak3: Never mind. I misunderstood the point.

Manifestival's `ApplyAll()` already have the function (`UpdateChanged()`) to decide if apply or not. But we only want to move into `Installing` state only when we need to apply things.

markusthoemmes: Maybe for now, we should just stay in Installed=True but apply the updates anyway?

markusthoemmes: Discussed this with @jcrossley3 in Slack.

Shall we try with the following pseudo-code flow?

```
func Apply(instance *servingv1alpha1.KnativeServing, api client.Client, scheme *runtime.Scheme) error {
  manifest := NewManifest(...)
  // apply transforms
  manifest.Apply()

  if err := checkDeployments(...); err != nil {
    instance.Status.MarkDependencyInstalling("Kourier")
  }
  instance.Status.MarkDependenciesInstalled()
  return api.Status().Update(context.TODO(), instance)
}
```

Is there any downside to this? 🤔

jcrossley3: did you mean _else_ `MarkDepsInstalled()`?

markusthoemmes: yes I did, edited!

nak3:
/retest

Failed tests due to the same reason with SRVKS-458 and SRVKS-459

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
```diff
@@ -61,23 +101,37 @@ var _ reconcile.Reconciler = &ReconcileKnativeServing{}
 type ReconcileKnativeServing struct {
 	// This client, initialized using mgr.Client() above, is a split client
 	// that reads objects from the cache and writes to the apiserver
-	client client.Client
-	scheme *runtime.Scheme
+	client         client.Client
+	scheme         *runtime.Scheme
+	namespacedName *types.NamespacedName
```
markusthoemmes: What do we need this for? Seems pretty much unused below.

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
```diff
@@ -51,6 +57,40 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
 
+	resources, err := kourier.Resources(ingressNamespace)
+	if err != nil {
+		return err
+	}
+	// Watch Kourier resources.
+	for _, u := range resources {
+		g := u.GroupVersionKind()
+		gvk := schema.GroupVersionKind{Group: g.Group, Kind: g.Kind, Version: g.Version}
+		u.SetGroupVersionKind(gvk)
```
markusthoemmes: These three lines seem to do a redundant thing?

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
```diff
@@ -51,6 +57,40 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+		name := u.GetName()
+		namespace := u.GetNamespace()
+		err = c.Watch(&source.Kind{Type: &u}, &handler.EnqueueRequestsFromMapFunc{
+			ToRequests: handler.ToRequestsFunc(func(obj handler.MapObject) []reconcile.Request {
+				requests := []reconcile.Request{}
+				requests = append(requests, reconcile.Request{
+					NamespacedName: types.NamespacedName{
+						Name:      name,
+						Namespace: namespace,
```
markusthoemmes: This appends the wrong resources.

Instead, we need to append `knative-serving/knative-serving` in any case today. In theory, we need to build a relationship between the kourier resources and the KnativeServing. We could do that today by dropping the `-ingress` suffix from the namespace and then list a KnativeServing in the resulting namespace.

Today though, I think we can live with just doing this:

```golang
return []reconcile.Request{{
  NamespacedName: types.NamespacedName{
    Namespace: "knative-serving",
    Name: "knative-serving",
  }
}}
```

Potentially taking the strings as constants from somewhere.

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
```diff
@@ -51,6 +57,40 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+					},
+				})
+				return requests
+			})}, cpredicate.Funcs{
+			CreateFunc: func(event event.CreateEvent) bool { return false },
+			// It might be better to reconcile on Update event.
+			// But there are many dynamic value like Service's spec.cluterIP so we don't reconcile on Update event.
```
markusthoemmes: Is a `predicate.GenerationChangedPredicate{}` not enough to silence those?

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
markusthoemmes: Per the comments above, I think this won't be needed if we enqueue the right resources.

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
markusthoemmes: We'll also need to filter all resources aggressively. As it stands, we're watching all deployments, secrets, configmaps etc. in the cluster. We need to filter those down to just the namespace/name pairs we'd like to watch.

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
```diff
@@ -51,6 +57,40 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	resources, err := kourier.Resources(ingressNamespace)
+	if err != nil {
+		return err
+	}
+	// Watch Kourier resources.
+	for _, u := range resources {
```
markusthoemmes: I **think** we might need to do some shuffling of the data first. A `Watch` watches all resources of the specified type. I think we want to watch all the relevant types and then filter by the resources we actually know we want to watch.

To do that, I think we need to build a `map[schema.GroupVersionKind][]types.NamespacedName`, then loop over all keys, setup a watch for each of the GroupVersionKinds and inside the `ToRequests` func, make sure that the current object matches one of the namespaced names in the list of the relevant GVK.

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
nak3: Done.

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
nak3: Done.

nak3:
/hold

Still wip

nak3:
/unhold

Verified that it works on my local.

knative-operator/pkg/controller/knativeserving/kourier/kourier.go
```diff
@@ -33,28 +33,19 @@ func Apply(instance *servingv1alpha1.KnativeServing, api client.Client, scheme *
...
-	if err := api.Status().Update(context.TODO(), instance); err != nil {
-		return err
-	}
-
 	log.Info("Installing Kourier Ingress")
 	if err := manifest.Apply(); err != nil {
 		return err
 	}
 	if err := checkDeployments(&manifest, instance, api); err != nil {
+		instance.Status.MarkDependencyInstalling("Kourier")
```
markusthoemmes: Do we need to guard this further by checking if we're already in this state? I think calling this will cause us to update the "LastTransitionTimestamp" all the time, which in turn will cause update conflicts.

nak3: Sure, I think it should be guarded. Updated.

knative-operator/pkg/controller/knativeserving/kourier/kourier.go
```diff
@@ -38,8 +38,13 @@ func Apply(instance *servingv1alpha1.KnativeServing, api client.Client, scheme *
 		return err
 	}
 	if err := checkDeployments(&manifest, instance, api); err != nil {
-		instance.Status.MarkDependencyInstalling("Kourier")
 		log.Error(err, "")
+		if !instance.Status.IsFullySupported() {
```
markusthoemmes: Isn't this always true from the get-go? Are we still setting the Installing state at all now?

knative-operator/pkg/controller/knativeserving/kourier/kourier.go
markusthoemmes: I think this was nonsense from my end.

The SetConditions function underneath looks like this

```golang
// SetCondition sets or updates the Condition on Conditions for Condition.Type.
// If there is an update, Conditions are stored back sorted.
func (r conditionsImpl) SetCondition(new Condition) {
	if r.accessor == nil {
		return
	}
	t := new.Type
	var conditions Conditions
	for _, c := range r.accessor.GetConditions() {
		if c.Type != t {
			conditions = append(conditions, c)
		} else {
			// If we'd only update the LastTransitionTime, then return.
			new.LastTransitionTime = c.LastTransitionTime
			if reflect.DeepEqual(&new, &c) {
				return
			}
		}
	}
	new.LastTransitionTime = VolatileTime{Inner: metav1.NewTime(time.Now())}
	conditions = append(conditions, new)
	// Sorted for convenience of the consumer, i.e. kubectl.
	sort.Slice(conditions, func(i, j int) bool { return conditions[i].Type < conditions[j].Type })
	r.accessor.SetConditions(conditions)
}
```

Note the early exit on "If we'd only update the LastTransitionTime then return". So we're actually good without a guard. Sorry for the noise.

knative-operator/pkg/controller/knativeserving/kourier/kourier.go
markusthoemmes: See my comment above, I screwed this up.

markusthoemmes:
Logged the flakes

/retest

nak3:
/test 4.3-e2e-aws-ocp-43

Failed by TestMinScale. (Already logged.)

nak3:
/test 4.4-e2e-aws-ocp-44

Failed by TestGRPCLoadBalancing which is already logged in SRVKS-463

PR #180, opened 2020-03-16
[SRVKS-460] Watch routes to reconcile them on changes.

markusthoemmes:
Since OCP routes live in a different namespace than it's parent ingress, ownerRefs do not work. Instead, we need to build a parent/child relationship using labels "manually".


markusthoemmes:

Minor flake of the GC logic.

PR #187, opened 2020-03-17
[SRVKS-468] Support nightly versions of OCP4.3 too.

markusthoemmes:
https://issues.redhat.com/browse/SRVKS-468 for more context. This allows any prerelease versions to pass the version check as well, which is what we want.

### DIFF
--- a/olm-catalog/serverless-operator/1.6.0/serverless-operator.v1.6.0.clusterserviceversion.yaml
+++ b/olm-catalog/serverless-operator/1.6.0/serverless-operator.v1.6.0.clusterserviceversion.yaml
@@ -270,7 +270,7 @@ spec:
                     - name: OPERATOR_NAME
                       value: "knative-serving-openshift"
                     - name: MIN_OPENSHIFT_VERSION
-                      value: "4.3.0"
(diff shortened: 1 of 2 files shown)


olm-catalog/serverless-operator/1.6.0/serverless-operator.v1.6.0.clusterserviceversion.yaml
```diff
@@ -270,7 +270,7 @@ spec:
                     - name: OPERATOR_NAME
                       value: "knative-serving-openshift"
                     - name: MIN_OPENSHIFT_VERSION
-                      value: "4.3.0"
+                      value: "4.3.0-0"
```
mgencur: Are you sure this will work on normal installations (not nightly builds) ?

markusthoemmes: The issue is a semver check for a version that is bigger than or equal to what we define here. By semver's spec, a prerelease is smaller than a "proper" release. So this should just be widening the allowed range. Going to add tests though!

PR #188, opened 2020-03-17
[WIP] [SRVKS-157] Create our own ingress class that has kourier as a child.

markusthoemmes:
To be able to collapse both our ingress state (encoded in Openshift Routes) and the Kourier ingress state (as surfaced in the ingress' status) we're creating a parent-child relationship between the two, where we are in charge of the "openshift" parent and create a non-Openshift child, that is reconciled as usual. Once all routes are ready and the child is ready too, the parent will reflect readiness.


markusthoemmes:
Abandoned for now.

PR #190, opened 2020-03-18
Add techpreview-4.3 and techpreview-4.4 channels

mgencur:
* make it possible to choose a channel in scripts

I added the "techpreview" prefix to the other channels because 1.6 is not yet GA.


markusthoemmes:
The upgrade test failure looks genuine.

mgencur:
Yeah. Pushed a fix.

mgencur:
Looks like it works now:)

PR #191, opened 2020-03-18
[SRVKS-478] Update manifestival to fix transformer bug.

markusthoemmes:
Manifestival 0.3.0 had a bug that caused it to pass the wrong reference to transformers which broke transformers using the scheme.Convert kind of doing things. See https://github.com/manifestival/manifestival/issues/33.

Added a test to prevent us from hitting this with Kourier in the future.


markusthoemmes:

Setup issue.

PR #193, opened 2020-03-18
Update status of KnativeServing at the end of each reconcile cycle.

markusthoemmes:
Instead of updating the status during the reconcile cycle multiple times, we unconditionally (regardless of error or not) do it at the end of each reconcile cycle for a much easier to reason about path of execution.

### DIFF
--- a/knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
+++ b/knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
@@ -71,8 +71,8 @@ func (r *ReconcileKnativeServing) Reconcile(request reconcile.Request) (reconcil
 	reqLogger.Info("Reconciling KnativeServing")
 
 	// Fetch the KnativeServing instance
-	instance := &servingv1alpha1.KnativeServing{}
-	err := r.client.Get(context.TODO(), request.NamespacedName, instance)
+	original := &servingv1alpha1.KnativeServing{}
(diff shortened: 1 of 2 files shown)


knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
```diff
@@ -212,3 +223,20 @@ func finalizerName() string {
...
+
+func (r *ReconcileKnativeServing) updateStatus(desired *servingv1alpha1.KnativeServing) error {
+	ks := &servingv1alpha1.KnativeServing{}
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: desired.Name, Namespace: desired.Namespace}, ks)
+	if err != nil {
+		return err
+	}
+
+	// If there's nothing to update, just return.
+	if equality.Semantic.DeepEqual(ks.Status, desired.Status) {
```
jcrossley3: There's something about always doing this `DeepEqual` after always doing the same `DeepEqual` (different object) in the caller that tickles my spidey sense.

markusthoemmes: Yeah, the second one is probably not needed here.

markusthoemmes:
/retest

Minscale and weirdo i/o timeouts.

markusthoemmes:
/test 4.4-e2e-aws-ocp-44

GRPCLoadbalancing, at this point I really want 0.13.1 to come out 😂

PR #194, opened 2020-03-18
Prevent more than one KnativeServing CR in any namespace [SRVKS-415]

jcrossley3:


knative-operator/pkg/webhook/knativeserving/webhook_validating.go
```diff
@@ -64,6 +64,7 @@ func (v *KnativeServingValidator) validate(ctx context.Context, ks *servingv1alp
 	stages := []func(context.Context, *servingv1alpha1.KnativeServing) (bool, string, error){
 		v.validateNamespace,
```
markusthoemmes: Do we need to validate the name too or do we not care about the name currently and just care about it being in the "knative-serving" namespace and alone?

markusthoemmes:
/test 4.3-gcp-images

infra flake.

knative-operator/pkg/webhook/knativeserving/webhook_validating.go
markusthoemmes: Answered in Slack, this seems fine.

markusthoemmes:
Only a minscale flake, we can ignore that one. I'm fixing that with an upgrade to 0.13.1 hopefully tomorrow. I'd just like the gcp images to work here.

PR #200, opened 2020-03-23
Enable knative-serving-operator's debug log for CI

nak3:
This patch adds config-logging configmap on openshift-operators
namespace to enable the knative-serving-operator pod's debug log.


markusthoemmes:

Setup issues.

PR #202, opened 2020-03-23
Add default behavior for HA

pmorie:
Adds following behaviors:

- New `KnativeServing` resources are defaulted to HA, replicas=2 setup
- Existing `KnativeServing` resources are migrated to the same HA setup on update, if one is not specified
- If you don't want scaled up control plane, just use `KnativeServing.Spec.HighAvailability.Replicas=1`


markusthoemmes:
The errors look genuine. Looks like a feedback loop between ours and the upstream operator.

PR #203, opened 2020-03-24
[SRVKS-484] Make knative-operator and ingress log with readable timestamps.

markusthoemmes:
- **knative-operator**:
  - Force set the zap-time-encoding flag.
- **serving/ingress**:
  - Force set the zap-time-encoding flag.
  - Adjust the "inner" logger to read Knative's default config, which has the correct time encoding set.
  - Add a "key" to each log in the reconciler to allow for easier debuggability.


markusthoemmes:
/retest

Infra issue.

PR #205, opened 2020-03-24
Upgrade controller-runtime and operator-sdk for our ingress.

markusthoemmes:
The versions we use are fairly old and the upgrade brings us the added benefit of fake-implementation improvements so we can finally get rid of our custom label filtering.

This also surfaced a bug in our unit tests where we asserted that we delete more than we should. Narrowed that down here.


markusthoemmes:
Logged the flake in https://issues.redhat.com/browse/SRVKS-486

PR #206, opened 2020-03-24
Build all of our components with Golang 1.13.

markusthoemmes:
As per title. We already do that in productization AFAIK so CI should so too.


warrenvw:
As mentioned above, we are using go 1.13 (specifically 1.13.4) for productization w/o issues..

PR #215, opened 2020-03-25
Bootstrap the 1.7.0 release.

markusthoemmes:
- Added a CSV for 1.7.0.
- Adjusted the upgrade test because we're not inherently shipping a Knative version upgrade.

### DIFF
--- a/olm-catalog/serverless-operator/serverless-operator.package.yaml
+++ b/olm-catalog/serverless-operator/serverless-operator.package.yaml
@@ -4,4 +4,4 @@ channels:
 - name: techpreview
   currentCSV: serverless-operator.v1.5.0
 - name: preview-4.3
-  currentCSV: serverless-operator.v1.6.0
+  currentCSV: serverless-operator.v1.7.0
--- a/test/lib.bash
+++ b/test/lib.bash
@@ -182,9 +182,6 @@ function run_knative_serving_rolling_upgrade_tests {
   PROBER_PID=$!
 
   if [[ $UPGRADE_SERVERLESS == true ]]; then
-    local serving_version
-    serving_version=$(oc get knativeserving.operator.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.version}")
-
     # Get latest CSV from the given channel
     local upgrade_to
     upgrade_to=$("${rootdir}/hack/catalog.sh" | sed -n '/channels/,$p;' | sed -n "/- name: ${CHANNEL}$/{n;p;}" | awk '{ print $2 }')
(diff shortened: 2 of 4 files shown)


olm-catalog/serverless-operator/serverless-operator.package.yaml
```diff
@@ -4,4 +4,4 @@ channels:
 - name: techpreview
   currentCSV: serverless-operator.v1.5.0
 - name: preview-4.3
```
aliok: I didn't really follow the discussions around channels recently, but do we need a GA channel here?

markusthoemmes: In a separate PR, yes. It's a JIRA on its own.

test/lib.bash
```diff
@@ -198,14 +195,9 @@ function run_knative_serving_rolling_upgrade_tests {
...
-      # Check KnativeServing still has the old version
-      [[ $(oc get knativeserving.operator.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.version}") == "$serving_version" ]] || return 1
     else
       approve_csv "$upgrade_to" || return 1
-      timeout 900 '[[ ! ( $(oc get knativeserving.operator.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.version}") != $serving_version && $(oc get knativeserving.operator.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.conditions[?(@.type==\"Ready\")].status}") == True ) ]]' || return 1
-
-      # Assert that the old image references eventually fade away
-      timeout 900 "oc get pod -n $SERVING_NAMESPACE -o yaml | grep image: | uniq | grep $serving_version" || return 1
+      timeout 900 '[[ ! ( $(oc get knativeserving.operator.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.conditions[?(@.type==\"Ready\")].status}") == True ) ]]' || return 1
```
mgencur: I fear that this will return immediately even before the operator upgrades will start. AFAIK, at the beginning it returns True, later it returns False, and then True again when the operators are updated. Not sure how to check that operator images were updated

markusthoemmes: This should pass fine for now as the upgrade is basically a noop. I agree we might need to enhance this a bit to not rely on a version upgrade as much. Shall we do that in a follow up?

markusthoemmes: Regarding the check:  We could poll the openshift-operators namespace for "1.6.0" based images and only go through once they have vanished?

mgencur: Yeah. We can do it in a follow up. The operator images are updated after we approve the installplan, right? In the approve_csv function we wait for the new CSV to be available. Maybe that's enough to check that we have new operators available. But we could also do the check as you suggested.

markusthoemmes:
Flake logged in https://issues.redhat.com/browse/SRVKS-488. This looks like a cheeky bug where we're creating more revisions than we anticipate.

/retest

markusthoemmes:
All these tests passed thrice now. Merging.

PR #222, opened 2020-03-27
Provide options to specify CHANNEL and SOURCE for upgrades

mgencur:
Upgrades might need switching channel (and for downstream testing also source from redhat-operators to serverless-operator).


mgencur:
Rebased.

PR #224, opened 2020-03-30
[SRVKS-444] Use different injection mechanism to get proxy certs as well.

markusthoemmes:
Per https://github.com/openshift/enhancements/blob/0d53973f356b4a31bef86bf015a24d06f27bae90/enhancements/proxy/global-cluster-egress-proxy.md this annotation seems to do exactly what we need in that the operator merges the cluster CA with the user provided CA into one bundle and uses that here.


markusthoemmes:
/retest

Upgrade test was a setup issue
E2E test was TestGRPCLoadbalancing

PR #231, opened 2020-04-02
[SRVKS-505] Verify the default is HA

mgencur:
I'm updating the knative-serving dependency to "master" similar to what is in knative-operator's Gopkg.toml


mgencur:
/retest
Unrelated flake

test/e2e/knative_serving_test.go
```diff
@@ -42,9 +43,13 @@ func TestKnativeServing(t *testing.T) {
 	})
 
 	t.Run("deploy knativeserving cr and wait for it to be ready", func(t *testing.T) {
-		if _, err := v1a1test.WithKnativeServingReady(caCtx, knativeServing, knativeServing); err != nil {
+		ks, err := v1a1test.WithKnativeServingReady(caCtx, knativeServing, knativeServing)
+		if err != nil {
 			t.Fatal("Failed to deploy KnativeServing", err)
 		}
+		if ks.Spec.HighAvailability == nil || ks.Spec.HighAvailability.Replicas != defaultHAReplicas {
```
markusthoemmes, line 50: Should we instead adjust the "verify correct deployment shape" test below and actually check that the respective deployments have the correct replicas set?

mgencur: I'm not sure. We'll need to test different deployments specifically (i.e. right now only autoscaler-hpa and controller, later also add activator). I don't have a preference here.

markusthoemmes: I'd rather we properly end-to-end test this. The test we have already is kinda already covered by unit tests anyway 🤔 .

mgencur: OK.

PR #234, opened 2020-04-02
Rename knative-serving-openshift -> knative-openshift

jcrossley3:
Now that eventing is a part of this operator, the name is confusing

### DIFF
--- a/knative-operator/Dockerfile
+++ b/knative-operator/Dockerfile
@@ -9,9 +9,9 @@ RUN go build -o /usr/bin/manager ./cmd/manager
 
 FROM openshift/origin-base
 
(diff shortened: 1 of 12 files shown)


jcrossley3:
> Why not `knative-operator` everywhere? Would make it consistent with the folder name

I considered it, and I agree the folder name is a problem. I like that 'openshift' is a part of the name, because everything the operator does is specific to OpenShift: it's most definitely an "openshift-specific operator". Is renaming the folder to 'knative-openshift' too stupid to bear?

markusthoemmes:
Let's roll with this then and think about renaming the folder separately.

markusthoemmes:
/retest

Failures due to git struggles.

PR #238, opened 2020-04-08
Make it possible to choose different repos in tests

mgencur:
Fixes: https://issues.redhat.com/browse/SRVCOM-753 
It is important to have this parameterized. This way we can re-use the scripts in internal Jenkins for running tests from knative-serving and serving-operator repos (either from github or internal dist-git repos).

### DIFF
--- a/hack/lib/vars.bash
+++ b/hack/lib/vars.bash
@@ -6,6 +6,10 @@ readonly BUILD_NUMBER=${BUILD_NUMBER:-$(uuidgen)}
 source "$(dirname "${BASH_SOURCE[0]}")/../../test/vendor/github.com/knative/test-infra/scripts/e2e-tests.sh"
 
 readonly KNATIVE_VERSION="${KNATIVE_VERSION:-v0.13.2}"
+readonly KNATIVE_SERVING_REPO="${KNATIVE_SERVING_REPO:-"https://github.com/openshift/knative-serving.git"}"
+readonly KNATIVE_SERVING_VERSION="${KNATIVE_SERVING_VERSION:-}"
+readonly KNATIVE_SERVING_OPERATOR_REPO="${KNATIVE_SERVING_OPERATOR_REPO:-"https://github.com/openshift-knative/serving-operator.git"}"
+readonly KNATIVE_SERVING_OPERATOR_VERSION="${KNATIVE_SERVING_OPERATOR_VERSION:-}"
 
 readonly CATALOG_SOURCE_FILENAME="${CATALOG_SOURCE_FILENAME:-catalogsource-ci.yaml}"
 readonly DOCKER_REPO_OVERRIDE="${DOCKER_REPO_OVERRIDE:-}"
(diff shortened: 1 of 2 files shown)


hack/lib/vars.bash
```diff
@@ -6,6 +6,10 @@ readonly BUILD_NUMBER=${BUILD_NUMBER:-$(uuidgen)}
 source "$(dirname "${BASH_SOURCE[0]}")/../../test/vendor/github.com/knative/test-infra/scripts/e2e-tests.sh"
 
 readonly KNATIVE_VERSION="${KNATIVE_VERSION:-v0.13.2}"
```
cardil, line 8: Let's remove that parameter. It might be the same for serving & serving-operator, but not necessary. Also, for eventing & eventing operator.

Or, maybe us it as a default in:

```bash
eadonly KNATIVE_SERVING_VERSION="${KNATIVE_SERVING_VERSION:-${KNATIVE_VERSION}}"
```

mgencur: Remove that parameter? It's used in two places but they use a different prefix for branches (`release-` and `openshift-`).

mgencur: The second option won't work because we need those different prefixes in different places.

cardil: What about:
```bash
readonly KNATIVE_VERSION="${KNATIVE_VERSION:-v0.13.2}"
readonly KNATIVE_SERVING_VERSION="${KNATIVE_SERVING_VERSION:-${KNATIVE_VERSION}}"
readonly KNATIVE_SERVING_REF="${KNATIVE_SERVING_REF:-release-${KNATIVE_SERVING_VERSION}}"
readonly KNATIVE_SERVING_OPERATOR_VERSION="${KNATIVE_SERVING_OPERATOR_VERSION:-${KNATIVE_VERSION}}"
readonly KNATIVE_SERVING_OPERATOR_REF="${KNATIVE_SERVING_OPERATOR_REF:-openshift-${KNATIVE_SERVING_OPERATOR_VERSION}}"
```

Then I'd add similar lines for eventing.

cardil: And also, remove of parameters from `run_knative_serving_e2e_and_conformance_tests` and `run_knative_serving_operator_tests` functions.

cardil: Also, we can merge it as it is. And, I will do those changes in my upcoming PR.

mgencur: I'm not sure. I'd probably prefer the minimum number of global variables. This solution works as well. It just moves some stuff from local functions to global vars.

cardil: That would be required as the versions for serving and eventing are different. Also, if using other repos, that may need to use other ref then: `openshift-${KNATIVE_VERSION}` - this might be completely diffrent.

mgencur:
The failure looks unrelated to the changes.

/retest

aliok:
/retest

Throttling issues

PR #241, opened 2020-04-09
[SRVKS-513] Manually combine configmaps for service-ca and trusted-ca.

markusthoemmes:
My assumption about being able to use both the serviceCA annotation and the trustedCA label on the same configMap have been proven wrong. This pulls their management apart into separate configMaps that are watched by the reconciler. The contents of both configMaps are unified into a single configMap, which is then mounted into the controller via the operator's mechanism.

**Note:** This retains "stomping" behavior from master. We should maybe think about if that's the right thing or if the user should be able to add even more certs to the configmap if needed. I'd like to tackle that separately though.

### DIFF
--- a/knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
+++ b/knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
@@ -234,11 +234,9 @@ func (r *ReconcileKnativeServing) ensureFinalizers(instance *servingv1alpha1.Kna
 func (r *ReconcileKnativeServing) ensureCustomCertsConfigMap(instance *servingv1alpha1.KnativeServing) error {
 	const (
 		// Docs: https://github.com/openshift/service-ca-operator
-		serviceCAKey     = "service.alpha.openshift.io/inject-cabundle"
-		serviceCAContent = "service-ca.crt"
+		serviceCAKey = "service.alpha.openshift.io/inject-cabundle"
 		// Docs: https://docs.openshift.com/container-platform/4.3/networking/configuring-a-custom-pki.html#certificate-injection-using-operators_configuring-a-custom-pki
-		trustedCAKey     = "config.openshift.io/inject-trusted-cabundle"
-		trustedCAContent = "tls-ca-bundle.pem"
+		trustedCAKey = "config.openshift.io/inject-trusted-cabundle"
 	)
 
 	certs := instance.Spec.ControllerCustomCerts
--- a/knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
+++ b/knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
@@ -242,6 +242,10 @@ func (r *ReconcileKnativeServing) ensureCustomCertsConfigMap(instance *servingv1
 	)
 
 	certs := instance.Spec.ControllerCustomCerts
+
+	// If the user doesn't specify anything else, this is set by the webhook/controller defaulter to
+	// cause us to automatically pull in the relevant ConfigMaps from the cluster. The user needs
+	// to specifically opt-out of this today by specifying an empty Name and ConfigMap explicitly.
 	if certs.Type != "ConfigMap" || certs.Name == "" {
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
```diff
@@ -224,49 +233,97 @@ func (r *ReconcileKnativeServing) ensureFinalizers(instance *servingv1alpha1.Kna
...
 func (r *ReconcileKnativeServing) ensureCustomCertsConfigMap(instance *servingv1alpha1.KnativeServing) error {
 	const (
-		// Using both the serviceCA annotation and the trustedCA label is safe. The serviceCA reconciler
-		// will only reconcile once and then short-circuit on an existing key, see https://git.io/JvNSr.
-		// The trustedCA label reconciles in an additive way, see https://git.io/JvNSV.
-
 		// Docs: https://github.com/openshift/service-ca-operator
-		serviceCAKey = "service.alpha.openshift.io/inject-cabundle"
+		serviceCAKey     = "service.alpha.openshift.io/inject-cabundle"
+		serviceCAContent = "service-ca.crt"
```
bbrowning, line 238: Where do we use this?

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
```diff
@@ -224,49 +233,97 @@ func (r *ReconcileKnativeServing) ensureFinalizers(instance *servingv1alpha1.Kna
...
 		// Docs: https://docs.openshift.com/container-platform/4.3/networking/configuring-a-custom-pki.html#certificate-injection-using-operators_configuring-a-custom-pki
-		trustedCAKey = "config.openshift.io/inject-trusted-cabundle"
+		trustedCAKey     = "config.openshift.io/inject-trusted-cabundle"
+		trustedCAContent = "tls-ca-bundle.pem"
```
bbrowning, line 241: And this?

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
```diff
@@ -224,49 +233,97 @@ func (r *ReconcileKnativeServing) ensureFinalizers(instance *servingv1alpha1.Kna
...
+	serviceCACM, err := r.reconcileConfigMap(instance, certs.Name+"-service-ca", map[string]string{serviceCAKey: "true"}, nil, nil)
+	if err != nil {
+		return fmt.Errorf("error reconciling serviceCACM: %w", err)
+	}
+	trustedCACM, err := r.reconcileConfigMap(instance, certs.Name+"-trusted-ca", nil, map[string]string{trustedCAKey: "true"}, nil)
+	if err != nil {
+		return fmt.Errorf("error reconciling serviceCACM: %w", err)
+	}
+
+	combinedContents := make(map[string]string, len(serviceCACM.Data)+len(trustedCACM.Data))
```
bbrowning, line 262: Just so I understand the flow here, the first time this happens we'll effectively create an empty combined ConfigMap.  And as the underlying ConfigMaps get their certs set by OpenShift, we'll then start combining the ConfigMaps. Does the Knative Serving controller automatically watch for and pick up any changes to the certs in its mounted ConfigMap volume?

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
markusthoemmes: Good point, not used anymore.

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
markusthoemmes: same here.

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
markusthoemmes: That's correct and no the controller actually does not watch this currently. I was poking upstream about the possibility to do that.

FWIW: This is a hole on master as well. The timing usually comes around nicely for us so far. I did some refactoring today too though to implement a "kick" from our operator. I plan to set an env var on the deployment with the resourceVersion of the configMap effectively. That should also give us cert rotation in general.

bbrowning: Let's open a JIRA issue if there's not one already to track the kick/env var/cert rotation work. Given that, this looks good.

markusthoemmes: wdm!

PR #242, opened 2020-04-09
Generalize environment variable setting to deployments.

markusthoemmes:
This generalizes the helper to update the proxy environment variables of the controller to be able to update arbitrary environment variables of arbitrary deployments.
It also adds a check to only run the respective API update if we even applied a change.

### DIFF
--- a/knative-operator/pkg/common/deployment_test.go
+++ b/knative-operator/pkg/common/deployment_test.go
@@ -88,7 +88,7 @@ func TestApplyEnvironmentToDeployment(t *testing.T) {
 					t.Fatalf("Deployment.Get = %v, want no error", err)
 				}
 				if !equality.Semantic.DeepEqual(test.expect, got) {
-					t.Fatalf("Deployment wasn't wat we expected: %#v, want %#v", got, test.expect)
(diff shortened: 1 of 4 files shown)


knative-operator/pkg/common/deployment_test.go
```diff
@@ -1,160 +1,96 @@
...
+				t.Fatalf("ApplyEnvironmentToDeployment = %v, want no error", err)
+			}
+
+			if test.expect != nil {
+				got := &appsv1.Deployment{}
+				if err := c.Get(context.TODO(), client.ObjectKey{Name: test.expect.Name, Namespace: test.expect.Namespace}, got); err != nil {
+					t.Fatalf("Deployment.Get = %v, want no error", err)
+				}
+				if !equality.Semantic.DeepEqual(test.expect, got) {
+					t.Fatalf("Deployment wasn't wat we expected: %#v, want %#v", got, test.expect)
```
savitaashture, line 91:
```suggestion
					t.Fatalf("Deployment wasn't what we expected: %#v, want %#v", got, test.expect)
```

PR #245, opened 2020-04-09
Sort environment variables to make unit test stable.

markusthoemmes:
I recently introduced a flaky unit test. Because of map semantics in the test, the order in which the env vars are added is non-deterministic.

Sorting them makes them deterministic without affecting what we want to test.

See error here https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift-knative_serverless-operator/241/pull-ci-openshift-knative-serverless-operator-master-unit-test/570

### DIFF
--- a/knative-operator/pkg/common/deployment_test.go
+++ b/knative-operator/pkg/common/deployment_test.go
@@ -2,6 +2,7 @@ package common_test
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


jmprusi:
/retest 

Setup issue

PR #246, opened 2020-04-09
Remove the spec.config block from KnativeServing example

bbrowning:
Don't give an example using `spec.config` since it's not something we plan to support users munging for the initial GA release.


jcrossley3:
I'm curious: are you going to support user edits to `knative-serving` ConfigMaps?

bbrowning:
No, we should not.

bbrowning:
If we want to support either spec.config or editing of configmaps, then we need to document all the fields, allowed values, which ones our operator default and/or override, and so on. It's a big can of worms and until we're ready to address these things, we can't support usage of this field or the configmaps.

markusthoemmes:
I'm okay merging this as we don't document any of it yet (which is okay imo too). Any objections @jcrossley3?

jcrossley3:
I think it's weird not supporting knative configuration, but I have no real objections...
/lgtm

PR #254, opened 2020-04-15
[WIP] Add a test that verifies HTTPS works.

markusthoemmes:
This is just for showcasing it works. @nak3 shall pick the commit into his PR once we think it's good.


markusthoemmes:
Hm, I suppose this test would pass today as well? 🤔 I guess we'll need to explicitly trust the actual certificate. I'll poke around to try to find out how to get that.

markusthoemmes:
This is done.

PR #255, opened 2020-04-15
Test route conflict behavior.

markusthoemmes:
As per SRVKS-524.


knative-operator/pkg/common/serving.go
```diff
@@ -12,8 +12,18 @@ import (
 
 var log = Log
 
+const (
+	// defaultDomainTemplate is a value for domainTemplate in config-network.
+	// As Knative on OpnShift uses OpenShift's wildcard cert the domain level must have "<sub>.domain", not "<sub1>.<sub2>.domain".
```
bbrowning, line 17: nit: OpnShift -> OpenShift but don't hold up merge for this unless other changes are needed

bbrowning:
I manually verified that `testKnativeServiceHTTPS` fails when run using the old route template to ensure the cert checking was validating names as expected even when we're manually adding to the rootCAs. So, a positive test from CI should give us confidence that this test confirms the updated route template makes the default OpenShift wildcart cert work.

markusthoemmes:
Unrelated issue, merging.

PR #256, opened 2020-04-15
Remove trailing whitespace in description

warrenvw:
Title says it all...


jcrossley3:
Our CI is almost as good as not having CI at all! ;)

PR #257, opened 2020-04-15
CSV updates for GA

jcrossley3:


olm-catalog/serverless-operator/1.7.0/serverless-operator.v1.7.0.clusterserviceversion.yaml
```diff
@@ -23,11 +23,10 @@ metadata:
           }
         }
       ]
-    capabilities: Seamless Upgrades
+    capabilities: Full Lifecycle
     categories: Networking,Integration & Delivery,Cloud Provider,Developer Tools
     certified: "false"
-    containerImage: quay.io/openshift-knative/serverless-operator:v1.7.0
-    createdAt: "2019-07-27T17:00:00Z"
+    createdAt: "2020-04-20T17:00:00Z"
```
warrenvw, line 29: I'm curious why we are changing `createdAt` and removing `containerImage`? I thought `createdAt` was the date the operator was created, and `containerImage` was displayed in the UI. Is the latter not required anymore?

jcrossley3: The image doesn't exist, so it seemed better to remove the field entirely. If you can give me a path to an accurate image, I'm happy to put it back.

I'm not actually sure which date `createdAt` should refer to. The docs aren't clear. But it seemed to me the date the specific version is released would be more useful to the user than the date the first version was released. I don't feel that strongly about it, though.

pmorie:
Hrm. The status visible in github test UI does not seem to match the status of the actual prow job.

Example:

- Github UI: ci/prow/4.4-e2e-aws-ocp-44 — Job cannot be started: pods "359b336a-8018-11ea-b463-0a58ac102063" already exists 
- Prow build shows _failed_, but the logs seem to indicate success: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift-knative_serverless-operator/257/pull-ci-openshift-knative-serverless-operator-master-4.4-e2e-aws-ocp-44/470

pmorie:
Closed and reopened to see if that has an effect.

olm-catalog/serverless-operator/1.7.0/serverless-operator.v1.7.0.clusterserviceversion.yaml
markusthoemmes: Have we checked if this has any impact on the UI and such (referring to that non-existing image). I also don't know the significance of it (how big can that be given the image doesn't even exist) but I'm not super keen on taking OLM gambles here, given it worked in the past.

Can we doublecheck this with the OLM team please?

jcrossley3: @dmesser thoughts?

markusthoemmes:
@warrenvw checked with the wider OLM team and we think this is fine. We can revert if not.

olm-catalog/serverless-operator/1.7.0/serverless-operator.v1.7.0.clusterserviceversion.yaml
warrenvw: I verified with Lance Galletti that `containerImage` is only used in the OperatorHub UI. It is not a field that drives any functionality. He verified this with the larger OLM team. We should be safe to remove it.

PR #258, opened 2020-04-15
Fix approve_csv to actually use the oc replace command

mgencur:
The `oc replace` command is supposed to be at the end of the pipe right after the sed command but the `|` was missing there. As a result, the following error was thrown in all upgrade tests and the changes were not applied:
```
+ approve_csv serverless-operator.v1.6.0 4.3
+ local csv_version install_plan channel
+ csv_version=serverless-operator.v1.6.0
+ channel=4.3
+ oc get subscription serverless-operator -n openshift-operators -oyaml
+ sed -e 's/\(.*channel:\).*/\1 4.3/' -e 's/\(.*source:\).*/\1 serverless-operator/' oc replace -f -
sed: file - line 2: unknown command: `k'
```


markusthoemmes:
Which issue does this actually fix? A short description would be great.

mgencur:
@markusthoemmes sorry, added description.

PR #262, opened 2020-04-17
[SRVKS-470 SRVKS-472 SRVKS-473] Enable HA test suite execution

mgencur:
This is to enable the test execution in the serverless-operator repo. After merging this PR we can mark the JIRAs as Done.

### DIFF
--- a/test/serving.bash
+++ b/test/serving.bash
@@ -71,6 +71,17 @@ function run_knative_serving_e2e_and_conformance_tests {
   go_test_e2e -tags=e2e -timeout=30m ./test/e2e -run "^(TestHelloWorld)$" \
     --resolvabledomain --kubeconfig "$KUBECONFIG" \
     --imagetemplate "image-registry.openshift-image-registry.svc:5000/serving-tests/{{.Name}}" || failed=2
+  
+  # Prevent HPA from scaling to make HA tests more stable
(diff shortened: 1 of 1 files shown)


mgencur:
Rebased.

test/serving.bash
```diff
@@ -71,6 +71,17 @@ function run_knative_serving_e2e_and_conformance_tests {
   go_test_e2e -tags=e2e -timeout=30m ./test/e2e -run "^(TestHelloWorld)$" \
     --resolvabledomain --kubeconfig "$KUBECONFIG" \
     --imagetemplate "image-registry.openshift-image-registry.svc:5000/serving-tests/{{.Name}}" || failed=2
+  
+  # Prevent HPA from scaling to make HA tests more stable
+  oc -n "$SERVING_NAMESPACE" patch hpa activator --patch '{"spec":{"maxReplicas":2}}' || failed=3
+
+  # Use sed as the -spoofinterval parameter is not available yet
+  sed "s/\(.*requestInterval =\).*/\1 10 * time.Millisecond/" -i test/vendor/knative.dev/pkg/test/spoof/spoof.go
```
cardil, line 79: I would add a restore of both of those changes as they are interrupting other tests that might run afterwards, and can introduce false positives.

markusthoemmes: Going to ignore for now as the activator change won't impact eventing tests and the in-code change is rolled "back" as we're removing the code after this test anyway.

PR #178, opened 2020-03-16
Add kn ConsoleCLIDownload for disconnected install

navidshaikh:
- Add kn-downloads deployment, service and route using kn-cli-artifacts
 image.
 - Create namespace `serverless-operator` and deploy mentioned resources in it.
 - The deployment runs a container namely; kn-downloads-server. This container
 runs a simple http server using python2, serving the kn cross platform
 artifacts in tar.gz or zip format.
 - Paths for respective platform downloads:
    - /amd64/linux/kn-linux-amd64.tar.gz
    - /amd64/linux/kn-macos-amd64.tar.gz
    - /amd64/linux/kn-windows-amd64.zip
- Removed kn ConsoleCLIDownload manifest, now populates it in the code.

PR #264, opened 2020-04-17
Bump deps of knative-serving to 0.13, openshift to 4.3 and use servingv1 API.

markusthoemmes:
As per title, this is long overdue and broken out of #178 to make that PR not overwhelmed with dep updates.


markusthoemmes:
Done.

PR #276, opened 2020-04-21
Restore the original maxReplicas for activator HPA

mgencur:
The Eventing test suite runs after the Serving test suite. Let's use the default settings there.

### DIFF
--- a/test/serving.bash
+++ b/test/serving.bash
@@ -73,6 +73,8 @@ function run_knative_serving_e2e_and_conformance_tests {
     --imagetemplate "image-registry.openshift-image-registry.svc:5000/serving-tests/{{.Name}}" || failed=2
   
   # Prevent HPA from scaling to make HA tests more stable
+  local max_replicas
+  max_replicas=$(oc get hpa activator -n "$SERVING_NAMESPACE" -ojsonpath='{.spec.maxReplicas}')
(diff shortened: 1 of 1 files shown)


test/serving.bash
```diff
@@ -73,6 +73,7 @@ function run_knative_serving_e2e_and_conformance_tests {
     --imagetemplate "image-registry.openshift-image-registry.svc:5000/serving-tests/{{.Name}}" || failed=2
   
   # Prevent HPA from scaling to make HA tests more stable
+  local max_replicas=$(oc get hpa activator -n "$SERVING_NAMESPACE" -ojsonpath='{.spec.maxReplicas}')
```
cardil, line 76: Define local separately from invocation - shellcheck will catch this :stuck_out_tongue: . Now, result will be okay even if `oc get ..` fails.

cardil: https://github.com/koalaman/shellcheck/wiki/SC2155

mgencur: We don't do that in this file in other places except for one. But ok

cardil: We should do it everywhere as it is dangerous.

Btw. `local abc=0` isn't a problem, as well as `local abc="${other:-defaults}"`. Dangerous is only `local abc=$(command-that-might-fail)`

markusthoemmes:
@mgencur 4.5 is still broken and won't become green. Want me to just merge?

markusthoemmes:
I stand corrected! The fix is supposed to have been merged, let's see if this passes then!

PR #280, opened 2020-04-22
Change value keys to image under spec.relatedImages

warrenvw:
Resolve https://issues.redhat.com/browse/SRVKS-544.

Change all the `value:` keys to `image:` under the `spec.relatedImages` field.


bbrowning:
/lgtm

Thanks!

PR #287, opened 2020-04-28
Image override with deployment name

aliok:
Support the image override with deployment names : https://github.com/knative/eventing-operator/pull/159

Following env var
```
- name: IMAGE_imc-controller_controller
  value: registry.svc.ci.openshift.org/openshift/knative-v0.14.0:knative-eventing-channel-controller
```

will result in 
```
spec:
  registry:
    override:
      imc-controller/controller: >-
        registry.svc.ci.openshift.org/openshift/knative-v0.14.0:knative-eventing-channel-controller
```

Old image override behavior is not changed.

### DIFF
--- a/knative-operator/pkg/common/util.go
+++ b/knative-operator/pkg/common/util.go
@@ -1,8 +1,8 @@
 package common
 
 import (
+	"fmt"
 	servingv1alpha1 "knative.dev/serving-operator/pkg/apis/serving/v1alpha1"
-	"os"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 	"strings"
 )
@@ -34,14 +34,27 @@ func IngressNamespace(servingNamespace string) string {
 	return servingNamespace + "-ingress"
 }
 
-// buildImageOverrideMapFromEnviron creates a map to overrides registry images
-func buildImageOverrideMapFromEnviron() map[string]string {
+// BuildImageOverrideMapFromEnviron creates a map to overrides registry images
+func BuildImageOverrideMapFromEnviron(environ []string) map[string]string {
 	overrideMap := map[string]string{}
 
-	for _, e := range os.Environ() {
+	for _, e := range environ {
 		pair := strings.SplitN(e, "=", 2)
 		if strings.HasPrefix(pair[0], "IMAGE_") {
-			name := strings.SplitN(pair[0], "_", 2)[1]
+			// convert
+			// "IMAGE_container=docker.io/foo"
+			// "IMAGE_deployment_container=docker.io/foo2"
+			// to
+			// container: docker.io/foo
+			// deployment/container: docker.io/foo2
+			var name string
+			parts := strings.SplitN(pair[0], "_", -1)
+			if len(parts) == 3 {
+				name = fmt.Sprintf("%s/%s", parts[1], parts[2])
+			} else {
+				name = parts[1]
+			}
+
 			if pair[1] != "" {
 				overrideMap[name] = pair[1]
 			}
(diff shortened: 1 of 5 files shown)


aliok:
We need this change since we don't want to change the container names in our fork. 

There were some duplicate container names, and https://github.com/knative/eventing-operator/pull/159 provided a mechanism to override those images using a deployment name.

We now needed to adapt our way of passing these overrides from CSV to serverless operator and then to upstream operator.

knative-operator/pkg/common/util.go
```diff
@@ -42,6 +42,11 @@ func buildImageOverrideMapFromEnviron() map[string]string {
 		pair := strings.SplitN(e, "=", 2)
 		if strings.HasPrefix(pair[0], "IMAGE_") {
 			name := strings.SplitN(pair[0], "_", 2)[1]
+			// convert
+			// "IMAGE_eventing-controller__eventing-controller=docker.io/foo2"
+			// to
+			// eventing-controller/eventing-controller: docker.io/foo2
+			name = strings.Replace(name, "__", "/", 1)
```
jcrossley3, line 49: I'm wondering whether you need `__` at all. If we split on a single underscore, then I'm either going to have `IMAGE_<container>` or `IMAGE_<deployment>_<container>` right? So why do I need the double underscore? I know what I have based on how many fields I get after the split. Or am I missing something?

aliok: @jcrossley3 no you're not missing anything. I wanted to have an obvious separator there.
Let me change it to a single `_`.

jcrossley3:
@aliok can you rebase this to squash the redundant commits and remove the double underscore references in the description, please?

knative-operator/pkg/common/util.go
```diff
@@ -34,14 +34,30 @@ func IngressNamespace(servingNamespace string) string {
...
+			// to
+			// fo0: docker.io/foo
+			// eventing-controller/eventing-controller: docker.io/foo2
+			suffixParts := strings.SplitN(suffix, "_", 2)
+			if len(suffixParts) == 2 {
+				name = fmt.Sprintf("%s/%s", suffixParts[0], suffixParts[1])
+			} else {
+				name = fmt.Sprintf("%s", suffix)
+			}
+
```
jcrossley3, line 60: How about something along these lines? To me using a negative for the count is more clear -- and we really should consider more than 2 underscores an error, since they're illegal in container names, I think. The comments aren't especially enlightening because you use the same name for both deployment and container -- I can't tell which is which! :)
```suggestion
			// convert
			// "IMAGE_container=docker.io/foo"
			// "IMAGE_deployment_container=docker.io/foo2"
			// to
			// container: docker.io/foo
			// deployment/container: docker.io/foo2
			var name string
			parts := strings.SplitN(pair[0], "_", -1)
			if len(parts) == 3 {
				name = fmt.Sprintf("%s/%s", parts[1], parts[2])
			} else {
				name = parts[1]
			}

```

markusthoemmes:
/retest

To see what happens on 4.5 now.

knative-operator/pkg/common/util_test.go
```diff
@@ -1,12 +1,100 @@
...
+				"bar":     "quay.io/myimage2",
+			},
+		},
+		{
+			name: "3 underscores",
+			envMap: map[string]string{
+				"IMAGE_foo_bar_baz": "quay.io/myimage",
+			},
+			expected: map[string]string{
+				"foo/bar_baz": "quay.io/myimage",
```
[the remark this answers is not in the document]
aliok: Ok, removing that test and leaving that validation to Kubernetes

PR #291, opened 2020-05-08
Remove unnecessary openshift/console-operator dep

jcrossley3:
After running `dep ensure`

Not sure why these are here.


navidshaikh:
lgtm

PR #289, opened 2020-05-07
Take 2: [WIP] CSV 1.8 and eventing 0.14

aliok:
1.8 CSV with eventing 0.14

* Do `make images dev`
* Go to OpenShift console, install serverless operator
* Install eventing (in knative-eventing namespace)

Image override for eventing 0.14 upgrade job work work. See
- Sandbox operator PR: https://github.com/knative-sandbox/operator/issues/76
- knative/eventing-operator PR: https://github.com/openshift-knative/eventing-operator/pull/27

PR #295, opened 2020-05-12
Fix image overrides for env vars that contain underscores

aliok:
Changes:
- Use double underscore to separate the deployment name from the container names. Otherwise the env vars we need to override with an underscore in its name is not really overridden.
- Added a test that aims to fail if there's any image override mistakes. After the deployments and other things are created, the test checks if there's any refs to `gcr.io` in container images, container env vars.

To see the test fails, use these overrides in CSV:
```
- name: IMAGE_eventing-controller
                      value: gcr.io/knative-releases/knative.dev/eventing/cmd/controller@sha256:c408dc0f092d31ef1169daa5541be2763f6d41d775c565329fd53b6babac8ba8
- name: IMAGE_CRONJOB_RA_IMAGE
                      value: gcr.io/knative-releases/knative.dev/eventing/cmd/cronjob_receive_adapter@sha256:206733e8b5fdfc8d775d8231cfd23835095967125d080881c6fea4e5aebe2955
```

cc @matzew --> I am gonna have to update my WIP PR [0](https://github.com/openshift-knative/serverless-operator/pull/289) for CSV 0.18 and eventing 0.14 once this is merged.

### DIFF
--- a/knative-operator/pkg/common/util.go
+++ b/knative-operator/pkg/common/util.go
@@ -1,13 +1,14 @@
 package common
 
 import (
-	"fmt"
 	servingv1alpha1 "knative.dev/serving-operator/pkg/apis/serving/v1alpha1"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 	"strings"
 )
 
 var Log = logf.Log.WithName("knative").WithName("openshift")
+
+const ImagePrefix = "IMAGE_"
 
 // Configure is a  helper to set a value for a key, potentially overriding existing contents.
 func Configure(ks *servingv1alpha1.KnativeServing, cm, key, value string) bool {
@@ -40,21 +41,19 @@ func BuildImageOverrideMapFromEnviron(environ []string) map[string]string {
 
 	for _, e := range environ {
 		pair := strings.SplitN(e, "=", 2)
-		if strings.HasPrefix(pair[0], "IMAGE_") {
+		if strings.HasPrefix(pair[0], ImagePrefix) {
 			// convert
 			// "IMAGE_container=docker.io/foo"
-			// "IMAGE_deployment_container=docker.io/foo2"
+			// "IMAGE_deployment__container=docker.io/foo2"
+			// "IMAGE_env_var=docker.io/foo3"
+			// "IMAGE_deployment__env_var=docker.io/foo4"
 			// to
 			// container: docker.io/foo
 			// deployment/container: docker.io/foo2
-			var name string
-			parts := strings.SplitN(pair[0], "_", -1)
-			if len(parts) == 3 {
-				name = fmt.Sprintf("%s/%s", parts[1], parts[2])
-			} else {
-				name = parts[1]
-			}
-
+			// env_var: docker.io/foo3
+			// deployment/env_var: docker.io/foo4
+			name := strings.TrimPrefix(pair[0], ImagePrefix)
+			name = strings.Replace(name, "__", "/", 1)
 			if pair[1] != "" {
 				overrideMap[name] = pair[1]
 			}
--- a/test/e2e/knative_eventing_test.go
+++ b/test/e2e/knative_eventing_test.go
@@ -49,6 +49,10 @@ func TestKnativeEventing(t *testing.T) {
 		}
 	})
 
+	t.Run("make sure no gcr.io references are there", func(t *testing.T) {
+		verifyNoDisallowedImageReference(t, caCtx, knativeEventing)
+	})
+
 	t.Run("remove knativeeventing cr", func(t *testing.T) {
 		if err := v1a1test.DeleteKnativeEventing(caCtx, knativeEventing, knativeEventing); err != nil {
 			t.Fatal("Failed to remove Knative Eventing", err)
--- a/knative-operator/pkg/common/eventing_test.go
+++ b/knative-operator/pkg/common/eventing_test.go
@@ -32,7 +32,7 @@ func TestMutateEventing(t *testing.T) {
 	// Setup image override
 	os.Setenv("IMAGE_foo", image1)
 	// Setup image override with deployment name
-	os.Setenv("IMAGE_bar_baz", image2)
+	os.Setenv("IMAGE_bar__baz", image2)
 
 	// Mutate for OpenShift
 	if err := common.MutateEventing(ke, client); err != nil {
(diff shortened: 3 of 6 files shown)


test/e2e/knative_eventing_test.go
```diff
@@ -49,6 +55,66 @@ func TestKnativeEventing(t *testing.T) {
 		}
 	})
 
+	t.Run("make sure no gcr.io references are there", func(t *testing.T) {
+		podSpecableTypes := []schema.GroupVersionResource{
```
markusthoemmes, line 59: nit:

```golang
types := []type{{
  Group: "apps",
  Version: "test",
  ...
}, {
  Group: "blah",
  ..
}}
```

allows for a bit denser packing.

test/e2e/knative_eventing_test.go
```diff
@@ -49,6 +55,66 @@ func TestKnativeEventing(t *testing.T) {
...
+		for i := range podSpecableTypes {
+			podSpecableType := podSpecableTypes[i]
+
+			result, err := caCtx.Clients.Dynamic.Resource(podSpecableType).Namespace(knativeEventing).List(metav1.ListOptions{})
+			if err != nil {
+				t.Fatalf("Error listing %v: %v", podSpecableType, err)
+			}
+
+			for _, u := range result.Items {
+				ps := &duckv1.WithPod{}
```
markusthoemmes, line 91: Great use of ducktyping!

test/e2e/knative_eventing_test.go
```diff
@@ -49,6 +55,66 @@ func TestKnativeEventing(t *testing.T) {
...
+			},
+			{
+				Group:    "apps",
+				Version:  "v1",
+				Resource: "replicasets",
+			},
+		}
+
+		for i := range podSpecableTypes {
+			podSpecableType := podSpecableTypes[i]
```
markusthoemmes, line 83:
```suggestion
		for _, podSpecableType := range podSpecableTypes {
```

aliok:
@jcrossley3 I am sorry I have to go back to double underscore solution. I don't want to go deep and come up with a bigger solution. Double underscores work for now.

knative-operator/pkg/common/util.go
```diff
@@ -40,21 +41,19 @@ func BuildImageOverrideMapFromEnviron(environ []string) map[string]string {
...
 			// "IMAGE_container=docker.io/foo"
-			// "IMAGE_deployment_container=docker.io/foo2"
+			// "IMAGE_deployment__container=docker.io/foo2"
+			// "IMAGE_env_var=docker.io/foo3"
+			// "IMAGE_deployment__env_var=docker.io/foo4"
 			// to
 			// container: docker.io/foo
 			// deployment/container: docker.io/foo2
-			var name string
-			parts := strings.SplitN(pair[0], "_", -1)
```
markusthoemmes, line 51: If we had set this to `2`, would we theoretically get the same behavior even with underscores?

Regardless: I prefer a "proper" sentinel, double underscores seem fine.

markusthoemmes:
We definitely should send this against Serving too. There might even be a JIRA somewhere to write something like this :joy:. Please do once this one merges :)

knative-operator/pkg/common/util.go
aliok: > If we had set this to 2, would we theoretically get the same behavior even with underscores?

`IMAGE_PING_IMAGE` might not work, depending on how you implement it.

That's the problematic case.
We receive the env var as `IMAGE_PING_IMAGE=docker.io/foo` and deployment name is not `PING` :)

> Regardless: I prefer a "proper" sentinel, double underscores seem fine.

Yeah, me too. Also makes human-reading of CSV easier.

aliok:
> We definitely should send this against Serving too. There might even be a JIRA somewhere to write something like this joy. Please do once this one merges :)

Yeah, let me do it for serving too!

knative-operator/pkg/common/util.go
markusthoemmes: Gotcha, yeah we need a proper sentinel then.

aliok:
Ok, enabled for serving too.

Commits are squashed

matzew:
ah, great - thanks Ali, I think I just got bitten by that :smile:

jcrossley3:
Personally, I'd prefer we didn't have to set TWO env vars to override an image, but since upstream requires it, and underscores are legal characters in env vars, here we are. I don't have a strong opinion on which sentinel we use, but because some fonts make '_' and '__' hard to distinguish, I might consider something else, something illegal in both env vars and k8s names, maybe ':' or '|'?

Some of this may change depending on how https://github.com/knative-sandbox/operator/issues/22 is resolved, too.

aliok:
@jcrossley3 

Which separator do you like? I will update my PR.

`__`
`|`
`:`

matzew:
I think I buy the argument of readability... so, I think `|` seems a good compromise ?

matzew:
Oh, perhaps `:` is the better choice ?

markusthoemmes:
Use `:` if that doesn't cause any havoc. I've seen that being used as a replacement character for `/` if that's not available. Anecdotal only, but at least there's precedence.

aliok:
Ok let me change it to : and let's merge this

aliok:
/hold

Just gave this a try:

```
message: 'install strategy failed: Deployment.apps "knative-openshift" is invalid:
 979   │         [spec.template.spec.containers[0].env[16].name: Invalid value: "IMAGE_eventing-controller:eventing-controller":
 980   │         a valid environment variable name must consist of alphabetic characters, digits,
 981   │         ''_'', ''-'', or ''.'', and must not start with a digit (e.g. ''my.env-name'',  or
 982   │         ''MY_ENV.NAME'',  or ''MyEnvName1'', regex used for validation is ''[-._a-zA-Z][-._a-zA-Z0-9]*''),
 983   │         spec.template.spec.containers[0].env[17].name: Invalid value: "IMAGE_eventing-webhook:eventing-webhook":
```

aliok:
`[-._a-zA-Z][-._a-zA-Z0-9]*'`

`-` is a no go.
`.` might look strange.

I am going back to `__`

markusthoemmes:
ugh. YOLO then.

aliok:
/test 4.5-e2e-aws-ocp-45

CheckSLO failed: SLI for "TestActivatorHA" = 0.944444, wanted >= 0.990000

PR #296, opened 2020-05-12
bump to 1.8.0, but keep images at 0.13

matzew:
as per title. just bump the csv version

/assign @markusthoemmes 
/assign @aliok

### DIFF
--- a/hack/lib/vars.bash
+++ b/hack/lib/vars.bash
@@ -46,7 +46,7 @@ readonly UPGRADE_SERVERLESS="${UPGRADE_SERVERLESS:-"true"}"
 readonly UPGRADE_CLUSTER="${UPGRADE_CLUSTER:-"false"}"
 
 readonly INSTALL_PREVIOUS_VERSION="${INSTALL_PREVIOUS_VERSION:-"false"}"
-export OLM_CHANNEL="${OLM_CHANNEL:-"4.3"}"
+export OLM_CHANNEL="${OLM_CHANNEL:-"preview-4.6"}"
 # Change this when upgrades need switching to a different channel
 export OLM_UPGRADE_CHANNEL="${OLM_UPGRADE_CHANNEL:-"$OLM_CHANNEL"}"
 export OLM_SOURCE="${OLM_SOURCE:-"$OPERATOR"}"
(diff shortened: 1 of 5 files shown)


aliok:
> om server (NotFound): knativeeventings.operator.knative.dev "knative-eventing" not found
I don't get why the upgrade test fails :( 
need to debug locally

markusthoemmes:
/hold

We can't actually do it like this. We shouldn't bump the 4.3 and 4.4 channels here.

matzew:
> We can't actually do it like this. We shouldn't bump the 4.3 and 4.4 channels here.

oh - I see ... so  4.5 channel only ?

markusthoemmes:
Not even, I think it'd be preview-4.6.

matzew:
@markusthoemmes I've changed the channels.

1.8.0 is now _only_ on the `preview-4.6` channel.

Feel free to unhold, if that makes sense for you

markusthoemmes:
This looks like a consistent failure between all versions in the upgrade. Let's please take a look at what happened there.

aliok:
I had a look at this. Nothing shoudl happen actually as the new channel is not used in the scripts. Checking it in detail now

markusthoemmes:
We should probably switch the default channel in the same go.

markusthoemmes:
The problem here is

```
DEBUG 14:37:35.336 Execution failed: [[ ! ( $(oc get knativeserving.operator.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.version}") != $serving_version && $(oc get knativeserving.operator.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.conditions[?(@.type==\"Ready\")].status}") == True ) ]]. Waiting 5 sec (905/900)...
```

A good old issue with our upgrade test if we don't upgrade serving versions in the same go. See my bootstrap from the 1.7.0 release for what to remove/comment out to make that pass: https://github.com/openshift-knative/serverless-operator/commit/a6f6f11dfbda73cd1451093c647570b5259759a5

@mgencur we should look at fixing this. It's... annoying :joy:

aliok:
/close

I created https://github.com/openshift-knative/serverless-operator/pull/300 because I cannot add commits here w/o bothering matzew

PR #302, opened 2020-05-15
[WIP] Trigger CI

mgencur:
Do not merge! Just trying if master branch CI is broken because this is what I'm seeing in my experiments.


ci
```diff
@@ -0,0 +1 @@
+Pá kvě 15 10:29:43 CEST 2020
```
matzew, line 1: :smiley:

mgencur: This is not supposed to be read by humans!!! :-D

PR #303, opened 2020-05-15
Allow upgrades to OCP that is not yet available via channel

mgencur:
* for example, to a nightly build of OCP (e.g. 4.5


mgencur:
/hold cancel

I've added `--allow-explicit-upgrade` which is required with the latest versions of `oc`

PR #310, opened 2020-05-19
[SRVKS-565] Create OpenShift Route in the same namespace with Knative Service instead of knative-serving-ingress

nak3:
This patch changes to create OpenShift Route in the same namespace
with Knative Service intead of knative-serving-ingress.

In order to create Route in each namespace, O-K-I makes service and
endpoints for Kourier Gateway in each namespace. Then, OpenShift
Route points to the service as the target.


nak3:
This needs https://github.com/knative/serving/commit/584d41b1723f3a7aa7ab4181638793f47cb2a430 as `config-leader-election` is conflicted.

Closing for now.

PR #313, opened 2020-05-20
Bump to proper version of knative.dev/pkg

jcrossley3:
Ours (10) was very old, and the latest (13) lacks the istio API's, so
we bring in istio.io now.


jcrossley3:
I can't tell why these are failing. Passed locally for me, but I'm on 4.3, which also passed here. If anybody cares to walk through how to quickly determine what failed with me, I'm all ears. Until then...

/retest

jcrossley3:
Hey, whaddaya know? It worked! So here's half-a-million new lines of code! ;)

PR #317, opened 2020-05-23
[BACKPORT][SRVKS-567] Set KnativeServing CR status to "Dependency installing" whenever Kourier installation failed

nak3:
This is a backport patch for https://github.com/openshift-knative/serverless-operator/pull/316

Currently `MarkDependencyInstalling("Kourier")` is called only when
`checkDeployments()` returned error.
Due to this, when kourier installation failed by `manifest.Apply()` or
other place, the correct status is not set.

Hence this patch fixes it by adding `MarkDependencyInstalling("Kourier")`
whenever `kourier.Apply()` returns error.

### DIFF
--- a/knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
+++ b/knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
@@ -395,8 +395,8 @@ func (r *ReconcileKnativeServing) reconcileConfigMap(instance *servingv1alpha1.K
 // Install Kourier Ingress Gateway
 func (r *ReconcileKnativeServing) installKourier(instance *servingv1alpha1.KnativeServing) error {
 	// install Kourier
+	instance.Status.MarkDependencyInstalling("Kourier")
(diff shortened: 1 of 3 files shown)


markusthoemmes:
/hold

Sorry I jumped the gun on this one. The release is already in verification. Let's hold this to see if we're spinning a new release, which could then include this.

nak3:
/retest

e2e test failed due to the same reason with https://github.com/openshift-knative/serverless-operator/pull/365...
upgrade test failed due to probe failure.

rhuss:
@nak3 @markusthoemmes Just to be clear, this PR is supposed to beads to 1.7.2, right ? 

Do I have something to do for adding this to some changelog etc. so that this fix appears in the release notes ? // @abrennan89

nak3:
@rhuss Thank you. Yes, it is better to add "Fixed issues" in the release note. I will update the JIRA ticket to request it to @abrennan89

PR #318, opened 2020-05-25
[SRVKS-569] Add KnativeServing tracing config to CRD schema

nak3:
This patch adds KnativeServing tracing config to CRD schema. 

After this patch, `oc explain` prints following config:

```
$ oc explain ks.spec.config.tracing
KIND:     KnativeServing
VERSION:  operator.knative.dev/v1alpha1

RESOURCE: tracing <Object>

DESCRIPTION:
     configuration for tracing

FIELDS:
   backend	<string>
     The backend for tracing. It may be "zipkin" or "none". The default is
     "none"

   debug	<string>
     URL to jaeger collector where traces are sent. This must be specified when
     backend is "zipkin".

   sample-rate	<string>
     Percentage (0-1) of requests to trace

   zipkin-endpoint	<string>
     URL to zipkin collector where traces are sent.
```

### DIFF
--- a/olm-catalog/serverless-operator/1.8.0/operator_v1alpha1_knativeserving_crd.yaml
+++ b/olm-catalog/serverless-operator/1.8.0/operator_v1alpha1_knativeserving_crd.yaml
@@ -55,12 +55,59 @@ spec:
                   type: object
               type: object
             config:
-              additionalProperties:
-                additionalProperties:
-                  type: string
-                type: object
               description: A means to override the corresponding entries in the upstream
                 configmaps
+              properties:
+                autoscaler:
+                  additionalProperties:
+                    type: string
+                  type: object
+                defaults:
+                  additionalProperties:
+                    type: string
+                  type: object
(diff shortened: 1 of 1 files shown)


nak3:
cc @maschmid 
Please let us know if this PR should wait for your verification.

markusthoemmes:
Have we talked about/proposed this upstream? That'd be value-add there as well and I'd feel more comfortable not running with a "patched" version of the CRD.

jcrossley3:
I'm not sure I love this idea. I view `spec.config` as a convenient *non-typesafe* way to consolidate configuration. It's just a stopgap until the "real" types reveal themselves. In this case, "tracing" has revealed itself, so we can now define `spec.tracing`. Because all the fields in the `spec` *are* configuration, having `spec.config.foo` seems redundant: it should just be `spec.foo`.

nak3:
Closing as this change needs a migration to `spec.xxx` from `spec.config` for CR.

PR #322, opened 2020-05-28
Run upgrade tests from multiple previous version in a single CI run

mgencur:
This PR adds a few scenarios to upgrade tests:
* upgrade from oldest CSV with the same minKubeVersion as the latest on the given channel
* upgrade from latest-2 CSV
Both of these scenarios use automatic upgrades/approval. However, at the beginning the Subscription has a manual approval strategy so that we can install the given old version and start upgrade from there. Then the approval strategy is changed to automatic and the first InstallPlan approved manually (because it stays manual), the next install plans are then automatically approved by OLM. These tests run pre-upgrade tests from Knative Serving after the first old version is installed, and then run post-upgrade tests after OLM upgrades to the latest version. We don't run Prober test in these scenarios because some upgrades break it and we can't control that with automatic upgrades.

The test container execution times are as follows:
* 16min - without these scenarios (meaning we test only upgrades from latest-1) (on OCP 4.3)
* 25min - when testing upgrades from latest-1 and from oldest compatible CSV (via minKubeVersion) (on OCP 4.3)
* 48min - when testing ugprade from latest-1, from latest-2, and from oldest compatible CSV (on OCP 4.4) - this execution time is strange but I couldn't verify it on other OCP version because the Prober test is currently very unstable (unrelated to this PR). Update: In the latest run https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift-knative_serverless-operator/322/pull-ci-openshift-knative-serverless-operator-master-4.3-upgrade-tests-aws-ocp-43/917 it took 31 minutes for the test container to run everything

### DIFF
--- a/test/upgrade-tests.sh
+++ b/test/upgrade-tests.sh
@@ -20,8 +20,21 @@ failed=0
 (( !failed )) && install_catalogsource || failed=3
 (( !failed )) && logger.success 'Cluster prepared for testing.'
 
-(( !failed )) && install_serverless_previous || failed=5
-(( !failed )) && run_knative_serving_rolling_upgrade_tests || failed=6
+# Test upgrades from CSV one version back, with manual approval.
+(( !failed )) && install_serverless_previous || failed=4
+(( !failed )) && run_knative_serving_rolling_upgrade_tests || failed=5
+(( !failed )) && teardown_serverless || failed=6
+
+# Test upgrades from CSV two versions back to the latest, with automatic approval.
+upgrade_from="$(latest_minus_two_csv)" || return $?
+(( !failed )) && INITIAL_CSV="$upgrade_from" install_serverless_previous || failed=7
+(( !failed )) && INITIAL_CSV="$upgrade_from" AUTO_UPGRADES=true run_knative_serving_rolling_upgrade_tests || failed=8
+(( !failed )) && teardown_serverless || failed=9
+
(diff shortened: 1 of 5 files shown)


test/upgrade-tests.sh
```diff
@@ -20,8 +20,21 @@ failed=0
...
 (( !failed )) && logger.success 'Cluster prepared for testing.'
 
-(( !failed )) && install_serverless_previous || failed=5
-(( !failed )) && run_knative_serving_rolling_upgrade_tests || failed=6
+# Test upgrades from CSV one version back, with manual approval.
+(( !failed )) && install_serverless_previous || failed=4
+(( !failed )) && run_knative_serving_rolling_upgrade_tests || failed=5
+(( !failed )) && teardown_serverless || failed=6
+
+# Test upgrades from CSV two versions back to the latest, with automatic approval.
```
nak3, line 28: Do we need to test both `from two versions back` and `from oldest`? I wonder that the `from oldest` covers `from two versions back`.

mgencur: Yeah. I think these two tests are different and we need them both. If you install something at version X and upgrade to latest, it is different from installing it at version X+1 and upgrading to latest. Some issues from the past remind me that upgrading from X to X+1 is not same as installing it freshly at version X+1.

PR #304, opened 2020-05-18
Use knative-serving-src image for running tests

mgencur:
* this image is used to run tests

Used in: https://github.com/openshift/release/pull/9143 (the PRs depends on each other so thay should be merged together)
Depends on: 
https://github.com/openshift/knative-serving/pull/474
https://github.com/openshift/release/pull/9141

PR #335, opened 2020-06-15
Do not clone Knative Eventing and remove third-party operator test suites

mgencur:
* the sources will be provided by ci-operator

Regarding operator test suites, the related discussion around removing those is here: https://coreos.slack.com/archives/CD87JDUB0/p1591185499029200
Summary: the tests do not provide much value as they're short, partially replicated by the serverless operator test suite, and they haven't revealed any bugs. It should be sufficient to run then as part of QE cycle rather than slowing down every single PR.

### DIFF
--- a/hack/lib/vars.bash
+++ b/hack/lib/vars.bash
@@ -12,18 +12,12 @@ fi
 source "$(dirname "${BASH_SOURCE[0]}")/../../test/vendor/knative.dev/test-infra/scripts/e2e-tests.sh"
 
 readonly KNATIVE_SERVING_VERSION="${KNATIVE_SERVING_VERSION:-v0.13.3}"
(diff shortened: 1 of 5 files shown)


mgencur:
/hold

Let's wait with merging until after https://github.com/openshift-knative/serverless-operator/pull/328

matzew:
/hold

wait for #328

mgencur:
The single failure is the well-known random fail in Probe test. It's apparent that the tests get to run which is important. IMO, no need to re-run further.

markusthoemmes:
I'd be game to merge this as-is! WDYT @matzew?

/lgtm
/approve

matzew:
/lgtm 

@markusthoemmes

cardil:
As #304 this change is responsible for CVP breakage.

PR #347, opened 2020-06-19
[SRVKS-511] Use spec.resources to bump webhook memory limit to 1G

jcrossley3:


knative-operator/pkg/common/serving.go
```diff
@@ -51,6 +54,22 @@ func defaultToHa(ks *servingv1alpha1.KnativeServing, c client.Client) error {
...
 }
 
+func defaultResources(ks *servingv1alpha1.KnativeServing, c client.Client) error {
+	if ks.Spec.Resources == nil {
+		ks.Spec.Resources = []servingv1alpha1.ResourceRequirementsOverride{
+			servingv1alpha1.ResourceRequirementsOverride{
+				Container: "webhook",
+				ResourceRequirements: corev1.ResourceRequirements{
+					Limits: corev1.ResourceList{
+						corev1.ResourceMemory: resource.MustParse("1024Mi"),
```
markusthoemmes, line 64: Do we want to fill this in even if Resources are set by the user? Maybe they're unaware of the webhook issues? :thinking:

jcrossley3: I would expect anyone who might set them would see that they're already in there. And if upon seeing them, they decide to remove them, I'm assuming they have a good reason, and I think it might surprise them if the operator presumes to put them back.

But I don't feel that strongly about it.

markusthoemmes: I'd love to merge with the user's settings. I'd only not set this if the user sets a different webhook requirement.

jcrossley3: Which user's settings? In the KnativeServing CR? Or somewhere else? If in the CR, I don't think the user will ever see a CR that doesn't already have the limit in there.

markusthoemmes: No I meant merge this webhook limit with whatever the user sets, but the user's settings win. I.e. they can adjust the controller but be oblivious to the webhook issues and only if they **also** specify the webhook limit, we of course take the user's settings because then they need to know what they're doing.

jcrossley3: I'm just questioning the likelihood of any user ever being oblivious to the webhook issue with the setting right there in the CR. You seem to be digging your heels into the idea that a user might create a CR from scratch, though, so ok.

markusthoemmes: Yeah, gitops and stuff like that. I'm not thinking about the edit use case too much at all, you're right. It's less of a concern there. Given we talk about declarative APIs though, I'd think the `oc apply` flow is the most common one. I might be wrong though but why not help them too?

jcrossley3:
Go ahead and list all the test fixtures you'll want to see before you'll
approve this PR, please.

markusthoemmes:
/retest

Lotsa infra stuff. Should be good though.

markusthoemmes:
Knative Serving tests mostly passed apart from one nit, gonna merge this.

PR #348, opened 2020-06-22
Patch config to use explicit broker class

matzew:

### DIFF
--- a/test/eventing.bash
+++ b/test/eventing.bash
@@ -9,13 +9,19 @@ function knative_eventing_tests {
 
   image_template="registry.svc.ci.openshift.org/openshift/knative-${KNATIVE_EVENTING_VERSION}:knative-eventing-test-{{.Name}}"
 
-  oc patch cm config-br-defaults -n knative-eventing -p '{"data":{"default-br-config":"clusterDefault:\n  brokerClass: ChannelBasedBroker\n  apiVersion: v1\n  kind: ConfigMap\n  name: config-br-default-channel\n  namespace: knative-eventing\n"}}' --type=merge || failed=1
+  oc patch KnativeEventing knative-eventing \
+    --namespace knative-eventing \
+    --type merge \
+    --patch '{"spec":{"defaultBrokerClass":"ChannelBasedBroker"}}' || failed=1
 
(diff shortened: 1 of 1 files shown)


test/eventing.bash
```diff
@@ -9,15 +9,17 @@ function knative_eventing_tests {
 
   image_template="registry.svc.ci.openshift.org/openshift/knative-${KNATIVE_EVENTING_VERSION}:knative-eventing-test-{{.Name}}"
 
+  oc patch cm config-br-defaults -n knative-eventing -p '{"data":{"default-br-config":"clusterDefault:\n  brokerClass: ChannelBasedBroker\n  apiVersion: v1\n  kind: ConfigMap\n  name: config-br-default-channel\n  namespace: knative-eventing\n"}}' --type=merge || failed=1
```
markusthoemmes, line 12: Should we do this via the KnativeEventing CR created by the tests?

matzew: @markusthoemmes you mean patching the `KnativeEventing` CR, instead of patching the `ConfigMap` ?

matzew: currently this matches how we apply the diffs on the upstream and downstream/midstream forks.

IMO it's nice to have them aligned :man_shrugging:

markusthoemmes: Well I guess we already had a patch in here anyway... I was more thinking to statically set it but as we change it during the tests that doesn't, work either.
Me still thinks it'd be nicer to patch through the CR to exercise the path users would take here. But your call.

matzew: Good point, I've changed this to use the `KnativeEventing` CR, since that's the customer path

test/eventing.bash
```diff
@@ -9,15 +9,23 @@ function knative_eventing_tests {
 
   image_template="registry.svc.ci.openshift.org/openshift/knative-${KNATIVE_EVENTING_VERSION}:knative-eventing-test-{{.Name}}"
 
+  oc patch KnativeEventing knative-eventing \
+    --namespace knative-eventing \
+    --type merge \
+    --patch '{"spec":{"defaultBrokerClass":"ChannelBasedBroker"}}' || failed=1
```
markusthoemmes, line 15: Beautiful <3

matzew:
/retest

failures from AWS

matzew:

was again a `TestProbe` issue :shrug:

matzew:
TestProbe...

PR #353, opened 2020-06-25
[WIP] Adjust rollout settings to minimize Kourier rollout impact.

markusthoemmes:
Playing with this to see if it helps our TestProbe flakes (and thus the users during rollout).

### DIFF
--- a/knative-operator/deploy/resources/kourier/kourier-release-0.14.yaml
+++ b/knative-operator/deploy/resources/kourier/kourier-release-0.14.yaml
@@ -92,6 +92,12 @@ spec:
   selector:
     matchLabels:
       app: 3scale-kourier-gateway
+  minReadySeconds: 5
+  strategy:
+    rollingUpdate:
+      maxSurge: 100%
+      maxUnavailable: 0%
+    type: RollingUpdate
   template:
     metadata:
       labels:


markusthoemmes:
/test 4.4-upgrade-tests-aws-ocp-44

Both passed, let's see if that was just luck.

markusthoemmes:

Infra issues!

markusthoemmes:

Infra issues.

markusthoemmes:
/retest

A few failures, but veeery low failure rates. Let's try to gather more data.

nak3:
I agree that it is worth to use the options. Should we add these options to upstream net-kourie as well? Or we just use it for the workaround of the tests in unsupported grace shutdown?
/cc @jmprusi

markusthoemmes:
@nak3 I was thinking the latter but obviously it doesn't make it 100% tight.

nak3:
I see. Then, let's merge this PR.

/lgtm

jmprusi:
> I agree that it is worth to use the options. Should we add these options to upstream net-kourie as well? Or we just use it for the workaround of the tests in unsupported grace shutdown?
> 
> /cc @jmprusi

In upstream we have a pre stop hook, I was planning on adding some sleep after setting the envoy as not ready and also tweak the rollout process, but I need to run some tests first to avoid making it worse.

PR #356, opened 2020-06-26
Rename CSV

warrenvw:
So it seems Koji does not support semver conventions in the version string. The version `1.8.0-rc1` is a valid semver, yet is not allowed by Koji. `1.8.0rc1` and `1.8.0.rc1` are not valid semver but work with Koji.

Further, there does not seem to be a consistent naming for release candidates within Red Hat. Some use the format: `1.8.0.rc1` while others use `1.8.0rc1`, and there may be other formats that I have not yet discovered.

In absence of a standard, we chose to use `1.8.0rc1`.

https://regex101.com/r/Ly7O1x/3/


bbrowning:
We must use valid semantic versions here and `1.8.0rc1` is not a valid semantic version. Specifically OLM uses the github.com/blang/semver library to parse semantic versions in Go and it will throw an error using `1.8.0rc1`.

bbrowning:
Run https://play.golang.org/p/50E39M2PhvG to see how this fails in the semver library.

warrenvw:
Essentially, this PR just renames the CSV to include "1.8.0-rc1". Release candidates are indicated in the errata, which no change to the brew version. We are setting OLM version to 1.8.0-rc1, and that is what will be placed in the `candidate` channel.

rhuss:
I this PR still needed ? If so, @warren could you please resolve the conflict ? (i.e. copy over the updated CSV file to this name ?

thanks !

rhuss:
I'm a bit lost about the current status for the 1.8.0 preparations. Is this PR still needed @warrenvw ?

markusthoemmes:
imma ditch this PR, don't think it's needed anymore. Please reopen if it is.

PR #357, opened 2020-06-29
Metadata about supported environment

cardil:
This file contains information about supported environment of Serverless, and it's used in release pipeline to ensure proper behavior of this operator versus other parts of software that it runs against.

### DIFF
--- a/olm-catalog/serverless-operator/operator.environment.yaml
+++ b/olm-catalog/serverless-operator/operator.environment.yaml
@@ -1,0 +1,24 @@
+---
+openshift:
+  knative:
+    serving:
+      branch: release-v0.13.3
+      ci-promotion-name: knative-v0.13.3
+    eventing:
+      branch: release-v0.13.0
+      ci-promotion-name: knative-v0.13.0
+    eventing-contrib:
+      branch: release-v0.15.0
+      ci-promotion-name: knative-v0.15.0
+    kafka:
+      branch: v0.15.0
+    camel:
+      branch: master
+    client:
+      branch: openshift-v0.13.2
+      url: https://mirror.openshift.com/pub/openshift-v4/clients/serverless/latest/
+  pipelines:
+    channel: preview
+knative:
+  operator:
+    branch: master


olm-catalog/serverless-operator/operator.environment.yaml
```diff
@@ -0,0 +1,24 @@
+---
+openshift:
```
mgencur, line 2: I'm trying to find any system in how these variables are named. It looks a bit random.
Perhaps we could follow how the repositories are really named but then I'm wondering how to add any random information.
I'd say this file needs some well-defined structure or needs to be more flat. But I don't know what structure to suggest right now.

cardil: I basically rewritten the variables that we were using to be nested.

```java
static final String OPENSHIFT_KNATIVE_SERVING_BRANCH = "release-v0.13.3"
static final String OPENSHIFT_KNATIVE_SERVING_CI_PROMOTION_NAME = "knative-v0.13.3"
```
bacame:

```yaml
openshift:
  knative:
    serving:
      branch: release-v0.13.3
      ci-promotion-name: knative-v0.13.3
```

with the idea that you'll fetch them with something like:

```bash
$ opskope registry.example.org/openshift/serverless:v1.8.0 
    metadata-get openshift.knative.serving.branch
```

Or maybe itarete on some keys, like:

```bash
$ opskope registry.example.org/openshift/serverless:v1.8.0 
    metadata-foreach --output json openshift.serverless.upgradepaths | ...
```

cardil: I basically rewritten the variables that we were using to be nested.

```java
static final String OPENSHIFT_KNATIVE_SERVING_BRANCH = "release-v0.13.3"
static final String OPENSHIFT_KNATIVE_SERVING_CI_PROMOTION_NAME = "knative-v0.13.3"
```
bacame:

```yaml
openshift:
  knative:
    serving:
      branch: release-v0.13.3
      ci-promotion-name: knative-v0.13.3
```

with the idea that you'll fetch them with something like:

```bash
$ opskope registry.example.org/openshift/serverless:v1.8.0 
    metadata-get openshift.knative.serving.branch
```

Or maybe itarete on some keys, like:

```bash
$ opskope registry.example.org/openshift/serverless:v1.8.0 
    metadata-foreach --output json openshift.serverless.upgradepaths | ...
```

mgencur: ok. Thanks. I was thinking about two ways to give it more well-defined structure:
1) Follow the repository hierarchy and names as we have them on github:
```
---
sources:
  openshift:
    knative-serving:
      branch: release-v0.13.3
      cirr-promotion-name: knative-v0.13.3
    knative-eventing:
      branch: release-v0.13.0
      ci-promotion-name: knative-v0.13.0
    knative-eventing-contrib:
      branch: release-v0.15.0
      ci-promotion-name: knative-v0.15.0
    knative-client:
      branch: openshift-v0.13.2
      url: https://mirror.openshift.com/pub/openshift-v4/clients/serverless/latest/
    tektoncd-pipeline-operator:
      channel: preview
  openshift-knative:
    knative-kafka-operator:
      branch: v0.15.0
    knative-camel-operator:
      branch: master
    operator:
      branch: master
custom:
  ???

```

2) Or use a flat structure like this:
```
openshift.knative-serving.branch: release-v0.13.3
openshift.knative-serving.ci-promotion-name: knative-v0.13.3
...
```

It would be good to see any other items that we want to put in this file.

cardil: I would prefer the first, nested, proposition as there are tools already to parse such structure. I'm thinking of [jsonpath](https://support.smartbear.com/alertsite/docs/monitors/api/endpoint/jsonpath.html).

olm-catalog/serverless-operator/operator.environment.yaml
```diff
@@ -0,0 +1,24 @@
...
+      branch: release-v0.13.3
+      ci-promotion-name: knative-v0.13.3
+    eventing:
+      branch: release-v0.13.0
+      ci-promotion-name: knative-v0.13.0
+    eventing-contrib:
+      branch: release-v0.15.0
+      ci-promotion-name: knative-v0.15.0
+    kafka:
+      branch: v0.15.0
```
maschmid, line 14: kafka/camel/pipelines settings aren't image-specific, they refer to current releases of their respective operators, so I don't think they should be maintained here...

cardil: It's not image specific parameters. Those specify serverless supported environment. So, in relation to kafka/camel/pipelines it describe what version we test that with.

maschmid: the kafka.branch and camel.branch parameters aren't actually used by jobs by default... (by default, the current versions of knative camel and kafka operators from the operatorhub are installed) .. the .branch parameters are only used when installing unreleased versions)

maschmid: we should probably just change that to just specify channels, as in the automatic triggers we will probably only want to test with released versions of those...

cardil: I think not necessarily. The goal is to be able to trigger all jobs that would test the release, against all we are supporting.

PR #366, opened 2020-07-06
[SRVKS-588] Fix churn on KnativeServing CR caused by constant spec overriding.

markusthoemmes:
The `resources` array not being part of the CRD validation causes K8s API to drop it upon being created. The controller is thus constantly overriding the spec because it seemingly needs to add it back in, only for it being dropped again.


olm-catalog/serverless-operator/1.8.0/operator_v1alpha1_knativeserving_crd.yaml
```diff
@@ -10,7 +10,7 @@ spec:
   - JSONPath: .status.conditions[?(@.type=="Ready")].status
     name: Ready
     type: string
-  - JSONPath: .status.conditions[?(@.type=="Ready")].reason
+  - JSONPath: ".status.conditions[?(@.type==\"Ready\")].reason"
```
markusthoemmes, line 13: This diff is fairly large because I opted to just copy upstream's CRD in here for easier management in the future.

jcrossley3:
I'm approving this based on my assumption that the changes (which i haven't studied) are due to the upstream CRD's corresponding to the versions in the CSV are being copied over, as they should whenever we update the versions in the CSV.
/approve
/lgtm

PR #373, opened 2020-07-13
Serverless 1.9.0, knative 0.15.2

jcrossley3:
Contains only the CSV and deps for the operator. Changes to support SRVLS-264 will be in a separate PR.


jcrossley3:
Retrying after removing istio crap: https://github.com/openshift-knative/operator/pull/22
/retest

nak3:
Looks good to me overall.
But there seems to be one issue. Serving and Eventing have the same name configmaps like `config-logging`, `config-tracing`, (and `config-leader-election`).

e.g. config-logging
https://github.com/openshift-knative/operator/blob/release-0.15/cmd/operator/kodata/knative-eventing/0.15.2/1-eventing.yaml#L403
https://github.com/openshift-knative/operator/blob/release-0.15/cmd/operator/kodata/knative-serving/0.15.2/2-serving-core.yaml#L673

It seems that they will be conflicted when both knativeserving and knativeeventing CR had same configuration but different values.

jcrossley3:
> It seems that they will be conflicted when both knativeserving and knativeeventing CR had same configuration but different values.

Great point, @nak3 -- and I'm not a little embarrassed to say I hadn't even considered that possibility. :)

I verified that these are the only conflicting resources (in 0.15 and 0.16):
* config-leader-election /v1, Kind=ConfigMap
* config-logging /v1, Kind=ConfigMap
* config-observability /v1, Kind=ConfigMap
* config-tracing /v1, Kind=ConfigMap

And there is definitely potential for any of those to be overwritten when either knative component CR is reconciled. In light of this, I'd like confirmation from arch/prod as to whether having serving and eventing share the same `openshift-serverless` namespace is a goal worth pursuing. /cc @pmorie @markito @bbrowning @markusthoemmes 

If it is, then I guess we can either work with upstream to ensure these configmap names are configurable or we can document that any modifications to the above CM's must be made not in the CM's themselves but in *both* the `KnativeServing` and `KnativeEventing` CR's because whichever is the last to reconcile will "win". Or something else?

mgencur:
OCP 4.3 and 4.4 checks are not part of pre-submit checks anymore. After all, the "tide" plugin says this PR is clean. The failures shown in this PR are from old runs (10 days+). We changed the way we run tests on 4.3 and 4.4 since then.
They now run in periodic jobs every twelve hours.

bbrowning:
@jcrossley3 From my perspective, ideally we'd push on running both Serving and Eventing in the same namespace in upstream. Do upstream Serving and Eventing support running in the same namespace, assuming no operator is involved? Is it just an oversight of the upstream/our operator about how to handle this conflict? Or is it a broader Serving and Eventing issue that running both in the same namespace doesn't work?

If it's just an operator issue, we should think through how to solve that properly. As you pointed out, having both KnativeServing and KnativeEventing reconciling the same configmaps isn't a good solution. Either way, it sounds like we should think about separating the Knative 0.15 upgrade out from combining namespaces so we can choose to deliver those separately if there's not an easy fix for running both in the same namespace.

markusthoemmes:
Agree with @bbrowning

jcrossley3:
The simplest solution may be to just never delete any shared ConfigMaps.

jcrossley3:
no clue
/retest

PR #374, opened 2020-07-13
Trying to prevent being locked out of my dev cluster

jcrossley3:


markusthoemmes:
Can you describe why this is needed? I didn't have issues with being logged out while extensively testing #397

PR #376, opened 2020-07-13
Script and make target for running all tests

mgencur:
https://issues.redhat.com/browse/SRVCOM-891

Will be used in https://github.com/openshift/release/pull/10210

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,11 @@ test-unit:
 	go test ./serving/ingress/...
 
 test-e2e:
-	./test/e2e-tests.sh
+	TEST_ALL=false ./test/e2e-tests.sh
+
+# Used by periodic CI runs
+test-all:
+	TEST_ALL=true ./test/e2e-tests.sh
 
 test-upgrade:
 	./test/upgrade-tests.sh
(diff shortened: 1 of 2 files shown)


mgencur:
The only failure was TestProbe on one platform. This proves that the new scripts runs all the tests we want. I've removed the temporary commit that was calling the script. In the next run the script won't be used so that failures will be unrelated to changes in this PR.

test/all-tests.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
```
nak3, line 1: This is almost same with `test/e2e-tests.sh`. Can we re-use it instead of creating this script?

mgencur: Can you suggest how to reuse it? I'm not able to find and elegant way. One way is to pass flag and change behaviour based on flags. But I'd rather have the behaviour clear from looking at the script.

nak3: Sure. How about using env variable like this?

Makefile:

```
test-e2e:
        TEST_ALL="false" ./test/e2e-tests.sh

test-all:
        TEST_ALL="true" ./test/e2e-tests.sh
```

test/e2e-tests.sh
```
if [[ $TEST_ALL == "true" ]]; then
  (( !failed )) && install_serverless_previous || failed=2
  (( !failed )) && run_knative_serving_rolling_upgrade_tests || failed=3
  (( !failed )) && teardown_serverless || failed=4
fi
```

mgencur: Done. Thanks for suggestion.

PR #377, opened 2020-07-14
[SRVKS-588] Avoid resourceVersion churn due to constant status resetting.

markusthoemmes:
The status conditions have a guard that avoid updating the condition's lastTransitionTime timestamp if you're trying to set it to the same state multiple times. As such, running `MarkDependenciesInstalled` on each reconcile is benign.

However, internally flipping the state on each reconcile is not. It'll cause us to constantly update the timestamps and thus churn through resourceVersions. This prevents that. The status will be "Unknown" before `Apply` returns for the first time, which is I think just as good as what we have today. If we want to be smarter with the status setting, we'd need to check if we *actually* need to run apply etc etc. Not sure it's worth it.


nak3:
/lgtm
/approve

Thank you!

mgencur:
It would be nice to have it on 1.8 branch as well ;-)

PR #379, opened 2020-07-20
Decrease SLO to 95% if the previous serving version is earlier than 0.14

nak3:
One of common flake test is TestProbe. Since Kourier does not support
Graceful shutdown before 0.14, it does not meet SLO.

This patch decreases the SLO from 100% to 95% when the previous
serving version is ealier than 0.14.

### DIFF
--- a/test/serving.bash
+++ b/test/serving.bash
@@ -95,7 +95,6 @@ function run_knative_serving_rolling_upgrade_tests {
   logger.info "Target success fraction is $PROBE_FRACTION"
 
   go_test_e2e -tags=preupgrade -timeout=20m ./test/upgrade \
-    -probe.success_fraction=$PROBE_FRACTION \
     --imagetemplate "$image_template" \
     --kubeconfig "$KUBECONFIG" \
     --resolvabledomain || return 1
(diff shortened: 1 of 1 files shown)


test/serving.bash
```diff
@@ -86,8 +86,17 @@ function run_knative_serving_rolling_upgrade_tests {
 
   failed=0
   image_template="registry.svc.ci.openshift.org/openshift/knative-${KNATIVE_SERVING_VERSION}:knative-serving-test-{{.Name}}"
+  PROBE_FRACTION=1.0
+  prev_serving_version=$(oc get knativeserving.operator.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.version}")
 
+  prev_version=$(echo ${prev_serving_version} | awk -F. '{printf "%2d%02d%02d", $1,$2,$3}') # Format serving version e.g. 0.13.2 to 01302
```
mgencur, line 92: I'm curious if this is necessary. Could we just compare the strings alphabetically? e.g. `if [[ $prev_version < "0.14.0" ]]`

nak3:
/hold

It seems that the test does not use proper branch https://github.com/openshift/knative-serving. The error says the option is missing.

nak3:

It seems that the source code is from an image built by https://github.com/openshift-knative/serverless-operator/tree/master/openshift/ci-operator/source-image/Dockerfile and it was built periodically? Try to re-test.

mgencur:
/lgtm 

Yeah, I overlooked it too.

PR #381, opened 2020-07-21
[SRVCLI-191] Change CLI download Deployment to a Knative Service

dsimansk:
This PR adds:
- CLI Download as a Knative Service
- A lot of vendor dir additions (I'd appreciate a feedback if the amount of changes could be decreased)

Ref: https://issues.redhat.com/browse/SRVCLI-191

/cc @rhuss @navidshaikh


knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
```diff
@@ -12,24 +12,26 @@ import (
...
-	knCLIDownload                   = "kn"
-	knDownloadServer                = "kn-download-server"
-	knConsoleCLIDownloadDeployRoute = "kn-cli-downloads"
+	knCLIDownload               = "kn"
+	knDownloadServer            = "kn-download-server"
+	knConsoleCLIDownloadService = "kn-cli-downloads"
 )
 
 var log = common.Log.WithName("consoleclidownload")
+var knServiceGVK = schema.GroupVersionKind{Group: "serving.knative.dev", Version: "v1", Kind: "Service"}
```
markusthoemmes, line 34: This is unused and more correctly constructed below now.

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
```diff
@@ -64,31 +66,24 @@ func reconcileKnCCDResources(instance *servingv1alpha1.KnativeServing, apiclient
...
 				return err
 			}
-			for _, c := range deployment.Status.Conditions {
-				if c.Type == appsv1.DeploymentAvailable && c.Status != corev1.ConditionTrue {
-					return fmt.Errorf("deployment %q/%q not ready yet", u.GetName(), u.GetNamespace())
+			for _, c := range service.Status.Conditions {
+				if c.Type == servingv1.ServiceConditionReady && c.Status != corev1.ConditionTrue {
+					return fmt.Errorf("Knative Service %q/%q not ready yet", u.GetName(), u.GetNamespace())
 				}
 			}
```
markusthoemmes, line 82: I believe you can replace this with `if !service.IsReady()`

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
```diff
@@ -64,31 +66,24 @@ func reconcileKnCCDResources(instance *servingv1alpha1.KnativeServing, apiclient
...
 func checkResources(manifest *mf.Manifest, instance *servingv1alpha1.KnativeServing, api client.Client) error {
 	log.Info("Checking deployments")
-	for _, u := range manifest.Filter(mf.ByKind("Deployment"), mf.ByKind("Route")).Resources() {
-		switch u.GetKind() {
-		case "Deployment":
-			deployment := &appsv1.Deployment{}
-			err := api.Get(context.TODO(), client.ObjectKey{Namespace: u.GetNamespace(), Name: u.GetName()}, deployment)
-			if err != nil {
+	for _, u := range manifest.Filter(mf.ByGVK(servingv1.SchemeGroupVersion.WithKind("Service"))).Resources() {
+		if u.GetKind() == "Service" {
```
markusthoemmes, line 73: This is redundant to the filter above it.

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
```diff
@@ -64,31 +66,24 @@ func reconcileKnCCDResources(instance *servingv1alpha1.KnativeServing, apiclient
...
+// Check for Knative Service and URL of kn ConsoleCLIDownload resources
 func checkResources(manifest *mf.Manifest, instance *servingv1alpha1.KnativeServing, api client.Client) error {
 	log.Info("Checking deployments")
-	for _, u := range manifest.Filter(mf.ByKind("Deployment"), mf.ByKind("Route")).Resources() {
-		switch u.GetKind() {
-		case "Deployment":
-			deployment := &appsv1.Deployment{}
-			err := api.Get(context.TODO(), client.ObjectKey{Namespace: u.GetNamespace(), Name: u.GetName()}, deployment)
-			if err != nil {
+	for _, u := range manifest.Filter(mf.ByGVK(servingv1.SchemeGroupVersion.WithKind("Service"))).Resources() {
```
markusthoemmes, line 72: Do we even need to do this at this point? You're hardcoding the Service name below so we might get away with the same thing here maybe?

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
dsimansk: It's a leftover from my unstructured PoC impl, I'll remove it, thanks.

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
dsimansk: Sounds like a great idea to me.

dsimansk:
> In general I'm wondering if we should drop the manifestival machinery for generating/applying the Service now that it's so "simple" to generate in Go code thinking. Some of the things would be quite a bit easier (for example "replacing" the image)

It seems to be working reasonably well with code generated Service in my tests so far. Stay tuned for the update.

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
```diff
@@ -154,75 +156,52 @@ func Delete(instance *servingv1alpha1.KnativeServing, apiclient client.Client, s
...
-					containers[i].Image = image
-					break
-				}
-			}
-
-			if err := scheme.Convert(deploy, u, nil); err != nil {
-				return fmt.Errorf("failed to convert Deployment obj to unstructured: %w", err)
-			}
+// populateKnService populates Knatie Service object and its SPEC with provided image
+func populateKnService(image string, instance *servingv1alpha1.KnativeServing) *servingv1.Service {
```
markusthoemmes, line 168:
```suggestion
func makeKnService(image string, instance *servingv1alpha1.KnativeServing) *servingv1.Service {
```

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
```diff
@@ -44,52 +41,57 @@ func Apply(instance *servingv1alpha1.KnativeServing, apiclient client.Client, sc
...
+		serviceFromClusterDC := serviceFromCluster.DeepCopy()
+		changed := false
+		if serviceFromCluster.Spec.Template.Spec.GetContainer().Image != service.Spec.Template.Spec.GetContainer().Image {
+			serviceFromClusterDC.Spec.Template.Spec.GetContainer().Image = service.Spec.Template.Spec.GetContainer().Image
+			changed = true
+		}
+		if !equality.Semantic.DeepEqual(serviceFromCluster.Spec.Template.Spec.GetContainer().Resources, service.Spec.Template.Spec.GetContainer().Resources) {
+			serviceFromClusterDC.Spec.Template.Spec.GetContainer().Resources = service.Spec.Template.Spec.GetContainer().Resources
+			changed = true
+		}
```
markusthoemmes, line 66: Why are we checking resources here?

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
```diff
@@ -154,75 +156,52 @@ func Delete(instance *servingv1alpha1.KnativeServing, apiclient client.Client, s
...
-					log.Info("Replacing", "deployment", container.Name, "image", image)
-					containers[i].Image = image
-					break
-				}
-			}
-
-			if err := scheme.Convert(deploy, u, nil); err != nil {
-				return fmt.Errorf("failed to convert Deployment obj to unstructured: %w", err)
-			}
+// populateKnService populates Knatie Service object and its SPEC with provided image
```
markusthoemmes, line 167:
```suggestion
// makeKnService populates Knative Service object and its SPEC with provided image
```

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
```diff
@@ -44,52 +41,57 @@ func Apply(instance *servingv1alpha1.KnativeServing, apiclient client.Client, sc
...
-			}
-			host := getCanonicalHost(route)
-			if host == "" {
-				return fmt.Errorf("route %q/%q not ready yet", u.GetName(), u.GetNamespace())
-			}
-		}
+	service := &servingv1.Service{}
+	if err := api.Get(context.TODO(), client.ObjectKey{Namespace: instance.GetNamespace(), Name: knConsoleCLIDownloadService}, service); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
dsimansk: :thinking: how'd I get updated Status fields then?

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
dsimansk: My thought was to check if fields needs to be updated when `Service` exists already.

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
dsimansk: Done in https://github.com/openshift-knative/serverless-operator/pull/381/commits/820abe6684430c955e29172ab12d388b9d979165.

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
dsimansk: Done in https://github.com/openshift-knative/serverless-operator/pull/381/commits/820abe6684430c955e29172ab12d388b9d979165.

dsimansk:
@markusthoemmes gentle ping. Could you please check the above comments if you've got a minute?

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
markusthoemmes: Ah, one would usually just check `Spec` for equality, but I guess this works too :shrug:

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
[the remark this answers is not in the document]
markusthoemmes: from the Get call above I'd think. On the first round, the status won't be set anyway.

knative-operator/Gopkg.toml
```diff
@@ -51,6 +51,10 @@ required = [
 #   name = "knative.dev/pkg"
 #   branch = "release-0.14"
 
+[[override]]
+  name = "knative.dev/serving"
+  branch = "release-0.14"
```
markusthoemmes, line 56: should probably be 0.16 now.

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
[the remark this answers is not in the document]
dsimansk: Ahh, sure I got it now. I get it fetched during reconciliation, just pass it & check the status then.

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
dsimansk: Sounds more reasonable and streamlined, thanks for the suggestion.

dsimansk:
@markusthoemmes as it turned out clean cluster isn't ready for ksvc before actual operator does its job first. Therefore I've tried my best to isolated the deployment. Let me know if it makes sense at all, https://github.com/openshift-knative/serverless-operator/pull/381/commits/7dad56fca1fe022290e0183dad1074a6bc0d1916.

knative-operator/Gopkg.toml
dsimansk: I opted for `0.15` to be on pair with ingress operator and also `knative.dev/operator`.

dsimansk:
@markusthoemmes I think it's complete and ready for the final review, just gentle ping hope you don't mind.

@jcrossley3 maybe a second pair of eyes to review? :)

rhuss:
would be very cool if we get this landed within this week as we are already late with this feature and the sprint closes next week.

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
```diff
@@ -154,75 +153,52 @@ func Delete(instance *servingv1alpha1.KnativeServing, apiclient client.Client, s
...
-				return fmt.Errorf("failed to convert Deployment obj to unstructured: %w", err)
-			}
+// makeKnService makes Knative Service object and its SPEC from provided image parameter
+func makeKnService(image string, instance *servingv1alpha1.KnativeServing) *servingv1.Service {
+	anno := make(map[string]string)
+	if instance != nil {
+		anno = map[string]string{
+			common.ServingOwnerName:      instance.Name,
+			common.ServingOwnerNamespace: instance.Namespace,
 		}
```
markusthoemmes, line 171: I don't think we need these as we can just set an OwnerReference to the KnativeServing instance (we're in the same namespace per below)

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
```diff
@@ -154,75 +153,52 @@ func Delete(instance *servingv1alpha1.KnativeServing, apiclient client.Client, s
...
-	manifest, err := manifest(instance, apiclient, scheme)
-	if err != nil {
-		return err
-	}
-
-	if err := manifest.Delete(); err != nil && !apierrors.IsNotFound(err) {
-		return fmt.Errorf("failed to delete kn ConsoleCLIDownload resources manifest: %w", err)
+	log.Info("Deleting kn ConsoleCLIDownload Service")
+	if err := apiclient.Delete(context.TODO(), makeKnService("", instance)); err != nil && !apierrors.IsNotFound(err) {
+		return fmt.Errorf("failed to delete kn ConsoleCLIDownload Service: %w", err)
```
markusthoemmes, line 158: If you go by my suggestion below, this deletion will be handled via the OwnerReference and we don't need to "manually" delete this service here.

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
```diff
@@ -44,52 +40,55 @@ func Apply(instance *servingv1alpha1.KnativeServing, apiclient client.Client, sc
...
+	case apierrors.IsNotFound(err):
+		tmpService := makeKnService(os.Getenv("IMAGE_KN_CLI_ARTIFACTS"), instance)
+		if err := apiclient.Create(context.TODO(), tmpService); err != nil {
+			return err
+		}
+	case err == nil:
+		tmpService := makeKnService(os.Getenv("IMAGE_KN_CLI_ARTIFACTS"), instance)
+		serviceFromClusterDC := service.DeepCopy()
+		changed := false
+		if !equality.Semantic.DeepEqual(service.Spec, tmpService.Spec.Template.Spec) {
```
markusthoemmes, line 66: This will always be true I believe since you're comparing the template spec to the top-level spec. Please use the top-level spec in both cases.

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
```diff
@@ -44,52 +40,55 @@ func Apply(instance *servingv1alpha1.KnativeServing, apiclient client.Client, sc
...
+			serviceFromClusterDC.Spec = tmpService.Spec
+			changed = true
```
markusthoemmes, line 68: The changed dance seems unnecessary here. I'd move the update call into the `!DeepEqual` if block for simplicity.

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
```diff
@@ -44,52 +40,55 @@ func Apply(instance *servingv1alpha1.KnativeServing, apiclient client.Client, sc
...
 
-// reconcileKnCCDResources reconciles required resources viz Deployment, Service, Route
+// reconcileKnCCDResources reconciles required resources viz Knative Service
 // which will serve kn cross platform binaries within cluster
 func reconcileKnCCDResources(instance *servingv1alpha1.KnativeServing, apiclient client.Client, scheme *runtime.Scheme) error {
-	log.Info("Installing kn ConsoleCLIDownload resources")
-	manifest, err := manifest(instance, apiclient, scheme)
-	if err != nil {
+	if !instance.Status.IsReady() {
+		return fmt.Errorf("Knative instance %q/%q not ready yet", instance.Name, instance.GetNamespace())
```
markusthoemmes, line 47: I don't think we need to return an error here since returning an error will retry immediately. Instead, we can just return `nil` and wait for the informer to kick us again once the KnativeServing became ready.

markusthoemmes (commented):
Mostly quick to fix nits throughout to make stuff a little cleaner/easier.

We do have E2E tests that check if the CLIs are actually downloadable, right?

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
```diff
@@ -99,21 +98,21 @@ func checkResources(manifest *mf.Manifest, instance *servingv1alpha1.KnativeServ
...
+		return fmt.Errorf("failed to find kn ConsoleCLIDownload Knative Service: %v", err)
 	}
-	knRoute := getCanonicalHost(route)
-	// re-verify if the route has a URL
-	if knRoute == "" {
-		return fmt.Errorf("found empty URL of route %q/%q", route.GetNamespace(), route.GetName())
+
+	knRouteURL := knService.Status.URL
+	if knRouteURL == nil || knRouteURL.String() == "" {
+		return fmt.Errorf("failed to get kn ConsoleCLIDownload Knative Service URL")
```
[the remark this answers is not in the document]
dsimansk: As we'd pass actual ksvc reference per the suggestion above. Does it make sense to have such a code? In other words make it more obvious. 
```
func Apply(instance *servingv1alpha1.KnativeServing, apiclient client.Client, scheme *runtime.Scheme) error {
	service := &servingv1.Service{}
	if err := reconcileKnCCDResources(instance, apiclient, scheme, service); err != nil {
		return err
	}
       //
       // check service status here  
       //
	if err := reconcileKnConsoleCLIDownload(apiclient, instance, service); err != nil {
		return err
	}

	return nil
}
```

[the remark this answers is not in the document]
markusthoemmes: I'd actually return the service from the `reconcileXXX` call instead tbh.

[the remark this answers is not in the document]
dsimansk: ...nevermind it shouldn't matter much :)

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
```diff
@@ -99,21 +98,21 @@ func checkResources(manifest *mf.Manifest, instance *servingv1alpha1.KnativeServ
...
-	route := &routev1.Route{}
 	ctx := context.TODO()
 
-	// find the route first
-	if err := apiclient.Get(ctx, client.ObjectKey{Namespace: instance.GetNamespace(), Name: knConsoleCLIDownloadDeployRoute}, route); err != nil {
-		return fmt.Errorf("failed to find kn ConsoleCLIDownload deployment route")
+	// find the Knative Service first
+	knService := &servingv1.Service{}
+	if err := apiclient.Get(ctx, client.ObjectKey{Namespace: instance.GetNamespace(), Name: knConsoleCLIDownloadService}, knService); err != nil {
+		return fmt.Errorf("failed to find kn ConsoleCLIDownload Knative Service: %v", err)
```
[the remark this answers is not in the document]
dsimansk: Done in https://github.com/openshift-knative/serverless-operator/pull/381/commits/f79521812b43c1d2cad86c240249dbb3633dbb16.

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
dsimansk: Per our chat, once switched to OwnerReference approach there's a race condition issue with control-plane deletion that causes route's finalizer to hang and block clean resouce removal. I've added a code comment with the explanation.

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
dsimansk: Keeping the "manual" delete due to reason above.

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
dsimansk: Done in https://github.com/openshift-knative/serverless-operator/pull/381/commits/f79521812b43c1d2cad86c240249dbb3633dbb16.

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
dsimansk: Done in https://github.com/openshift-knative/serverless-operator/pull/381/commits/f79521812b43c1d2cad86c240249dbb3633dbb16.

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
dsimansk: Done in https://github.com/openshift-knative/serverless-operator/pull/381/commits/f79521812b43c1d2cad86c240249dbb3633dbb16.

dsimansk:
> We do have E2E tests that check if the CLIs are actually downloadable, right?

Yes, there's such a [test](https://github.com/openshift-knative/serverless-operator/blob/master/test/servinge2e/verify_deploy_resources_test.go#L18).

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
```diff
@@ -29,29 +29,28 @@ var log = common.Log.WithName("consoleclidownload")
...
 	return nil
 }
 
 // reconcileKnCCDResources reconciles required resources viz Knative Service
 // which will serve kn cross platform binaries within cluster
-func reconcileKnCCDResources(instance *servingv1alpha1.KnativeServing, apiclient client.Client, scheme *runtime.Scheme) error {
+func reconcileKnCCDResources(instance *servingv1alpha1.KnativeServing, apiclient client.Client, scheme *runtime.Scheme, service *servingv1.Service) error {
 	if !instance.Status.IsReady() {
-		return fmt.Errorf("Knative instance %q/%q not ready yet", instance.Name, instance.GetNamespace())
+		// Don't return error, wait silently until Serving instance is ready
```
markusthoemmes, line 46: This now won't have an effect as you'll pass an empty KnativeService down into the second function, which will then error because the service isn't ready yet.

I think we should pull this check to the top-level `Apply`

dsimansk: I should follow my idea from https://github.com/openshift-knative/serverless-operator/pull/381#discussion_r497468806. :)

dsimansk: Addressed in https://github.com/openshift-knative/serverless-operator/pull/381/commits/1725e378dd5bf945ee857089fffbd1d210be922c.

knative-operator/deploy/resources/console_cli_download_kn_resources.yaml
```diff
@@ -1,77 +0,0 @@
-apiVersion: apps/v1
```
markusthoemmes, line 1: I just noticed (sorry for the late comment): What happens to the old things on update? I don't think we're removing them anywhere, are we?

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
```diff
@@ -138,6 +145,25 @@ func Delete(instance *servingv1alpha1.KnativeServing, apiclient client.Client, s
...
 
+// deleteDeprecatedResources removes deprecated resources created by previous versions
+func deleteDeprecatedResources(instance *servingv1alpha1.KnativeServing, apiclient client.Client) error {
+	metaName := metav1.ObjectMeta{
+		Name:      deprecatedResourceName,
+		Namespace: instance.Namespace,
+	}
+	toDelete := []runtime.Object{
+		&appsv1.Deployment{ObjectMeta: metaName},
+		&corev1.Service{ObjectMeta:	metaName},
```
markusthoemmes, line 156: For posterity from discussion in Slack: This will probably remove the K8s Service created by the Knative Service to result in an error loop.

@dsimansk and I agreed to rename the new Knative Service to bypass this issue.

dsimansk: I pushed an update with new name `kn-cli`. Meanwhile I've tested upgrade scenario `1.9->1.10` that passed successfully with old resources deleted, new created and URL in `ConsoleCLIDownload` CR updated.

markusthoemmes (approved):
/lgtm
/approve

PR #382, opened 2020-07-21
SRVCOM-921 Separate serverless tests from third-party test suites

mgencur:
* new script for running just serverless operator and downstream Serving
tests from the same repo
* new Makefile target test-operator which runs the above script
* remove redundant creating of users for upgrade tests
* rename function for running eventing tests to indicate running
upstream tests

The will also be a slight change on ci-operator side to run the "operator-tests" in a Required prow check, the third part test suites will run as optional checks.

The https://github.com/openshift/release/pull/10365 should be merged right after this one so that the operator tests are actually run.


test/e2e-tests.sh
```diff
@@ -21,23 +21,25 @@ failed=0
 (( !failed )) && install_catalogsource || failed=1
 (( !failed )) && logger.success '🚀 Cluster prepared for testing.'
 
+# TEST_ALL used in nightly CI where all tests are run in one go
 if [[ $TEST_ALL == true ]]; then
```
jcrossley3, line 25: Curious why you still need the the TEST_ALL var in this script now that there's the `operator-tests.sh` script?

mgencur: Because the nightly run on OCP 4.3 and 4.4 runs all tests at once: upgrade, operator tests, serving, eventing. Without that we'd need to create one more script just for that purpose.

mgencur: With that being said, we could remove TEST_ALL and create one more script: `all-tests.sh`. That's what I was doing with my previous PR where @nak3 suggested using this TEST_ALL variable. Both solutions are possible.

jcrossley3: I was under the impression that e2e-tests.sh with TEST_ALL=false is basically what operator-tests.sh does.

mgencur: There are these levels (an Makefile targets):
* test-operator - only tests from serverless-operator repo
* test-e2e - only Serving, Eventing test suites (external), but without upgrades
* test-upgrade - only Serving upgrades
* test-all - all of the above together (this one is achieved by the TEST_ALL flag), but it could have a separate script too.

What's your suggestion?

jcrossley3: I think of any given repo as having two types of tests: unit (tests requiring only `go`) and e2e (tests requiring external infrastructure). Our operator has both, so it's a little confusing to have `test-operator` and `test-e2e` targets. The latter I might call `test-upstream` instead. And though I can see how it might be useful to treat `test-upgrade` as its own thing, I really see it as a subset of `test-upstream`.

So in terms of makefile targets, I was thinking:
- test-unit
- test-e2e
- test-upstream

All other targets could be defined in terms of those, e.g. `test-operator` would depend on the first two, and `test-all` would depend on all three. These would just be normal makefile deps between targets, so no need for any extra scripts.

At least, that's kinda how I was imagining it.

jcrossley3: So I was thinking you were using the TEST_ALL var as a way to distinguish between upstream e2e's and downstream (our) e2e's.

mgencur: Yeah. It makes sense. We just needs to distinguish upgrade tests and the other scenarios for CI where they need to run as separate Prow checks. I'm also not sure how the deps between targets will work in CI. The scripts will have to work when called separately and also when called one after another. I'll see how we can get closer to what you're suggesting and still satisfity ci-operator etc.

jcrossley3: so prow can run bash but not make?

mgencur:
/hold

The latest changes will require other changes in ci-operator config. Let's see what we agree on before moving forward with the ci-operator changes.

test/e2e-tests.sh
mgencur: I've tried to address most of your concerns. Had to add some additional make targets for CI but overall I think this is pretty much what you wanted.

mgencur:
@markusthoemmes your review would be welcome as well

test/upstream-e2e-tests.sh
```diff
@@ -0,0 +1,41 @@
...
+(( !failed )) && teardown_serverless || failed=1
+(( !failed )) && install_catalogsource || failed=2
+(( !failed )) && logger.success '🚀 Cluster prepared for testing.'
+
+# Run upgrade tests
+if [[ $TEST_KNATIVE_UPGRADE == true ]]; then
+  (( !failed )) && install_serverless_previous || failed=3
+  (( !failed )) && run_knative_serving_rolling_upgrade_tests || failed=4
+  (( !failed )) && trigger_gc_and_print_knative || failed=5
+  (( !failed )) && teardown_serverless || failed=6
```
mgencur, line 28: I guess this teardown could be called only if `$TEST_KNATIVE_E2E == true` because otherwise we're at the end of the test and we just want to exit. The CI will teardown the cluster. Will fix it

mgencur:
OK. The previous run https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift-knative_serverless-operator/382/pull-ci-openshift-knative-serverless-operator-master-4.5-e2e-aws-ocp-45/1285863607493464064 showed that running just the operator-e2e works. I'll do some temporary changes to exercise the other test scripts (without changing ci-operator) configuration.

jcrossley3:
/hold
I'd like to get #373 merged before I have to rebase these changes, thanks.

PR #385, opened 2020-07-22
Remove diff to avoid panic due to comparing unexported fields.

markusthoemmes:
```
E0722 14:47:03.617271       1 runtime.go:67] Observed a panic: cannot handle unexported field at {v1alpha1.KnativeServingSpec}.CommonSpec.Resources[0].ResourceRequirements.Limits["memory"].i:
	"github.com/openshift-knative/serverless-operator/knative-operator/vendor/k8s.io/apimachinery/pkg/api/resource".Quantity
consider using a custom Comparer; if you control the implementation of type, you can also consider using an Exporter, AllowUnexported, or cmpopts.IgnoreUnexported
```

This can happen if a component races the webhook and we need to add resources from the reconciler. Not good!

### DIFF
--- a/knative-operator/Gopkg.lock
+++ b/knative-operator/Gopkg.lock
@@ -1052,6 +1052,7 @@
     "k8s.io/apimachinery/pkg/api/equality",
     "k8s.io/apimachinery/pkg/api/errors",
     "k8s.io/apimachinery/pkg/api/meta",
+    "k8s.io/apimachinery/pkg/api/resource",
     "k8s.io/apimachinery/pkg/apis/meta/v1",
     "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured",
     "k8s.io/apimachinery/pkg/runtime",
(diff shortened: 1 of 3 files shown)


jcrossley3:
Probably need to run `dep ensure` to remove `cmp` from the lock file.

markusthoemmes:
It didn't do that, but it produced a diff anyway :joy:

jcrossley3:
> It didn't do that, but it produced a diff anyway

Weird. It removed it on mine, and I didn't think master was ahead of me.

markusthoemmes:
/test 4.4-upgrade-tests-aws-ocp-44

Failed during killing the Openshift APIServers

markusthoemmes:
@rhuss see the error message, it's K8s Quantity type: https://github.com/kubernetes/apimachinery/blob/3f0ac6a8e1122b68c04c77e8947563cc8a8896bc/pkg/api/resource/quantity.go#L88-L99

rhuss:
I see. However, I would expect any diff algorithm just to ignore private fields.

PR #386, opened 2020-07-24
[SRVLS-264] Create knative components in openshift-serverless

jcrossley3:
Most of the complexity is to account for existing installations in the old namespaces being upgraded with minimal service disruption. See commits for details.

### DIFF
--- a/hack/lib/serverless.bash
+++ b/hack/lib/serverless.bash
@@ -3,6 +3,10 @@
 function ensure_serverless_installed {
   logger.info 'Check if Serverless is installed'
   local prev=${1:-false}
+  if [[ $prev == "true" ]]; then
+    export SERVING_NAMESPACE="knative-serving"
+    export EVENTING_NAMESPACE="knative-eventing"
+  fi
   if oc get knativeserving.operator.knative.dev knative-serving -n "${SERVING_NAMESPACE}" >/dev/null 2>&1 && \
      oc get knativeeventing.operator.knative.dev knative-eventing -n "${EVENTING_NAMESPACE}" >/dev/null 2>&1
   then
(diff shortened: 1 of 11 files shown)


jcrossley3:
Copying over @nak3 's review...

> Looks good to me overall.
> But there seems to be one issue. Serving and Eventing have the same name configmaps like `config-logging`, `config-tracing`, (and `config-leader-election`).
> 
> e.g. config-logging
> https://github.com/openshift-knative/operator/blob/release-0.15/cmd/operator/kodata/knative-eventing/0.15.2/1-eventing.yaml#L403
> https://github.com/openshift-knative/operator/blob/release-0.15/cmd/operator/kodata/knative-serving/0.15.2/2-serving-core.yaml#L673
> 
> It seems that they will be conflicted when both knativeserving and knativeeventing CR had same configuration but different values.

Great point, @nak3 -- and I'm not a little embarrassed to say I hadn't even considered that possibility. :)

I verified that these are the only conflicting resources (in 0.15 and 0.16):
* config-leader-election /v1, Kind=ConfigMap
* config-logging /v1, Kind=ConfigMap
* config-observability /v1, Kind=ConfigMap
* config-tracing /v1, Kind=ConfigMap

And there is definitely potential for any of those to be overwritten when either knative component CR is reconciled. In light of this, I'd like confirmation from arch/prod as to whether having serving and eventing share the same `openshift-serverless` namespace is a goal worth pursuing. /cc @pmorie @markito @bbrowning @markusthoemmes 

If it is, then I guess we can either work with upstream to ensure these configmap names are configurable or we can document that any modifications to the above CM's must be made not in the CM's themselves but in *both* the `KnativeServing` and `KnativeEventing` CR's because whichever is the last to reconcile will "win". Or something else?

jcrossley3:
nfi

jcrossley3:
nfi
/retest

jcrossley3:
failed to acquire lease? what lease?
/retest

jcrossley3:
can't find config-autoscaler CM?
/retest

jcrossley3:
i think I got past this point before, so...
/retest

markusthoemmes:
So the config-map clashes are *really* unfortunate :thinking:. I think the operator would have to be smart to merge the two together and error smartly on conflicts (i.e. if serving specifies a different logging level than eventing).

markusthoemmes:
I guess we have kinda dropped the ball on this one and a ton has changed meanwhile. @jcrossley3 any idea on how to move forward with this? Moving stuff into the -serverless namespace ain't a prio right now.

matzew:
Do we want to keep this? or are we fine in closing this PR ?

matzew:
@jcrossley3 should we close this ? :smile: :see_no_evil:

jcrossley3:
Yes. Probably making incorrect assumptions about both upstream and downstream at this point.

PR #388, opened 2020-07-27
[SRVKS-495] Add readiness metrics and dashboard

skonto:
-  Adds a dashboard for monitoring status for serving and eventing:

image

Killing some pods: `oc delete pods --all --grace-period=0 --force -n knative-serving`. The kantive-serving operator will re-launch the pods and we should see some down time period.

image

Notes: 
 * Right now wrt resource ownership, managed cm, role, rolebindings are owned by the serverless operator deployment. Although cross-namespace ownership is not allowed in general it does not create issues and the related dashboard cm is removed with uninstall (might need to change this but need better alternatives). For the roles etc these could be owned by the related service which owns also the service monitor, but the idea was to have these resources independent of the monitor in case the latter is re-installed for some reason.
* Spotted some issues with Grafana on Openshift as a) the custom dashboards for knative dont show in the Grafana UI, this refers also to the pre-existed knative grafana dashboard, b) applying changes to the dashboard for tuning the y-axis `show` setting has no effect. The latter works fine in a local grafana instance as expected. Tried `Stat` didnt work either (better for showing Up/Down status,  although you the lose timeline).

cc @markusthoemmes

### DIFF
--- a/knative-operator/cmd/manager/main.go
+++ b/knative-operator/cmd/manager/main.go
@@ -134,24 +134,24 @@ func main() {
 func setupMonitoring(ctx context.Context, cfg *rest.Config) error {
 	config, err := rest.InClusterConfig()
 	if err != nil {
-		return fmt.Errorf("failed to create cluster config", err)
+		return fmt.Errorf("failed to create cluster config: %w", err)
 	}
 	cl, err := client.New(config, client.Options{})
 
 	if err != nil {
-		return fmt.Errorf("failed to create a client", err)
+		return fmt.Errorf("failed to create a client: %w", err)
 	}
 
 	if err = common.SetUpMonitoringRequirements(cl); err != nil {
-		return fmt.Errorf("failed to setup monitoring resources", err)
+		return fmt.Errorf("failed to setup monitoring resources: %w", err)
 	}
 
-	if err = common.SetupServiceMonitor(ctx, cfg, cl, metricsPort, operatorMetricsPort); err != nil {
-		return fmt.Errorf("failed to setup the Service monitor", err)
+	if err := common.SetupServiceMonitor(ctx, cfg, cl, metricsPort, operatorMetricsPort); err != nil {
+		return fmt.Errorf("failed to setup the Service monitor: %w", err)
 	}
 
-	if err = common.InstallHealthDashboard(cl); err != nil {
-		return fmt.Errorf("failed to setup the Knative Health Status Dashboard", err)
+	if err := common.InstallHealthDashboard(cl); err != nil {
+		return fmt.Errorf("failed to setup the Knative Health Status Dashboard: %w", err)
 	}
 	return nil
 }
--- a/knative-operator/cmd/manager/main.go
+++ b/knative-operator/cmd/manager/main.go
@@ -8,20 +8,18 @@ import (
 	"runtime"
 
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/apis"
+	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/common"
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/controller"
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/webhook"
-	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
-	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/common"
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 	"github.com/operator-framework/operator-sdk/pkg/leader"
 	"github.com/operator-framework/operator-sdk/pkg/log/zap"
 	"github.com/operator-framework/operator-sdk/pkg/restmapper"
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
 	"github.com/spf13/pflag"
+	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
-
 	"k8s.io/client-go/rest"
-
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -153,7 +151,7 @@ func setupMonitoring(ctx context.Context, cfg *rest.Config) error {
 	}
 
 	if err = common.InstallHealthDashboard(cl); err != nil {
-		return fmt.Errorf( "failed to setup the Knative Health Status Dashboard", err)
+		return fmt.Errorf("failed to setup the Knative Health Status Dashboard", err)
 	}
 	return nil
 }
--- a/knative-operator/cmd/manager/main.go
+++ b/knative-operator/cmd/manager/main.go
@@ -7,24 +7,22 @@ import (
 	"os"
 	"runtime"
 
-	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
-	_ "k8s.io/client-go/plugin/pkg/client/auth"
-	"k8s.io/client-go/rest"
-
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/apis"
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/controller"
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/webhook"
-
+	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
+	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/common"
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
-	kubemetrics "github.com/operator-framework/operator-sdk/pkg/kube-metrics"
 	"github.com/operator-framework/operator-sdk/pkg/leader"
 	"github.com/operator-framework/operator-sdk/pkg/log/zap"
-	"github.com/operator-framework/operator-sdk/pkg/metrics"
 	"github.com/operator-framework/operator-sdk/pkg/restmapper"
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
 	"github.com/spf13/pflag"
-	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/util/intstr"
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
+
+	"k8s.io/client-go/rest"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
@@ -122,34 +120,8 @@ func main() {
 		os.Exit(1)
 	}
 
-	// Commented below to avoid a stream of these errors at startup:
-	// E1021 22:50:03.372487       1 reflector.go:134] github.com/operator-framework/operator-sdk/pkg/kube-metrics/collector.go:67: Failed to list *unstructured.Unstructured: the server could not find the requested resource
-	// if err = serveCRMetrics(cfg); err != nil {
-	// 	log.Info("Could not generate and serve custom resource metrics", "error", err.Error())
-	// }
-
-	// Add to the below struct any other metrics ports you want to expose.
-	servicePorts := []v1.ServicePort{
-		{Port: metricsPort, Name: metrics.OperatorPortName, Protocol: v1.ProtocolTCP, TargetPort: intstr.IntOrString{Type: intstr.Int, IntVal: metricsPort}},
-		{Port: operatorMetricsPort, Name: metrics.CRPortName, Protocol: v1.ProtocolTCP, TargetPort: intstr.IntOrString{Type: intstr.Int, IntVal: operatorMetricsPort}},
-	}
-	// Create Service object to expose the metrics port(s).
-	service, err := metrics.CreateMetricsService(ctx, cfg, servicePorts)
-	if err != nil {
-		log.Info("Could not create metrics Service", "error", err.Error())
-	}
-
-	// CreateServiceMonitors will automatically create the prometheus-operator ServiceMonitor resources
-	// necessary to configure Prometheus to scrape metrics from this operator.
-	services := []*v1.Service{service}
-	_, err = metrics.CreateServiceMonitors(cfg, namespace, services)
-	if err != nil {
-		log.Info("Could not create ServiceMonitor object", "error", err.Error())
-		// If this operator is deployed to a cluster without the prometheus-operator running, it will return
-		// ErrServiceMonitorNotPresent, which can be used to safely skip ServiceMonitor creation.
-		if err == metrics.ErrServiceMonitorNotPresent {
-			log.Info("Install prometheus-operator in your cluster to create ServiceMonitor objects", "error", err.Error())
-		}
+	if err := setupMonitoring(ctx, cfg); err != nil {
+		log.Error(err, "Failed to start monitoring")
 	}
 
 	log.Info("Starting the Cmd.")
@@ -161,26 +133,27 @@ func main() {
 	}
 }
 
-// serveCRMetrics gets the Operator/CustomResource GVKs and generates metrics based on those types.
-// It serves those metrics on "http://metricsHost:operatorMetricsPort".
-func serveCRMetrics(cfg *rest.Config) error {
-	// Below function returns filtered operator/CustomResource specific GVKs.
-	// For more control override the below GVK list with your own custom logic.
-	filteredGVK, err := k8sutil.GetGVKsFromAddToScheme(apis.AddToScheme)
+func setupMonitoring(ctx context.Context, cfg *rest.Config) error {
+	config, err := rest.InClusterConfig()
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to create cluster config", err)
 	}
-	// Get the namespace the operator is currently deployed in.
-	operatorNs, err := k8sutil.GetOperatorNamespace()
+	cl, err := client.New(config, client.Options{})
+
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to create a client", err)
 	}
-	// To generate metrics in other namespaces, add the values below.
-	ns := []string{operatorNs}
-	// Generate and serve custom resource specific metrics.
-	err = kubemetrics.GenerateAndServeCRMetrics(cfg, ns, filteredGVK, metricsHost, operatorMetricsPort)
-	if err != nil {
-		return err
+
+	if err = common.SetUpMonitoringRequirements(cl); err != nil {
+		return fmt.Errorf("failed to setup monitoring resources", err)
+	}
+
+	if err = common.SetupServiceMonitor(ctx, cfg, cl, metricsPort, operatorMetricsPort); err != nil {
+		return fmt.Errorf("failed to setup the Service monitor", err)
+	}
+
+	if err = common.InstallHealthDashboard(cl); err != nil {
+		return fmt.Errorf( "failed to setup the Knative Health Status Dashboard", err)
 	}
 	return nil
 }
--- a/knative-operator/cmd/manager/main.go
+++ b/knative-operator/cmd/manager/main.go
@@ -142,7 +142,7 @@ func setupMonitoring(ctx context.Context, cfg *rest.Config) error {
 		return fmt.Errorf("failed to create a client: %w", err)
 	}
 
-	if err = common.SetUpMonitoringRequirements(cl); err != nil {
+	if err = common.SetupMonitoringRequirements(cl); err != nil {
 		return fmt.Errorf("failed to setup monitoring resources: %w", err)
 	}
 
--- a/knative-operator/deploy/resources/dashboards/grafana-dash-knative-health.yaml
+++ b/knative-operator/deploy/resources/dashboards/grafana-dash-knative-health.yaml
@@ -1,17 +1,3 @@
-# Copyright 2018 The Knative Authors
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#     https://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
 apiVersion: v1
 kind: ConfigMap
 metadata:
--- a/knative-operator/deploy/resources/dashboards/grafana-dash-knative-health.yaml
+++ b/knative-operator/deploy/resources/dashboards/grafana-dash-knative-health.yaml
@@ -28,6 +28,164 @@ data:
       "links": [],
       "panels": [
         {
+          "cacheTimeout": null,
+          "colorBackground": false,
+          "colorValue": false,
+          "colors": [
+           "rgba(245, 54, 54, 0.9)",
+           "rgba(237, 129, 40, 0.89)",
+           "rgba(50, 172, 45, 0.97)"
+          ],
+          "datasource": "prometheus",
+          "editable": false,
+          "error": false,
+           "format": "none",
+           "gauge": {
+             "maxValue": 100,
+             "minValue": 0,
+             "show": false,
+             "thresholdLabels": false,
+             "thresholdMarkers": true
+           },
+           "id": 28,
+           "interval": null,
+           "isNew": true,
+           "links": [],
+           "mappingType": 1,
+           "mappingTypes": [
+           {
+             "name": "value to text",
+             "value": 1
+           },
+           {
+             "name": "range to text",
+             "value": 2
+           }
+           ],
+           "maxDataPoints": 100,
+           "nullPointMode": "connected",
+           "nullText": null,
+           "postfix": "",
+           "postfixFontSize": "50%",
+           "prefix": "",
+           "prefixFontSize": "50%",
+           "rangeMaps": [
+           {
+             "from": "null",
+             "text": "N/A",
+             "to": "null"
+           }
+           ],
+           "span": 6,
+           "sparkline": {
+             "fillColor": "rgba(31, 118, 189, 0.18)",
+             "full": false,
+             "lineColor": "rgb(31, 120, 193)",
+             "show": false
+           },
+           "targets": [
+           {
+             "expr": "sum(count(kube_pod_status_ready{namespace=\"knative-serving\", pod=~\".*\", condition=\"true\"} == 1))",
+             "intervalFactor": 2,
+             "legendFormat": "",
+             "metric": "etcd_server_has_leader",
+             "refId": "A",
+             "step": 20
+           }
+           ],
+           "thresholds": "",
+           "title": "Up (Knative Serving)",
+           "type": "singlestat",
+           "valueFontSize": "200%",
+           "valueMaps": [
+           {
+             "op": "=",
+             "text": "N/A",
+             "value": "null"
+           }
+           ],
+           "valueName": "avg"
+         },
+         {
+           "cacheTimeout": null,
+           "colorBackground": false,
+           "colorValue": false,
+           "colors": [
+             "rgba(245, 54, 54, 0.9)",
+               "rgba(237, 129, 40, 0.89)",
+               "rgba(50, 172, 45, 0.97)"
+             ],
+           "datasource": "prometheus",
+           "editable": false,
+           "error": false,
+           "format": "none",
+           "gauge": {
+             "maxValue": 100,
+             "minValue": 0,
+             "show": false,
+             "thresholdLabels": false,
+             "thresholdMarkers": true
+           },
+           "id": 29,
+           "interval": null,
+           "isNew": true,
+           "links": [],
+               "mappingType": 1,
+               "mappingTypes": [
+               {
+                 "name": "value to text",
+                 "value": 1
+               },
+               {
+                 "name": "range to text",
+                 "value": 2
+               }
+               ],
+               "maxDataPoints": 100,
+               "nullPointMode": "connected",
+               "nullText": null,
+               "postfix": "",
+               "postfixFontSize": "50%",
+               "prefix": "",
+               "prefixFontSize": "50%",
+               "rangeMaps": [
+               {
+                 "from": "null",
+                 "text": "N/A",
+                 "to": "null"
+               }
+               ],
+               "span": 6,
+               "sparkline": {
+                 "fillColor": "rgba(31, 118, 189, 0.18)",
+                 "full": false,
+                 "lineColor": "rgb(31, 120, 193)",
+                 "show": false
+               },
+               "targets": [
+               {
+                 "expr": "sum(count(kube_pod_status_ready{namespace=\"knative-eventing\", pod=~\".*\", condition=\"true\"} == 1))",
+                 "intervalFactor": 2,
+                 "legendFormat": "",
+                 "metric": "etcd_server_has_leader",
+                 "refId": "A",
+                 "step": 20
+               }
+               ],
+               "thresholds": "",
+               "title": "Up (Knative Eventing)",
+               "type": "singlestat",
+               "valueFontSize": "200%",
+               "valueMaps": [
+               {
+                 "op": "=",
+                 "text": "N/A",
+                 "value": "null"
+               }
+               ],
+               "valueName": "avg"
+             },
+        {
           "aliasColors": {},
           "bars": false,
           "dashLength": 10,
@@ -60,6 +218,7 @@ data:
           },
           "lines": true,
           "linewidth": 1,
+          "span": 6,
           "nullPointMode": "null",
           "percentage": false,
           "pluginVersion": "7.1.0",
@@ -72,10 +231,10 @@ data:
           "steppedLine": false,
           "targets": [
             {
-              "expr": "knative_serving_ready{namespace=\"$namespace\", service=\"knative-openshift-metrics\"}",
+              "expr": "knative_up{namespace=\"$namespace\", service=\"knative-openshift-metrics\", type=\"serving_status\"}",
               "format": "time_series",
               "interval": "",
-              "legendFormat": "Ready status - 1 is up, 0 is down",
+              "legendFormat": "Serving status - 1 is up, 0 is down",
               "refId": "A"
             }
           ],
@@ -161,16 +320,17 @@ data:
        "pointradius": 2,
        "points": false,
        "renderer": "flot",
+       "span": 6,
        "seriesOverrides": [],
        "spaceLength": 10,
        "stack": false,
        "steppedLine": false,
        "targets": [
          {
-           "expr": "knative_eventing_ready{namespace=\"$namespace\", service=\"knative-openshift-metrics\"}",
+           "expr": "knative_up{namespace=\"$namespace\", service=\"knative-openshift-metrics\", type=\"eventing_status\"}",
            "format": "time-series",
            "interval": "",
-           "legendFormat": "Ready status - 1 is up, 0 is down",
+           "legendFormat": "Eventing status - 1 is up, 0 is down",
            "refId": "B"
          }
        ],
@@ -231,7 +391,7 @@ data:
             "multi": false,
             "name": "namespace",
             "options": [],
-            "query": "label_values(knative_serving_ready{namespace!=\"unknown\"}, namespace)",
+            "query": "label_values(knative_up{namespace!=\"unknown\"}, namespace)",
             "refresh": 1,
             "regex": "",
             "skipUrlSync": false,
--- a/knative-operator/deploy/resources/dashboards/grafana-dash-knative-health.yaml
+++ b/knative-operator/deploy/resources/dashboards/grafana-dash-knative-health.yaml
@@ -27,85 +27,6 @@ data:
       "graphTooltip": 0,
       "links": [],
       "panels": [
-        {
-          "cacheTimeout": null,
-          "colorBackground": false,
-          "colorValue": false,
-          "colors": [
-           "rgba(245, 54, 54, 0.9)",
-           "rgba(237, 129, 40, 0.89)",
-           "rgba(50, 172, 45, 0.97)"
-          ],
-          "datasource": "prometheus",
-          "editable": false,
-          "error": false,
-           "format": "none",
-           "gauge": {
-             "maxValue": 100,
-             "minValue": 0,
-             "show": false,
-             "thresholdLabels": false,
-             "thresholdMarkers": true
-           },
-           "id": 28,
-           "interval": null,
-           "isNew": true,
-           "links": [],
-           "mappingType": 1,
-           "mappingTypes": [
-           {
-             "name": "value to text",
-             "value": 1
-           },
-           {
-             "name": "range to text",
-             "value": 2
-           }
-           ],
-           "maxDataPoints": 100,
-           "nullPointMode": "connected",
-           "nullText": null,
-           "postfix": "",
-           "postfixFontSize": "50%",
-           "prefix": "",
-           "prefixFontSize": "50%",
-           "rangeMaps": [
-           {
-             "from": "null",
-             "text": "N/A",
-             "to": "null"
-           }
-           ],
-           "span": 6,
-           "sparkline": {
-             "fillColor": "rgba(31, 118, 189, 0.18)",
-             "full": false,
-             "lineColor": "rgb(31, 120, 193)",
-             "show": false
-           },
-           "targets": [
-           {
-             "expr": "sum(count(kube_pod_status_ready{namespace=\"knative-serving\", pod=~\".*\", condition=\"true\"} == 1))",
-             "intervalFactor": 2,
-             "legendFormat": "",
-             "metric": "etcd_server_has_leader",
-             "refId": "A",
-             "step": 20
-           }
-           ],
-           "thresholds": "",
-           "title": "Up (Knative Serving)",
-           "type": "singlestat",
-           "valueFontSize": "200%",
-           "valueMaps": [
-           {
-             "op": "=",
-             "text": "N/A",
-             "value": "null"
-           }
-           ],
-           "valueName": "avg"
-         },
          {
            "cacheTimeout": null,
            "colorBackground": false,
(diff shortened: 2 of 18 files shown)


knative-operator/cmd/manager/main.go
```diff
@@ -161,26 +133,27 @@ func main() {
...
-	filteredGVK, err := k8sutil.GetGVKsFromAddToScheme(apis.AddToScheme)
+func setupMonitoring(ctx context.Context, cfg *rest.Config) error {
+	config, err := rest.InClusterConfig()
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to create cluster config", err)
 	}
-	// Get the namespace the operator is currently deployed in.
-	operatorNs, err := k8sutil.GetOperatorNamespace()
+	cl, err := client.New(config, client.Options{})
```
skonto, line 141: Need to do this since manager's client wont work.

knative-operator/pkg/common/health_dashboard.go
```diff
@@ -0,0 +1,89 @@
...
+	} else if err != nil {
+		return fmt.Errorf("failed to get namepsace %q: %w", ConfigManagedNamespace, err)
+	}
+
+	manifest, err := manifest(instance, api, namespace)
+	if err != nil {
+		return fmt.Errorf("failed to load dashboard manifest: %w", err)
+	}
+	logh.Info("Installing dashboard")
+	logh.Info(fmt.Sprintf("Manifest before apply: %v in ns", manifest, namespace))
```
markusthoemmes, line 43: Isn't this logline fairly noisy?

knative-operator/pkg/common/health_dashboard.go
```diff
@@ -0,0 +1,89 @@
...
+
+	instance.SetNamespace(namespace)
+
+	transforms := []mf.Transformer{mf.InjectOwner(instance)}
+	if len(ConfigManagedNamespace) > 0 {
+		transforms = append(transforms, mf.InjectNamespace(ConfigManagedNamespace))
+	}
+
+	if manifest, err = manifest.Transform(transforms...); err != nil {
+		log.Error(err, "Unable to transform role and roleBinding serviceMonitor manifest")
```
markusthoemmes, line 72: No need to log the error here I believe, we should log at the topmost call just once.

knative-operator/pkg/common/monitoring.go
```diff
@@ -0,0 +1,124 @@
...
+
+const (
+	// OpenshiftServerlessInstalledNamespace is the ns where knative serving operator has been installed
+	OpenshiftServerlessInstalledNamespace = "NAMESPACE"
+	// service monitor created successfully when monitoringLabel added to namespace
+	monitoringLabel = "openshift.io/cluster-monitoring"
+	rolePath        = "deploy/role_service_monitor.yaml"
+)
+
+func AddMonitoringLabelToNamespace(namespace string, api client.Client) error {
```
markusthoemmes, line 28: Need to export?

knative-operator/pkg/common/monitoring.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	ns.Labels[monitoringLabel] = "true"
+	if err := api.Update(context.TODO(), ns); err != nil {
+		log.Error(err, fmt.Sprintf("could not add label %q to namespace %q", monitoringLabel, namespace))
+		return err
+	}
+	return nil
+}
+
+func CreateRoleAndRoleBinding(instance *appsv1.Deployment, namespace, path string, client client.Client) error {
```
markusthoemmes, line 44: Need to export?

knative-operator/pkg/common/monitoring.go
```diff
@@ -0,0 +1,124 @@
...
+	instance.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "apps",
+		Version: "v1",
+		Kind:    "Deployment",
+	})
+	// this is typical probably not needed as uid is enough for ownership
+	instance.SetNamespace(namespace)
+
+	transforms := []mf.Transformer{mf.InjectOwner(instance)}
+	if len(namespace) > 0 {
```
markusthoemmes, line 59:
```suggestion
	if namespace != "" {
```

knative-operator/pkg/common/monitoring.go
```diff
@@ -0,0 +1,124 @@
...
+
+func GetOperatorNamespace() (string, error) {
+	ns, found := os.LookupEnv(OpenshiftServerlessInstalledNamespace)
+	if !found {
+		return "", fmt.Errorf("%s must be set", OpenshiftServerlessInstalledNamespace)
+	}
+	return ns, nil
+}
+
+func SetUpMonitoringRequirements(api client.Client) error {
```
markusthoemmes, line 82: Need to export?

knative-operator/cmd/manager/main.go
```diff
@@ -161,26 +133,27 @@ func main() {
...
+		return fmt.Errorf("failed to create a client", err)
 	}
-	// To generate metrics in other namespaces, add the values below.
-	ns := []string{operatorNs}
-	// Generate and serve custom resource specific metrics.
-	err = kubemetrics.GenerateAndServeCRMetrics(cfg, ns, filteredGVK, metricsHost, operatorMetricsPort)
-	if err != nil {
-		return err
+
+	if err = common.SetUpMonitoringRequirements(cl); err != nil {
```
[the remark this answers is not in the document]
skonto: @markusthoemmes what is the comment about, I see no difference in the suggested change.

knative-operator/pkg/common/health_dashboard.go
```diff
@@ -0,0 +1,89 @@
...
+	instance.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "apps",
+		Version: "v1",
+		Kind:    "Deployment",
+	})
+
+	instance.SetNamespace(namespace)
+
+	transforms := []mf.Transformer{mf.InjectOwner(instance)}
+	if len(ConfigManagedNamespace) > 0 {
```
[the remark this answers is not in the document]
skonto: This pre-existed.

knative-operator/pkg/common/monitoring.go
```diff
@@ -0,0 +1,124 @@
...
+	if resource.GetKind() != "RoleBinding" {
+		return nil
+	}
+	var rb = &rbacv1.RoleBinding{}
+	if err := scheme.Scheme.Convert(resource, rb, nil); err != nil {
+		return err
+	}
+	sub := rb.Subjects[0]
+	sub.Namespace = "openshift-monitoring"
+	rb.Subjects = []rbacv1.Subject{sub}
```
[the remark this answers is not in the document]
skonto: It is already in the related yaml file but it was being removed somehow, let me double check.

knative-operator/pkg/common/health_dashboard.go
skonto: correct leftover from my debugging.

knative-operator/pkg/common/monitoring.go
skonto: yes, it is used from another package.

knative-operator/pkg/common/monitoring.go
```diff
@@ -0,0 +1,124 @@
...
+		return err
+	}
+	err = CreateRoleAndRoleBinding(d, ns, rolePath, api)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func GetServerlessOperatorDeployment(api client.Client, namespace string) (*appsv1.Deployment, error) {
```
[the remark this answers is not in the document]
skonto: > Why not? Maybe add a comment.
You mean why I use a new client?

knative-operator/pkg/common/monitoring.go
```diff
@@ -0,0 +1,124 @@
...
+	deployment := &appsv1.Deployment{}
+	key := types.NamespacedName{Name: "knative-openshift", Namespace: namespace}
```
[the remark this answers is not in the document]
skonto: Yes doable.

knative-operator/pkg/common/monitoring.go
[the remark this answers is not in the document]
skonto: Ok so the manifest transformation patches everything, wherever it finds a `namespace` it will replace it. But the problem here is that the subresource needs a different namespace than the role. If I remove that thing I get:
```
$ oc describe rolebindings knative-serving-prometheus-k8s -n openshift-serverless
Name:         knative-serving-prometheus-k8s
Labels:       <none>
Annotations:  manifestival: new
Role:
  Kind:  Role
  Name:  knative-serving-prometheus-k8s
Subjects:
  Kind            Name            Namespace
  ----            ----            ---------
  ServiceAccount  prometheus-k8s  openshift-serverless
```

knative-operator/cmd/manager/main.go
skonto: Let me double check because for some reason it didnt bring back the namespace object.

skonto: The client coming from the manager is a [split client](https://sdk.operatorframework.io/docs/golang/references/client/#default-client) so it reads from cache, the error I get is:

```
{"level":"error","ts":"2020-07-30T11:20:54.750Z","logger":"cmd","msg":"Failed to start monitoring","error":"failed to setup monitoring resources: Namespace \"openshift-serverless\" not found","stacktrace":"github.com/openshift-knative/serverless-operator/knative-operator/vendor/github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/src/github.com/openshift-knative/serverless-operator/knative-operator/vendor/github.com/go-logr/zapr/zapr.go:128\nmain.main\n\t/go/src/github.com/openshift-knative/serverless-operator/knative-operator/cmd/manager/main.go:122\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:203"}
```
I will try check if cache sync works here, because I suspect cache prohibits getting the resource, although it shouldnt.

knative-operator/pkg/common/monitoring.go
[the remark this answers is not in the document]
markusthoemmes: Yep

knative-operator/cmd/manager/main.go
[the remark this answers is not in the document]
markusthoemmes: Lowercase "u" in setup.

knative-operator/cmd/manager/main.go
skonto: @markusthoemmes  Ok now I recall at this point cache has not been started so there is no way to make it work. Usually clients are used in reconciliation funcs where things are setup for you. Since mgr.Start() is blocking I cannot use this client after.

olm-catalog/serverless-operator/1.8.0/serverless-operator.v1.8.0.clusterserviceversion.yaml
```diff
@@ -325,6 +325,12 @@ spec:
                   env:
                     - name: WATCH_NAMESPACE
                       value: ""
+                    - name: NAMESPACE
+                      valueFrom:
+                        fieldRef:
+                          fieldPath: metadata.namespace
+                    - name: DEPLOYMENT_NAME
+                      value: "knative-openshift"
```
skonto, line 333: injecting it here since its convenient.

knative-operator/pkg/common/monitoring.go
```diff
@@ -0,0 +1,135 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// installedNamespaceEnvKey is the ns where Openshift serverless operator has been installed
+	installedNamespaceEnvKey = "NAMESPACE"
+	// installedNamespaceEnvKey is the ns where Openshift serverless operator has been installed
```
markusthoemmes, line 21: wrong comment :)

knative-operator/pkg/common/health_dashboard.go
```diff
@@ -0,0 +1,86 @@
...
+		Version: "v1",
+		Kind:    "Deployment",
+	})
+
+	instance.SetNamespace(namespace)
+
+	transforms := []mf.Transformer{mf.InjectOwner(instance)}
+	if ConfigManagedNamespace != "" {
+		transforms = append(transforms, mf.InjectNamespace(ConfigManagedNamespace))
+	}
```
markusthoemmes, line 68: Can you elaborate why this cross-namespace ownerRef is benign? We've been bitten **very** hard by wrong ownerRefs in the past, so I'd rather not set wrong ones again.

knative-operator/pkg/common/monitoring.go
```diff
@@ -0,0 +1,135 @@
...
+
+func getOperatorDeploymentName() (string, error) {
+	ns, found := os.LookupEnv(operatorDeploymentNameEnvKey)
+	if !found {
+		return "", fmt.Errorf("%s must be set", operatorDeploymentNameEnvKey)
+	}
+	return ns, nil
+}
+
+func injectNameSpace(namespace string) mf.Transformer {
```
markusthoemmes, line 127: Ah comment as to why this custom version is needed would be helpful I believe.

knative-operator/pkg/common/health_dashboard.go
skonto: I set this cross namespace reference so the dashboard cm is removed when the serverless operator deployment is removed (this is managed externally). I know in general there can be bugs due to this (I checked the history on the K8s project), but the only alternative I see is when we start the serverless operator to check if that resource exists and overwrite it. However, I cant think of a scenario where this can be harmful.

knative-operator/pkg/common/monitoring.go
skonto: Added.

knative-operator/pkg/common/health_dashboard.go
```diff
@@ -0,0 +1,86 @@
...
+func InstallHealthDashboard(api client.Client) error {
+	namespace, err := getOperatorNamespace()
+	if err != nil {
+		return err
+	}
+	instance, err := getServerlessOperatorDeployment(api, namespace)
+	if err != nil {
+		return err
+	}
+	err = api.Get(context.TODO(), client.ObjectKey{Name: ConfigManagedNamespace}, &corev1.Namespace{})
```
nak3, line 30: The `openshift-config-managed` namespace verification should move up at L#22.

knative-operator/pkg/common/health_dashboard.go
```diff
@@ -0,0 +1,86 @@
...
+	instance.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "apps",
+		Version: "v1",
+		Kind:    "Deployment",
+	})
+
+	instance.SetNamespace(namespace)
+
+	transforms := []mf.Transformer{mf.InjectOwner(instance)}
+	if ConfigManagedNamespace != "" {
```
nak3, line 66: Is this check necessary even though `ConfigManagedNamespace` is const value? But anyway as @markusthoemmes pointed out cross-namespace ownerRef injection must be solved.

knative-operator/pkg/common/monitoring.go
```diff
@@ -0,0 +1,137 @@
...
+	ns, found := os.LookupEnv(testRolePath)
+	if found {
+		return ns
+	} else {
+		return rolePath
+	}
+}
+
+func getOperatorNamespace() (string, error) {
+	ns, found := os.LookupEnv(installedNamespaceEnvKey)
```
nak3, line 60: `os.LookupEnv` returns `true` even when empty value (e.g. `NAMESPACE: ""`) sets.

We should use  `os.Getenv` instead?
```go
	ns := os.Getenv("NAMESPACE")
	if ns == "" {
		return "", fmt.Errorf("%s must be set", installedNamespaceEnvKey)
	}
```

knative-operator/pkg/common/monitoring.go
```diff
@@ -0,0 +1,137 @@
...
+	if !found {
+		return "", fmt.Errorf("%s must be set", installedNamespaceEnvKey)
```
nak3, line 62:
```suggestion
		return "", fmt.Errorf("The environment variable %q must be set", installedNamespaceEnvKey)
```

nak3:
It seems that the dashboard will never reconcile. If I deleted the configmap, it does not come back even after restarting operator pods.

```
oc delete -n openshift-config-managed  cm grafana-dashboard-definition-knative-health
```

skonto:
@nak3 the dashboard is not going to be reconciled because it is not part of knative serving lifecycle or eventing lifecycle its part of the serverless operator lifecycle.

knative-operator/pkg/common/health_dashboard.go
skonto: I also described the cross-namespace issue in the description of this PR,  I agree we need a solution for that to be on the safe side. Do you see other alternatives beyond the ones I listed?

skonto: I will remove the check.

knative-operator/pkg/common/health_dashboard.go
skonto: fixed

skonto:
Removed the owner ref and added a controller for managing the lifecycle of the dashboard.

knative-operator/pkg/controller/healthdashboard/healthdashboard_controller.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// add adds a new Controller to mgr with r as the reconcile.Reconciler
+func add(mgr manager.Manager, r reconcile.Reconciler) error {
+	// Create a new controller
+	c, err := controller.New("healthdashboard-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
```
skonto, line 41: probably I cold run reconcile here to install the dashboard instead of doing that in `main`.

skonto:
@nak3 have you seen this before?
```{"component":"entrypoint","file":"prow/entrypoint/run.go:168","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Entrypoint received interrupt: terminated","severity":"error","time":"2020-08-03T01:12:39Z"}
```

nak3:
@skonto It seems a CI's infra issue. Today's CI is not stable somehow :crying_cat_face:

knative-operator/pkg/common/monitoring.go
```diff
@@ -0,0 +1,136 @@
...
+	// meant for testing only
+	ns, found := os.LookupEnv(testRolePath)
+	if found {
+		return ns
+	} else {
+		return rolePath
+	}
+}
+
+func getOperatorNamespace() (string, error) {
```
nak3, line 59: Sorry for my ignorance. But this func is different from `k8sutil.GetOperatorNamespace()`?

zroubalik:
> Removed the owner ref and added a controller for managing the lifecycle of the dashboard.
> Still the dashboard needs to be cleaned up manually after serverless uninstall. I will add a command to remove it in a preStop hook. Btw I discussed the limitation in OLM with @jcrossley3, there might be an alternative there, @zroubalik might know more.

If we are talking about the finalization/cleanup part, unfortunately there isn't any better solution in OLM that I am aware of 😞

skonto:
@zroubalik is there a way to add a configmap as part of the olm managed operator lifecycle beyond crds and deployments?

zroubalik:
@skonto I haven't tried ConfigMaps (but I have tried other k8s resources), but you could try to put yaml files, that contains CMs definitions in the directory next to the CSV. For example for 1.9.0 in this directory: https://github.com/openshift-knative/serverless-operator/tree/master/olm-catalog/serverless-operator/1.9.0

OLM will create this resources during the installation of an operator.

skonto:
@zroubalik I tried that but the CSV spec has only a limited number of fields defined in the configmap which stores its data.

zroubalik:
> @zroubalik I tried that without success, The CSV spec has only a limited number of fields defined in the [configmap](https://github.com/openshift-knative/serverless-operator/blob/e5228d7eab2e6793e6047722bbe25983840867db/hack/catalog.sh#L43-L53) which stores its data. How will the yaml be picked up during installation? Should I hack it so I pass an arbitrary yaml as a crd?

@skonto pardon my ignorance. But I am not sure that I fully understand you problem/question. I might be wrong, but the ConfigMap linked here doesn't have anything to do with OLM itself.

If you question is, "Is it possible to deploy/create any ConfigMap via OLM during the installation of an operator?", then the answer is yes. You need to put yaml file with the ConfigMap definition next to the CSV file in the same directory and OLM will handle this, as I mentioned above.

skonto:
I had a discussion with @zroubalik and @mgencur this an be addressed with the new [bundle format](https://github.com/operator-framework/operator-registry#manifest-format) which supports more type of resources or at least try to.

> A bundle typically includes a ClusterServiceVersion and the CRDs that define the owned APIs of the CSV in its manifest directory, though additional objects may be included.

Meanwhile there is ongoing work to add support for the new format: https://github.com/openshift-knative/serverless-operator/pull/389 and create the UX for installing the operator. After this is ready we can try fixing it.

skonto:
@mgencur @nak3 still getting 

```{"component":"entrypoint","file":"prow/entrypoint/run.go:168","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Entrypoint received interrupt: terminated","severity":"error","time":"2020-08-03T01:12:39Z"}``` 
Any ideas? Is there a way to fix this?

**Comment by skonto** [comment]
@openshift/openshift-team-monitoring fyi.

**Comment by s-urbaniak** [comment]
@skonto can you give us a hint where the underlying metrics are actually defined? I suggest to open a PR a la https://github.com/openshift/cluster-monitoring-operator/pull/892 against the cluster-monitoring-operator, as in this PR I don't seem to recognize the concrete metrics you are trying to send via telemetry.

File: knative-operator/pkg/common/metrics.go [hunk 23]
```diff
@@ -0,0 +1,26 @@
...
+		prometheus.GaugeOpts{
+			Name: "knative_serving_ready",
+			Help: "Reports if Knative Serving is up",
+		},
+	)
+	KnativeEventingReadyG = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "knative_eventing_ready",
+			Help: "Reports if Knative Eventing is up",
+		},
```
skonto, line 19: @s-urbaniak these are the metrics.

s-urbaniak: nit: since the help text says "Reports if Knative Eventing is up", why not call it `knative_eventing_up` ?
Additionally you could merge it into one metric called `knative_up{type="eventing"}` and `knative_up{type="serving"}`.

After the naming changes this lgtm as the amount of series is 2. However pinging @lilic and/or @paulfantom for another best-practice pair of eyes.

And pinging @smarterclayton for approval as those seem to be the candidates to be sent via telemetry?

s-urbaniak: > Is it true that only the ones that are part of the cluster-monitoring-operator are pushing data to telemetry?
currently, yes

skonto: @s-urbaniak thanks for the suggestions I will update as needed. 
Currently there was a request in SRVKE-357 to show the relevant info for eventing to telemeter, so I guess these are going to be the appropriate metrics, still need to discuss internally but wanted to do a dry run as well.

lilic: Can you explain what are you trying to observe with this metric? Looks like a health metric to me?

Are you installed via OLM, if yes then integrating into cluster monitoring operator should be easy enough.

knative-operator/pkg/common/health_dashboard.go
bbrowning: We should not ever do any cross-namespace owner references. They only appear benign until they fail hard under load when garbage collection kicks in.

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
```diff
@@ -184,6 +183,12 @@ func (r *ReconcileKnativeServing) Reconcile(request reconcile.Request) (reconcil
 			return reconcile.Result{}, fmt.Errorf("failed to update status: %w", err)
 		}
 	}
+
+	if instance.Status.IsReady() {
+		common.KnativeServingReadyG.Set(1)
+	} else {
+		common.KnativeServingReadyG.Set(0)
+	}
```
bbrowning, line 191: This will break if a second `KnativeServing` instance gets installed in a cluster, right? Like if the user creates it in the wrong namespace as we've had users do in the past?

Also, does `reconcileErr` need to be taken into account in case we had a reconcilation error? Do we want to unconditionally set the metric value even if reconciling failed?

Tagging @jcrossley3 in because he may know about the multiple `KnativeServing` issue, at least.

knative-operator/pkg/common/monitoring.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	return deployment, nil
+}
+
+func createRoleAndRoleBinding(instance *appsv1.Deployment, namespace, path string, client client.Client) error {
+	manifest, err := mf.NewManifest(path, mf.UseClient(mfclient.NewClient(client)))
+	if err != nil {
+		return fmt.Errorf("unable to create role and roleBinding ServiceMonitor install manifest: %w", err)
+	}
+	instance.SetGroupVersionKind(schema.GroupVersionKind{
```
bbrowning, line 101: This feels odd - why do we need the Deployment here to create a Role and RoleBinding?

knative-operator/pkg/common/health_dashboard.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// manifest returns dashboard deploymnet resources manifest
+func manifest(instance *appsv1.Deployment, apiclient client.Client, namespace string) (mf.Manifest, error) {
+	manifest, err := mfc.NewManifest(manifestPath(), apiclient, mf.UseLogger(logh.WithName("mf")))
+	if err != nil {
+		return mf.Manifest{}, fmt.Errorf("failed to read dashboard manifest: %w", err)
+	}
+
+	instance.SetGroupVersionKind(schema.GroupVersionKind{
```
bbrowning, line 57: Same as other comment - why do we need the Deployment here?

knative-operator/pkg/common/metrics.go
bbrowning: Knowing if it's up or down is helpful, but what action will we take if we observe some are down? Should we be reporting something about why it's down instead of just whether it's up or down? If we ever want to report on specific reasons why it's down, then should we just do that from the outset instead of having an up/down metric that gets obsoleted by one with more information?

knative-operator/pkg/common/monitoring.go
```diff
@@ -0,0 +1,136 @@
...
+	rolePath        = "deploy/role_service_monitor.yaml"
+	testRolePath    = "TEST_ROLE_PATH"
+)
+
+func SetupMonitoringRequirements(api client.Client) error {
+	ns, err := getOperatorNamespace()
+	if err != nil {
+		return err
+	}
+	err = addMonitoringLabelToNamespace(ns, api)
```
bbrowning, line 34: The monitoring label can automatically get added to our namespace with recent OLM integration for monitoring with the console. The same goes for the appropriate roles. Are we sure we want to handle these separately in our operator instead of letting OLM do this on our behalf?

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
jcrossley3: Currently, we have a validating webhook that rejects any create/update request that's not in a specific namespace (configured via env var), and there can only be one in that namespace.

In the future, when knative compoonents will all be installed in a single `openshift-serverless` namespace, we need slightly different behavior: https://github.com/openshift-knative/serverless-operator/pull/386/commits/196765214b6646da3b76132a35c39679a335c3bb

But the bottom line is that we should always be using a validating webhook of some sort to ensure that only one instance is installed per cluster. HTH

knative-operator/pkg/common/monitoring.go
skonto: Please read my comment bellow.

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
skonto: My assumption was that metrics have namespace as an attribute (at least they show this way in prometheus when I checked them) and since we could have openshift-serverless-1, openshift-serverless-2 why this would break?

knative-operator/pkg/common/health_dashboard.go
skonto: Please read my comment bellow.

knative-operator/pkg/common/monitoring.go
skonto: Please read my comment bellow.

skonto:
@bbrowning 

> why do we need to reference our own deployment from within the operator code,
This is adapted from this previous [PR](https://github.com/openshift-knative/knative-serving-operator/pull/54). If OLM (with the new bundle format) can handle this then pls provide some pointers. The concpet here is that when the deployment is removed I would like the roles to disappear.

>  are we sure a simple boolean health metric is useful for us, 
This is discussed in the related internal story: "...knative serving is working (overall health) after being installed". This is like a smoke test that is my POV.

bbrowning:
@skonto Typically, when setting an Owner Reference from a controller you'd use something like https://github.com/openshift-knative/serverless-operator/blob/5110d1b02e8585b7b4bc240d170e1f7110b511b9/knative-operator/vendor/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil/controllerutil.go#L56 to do this instead of needing to lookup a Deployment. We'd set a controller reference on the roles and role bindings from this controller. The trick, as you've discovered, is that we don't have a custom resource that we're reconciling to drive all of this.

That makes me wonder if this is a "smell" that we should consider. Should we have a CRD that represents Serverless as a whole, instead of just the individual Serving and Eventing components? So things can be configured as a whole like this?

knative-operator/pkg/common/metrics.go
lilic: I would scrape the components not the operator, that way you get the up metric out of the box.

knative-operator/deploy/resources/dashboards/grafana-dash-knative-health.yaml
```diff
@@ -0,0 +1,276 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: grafana-dashboard-definition-knative-health
```
matzew, line 4: does this file match with an upstream version?

knative-operator/pkg/common/metrics.go
skonto: @lilic I am not scraping the operator, I am reporting the status of the components which is stated in the operator CR.

knative-operator/deploy/resources/dashboards/grafana-dash-knative-health.yaml
skonto: @matzew no this is specific to Openshift at the moment.

skonto:
Working with W. Oliveira on how this should look like. A current update not pushed yet is:
image

skonto:
> @skonto Typically, when setting an Owner Reference from a controller you'd use something like
> 
> https://github.com/openshift-knative/serverless-operator/blob/5110d1b02e8585b7b4bc240d170e1f7110b511b9/knative-operator/vendor/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil/controllerutil.go#L56
> 
> to do this instead of needing to lookup a Deployment. We'd set a controller reference on the roles and role bindings from this controller. The trick, as you've discovered, is that we don't have a custom resource that we're reconciling to drive all of this.
> That makes me wonder if this is a "smell" that we should consider. Should we have a CRD that represents Serverless as a whole, instead of just the individual Serving and Eventing components? So things can be configured as a whole like this?

@bbrowning you are right I wanted to use the provided API but had to do this hack for that reason. I think a global CRD would be useful to be able to detect what components are installed from that spec. 
Right now since components are installed independently you have to dig into the UI or search the corresponding namespace to find what is installed.

knative-operator/deploy/resources/dashboards/grafana-dash-knative-health.yaml
matzew: Asking, b/c there is something on upstream: https://github.com/knative/eventing/tree/master/config/monitoring/metrics

Also, @skonto I have this PR now installed, where do I see the eventing dashboard? 

I do see **Knative Serving - Revision CPU and Memory Usage** in the drop-down, but nothing for eventing :thinking:

skonto: @matzew this is coming is not part of this PR. Adding that dashboard is part of another story.

knative-operator/pkg/common/metrics.go
```diff
@@ -6,21 +6,23 @@ import (
...
+			ConstLabels: map[string]string{"type": "serving_status"},
 		},
 	)
-	KnativeEventingReadyG = prometheus.NewGauge(
+	KnativeEventingUpG = prometheus.NewGauge(
 		prometheus.GaugeOpts{
-			Name: "knative_eventing_ready",
-			Help: "Reports if Knative Eventing is up",
+			Name:        "knative_up",
+			Help:        "Reports if a Knative component is up",
```
skonto, line 19: Help msg needs to be the same otherwise things fail.

skonto:
New dashboard and metric names:
image

image
@markusthoemmes should we wait for the bundle format to fix the configmap lifecycle or merge it and update it later?

knative-operator/deploy/resources/dashboards/grafana-dash-knative-health.yaml
```diff
@@ -0,0 +1,436 @@
...
+             "full": false,
+             "lineColor": "rgb(31, 120, 193)",
+             "show": false
+           },
+           "targets": [
+           {
+             "expr": "sum(count(kube_pod_status_ready{namespace=\"knative-serving\", pod=~\".*\", condition=\"true\"} == 1))",
+             "intervalFactor": 2,
+             "legendFormat": "",
+             "metric": "etcd_server_has_leader",
```
skonto, line 91: will update this

skonto:
@s-urbaniak I am trying to use valueMaps to map values to Up/Down text any idea why they are not working?

skonto:
@paulfantom @anpicker I am trying to add a valueMaps field for a singlestat panel as in this example:
```
            "valueMaps": [
                 {
                   "op": "=",
                   "text": "N/A",
                   "value": "null"
                 },
                 {
                   "op": "=",
                   "text": "Healthy",
                   "value": "1"
                 },
                 {
                   "op": "=",
                   "text": "Unhealthy",
                   "value": "0"
                 }
            ],
```
Do we support this with the native dashboard spec? In Grafana this is supported.

skonto:
After discussing the design with W. Oliveira, we need a single stat to show status.

paulfantom:
As @kyoto already pointed out on slack, `valueMaps` are not supported.

FWIW having a status doesn't give the operator much meaningful data as it only shows the current state and not historical trend.

skonto:
I am wondering why ` TestRunLatestService` fails the first time.

markito:
Hi folks, seems like this a limitation then and we will have to settle for having the number of containers.  Is there a way to at least display something like {actual}/{required} ?   As in "8/9" would at least give the user a clue that not all containers are running and that the system may not be healthy while "9/9" would then indicate that everything is ok ?   If this is possible, we can keep the chart (trend) and the number of containers as you suggested Stravos. 

Displaying a histogram or trend of some kind is helpful, but "1" or "0" would not be informative enough for a user on a visual dashboard to indicate that health of the overall system, not to mention that it's not visually that interesting. I do like the idea of having the trend and specially if we could make it span over a month or more to indicate any blips on the health of the system, but I'd like to inform a customer/sys admin very clearly of what that blip mean as in "System was down for 10 mins on Dec 21". 

Thanks for your patience.

skonto:
@markito it seems that this is [NOT supported](https://github.com/grafana/grafana/issues/4199) by grafana or native dashboards. @paulfantom are you aware of this capability?
Regarding the trend I dont see many options here. If we want something more elegant we need something like a plugin check this [status plugin](https://medium.com/flant-com/statusmap-grafana-plugin-to-visualize-status-over-time-fe6ced391853) or somehow utilize a [heatmap](https://grafana.com/docs/grafana/latest/panels/visualizations/heatmap/). @paulfantom @spadgett what are the options here?

skonto:
@markito There is a PR for support valuemaps: https://github.com/openshift/console/pull/6328 so will update this when that is available.

knative-operator/pkg/common/health_dashboard.go
```diff
@@ -0,0 +1,92 @@
...
+	namespace, err := getOperatorNamespace()
+	if err != nil {
+		return err
+	}
+	err = api.Get(context.TODO(), client.ObjectKey{Name: ConfigManagedNamespace}, &corev1.Namespace{})
+	if apierrors.IsNotFound(err) {
+		logh.Info(fmt.Sprintf("namespace %q not found. Skipping to create dashboard.", ConfigManagedNamespace))
+		return nil
+	} else if err != nil {
+		return fmt.Errorf("failed to get namepsace %q: %w", ConfigManagedNamespace, err)
```
simonpasquier, line 32: s/namepsace/namespace/

knative-operator/pkg/common/metrics.go
simonpasquier: Rather than declaring 2 gauge metrics, you can use a `prometheus.GaugeVec`.

```
knativeUp = prometheus.NewGaugeVec(prometheus.GaugeOpts{...}, []string{"type"})

KnativeEventingUpG = knativeUp.WithLabelValues("eventing_status")
KnativeServingUpG = knativeUp.WithLabelValues("serving_status")
```

skonto:
This requires a 4.6 crcs env that comes with the bug fix, this is not available yet.

skonto:
@markito @markusthoemmes I updated the console, tested the fix in the console above and now the health dashboard looks like this:

image
image
image
image

@markito let me know if this is now ok to be merged from a ux perspective.

skonto:
CI fails on unrelated stuff:

```
Link to job on registry info site: https://steps.ci.openshift.org/job?org=openshift-knative&repo=serverless-operator&branch=master&test=upstream-e2e-aws-ocp-45&variant=4.5, failed to create or restart "upstream-e2e-aws-ocp-45-gather-must-gather" pod: unable to create pod: timed out waiting for the condition, failed to create or restart "upstream-e2e-aws-ocp-45-gather-extra" pod: unable to create pod: timed out waiting for the condition, failed to create or restart "upstream-e2e-aws-ocp-45-gather-audit-logs" pod: unable to create pod: timed out waiting for the condition, failed to create or restart "upstream-e2e-aws-ocp-45-ipi-deprovision-deprovision" pod: unable to create pod: timed out waiting for the condition]]```

**Comment by skonto** [comment]
@markito @markusthoemmes gentle ping. I would like to move forward as other stuff are coming based on this PR.
Btw not sure what is the problem with CI:
```
Container test is not ready with reason ImagePullBackOff and message Back-off pulling image "docker-registry.default.svc:5000/ci-op-xfwdjkcb/pipeline@sha256:fc4a234b91cc4b542bac8a6ad23b2ddcee60ae68fc4dbd4a52efb5f1b0baad71"
Link to step on registry info site: https://steps.ci.openshift.org/reference/ipi-conf
```

File: knative-operator/pkg/common/service_monitor.go [hunk 32]
```diff
@@ -0,0 +1,77 @@
+package common
+
+import (
+	"golang.org/x/net/context"
```
markusthoemmes, line 4: I don't think this is the correct import. It should be just `"context"`

knative-operator/pkg/controller/healthdashboard/healthdashboard_controller.go
```diff
@@ -0,0 +1,89 @@
...
+	// in any case restore the current health dashboard, since the configmap shouldnt
+	// be modified, if the configmap has not changed this will not trigger a real update
+	err := common.InstallHealthDashboard(r.client)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	return reconcile.Result{}, nil
+}
+
+type skipCreatePredicats struct {
```
markusthoemmes, line 82:
```suggestion
type skipCreatePredicate struct {
```

knative-operator/pkg/common/constants.go
```diff
@@ -2,6 +2,8 @@ package common
 
 // Annotation keys being used to tag the owned resources by instance
 const (
-	ServingOwnerName      = "serving.knative.openshift.io/ownerName"
-	ServingOwnerNamespace = "serving.knative.openshift.io/ownerNamespace"
+	ServingOwnerName            = "serving.knative.openshift.io/ownerName"
+	ServingOwnerNamespace       = "serving.knative.openshift.io/ownerNamespace"
+	ServerlessOperatorName      = "operator.knative.openshift.io/ownerName"
```
markusthoemmes, line 7:
```suggestion
	ServerlessOperatorOwnerName      = "operator.knative.openshift.io/ownerName"
```

knative-operator/pkg/common/constants.go
```diff
@@ -2,6 +2,8 @@ package common
...
+	ServerlessOperatorNamespace = "operator.knative.openshift.io/ownerNamespace"
```
markusthoemmes, line 8:
```suggestion
	ServerlessOperatorOwnerNamespace = "operator.knative.openshift.io/ownerNamespace"
```

knative-operator/pkg/common/health_dashboard.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		return mf.Manifest{}, fmt.Errorf("failed to read dashboard manifest: %w", err)
+	}
+
+	instance.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "apps",
+		Version: "v1",
+		Kind:    "Deployment",
+	})
+	instance.SetNamespace(namespace)
```
markusthoemmes, line 62: Isn't this overkill? Don't we know the namespace and name of the deployment from environment anyway? Seems like it's only needed to set the annotations so I'd guess we don't even need to query the deployment here?

markusthoemmes (commented):
Just a few nits, nothing major. Looks okay overall, though it's fairly big so I can't load it all into L1 cache :joy:. Fine for me to get in and get baking though.

knative-operator/pkg/common/health_dashboard.go
skonto: This is how namespace is fetched, from the container's env:
```
func getOperatorNamespace() (string, error) {
	ns := os.Getenv(installedNamespaceEnvKey)
	if ns == "" {
		return "", fmt.Errorf("the environment variable %q must be set", installedNamespaceEnvKey)
	}
	return ns, nil
}
```

markusthoemmes: You are I believe. `instance` is a deployment and that's not really needed in this case I believe.

knative-operator/pkg/common/service_monitor.go
skonto: ok my IDE played tricks on my will update.

markusthoemmes (approved):
/lgtm
/approve

Let's get this baking.

PR #389, opened 2020-07-28
Remove appregistry files and replace with operator-registry bundle files

warrenvw:
With the operator-registry bundle format, we no longer need to keep around the old app registry metadata. So remove these old files and replace with only the metadata required for the current release.


mgencur:
Just a heads up, this will make the upgrade tests fail because we're using one of the old CSV to install the previous version of Serverless. The tests then upgrade it to the latest. We'll probably need to install the old version from live operatorhub.

warrenvw:
I should also call out that I've added `olm.skipRange` in this PR. Please pay attention to the versions in the skipRange.

markusthoemmes:
Let's please not rely on the version in Operatorhub to be "the last". It'll prevent us from moving forward more quickly than we release.

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
```diff
@@ -32,7 +32,8 @@ metadata:
       of serverless applications and functions.
     repository: https://github.com/openshift-knative/serverless-operator
     support: Red Hat, Inc.
-  name: serverless-operator.v1.8.0-rc1
+    olm.skipRange: '>1.7.2 <1.8.0'
```
maschmid, line 35: I we are to support upgrade from 1.7.2, it should be >=1.7.2 perhaps?

mgencur:
At this point the E2E fail to install the catalogsource: 
```
Installing CatalogSource
cat: /go/src/github.com/openshift-knative/serverless-operator/olm-catalog/serverless-operator/*package.yaml: No such file or directory
```

olm-catalog/serverless-operator/metadata/annotations.yaml
```diff
@@ -0,0 +1,7 @@
+annotations:
+    operators.operatorframework.io.bundle.channel.default.v1: "4.3"
```
maschmid, line 2: I am not sure how this works, but isn't it weird that the default channel here is not one of the "channels"? ("operators.operatorframework.io.bundle.channels.v1" only have "4.5" ?

mgencur:
I've tried to build the bundle image from the above Dockerfile, build the index image (catalog), create the catalog source from it and install Serverless. Serverless 1.8.0 is getting installed but then I see errors pulling some images:
```
knative-openshift: Failed to apply default image tag "$IMAGE_KNATIVE_OPERATOR": couldn't parse image reference "$IMAGE_KNATIVE_OPERATOR": invalid reference format: repository name must be lowercase
knative-openshift-ingress: Failed to apply default image tag "$IMAGE_KNATIVE_OPENSHIFT_INGRESS": couldn't parse image reference "$IMAGE_KNATIVE_OPENSHIFT_INGRESS": invalid reference format: repository name must be lowercase
```

warrenvw:
Those images are not yet productized. This is the upstream version which has yet to be run through the productization process. During productization, images are built and the image refs are updated in the CSV.

warrenvw:
Replaced by #409.

PR #399, opened 2020-08-03
[WIP] Test new operator bundle format in CI

bbrowning:
This time with the correct base branch.


mgencur:
@bbrowning If you want to make upgrade tests work they you'll probably need what @jcrossley3 did on master branch on this line: https://github.com/openshift-knative/serverless-operator/blob/master/test/serving.bash#L122 . This can save you a bit of time investigating.

mgencur:
@bbrowning and we might need to hardcode current_csv and previous_csv here: https://github.com/openshift-knative/serverless-operator/blob/release-1.8/hack/lib/serverless.bash#L24 (or use a different approach to get it)

bbrowning:
Continued in #402

PR #400, opened 2020-08-04
Bump to new version 1.7.3

matzew:
naive try to simply bump the CSV version

/assign @warrenvw

### DIFF
--- a/olm-catalog/serverless-operator/serverless-operator.package.yaml
+++ b/olm-catalog/serverless-operator/serverless-operator.package.yaml
@@ -2,8 +2,8 @@ packageName: serverless-operator
 defaultChannel: "4.3"
 channels:
 - name: "4.3"
-  currentCSV: serverless-operator.v1.7.2
+  currentCSV: serverless-operator.v1.7.3
 - name: "4.4"
-  currentCSV: serverless-operator.v1.7.2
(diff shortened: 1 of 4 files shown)


olm-catalog/serverless-operator/serverless-operator.package.yaml
```diff
@@ -2,8 +2,8 @@ packageName: serverless-operator
...
 channels:
 - name: "4.3"
-  currentCSV: serverless-operator.v1.7.2
+  currentCSV: serverless-operator.v1.7.3
 - name: "4.4"
-  currentCSV: serverless-operator.v1.7.2
+  currentCSV: serverless-operator.v1.7.3
 - name: "4.5"
-  currentCSV: serverless-operator.v1.7.2
+  currentCSV: serverless-operator.v1.7.3
```
mgencur, line 9: Looks like if we publish Serverless 1.8 first in the 4.5 channel, we won't be able to release 1.7.3 in the same channel later and have 1.7.3 replace 1.7.2. Maybe it would make sense to publish 1.7.3 only to 4.3 and 4.4 channels. But that needs a discussion.

warrenvw: We'll only release 1.7.3 to 4.3 and 4.4 channels.

PR #402, opened 2020-08-04
[WIP] The new bundle format.

markusthoemmes:
as per title.


olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
```diff
@@ -553,4 +553,5 @@ spec:
       image: registry.svc.ci.openshift.org/openshift/knative-v0.14.2:knative-eventing-channel-dispatcher
     - name: IMAGE_KN_CLI_ARTIFACTS
       image: registry.svc.ci.openshift.org/openshift/knative-v0.14.0:kn-cli-artifacts
+  replaces: serverless-operator.v1.7.2
```
mgencur, line 556: Why this? I thought we were using the skipRange instead of this.

markusthoemmes: Eh, precautionary. I can try if it works without it. BRB

olm-catalog/serverless-operator/Dockerfile
```diff
@@ -1,12 +1,12 @@
-FROM alpine
+FROM centos
```
bbrowning, line 1: If we can get this thing green, before merging, ideally we'd move this to ubi8 or whatever base image we're using for all our other operator images. For the purpose of hacking away, we can ignore it for now.

markusthoemmes: I believe this is replaced in CI anyway by something else. I chose something that comes close to what CI will replace it with but the actual CI replacement has some RHEL dependencies which prevented `yum` to work locally.

hack/lib/vars.bash
```diff
@@ -48,7 +48,10 @@ readonly UPGRADE_CLUSTER="${UPGRADE_CLUSTER:-"false"}"
...
 
 readonly INSTALL_PREVIOUS_VERSION="${INSTALL_PREVIOUS_VERSION:-"false"}"
-export OLM_CHANNEL="${OLM_CHANNEL:-"candidate"}"
+export OLM_CHANNEL="${OLM_CHANNEL:-"4.5"}"
 # Change this when upgrades need switching to a different channel
 export OLM_UPGRADE_CHANNEL="${OLM_UPGRADE_CHANNEL:-"$OLM_CHANNEL"}"
 export OLM_SOURCE="${OLM_SOURCE:-"$OPERATOR"}"
+
+export CURRENT_CSV="serverless-operator.v1.8.0"
+export PREVIOUS_CSV="serverless-operator.v1.7.2"
```
jcrossley3, line 57: It'd be nice if you could derive the previous value from the CSV, using either `replaces` or `skipRange`. And if you can do that, grab the current one from there, too.

hack/lib/serverless.bash
```diff
@@ -108,8 +103,7 @@ function approve_csv {
 function find_install_plan {
   local csv=$1
   for plan in `oc get installplan -n ${OPERATORS_NAMESPACE} --no-headers -o name`; do 
-    [[ $(oc get $plan -n ${OPERATORS_NAMESPACE} -o=jsonpath='{.spec.clusterServiceVersionNames}' | grep -c $csv) -eq 1 && \
-       $(oc get $plan -n ${OPERATORS_NAMESPACE} -o=jsonpath="{.status.catalogSources}" | grep -c $OPERATOR) -eq 1 ]] && echo $plan && return 0
```
mgencur, line 112: I'm not sure why we had the second condition here. I think it was there because the CSV was sometimes present in multiple installplans so we were making it clear which one we want. If we hit that again we might need to change this to: 
```for plan in `oc get installplan -n ${OPERATORS_NAMESPACE} --no-headers -o name`; do 
    [[ $(oc get $plan -n ${OPERATORS_NAMESPACE} -o=jsonpath='{.spec.clusterServiceVersionNames}' | grep -c $csv) -eq 1 && \
       $(oc get $plan -n ${OPERATORS_NAMESPACE} -o=jsonpath="{.status.bundleLookups[0].identifier}") == $csv ]] && echo $plan && return 0
  done
```

PR #404, opened 2020-08-04
Restructure tests to avoid installing knative more than once

jcrossley3:
It's silly to re-install both serving *and* eventing only to run a few more serving tests (but no eventing ones!!!). We should run all the e2e tests in a single installation of each component.

TODO: Refactor the subtests to allow running single tests by adding a "-run" arg to TEST_ARGS. You can do this, currently:
```
  $ TEST_ARGS="-run Serving" SCALE_UP=0 make test-e2e
```
But we'd like to be able to do this, too:
```
  $ TEST_ARGS="-run Serving/downstream/https" make test-e2e
```
That requires a guard to ensure the CR is installed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ test-unit:
 
 # Run only SERVING/EVENTING E2E tests from the current repo.
 test-e2e:
-	./test/e2e-tests.sh
+	./test/e2e-tests.sh $(ARGS)
 
 # Run E2E tests from the current repo for serving+eventing+knativeKafka
 test-e2e-with-kafka:
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -13,6 +13,7 @@ debugging.setup
 
 scale_up_workers || exit $?
 create_namespaces || exit $?
+add_systemnamespace_label || exit $?
 create_htpasswd_users && add_roles || exit $?
 
 failed=0
--- a/test/e2e/knative_eventing_test.go
+++ b/test/e2e/knative_eventing_test.go
@@ -25,8 +25,8 @@ var knativeControlPlaneDeploymentNames = []string{
 
 func TestKnativeEventing(t *testing.T) {
 	caCtx := test.SetupClusterAdmin(t)
-
 	test.CleanupOnInterrupt(t, func() { test.CleanupAll(t, caCtx) })
+	defer test.CleanupAll(t, caCtx)
 
 	t.Run("create subscription and wait for CSV to succeed", func(t *testing.T) {
 		if _, err := test.WithOperatorReady(caCtx, "serverless-operator-subscription"); err != nil {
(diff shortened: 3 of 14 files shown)


markusthoemmes:
I like it but we do lose the ability to run those tests that require an installed system one-by-one (i.e. debugging one single test without running the entire suite).

Can we get that back somehow? Can we actually just leave all the `Test` prefixes but never call `go test` on the package to achieve the same effect and to do your change at the same time?

jcrossley3:
> Can we get that back somehow? Can we actually just leave all the `Test` prefixes but never call `go test` on the package to achieve the same effect and to do your change at the same time?

It's not just the prefix. We would have to handle the cleanup of the cluster admin context somehow. The subtests would have to know whether they should cleanup or whether to leave it up to their caller.

How do you typically do it? Something like this?
```
  go test ./test/servinge2e -run KnativeServiceHTTPS
```
My plan was to achieve the same result like this:
```
  TEST_ARGS="-run Serving/downstream/https" make test-e2e
```
I don't feel strongly either way. I'm happy to go with whatever you normally do.

jcrossley3:
Well, I did some more experiments, and I don't think it's possible to preserve the `Test*` prefix *and* invoke the "downstream" tests from within the primary test, mostly because go requires that `Test*` functions reside in a `*_test.go` file, which I can't import.

We can still accomplish what you want, though. It's just that instead of this:
```
    go test ./test/servinge2e -run KnativeServiceHTTPS
```
You'll have to do this:
```
    go test ./test/e2e -run Serving/downstream/https
```
To me, trading that for essentially cutting e2e test time in half is worth it.

markusthoemmes:
Fell between the cracks. Your last suggestion SGTM @jcrossley3 .

markusthoemmes:
Can we get this rebased?

mgencur:
I'm curious if it would make sense to move the "create subscription" out from Golang to Bash. Then we could run this only once at the beginning and not over and over again. We create and delete the subscription in 4 places: Golang test for Serving, Golang test for Eventing, Golang test for Kafka, and then in Bash in `ensure_serverless_installed` . We could extract that and call it only once at the beginning. This is anyway a functionality of OLM (as Jim menioned before?) so we don't need to test it multiple times.

jcrossley3:
> I'm curious if it would make sense to move the "create subscription" out from Golang to Bash. Then we could run this only once at the beginning and not over and over again. We create and delete the subscription in 4 places: Golang test for Serving, Golang test for Eventing, Golang test for Kafka, and then in Bash in `ensure_serverless_installed` .

Right now, this PR only addresses `KnativeServing` but it should be straightforward to add similar `EnsureKnativeEventing` and `EnsureKafka` functions to prevent the subscription churn. That would handle 3/4 of your cases and obviate the need for a developer to invoke any shell script prior to running his test. The goal is this:
```
  $ go test ./test/e2e -run Serving/downstream/https
```

jcrossley3:
Rebased, but untested. I left `WaitForOperatorDepsDeleted` in there because the kafka test is calling it, but I have a vague memory of deleting it on my first pass at this.

matzew:
Do we want to keep this? or are we fine in closing this PR ?

PR #409, opened 2020-08-05
[RELEASE-1.8] Move to the new bundle format.

markusthoemmes:
This replaces the "old" way of doing OLM manifests with the new bundle format.

## Approach

The way our CatalogSource is built has been vastly revamped. Instead of relying on a ConfigMap that hosts a list of CSVs, we now rely on a gRPC based source. During the run, a new bundle image (containing the CSV and metadata for the **current** version) will be built in the cluster-internal registry of the cluster under test. We'll then launch a `serverless-index` deployment, which contains the gRPC registry, which will pull the respective bundle image from the internal registry. In addition to that, it's also fed a bundle image for the version we want to upgrade from (1.7.2 in this case). This has been built [manually for this version](https://github.com/openshift-knative/serverless-operator/pull/403), seeing as 1.7.2 is still on the old format. For 1.9.0 onwards, we'll be able to use the promoted images by CI as 1.8 will be building a serverless-bundle as part of its images.

## Notable changes

1. We now have to define which is the current and the next CSV. Especially the former can likely be inferred from the files themselves. I'd like to leave that to a followup PR.
2. The in-repository CSV is now a "working" CSV by default. It contains the to-be-promoted images so that the image built by CI (which won't replace any image coordinates) has the right image coordinates built into it.


olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
```diff
@@ -552,5 +553,5 @@ spec:
       image: registry.svc.ci.openshift.org/openshift/knative-v0.14.2:knative-eventing-channel-dispatcher
     - name: IMAGE_KN_CLI_ARTIFACTS
       image: registry.svc.ci.openshift.org/openshift/knative-v0.14.0:kn-cli-artifacts
-  replaces: serverless-operator.v1.7.1
-  version: "1.8.0-rc1"
+  replaces: serverless-operator.v1.7.2
```
markusthoemmes, line 556: This is needed, otherwise the catalog is broken.

mgencur: Oh, so when we have skipRange we also need `replaces:` ? Then I'm curious what will happen when we have 1.7.3 which also `replaces` 1.7.2. I think that's not allowed by OLM.

warrenvw: I've seen some CSV's with `skipRange` _and_ `replaces`, and some with just `skipRange`, and obviously, ones with just `replaces`. I don't know which one is right. I'll dig in.

warrenvw: According to Shawn, replaces is not required if skipRange is specified.

hack/lib/catalogsource.bash
```diff
@@ -1,22 +1,101 @@
...
+
+  # Build the bundle image in the cluster-internal registry.
+  oc -n "$OLM_NAMESPACE" new-build --binary --strategy=docker --name serverless-bundle
+  oc -n "$OLM_NAMESPACE" start-build serverless-bundle --from-dir olm-catalog/serverless-operator -F
+
+  # Undo potential changes to the CSV to not pollute the repository.
+  mv "${rootdir}/bkp.yaml" "$csv"
+
+  # HACK: Allow to run the index pod as root so it can create the necessary directories.
+  oc -n "$OLM_NAMESPACE" adm policy add-scc-to-user anyuid -z default
```
cardil, line 35: Why we need this? Couldn't it operate on `/home` or `/tmp` subdirs?

hack/lib/catalogsource.bash
```diff
@@ -1,22 +1,101 @@
...
-  if oc get catalogsource "$OPERATOR" -n "$OLM_NAMESPACE" >/dev/null 2>&1; then
-    logger.success 'CatalogSource is already installed.'
-    return 0
-  fi
-  install_catalogsource
-}
-
 function install_catalogsource {
   logger.info "Installing CatalogSource"
+  local rootdir="$(dirname "$(dirname "$(dirname "$(realpath "${BASH_SOURCE[0]}")")")")"
```
cardil, line 5: Better to define local variable on separate line to avoid [SC2155](https://github.com/koalaman/shellcheck/wiki/SC2155)

olm-catalog/serverless-operator/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
+LABEL operators.operatorframework.io.bundle.package.v1=serverless-operator
+LABEL operators.operatorframework.io.bundle.channel.default.v1="4.5"
+LABEL operators.operatorframework.io.bundle.channels.v1="4.5"
+
+LABEL \
+      com.redhat.component="openshift-serverless-1-serverless-rhel8-operator-container" \
+      name="openshift-serverless-1/serverless-rhel8-operator" \
+      version="1.8.0" \
+      summary="Red Hat OpenShift Serverless 1 Serverless Operator" \
```
mgencur, line 17: This summary could probably be more user friendly. I don't know if it makes sense to mantion "Serverless Operator" twice and the `1`. Similarly for the `description` and `display-name`. That's probably a question for @warrenvw why they have this format.

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
```diff
@@ -32,7 +32,8 @@ metadata:
       of serverless applications and functions.
     repository: https://github.com/openshift-knative/serverless-operator
     support: Red Hat, Inc.
-  name: serverless-operator.v1.8.0-rc1
+    olm.skipRange: '>=1.7.2 <1.8.0'
```
mgencur, line 35: This `>=` probably means that even users being on 1.7.1 might be upgraded to 1.8.0? Maybe only for "manual" upgrades. I'm not sure about that.

hack/lib/catalogsource.bash
markusthoemmes: To be frank, I didn't have a ton of time to dig into why this image is behaving like this specifically. I"d personally go with this an fix in a followup to `master`. Not the end of the world for CI anyway imo.

hack/lib/catalogsource.bash
```diff
@@ -1,22 +1,101 @@
 #!/usr/bin/env bash
 
-function ensure_catalogsource_installed {
```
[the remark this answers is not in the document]
markusthoemmes: Well first of all, it's not really idempotent. It only checks the catalogsource being present and doesn't reconcile all the resources that are needed to make the catalogsource actually work properly.

It's also only called once per test path anyway and has been an issue when developing locally, if I needed to change stuff.

All in all: I'd say the check is overkill, stuff runs quick enough imo and I'd value reactivity to changes over idempotency here.

hack/lib/catalogsource.bash
markusthoemmes: Copy & pasted from what was already there. Let's fix this afterwards if we feel it's important.

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
markusthoemmes: how so? I'd read this as "any user between (and including) 1.7.2 and 1.8.0" wheres before it would've been "between 1.7.3 and 1.8.0" I believe, which is not what we want? :thinking:

olm-catalog/serverless-operator/Dockerfile
markusthoemmes: Yeah, deferring this to @warrenvw. I dunno where this is even used at all.

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
mgencur: Oh. I think you're right. I agree

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
bbrowning: Just too loop back on this, based on Slack threads, it seems `replaces` IS required.

bbrowning:
Outside of the TODOs already in the code (that I assume we'll address as follow-ups with less immediate urgency) and the questions already raised, this looks good to me.

Because of the amount of changes, we should keep a special eye out for any new CI flakes and track that down. And, we need to also get this ported forwards on the master branch.

markusthoemmes:
@bbrowning yes, my thoughts about fixing the TODOs are mostly towards the master port seeing as a lot of the test scripts has changed over there already and imo it doesn't make a ton of sense to invest as much time for polishing on the 1.8 branch as we'll prolly only do security respins for it anyway.

hack/lib/catalogsource.bash
cardil: In this exact case, not important. It's just good to know, and make a good habit, as it may save you a huge pain someday.

markusthoemmes:
/hold

Once LGTM, I want to merge manually to squash meaningfully.

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
warrenvw: That thread was updated to say just having skipRange is "10000%" a valid thing. `replaces` is not required.

markusthoemmes: @warrenvw that is not true. That is only true (as said in the thread) if the respective CSV is the **only** one in a channel!

olm-catalog/serverless-operator/Dockerfile
warrenvw: This was setup in part before I joined the team. I kept the same format for all the repos I requested. In the interest of merging this PR, let's leave as is. If necessary, we can fix in a future PR.

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
warrenvw: 👍

PR #413, opened 2020-08-11
CI was here...:fire:

matzew:


matzew:
/close

was a "test" to see if the image promotion of the SOURCE image did actually work, after it was - by accident - disabled

PR #414, opened 2020-08-13
[RELEASE-1.7] Move to the new bundle format

warrenvw:
This replaces the "old" way of doing OLM manifests with the new bundle format. (#409)

Approach

The way our CatalogSource is built has been vastly revamped. Instead of relying on a ConfigMap that hosts a list of CSVs, we now rely on a gRPC based source. During the run, a new bundle image (containing the CSV and metadata for the **current** version) will be built in the cluster-internal registry of the cluster under test. We'll then launch a `serverless-index` deployment, which contains the gRPC registry, which will pull the respective bundle image from the internal registry. In addition to that, it's also fed a bundle image for the version we want to upgrade from (1.7.2 in this case). This has been built [manually for this version](https://github.com/openshift-knative/serverless-operator/pull/403), seeing as 1.7.2 is still on the old format. For 1.9.0 onwards, we'll be able to use the promoted images by CI as 1.8 will be building a serverless-bundle as part of its images.

Notable changes

1. We now have to define which is the current and the next CSV. Especially the former can likely be inferred from the files themselves. I'd like to leave that to a followup PR.
2. The in-repository CSV is now a "working" CSV by default. It contains the to-be-promoted images so that the image built by CI (which won't replace any image coordinates) has the right image coordinates built into it.

Co-authored-by: Markus Thömmes <markusthoemmes@me.com>
Co-authored-by: Ben Browning <bbrownin@redhat.com>


hack/lib/vars.bash
```diff
@@ -11,25 +11,29 @@ fi
...
+# Adjust these when upgrading the knative versions.
 readonly KNATIVE_SERVING_VERSION="${KNATIVE_SERVING_VERSION:-v0.13.3}"
-readonly KNATIVE_SERVING_OPERATOR_VERSION="${KNATIVE_SERVING_OPERATOR_VERSION:-v0.13.3}"
 readonly KNATIVE_EVENTING_VERSION="${KNATIVE_EVENTING_VERSION:-v0.13.0}"
-readonly KNATIVE_EVENTING_OPERATOR_VERSION="${KNATIVE_EVENTING_OPERATOR_VERSION:-v0.13.3}"
+
+# Adjust these when cutting a new CSV.
+# TODO: Read these from metadata.
+export CURRENT_CSV="serverless-operator.v1.7.2"
+export PREVIOUS_CSV="serverless-operator.v1.7.3"
```
navidshaikh, line 21: should the values be swapped here ?

warrenvw: Yes, they most definitely should be! Thanks!

nak3:
It seems that the following issue is a crttical error on upgrade test.

in build.log
```
/go/src/github.com/openshift-knative/serverless-operator/test/serving.bash: line 82: cd: /go/src/knative.dev/serving: No such file or directory
```

`checkout_knative_serving` func was removed in this commit as:

https://github.com/openshift-knative/serverless-operator/pull/414/files#diff-be1a2000487b2e3da5d0a88d6b55af92L13
https://github.com/openshift-knative/serverless-operator/pull/414/files#diff-be1a2000487b2e3da5d0a88d6b55af92L47

But I remember that it was removed by https://github.com/openshift-knative/serverless-operator/commit/e4b6518229985ba9527faff2a6f5522635c120a5 so it is not available on 1.7 branch?

cardil:
Just to note, that change will break CVP for 1.7 line as well. See: https://issues.redhat.com/browse/SRVCOM-885 as related.

hack/lib/vars.bash
mgencur: Please see how we infer the values on master branch now. This commit brought the changes: https://github.com/openshift-knative/serverless-operator/commit/2a9e1adc339c5af323b914533358f26bcfafb001

olm-catalog/serverless-operator/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+COPY manifests /manifests
+COPY metadata/annotations.yaml /metadata/annotations.yaml
+
+LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
+LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
+LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
+LABEL operators.operatorframework.io.bundle.package.v1=serverless-operator
+LABEL operators.operatorframework.io.bundle.channel.default.v1="4.3"
+LABEL operators.operatorframework.io.bundle.channels.v1="4.3,4.4,4.5"
```
mgencur, line 11: Oh. Are we releasing into the 4.5 channel? Technically, I don't think it's possible because we already have a newer version there (1.8.0, 1.9.0) and 1.7.3 does not "replace" 1.9.0 in the CSV.

olm-catalog/serverless-operator/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+LABEL \
+      com.redhat.component="openshift-serverless-1-serverless-rhel8-operator-container" \
+      name="openshift-serverless-1/serverless-rhel8-operator" \
+      version="1.7.3" \
+      summary="Red Hat OpenShift Serverless 1 Serverless Operator" \
+      maintainer="serverless-support@redhat.com" \
+      description="Red Hat OpenShift Serverless 1 Serverless Operator" \
+      io.k8s.display-name="Red Hat OpenShift Serverless 1 Serverless Operator" \
+      com.redhat.openshift.versions="v4.5" \
```
mgencur, line 21: This should probably not be 4.5 (see my previous comment on versions)

olm-catalog/serverless-operator/Dockerfile
mgencur: I guess we will see if the skipRange alows for this setup. There's already 1.8.0 which replaces 1.7.2. It may work. Not sure

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
```diff
@@ -393,7 +394,8 @@ spec:
               serviceAccountName: knative-openshift-ingress
               containers:
                 - name: knative-openshift-ingress
-                  image: $IMAGE_KNATIVE_OPENSHIFT_INGRESS
+                  # This reference will be replaced in local builds and CI via hack/lib/catalogsource.bash
+                  image: registry.svc.ci.openshift.org/openshift/openshift-serverless-nightly:knative-openshift-ingress
```
markusthoemmes, line 398:
```suggestion
                  image: registry.svc.ci.openshift.org/openshift/openshift-serverless-v1.7.3:knative-openshift-ingress
```

(will need a change to openshift/release first I believe)

Same applies to other image refs.

olm-catalog/serverless-operator/Dockerfile
warrenvw: As I understand it, this refers to the OCP version.

markusthoemmes:
@warrenvw are we still working on this? Need any help?

hack/lib/catalogsource.bash
```diff
@@ -11,12 +11,91 @@ function ensure_catalogsource_installed {
...
+
+  # Build the bundle image in the cluster-internal registry.
+  oc -n "$OLM_NAMESPACE" new-build --binary --strategy=docker --name serverless-bundle
+  oc -n "$OLM_NAMESPACE" start-build serverless-bundle --from-dir olm-catalog/serverless-operator -F
+
+  # Undo potential changes to the CSV to not pollute the repository.
+  mv "${rootdir}/bkp.yaml" "$csv"
+
+  # HACK: Allow to run the index pod as root so it has necessary access.
+  oc -n "$OLM_NAMESPACE" adm policy add-scc-to-user anyuid -z default
```
markusthoemmes, line 44: Doesn't this have to be `add-scc-to-user privileged`?

olm-catalog/serverless-operator/Dockerfile
mgencur: I'm still wondering why this line has "v4.5". IMO, this line should list the same versions as `operators.operatorframework.io.bundle.channels.v1` above. For example, Serverless 1.10 had this: https://github.com/openshift-knative/serverless-operator/blob/release-1.10/olm-catalog/serverless-operator/Dockerfile
According to https://docs.google.com/spreadsheets/d/1HTxR37_MM03_JQImW-KmAFk1oEQ83fXzfGxdfWQSX6o/edit#gid=0 , Serverless 1.7.3 will be supported on 4.3, 4.4 ( but it must also work on 4.5 because that's the upgrade path to 1.8).

mgencur:
If you're having problems with resources in the cluster try increasing the SCALE_UP to e.g. 8 https://github.com/openshift-knative/serverless-operator/blob/release-1.7/hack/lib/vars.bash#L34 
It seems increasing the number of workers helped @dsimansk

mgencur:
Increasing helped a little but it's still not ideal.

olm-catalog/serverless-operator/Dockerfile
mgencur: Alright, https://access.redhat.com/product-life-cycles?product=Red%20Hat%20OpenShift%20Serverless mentions only 4.4 as the supported version. So I think this line should have "v4.4" and the channels above should list only "4.4"

gallettilance: @warrenvw @mgencur 

This is the OCP version. Here is the list of supported versions: https://docs.engineering.redhat.com/display/CFC/Delivery

"v4.5" includes <= 4.4 because everything pushed to the 4.5 index gets backported to app-registry and served to all clusters <= 4.5. In a couple weeks you may be able to use the new version scheme in the labels and have the label be set to "=v4.5" which will prevent these versions from being pulled forward to the 4.6 and 4.7 index images. Today, this new versioning scheme is not available and you can instead mark operator versions as deprecated using the flow described in the above doc.

olm-catalog/serverless-operator/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM scratch
+
+COPY manifests /manifests
+COPY metadata/annotations.yaml /metadata/annotations.yaml
+
+LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
+LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
+LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
+LABEL operators.operatorframework.io.bundle.package.v1=serverless-operator
+LABEL operators.operatorframework.io.bundle.channel.default.v1="4.3"
```
gallettilance, line 10: Given that this bundle will be published to all OCP clusters, if this bundle gets added last it will set the default channel to be 4.3. Is this desired? Is there another channel that should be default in 4.5 or 4.6? Just double checking

warrenvw: Shawn had always mentioned the 4.3 channel should be set as the default as long as we support 4.3 in any of our active releases.

test/serving.bash
```diff
@@ -128,14 +125,10 @@ function run_knative_serving_rolling_upgrade_tests {
   PROBER_PID=$!
 
   if [[ $UPGRADE_SERVERLESS == true ]]; then
-    prev_serving_version=$(oc get knativeserving.operator.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.version}")
-    # This is ugly hack. Use KNATIVE_SERVING_VERSION if issues/361 was solved.
-    latest_serving_version=$(sed -n 's/^.*Version.*"\(.*\)".*$/\1/p' ${rootdir}/knative-operator/vendor/knative.dev/serving-operator/version/version.go)
```
mgencur, line 133: latest_serving_version needs to be defined in the following way here:
```
latest_serving_version="${KNATIVE_SERVING_VERSION/v/}"
```
The upgrade tests currently fail because we use it a few lines below and it's undefined.

markusthoemmes:
/test 4.4-e2e-aws-ocp-44

CIIIIII

PR #420, opened 2020-08-28
Scale up workers only for upstream E2E tests

mgencur:
This properly sets the SCALE_UP variable to 6 and scales up only for upstream E2E tests

/cc @maschmid @rhuss @matzew

### DIFF
--- a/hack/lib/scaleup.bash
+++ b/hack/lib/scaleup.bash
@@ -8,6 +8,11 @@ function scale_up_workers {
     return 0
   fi
 
+  if ! cluster_scalable; then
+    logger.info 'Skipping scaling up, the cluster is not scalable.'
+    return 0
+  fi
+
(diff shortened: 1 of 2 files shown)


mgencur:
/hold
Need to check what to do with vsphere tests where we need SCALE_UP=-1

nak3:
/retest

All setup issue.

nak3:
/retest

CI infra issue again.

mgencur:
TestSinkBinding* tests fail from Eventing. It whould be fine to merge this PR. The scaling seems is set correctly.

rhuss:
I agree, so let's get it merged manually.

mgencur:
Thanks. You could still merge it by putting lgtm on it. The particular Prow check is not mandatory. Just FYI.

PR #421, opened 2020-09-01
Include test workaround for SRVCOM-1008

mgencur:
https://issues.redhat.com/browse/SRVCOM-1008

This real fix should be refactoring the operator to be an Extension, thereby having a single deployment monitoring that KS resource, avoiding the potential of multiple operators having different (cached?) views of the same CR.

### DIFF
--- a/test/e2e/knative_serving_test.go
+++ b/test/e2e/knative_serving_test.go
@@ -62,6 +62,11 @@ func TestKnativeServing(t *testing.T) {
 		}
 		// Check the status of deployments in the ingress namespace.
 		for _, deployment := range []string{"3scale-kourier-control", "3scale-kourier-gateway"} {
+			// Workaround for https://issues.redhat.com/browse/SRVCOM-1008 - wait for Kourier deployments to
+			// be ready before checking their scales.
+			if _, err := test.WithDeploymentReady(caCtx, deployment, servingNamespace+"-ingress"); err != nil {
+				t.Fatal("Failed", err)
+			}
 			if err := test.CheckDeploymentScale(caCtx, servingNamespace+"-ingress", deployment, haReplicas); err != nil {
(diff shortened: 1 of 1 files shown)


markusthoemmes:
Alternatively, I believe we could work on our readiness of the serverless-operator itself as I believe this issue is due to the fact that serving-operator and serverless-operator start separately. If we can wait with the creation of the `KnativeServing` resource until our serverless-operator webhook is truly active (which we should somehow ensure via OLM means anyway) we should be consistently passing this.

mgencur:
@markusthoemmes is this a real fix what you're proposing or a workaround?

markusthoemmes:
@mgencur the real deal will be what @jcrossley3 alluded to, but us understanding how OLM propagates readiness for things it deploys would benefit that as well so I believe investing in that rather than a workaround would be beneficial. Specifically, maybe we need a readiness probe for the serverless-operator and its webhook here to make it more consistent?

mgencur:
Maybe I'm wrong here but what you're proposing still seems like a workaround...

markusthoemmes:
Well, kind of yes but then don't we want "READY" from OLM to mean really ready? Currently that doesn't seem to be the case.

mgencur:
OK. Yeah. We want "ready" to mean ready. If the real solution (operator being an Extension) is far away then it probably makes sense to fix this. It's a matter of a few more seconds after the KnativeServing reports ready and before Serverless is really ready. Not sure how likely it is to hit in production so I thought to just "fix" the test for the mean time. But I'll defer to you how/whether to fix it now.

markusthoemmes (approved):
/lgtm
/approve

We agreed this is fine for now, will comment on a "better" fix to the issue if we decide to pick that up again.

PR #422, opened 2020-09-01
[WIP] Delete hanging resources

aliok:
https://issues.redhat.com/browse/SRVKE-473

`pingsource-jobrunner` deployment is hanging as described in the ticket.

I am trying to delete it afterwards. Having some problems though. Will comment inline

### DIFF
--- a/knative-operator/pkg/controller/knativeeventing/knativeeventing_controller.go
+++ b/knative-operator/pkg/controller/knativeeventing/knativeeventing_controller.go
@@ -3,12 +3,15 @@ package knativeeventing
 import (
 	"context"
 	"fmt"
-
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/common"
 	"github.com/operator-framework/operator-sdk/pkg/predicate"
+	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


knative-operator/pkg/controller/knativeeventing/knativeeventing_controller.go
```diff
@@ -109,3 +122,68 @@ func (r *ReconcileKnativeEventing) configure(instance *eventingv1alpha1.KnativeE
...
+func (r *ReconcileKnativeEventing) delete(instance *eventingv1alpha1.KnativeEventing) error {
+	finalizers := sets.NewString(instance.GetFinalizers()...)
+
+	if !finalizers.Has(finalizerName) {
+		log.Info("Finalizer has already been removed from KnativeEventing, nothing to do")
+		return nil
+	}
+
+	log.Info("Running cleanup logic for KnativeEventing")
+	if err := r.deleteHangingResources(instance); err != nil {
```
aliok, line 148: This deletes the deployment but the controller that watches this deployment, pingsource-controller, recreates it.
In this stage, that controller is not deleted yet.

Looking for help to solve this.

@markusthoemmes @jcrossley3

aliok:
/close

There's a race here and I can't fix it:
- Operator deletes the pingsource-jobrunner deployment
- Pingsource controller recreates the deployment
- Operator deletes all resources in the manifest (jobrunner deployment above is not in the manifest)
- Jobrunner deployment hangs

See the chat: https://coreos.slack.com/archives/CD87JDUB0/p1598950649040900

I will check other ways of fixing it.

PR #423, opened 2020-09-01
Migrate master to the new bundle format

warrenvw:
Migrate master to new bundle format.


olm-catalog/serverless-operator/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM scratch
+
+COPY manifests /manifests
+COPY metadata/annotations.yaml /metadata/annotations.yaml
+
+LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
+LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
+LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
+LABEL operators.operatorframework.io.bundle.package.v1=serverless-operator
+LABEL operators.operatorframework.io.bundle.channel.default.v1="4.5"
```
mgencur, line 10: Previously, we had the default channel `4.3` . Why is it changing now? We're still targeting 4.5 ?

hack/lib/vars.bash
```diff
@@ -11,13 +11,19 @@ fi
...
-export KNATIVE_SERVING_BRANCH="${KNATIVE_SERVING_BRANCH:-release-${KNATIVE_SERVING_VERSION}}"
-export KNATIVE_SERVING_REPO="${KNATIVE_SERVING_REPO:-"https://github.com/openshift/knative-serving.git"}"
-export KNATIVE_EVENTING_BRANCH="${KNATIVE_EVENTING_BRANCH:-release-${KNATIVE_EVENTING_VERSION}}"
-export KNATIVE_EVENTING_REPO="${KNATIVE_EVENTING_REPO:-"https://github.com/openshift/knative-eventing.git"}"
+# Adjust these when cutting a new CSV.
+# TODO: Read these from metadata.
+export CURRENT_CSV="serverless-operator.v1.9.0"
+export PREVIOUS_CSV="serverless-operator.v1.8.0"
+
+readonly KNATIVE_SERVING_BRANCH="${KNATIVE_SERVING_BRANCH:-release-${KNATIVE_SERVING_VERSION}}"
```
mgencur, line 23: I'd say these Serving and Eventing branches and repos are not used anywhere. We now take the sources from the src image, not by cloning those repos.

hack/lib/vars.bash
```diff
@@ -11,13 +11,19 @@ fi
 # shellcheck disable=SC1091,SC1090
 source "$(dirname "${BASH_SOURCE[0]}")/../../test/vendor/knative.dev/test-infra/scripts/e2e-tests.sh"
 
-export KNATIVE_SERVING_VERSION="${KNATIVE_SERVING_VERSION:-v0.15.2}"
-export KNATIVE_EVENTING_VERSION="${KNATIVE_EVENTING_VERSION:-v0.15.2}"
+# Adjust these when upgrading the knative versions.
+readonly KNATIVE_SERVING_VERSION="${KNATIVE_SERVING_VERSION:-v0.15.2}"
```
mgencur, line 15: Can we use "export" everywhere rather than "readonly" ? The readonly variables cause problems for example when we want to re-use code from other repos.

hack/lib/vars.bash
markusthoemmes: Unrelated to this change though?

mgencur: Yeah. Unrelated. I'm curious why we're changing that in this PR though.

hack/lib/catalogsource.bash
```diff
@@ -13,8 +13,97 @@ function install_catalogsource {
...
+
+  # Build the bundle image in the cluster-internal registry.
+  oc -n "$OLM_NAMESPACE" new-build --binary --strategy=docker --name serverless-bundle
+  oc -n "$OLM_NAMESPACE" start-build serverless-bundle --from-dir olm-catalog/serverless-operator -F
+
+  # Undo potential changes to the CSV to not pollute the repository.
+  mv "${rootdir}/bkp.yaml" "$csv"
+
+  # HACK: Allow to run the index pod as root so it can create the necessary directories.
+  oc -n "$OLM_NAMESPACE" adm policy add-scc-to-user anyuid -z default
```
markusthoemmes, line 45: This shoudln't be needed anymore.

markusthoemmes (approved):
/lgtm
/approve
/hold

For the nit! Great job! :tada:

olm-catalog/serverless-operator/Dockerfile
mgencur: OK. This is fixed. Thanks

hack/lib/vars.bash
mgencur: Thanks for fixing.

markusthoemmes (approved):
/lgtm

jcrossley3 (changes requested):
Can you include a detailed description of the steps required to run the test suite against a development cluster, e.g. CRC, after making changes to the code beneath `knative-operator/`?

markusthoemmes:
@jcrossley3 the steps haven't changed from "before the bundle format" there. CI essentially has the same requirement in that we need to replace images "on the fly" there. That's the whole point of building the bundle image during the actual catalogsource creation with an Openshift build. In `catalogsource.bash` we replace image references in the CSV based on whether we're local or in CI, just as we did before.

The "hardcoded" references (pointing to nightly images) are now in the checked-in CSV to allow the ci-operator to generate a working index image that is promoted. Note how the index image refers to `registry.svc.ci.openshift.org/openshift/openshift-serverless-v1.8.0:serverless-bundle`. That is exactly the image that's being promoted by the 1.8 branch currently and that needs a working CSV in it. Since we can't do replace stuff during build-time, I changed the replacement methodology when implementing this in 1.8 from env-var replacements to `sed` based replacements to have a fully working version regardless.

---

To keep myself honest and make sure i'm not just rambling away, I just ran the tests in CRC as documented in the README already and stuff "just works" (tm). I also doublechecked that the operators deployed in `openshift-serverless` are indeed based on the images I've just built since they are running `docker.io` based references.

markusthoemmes:
Issue solved

/unhold

PR #424, opened 2020-09-01
cluster-local variants of the servicemesh test

maschmid:
Additional (cluster-local) scenarios for TestKsvcWithServiceMeshSidecar

### DIFF
--- a/test/servinge2e/servicemesh_test.go
+++ b/test/servinge2e/servicemesh_test.go
@@ -2,16 +2,32 @@ package servinge2e
 
 import (
 	"github.com/openshift-knative/serverless-operator/test"
+	core "k8s.io/api/core/v1"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	pkgTest "knative.dev/pkg/test"
+	"knative.dev/serving/pkg/apis/autoscaling"
+	"knative.dev/serving/pkg/apis/serving"
+	servingv1 "knative.dev/serving/pkg/apis/serving/v1"
 	"testing"
 )
 
 const (
 	// A test namespace that is part of the ServiceMesh (setup by "make install-mesh")
 	serviceMeshTestNamespaceName = "default"
-	serviceMeshTestImage = "gcr.io/knative-samples/helloworld-go"
+	serviceMeshTestImage         = "gcr.io/knative-samples/helloworld-go"
+	serviceMeshTestProxyImage    = "registry.svc.ci.openshift.org/openshift/knative-v0.15.2:knative-serving-test-httpproxy"
 )
+
+// A knative service acting as an "http proxy", redirects requests towards a given "host". Used to test cluster-local services
+func httpProxyService(name, host string) *servingv1.Service {
+	proxy := test.Service(name, serviceMeshTestNamespaceName, serviceMeshTestProxyImage, nil)
+	proxy.Spec.Template.Spec.Containers[0].Env = append(proxy.Spec.Template.Spec.Containers[0].Env, core.EnvVar{
+		Name:  "TARGET_HOST",
+		Value: host,
+	})
+
+	return proxy
+}
 
 // Skipped unless ServiceMesh has been installed via "make install-mesh"
 func TestKsvcWithServiceMeshSidecar(t *testing.T) {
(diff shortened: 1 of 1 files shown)


test/service.go
```diff
@@ -16,17 +16,18 @@ import (
 	servingv1 "knative.dev/serving/pkg/apis/serving/v1"
 )
 
-func Service(name, namespace, image string, annotations map[string]string) *servingv1.Service {
+func Service(name, namespace, image string, ksvcLabels map[string]string, revisionTemplateAnnotations map[string]string) *servingv1.Service {
```
nak3, line 19: Umm... I feel that argument is too many. How about adding label in servicemesh_test.go like this?

```
svc := test.Service(scenario.name, serviceMeshTestNamespaceName, serviceMeshTestImage, scenario.annotations))
svc.ObjectMeta.Labels = scenario.labels
```

This is just a preference, though.

test/servinge2e/servicemesh_test.go
```diff
@@ -32,41 +46,67 @@ func TestKsvcWithServiceMeshSidecar(t *testing.T) {
...
 		annotations: map[string]string{
-			"sidecar.istio.io/inject":					"false",
+			"sidecar.istio.io/inject": "false",
 		},
 		expectIstioSidecar: false,
+	},{
+		// A cluster-local variant of the "sidecar-via-activator" scenario
+		name: "local-sidecar-via-activator",
+		labels: map[string]string{
+			"serving.knative.dev/visibility": "cluster-local",
```
nak3, line 82: You can use https://github.com/knative/serving/blob/master/pkg/apis/serving/register.go#L97. (It will move to networking repo in the future, though.)

test/servinge2e/servicemesh_test.go
```diff
@@ -32,41 +46,67 @@ func TestKsvcWithServiceMeshSidecar(t *testing.T) {
...
+		},
+		annotations: map[string]string{
+			"sidecar.istio.io/inject":        "true",
+		},
+		expectIstioSidecar: true,
+		useProxy: true,
```
nak3, line 88: `useProxy: true` is used only when cluster-local test, correct? If so, we can remove it but just determine it by the label like https://github.com/knative/serving/blob/master/pkg/reconciler/route/resources/filters.go#L25.

test/service.go
maschmid: I tend to agree.

test/servinge2e/servicemesh_test.go
maschmid: Right, the original idea was to have the internal test func as dumb as possible, with all the logic defined in the scenario struct, but I guess in this case that's just unnecessary clutter....

nak3:
@maschmid And could you fix gofmt? It has some errors when I run `gofmt -d test/servinge2e/servicemesh_test.go`.

test/servinge2e/servicemesh_test.go
```diff
@@ -85,11 +125,39 @@ func TestKsvcWithServiceMeshSidecar(t *testing.T) {
...
 
+			serviceURL := service.Status.URL.URL()
+
+			// For cluster-local ksvc, we deploy an "HTTP proxy" service, and request that one instead
+			if service.GetLabels()[serving.VisibilityLabelKey] == serving.VisibilityClusterLocal {
+				// Deploy an "HTTP proxy" towards the ksvc (using an httpproxy image from knative-serving testsuite)
+				httpProxy, err := caCtx.Clients.Serving.ServingV1().Services(serviceMeshTestNamespaceName).Create(
+					httpProxyService(scenario.name+"-proxy", service.Status.URL.Host))
+				if err != nil {
+					t.Errorf("error creating ksvc: %v", err)
```
mgencur, line 136: Would it be possible to use t.Fatalf instead of t.Errorf plus return?

mgencur (commented):
Looks good. Just one question.

test/servinge2e/servicemesh_test.go
maschmid: We want the other scenarios to continue in case of error.. (and I guess we'd lose that if we did Fatalf, but I am not sure tbh.. )

mgencur: Hmm. It should end just the current test. The Fatalf is equivalent to Logf followed by FailNow. And FailNow docs: https://golang.org/src/testing/testing.go?s=23520:23546#L682

maschmid: FailNow would call runtime.goexit , so this would skip all the other TestKsvcWithServiceMeshSidecar scenarios...

maschmid: (we run all TestKsvcWithServiceMeshSidecar scenarios in a single goroutine)

maschmid: hmm..or don't we? what exactly does t.Run  do  if we are not running parallel...

maschmid: ok, maybe we do then... let's try it then...

mgencur: I don't know :-D

PR #427, opened 2020-09-04
Use spec.rule.Visibility instead of spec.Visibility in KIngress

nak3:
Since upstream deprecates spec.Visibility by https://github.com/knative/networking/commit/f8dacf70c1e8a7daa7dd8847f42f3eb70fc5496a
This patch changes to use spec.rule.Visibility instead of spec.Visibility to create OpenShift Route.

For the context of `spec.visibility`:
- Since Knative v0.13, the `spec.visibility` always sets to `External` (default value) by [this commit](https://github.com/knative/serving/commit/b0c2013b48e86d9ebefb5f5a9c773d0c1cee1c0f#diff-7aca98cf2c25eb7a257bbe27cdfc7682L136.)
- The route controller always set `spec.rules.visibility` instead.
- Hence downstream (based on serving v0.16) is safe to use `spec.rules.visibility`.

### DIFF
--- a/serving/ingress/pkg/controller/ingress/ingress_controller.go
+++ b/serving/ingress/pkg/controller/ingress/ingress_controller.go
@@ -156,37 +156,30 @@ func (r *ReconcileIngress) ReconcileIngress(ctx context.Context, ing *networking
 		return r.reconcileDeletion(ctx, ing)
 	}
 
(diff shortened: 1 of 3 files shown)


markusthoemmes (approved):
/lgtm
/approve

Per discussion in Slack this field is always External since 0.13 anyway. The visibility of the rules is what matters. This change is thus safe.

PR #429, opened 2020-09-04
[POC] Prototype to automatically connect Knative Serving with skupper.

markusthoemmes:
This is a prototype on how Knative Serving integration with Skupper can look like.

I added a new annotation `serving.knative.openshift.io/multi-cloud-routing` which can be added to a Knative Service. If it's added, our Ingress will not create routes the "usual way", but instead create a K8s Service/Route combination that routes all requests via skupper so they automatically get spread across the entire skupper network and thus across multiple cloud providers in theory.

To set it up:

```console
$ make images install
$ kubectl apply -k skupper/rbac
$ kubectl apply -k skupper/controller
# Wait for everything to be up just to be sure.

$ kubectl apply -f skupper/service.yaml
$ curl $URL_OF_THE_SERVICE
```

**Caveats:** When first deploying things, it might take a while for the service to actually work but once it does it works great. Haven't looked deeper into what's the root cause of that just yet.

### DIFF
--- a/serving/ingress/pkg/controller/ingress/ingress_controller.go
+++ b/serving/ingress/pkg/controller/ingress/ingress_controller.go
@@ -29,6 +29,8 @@ import (
 	"github.com/openshift-knative/serverless-operator/serving/ingress/pkg/controller/ingress/resources"
 )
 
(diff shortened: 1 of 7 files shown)


markusthoemmes:
Closing for now.

PR #431, opened 2020-09-04
SRVKS-595 TestKsvcWithServiceMeshJWTDefaultPolicy

maschmid:


test/servinge2e/servicemesh_test.go
```diff
@@ -190,3 +231,295 @@ func TestKsvcWithServiceMeshSidecar(t *testing.T) {
...
+		Name:  "RESPONSE",
+		Value: jwks,
+	})
+	jwksKsvc.ObjectMeta.Labels = map[string]string{
+		serving.VisibilityLabelKey: serving.VisibilityClusterLocal,
+	}
+	jwksKsvc = withServiceReadyOrFail(caCtx, jwksKsvc)
+
+	// Create a Policy
+	authentication := &unstructured.Unstructured{
```
mgencur, line 362: probably, a better name would be `authPolicy` . ...authentication is too general and overloaded

test/servinge2e/servicemesh_test.go
```diff
@@ -190,3 +231,295 @@ func TestKsvcWithServiceMeshSidecar(t *testing.T) {
 		})
 	}
 }
+
+// formats an RSA public key as JWKS
+func rsaPublicKeyAsJwks(key rsa.PublicKey, kid string) (string, error) {
+	eString := base64.RawURLEncoding.EncodeToString(big.NewInt(int64(key.E)).Bytes())
+	nString := base64.RawURLEncoding.EncodeToString(key.N.Bytes())
+	kidString := base64.RawURLEncoding.EncodeToString([]byte(kid))
```
mgencur, line 239: What about `keyIdString` and the function argument named as `keyId` ? I was first thinking whose KID is it? :)

test/servinge2e/servicemesh_test.go
```diff
@@ -18,6 +34,18 @@ const (
 	serviceMeshTestProxyImage    = "registry.svc.ci.openshift.org/openshift/knative-v0.15.2:knative-serving-test-httpproxy"
 )
 
+func assumeServiceMeshInstalled(ctx *test.Context) {
+	// Verify the serviceMeshTestNamespace is part of the mesh
+	namespace, err := ctx.Clients.Kube.CoreV1().Namespaces().Get(serviceMeshTestNamespaceName, meta.GetOptions{})
+	if err != nil {
+		ctx.T.Fatalf("failed to verify %q namespace labels: %v", serviceMeshTestNamespaceName, err)
```
mgencur, line 41: AFAIK, whenever we call `Fatalf` or `Fatal` we start the message with a capital letter. That's how it's done in most tests in this repo.

test/servinge2e/servicemesh_test.go
```diff
@@ -190,3 +231,295 @@ func TestKsvcWithServiceMeshSidecar(t *testing.T) {
...
+
+	// Now verify a few scenarios of invalid tokens
+
+	// Generate a new key for a "wrong key" scenario
+	wrongKey, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		t.Fatalf("error generating private key: %v", err)
+	}
+
+	tests := []struct {
```
mgencur, line 464: I'd say the test above could be added to the table test below as well. The table can also define whether it expects success or failure so all tests could be there.

mgencur (commented):
Overall looks good, just a few nits and suggest to put the standalone test into the table test as well.

mgencur:
It might be good to let the test pass once (by installing service-mesh) so that we know it passes properly and then remove the service-mesh install in the next run. 
Right now it fails for some strange reason while it should be skipping the test, IMO

test/servinge2e/servicemesh_test.go
maschmid: That's how it originally was... however, the problem with that is if something is broken in such way that the token is not accepted at all,  all the negative tests will trivially Pass  (as they get the expected 401).

So then you see 3/4 tests passing, with just one failure (the test with the valid token)... 

So IMHO it is better to fail fully if the valid token is not accepted...

mgencur: In that case you could probably just call `return` or `break` in the given iteration and skip the other tests? Not sure whether we can check if the token was accepted at all. If yes, it would be good to add the check.

test/servinge2e/servicemesh_test.go
maschmid: is that really our convention? I thought all those cases are just violations of the https://github.com/golang/go/wiki/CodeReviewComments#error-strings convention...

test/servinge2e/servicemesh_test.go
maschmid: not in the given iteration (t.Run runs in a separate goroutine) . We could conditionally not run the t.Run, but that would mean breaking the separation if the individual tests (leaking test result into the upper test logic), which I think would be an antipattern...

maschmid: Basically, in case the valid token is not accepted, testing whether any other _invalid_ token is accepted is pointless... that's why we skip the whole "testing of invalid tokens" section in that case...

maschmid: One possible alternative I can see would be to move the "valid token" test into each of the "invalid token" test scenario, instead of being a separate test... that would make all of the "invalid token" scenarios valid even on their own...

maschmid:
fixed the test skipping, see https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift-knative_serverless-operator/431/pull-ci-openshift-knative-serverless-operator-master-4.5-operator-e2e-aws-ocp-45/1303023758998507520 

let's try a run with the mesh installed then...

maschmid:
/hold

forcing the install-mesh target for the purpose of testing

maschmid:
Passed with service mesh installed, https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift-knative_serverless-operator/431/pull-ci-openshift-knative-serverless-operator-master-4.5-operator-e2e-aws-ocp-45/1303048390715641856

test/servinge2e/servicemesh_test.go
mgencur: I see a little difference between using `fmt.Error` and `t.Fatal` . The strings printed through `t.Fatal` are usually (always?) at the beginning right after the test name: `blue_green_test.go:163: Error sending requests: ....` 
Everywhere I look in Knative there's a capital letter used with `t.Fatal`

test/servinge2e/servicemesh_test.go
mgencur: IMO, it's not necessary to skip all negative tests if the "valid token" test fails. You see them dependent on each other but they're still valid individually. If the valid token test fails you still want the other negative tests to pass and make sure that invalid tokens are not accepted. At this point you don't care about the valid token but the invalid ones. I don't see those tests to be so tight together. This scenario is even more unlikely but: let's see the valid token test fails, and then the system accepts invalid token and returns success but we're skipping the test so we don't know that. I'd say it's better to keep all tests running. But don't insist one that much.

mgencur:
Thanks for running the tests with Service Mesh enabled.

test/servinge2e/servicemesh_test.go
maschmid: Knative use capital letter in t.Errorf as well...  I don't think a convention where Errorf => lowercase , Fatalf => capital case  would make sense... we should either follow https://github.com/golang/go/wiki/CodeReviewComments#error-strings  (all starting lowercase)  ,  or follow the Knative convention (all capital case)     ( I don't care either way, just want to have it consistent)

maschmid: OK, never mind,  apparently those cases in Knative are just outliers...

maschmid: I see your point of Fatal always appearing at the beginning...

test/servinge2e/servicemesh_test.go
maschmid: OK, I have moved the valid token scenario to the list.

test/servinge2e/servicemesh_test.go
maschmid: ok, capitalized all Fatalf messages..

mgencur: Thank you

PR #433, opened 2020-09-08
fix ctx.T not referencing current testing.T

maschmid:
We cannot create Context once, as T needs to reference current testing.T.  Instead, only create the Clients once and create a new Context per each test.

### DIFF
--- a/test/clients.go
+++ b/test/clients.go
@@ -48,56 +48,56 @@ type Clients struct {
 // and register with the Context
 type CleanupFunc func() error
 
-var contexts []*Context
+var clients []*Clients
 
-// setupContextsOnce creates context objects for all kubeconfigs passed from the command line
-func setupContextsOnce(t *testing.T) {
-	if len(contexts) == 0 {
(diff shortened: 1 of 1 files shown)


mgencur:
I'm curious. How did you find out that ctx.T doesn't reference current testing.T? What was the error?

test/clients.go
```diff
@@ -48,56 +48,56 @@ type Clients struct {
 // and register with the Context
 type CleanupFunc func() error
 
-var contexts []*Context
+var clientsList []*Clients
```
mgencur, line 51: I'd say the word `List` is quite redundant here, isn't that an impl detail?. Using just `clients` would be just fine, IMO. Same for the other functions which include "clientsList".

Otherwise looks good.

maschmid: it is a list of "Clients", which are already plural, but sure...

maschmid:
It actually panicked if the test would call ctx.T.Failf or ctx.T.Skipf , instead of just failing the test.

PR #432, opened 2020-09-08
Bump deps in test and serving/ingress to knative 0.15

nak3:
This patch bumps up dependencies as mentioned in https://github.com/openshift-knative/serverless-operator/issues/426.

Fixes https://github.com/openshift-knative/serverless-operator/issues/426

PR #434, opened 2020-09-08
fix rerunning servicemesh test

maschmid:


test/servinge2e/servicemesh_test.go
```diff
@@ -336,6 +337,16 @@ func jwtHttpGetRequestBytes(url string, token *string) (*http.Response, []byte,
 	return resp, body, err
 }
 
+// randomSuffix generates a random suffix suitable for making unique k8s resource names
+func randomSuffix(t *testing.T) string {
```
mgencur, line 341: We could probably reuse https://github.com/knative/pkg/blob/master/test/helpers/name.go#L59

maschmid: OK, I'll wait for https://github.com/openshift-knative/serverless-operator/pull/432 first then, as that would conflict with the Gopkg and vendoring changes there.

rhuss:
From my naive perspective that looks good (tests are passing, too), so I'm fine if @markusthoemmes has no objections.

matzew:
@maschmid do you need this also on the (fresh) `release-1.9` branch ? 

Currently the PR for "CI runs" on that branch has _not_ been merged, yet /cc @mgencur

maschmid:
@matzew it is not critical to backport this for CI, these tests are skipped by default, as we don't install servicemesh on CI (and the problem this PR fixes manifests only when rerunning the test on the same cluster, so it won't affect CI either way... )

I can open a new backport PR for release-1.9 once this one lands in master.

PR #449, opened 2020-09-15
Remove unused files and roles.

markusthoemmes:
As per title, this removes files we don't use and reduces the role captured for the ingress component to a minimum.


markusthoemmes:
I'll redo this later.

PR #450, opened 2020-09-15
Replace value with image in relatedImages

warrenvw:
Should use `image` instead of `value`. Fixed.


markusthoemmes (approved):
/lgtm
/approve

PR #452, opened 2020-09-16
Fix typo in serverless-operator.clusterserviceversion.yaml

nak3:
This patch fixes a typo `s/commangd/command/g`.

Although it seems we can drop the `command` as it is not used actually, this patch fixes it.

/cc @warrenvw @markusthoemmes


matzew (approved):
/lgtm
/approve

markusthoemmes (approved):
/lgtm
/approve

PR #455, opened 2020-09-16
Disable cr metrics port in master

skonto:
- comments out the cr metrics port since it breaks the release build

/cc @markusthoemmes


markusthoemmes (approved):
/lgtm
/approve

PR #456, opened 2020-09-16
Remove controller-runtime client where not necessary.

markusthoemmes:
No behavioral change intended, this just reduces the surface area of the controller-runtime client a bit to see more clearly where it's actually needed.


skonto:
@markusthoemmes I am also refactoring part of this in the PR that is coming soon.

PR #457, opened 2020-09-17
Create serverless OperatorGroup only if there are no other operator g…

mgencur:
…roups

* This is important for upgrades (downstream) where we keep using openshift-operators NS and
there's already another operator group in that NS. Having multiple OGs
breaks CSV installation.

### DIFF
--- a/hack/lib/namespaces.bash
+++ b/hack/lib/namespaces.bash
@@ -7,13 +7,16 @@ function create_namespaces {
       oc create ns "${ns}"
     fi
   done
-  cat <<EOF | oc apply -f - || return $?
+  # Create an OperatorGroup if there are no other ones in the namespace.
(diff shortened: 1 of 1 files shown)


jcrossley3:
I'm surprised that breaks OLM. Why does it care?

mgencur:
@jcrossley3 The error message is like this: 
```
Danger alert:Failed: csv created in namespace with multiple operatorgroups, can't pick one automatically
```

markusthoemmes (commented):
Why do we even need these OperatorGroups in all of the namespaces we create?

mgencur:
That's required by OLM. It specifies which namespaces the operators targets etc. https://docs.openshift.com/container-platform/4.5/operators/understanding_olm/olm-understanding-operatorgroups.html#olm-operatorgroups-about_olm-understanding-operatorgroups

markusthoemmes (approved):
/lgtm
/approve

PR #458, opened 2020-09-17
Remove diff to avoid panic due to comparing unexported fields. (#385) - master

mgencur:
* Remove diff to avoid panic due to comparing unexported fields.

* Rerun dep ensure.


markusthoemmes (approved):
/lgtm
/approve

PR #459, opened 2020-09-17
Remove diff to avoid panic due to comparing unexported fields. (#385) - release-1.9

mgencur:
* Remove diff to avoid panic due to comparing unexported fields.

* Rerun dep ensure.


markusthoemmes (approved):
/lgtm
/approve

PR #472, opened 2020-09-24
Initial KnativeKafka type and controller

aliok:
Some limitations first:
- We can't run `operator-sdk` CLI on the serverless operator (actually the `knative-operator` subdir). It is not an operator-sdk-based project.
- I tried to setup Knative's generaration and injection tooling in serverless-operator. It doesn't work because we're still using `dep` here and Knative tooling doesn't work without Go modules. I think I could've make it work with 0.14 version of the tooling which stil uses `dep`, but that's a bad idea.

As per discussion with @markusthoemmes it is probably the best for now to create the generated stuff in a dummy project. I created this project for that: https://github.com/aliok/dummy-operator

Contents of this PR:
- Type
- Controller
- RBAC adjustment

CSV (or whatever it is called now) is not done yet. I won't do it until the controller actually does something.

Thus, I disabled the controller in a separate commit which can be reverted later.

In https://github.com/aliok/dummy-operator, the commit messages are the command I ran, if it helps.


knative-operator/deploy/crds/operator_v1alpha1_knativekafka_cr.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: operator.serverless.openshift.io/v1alpha1
+kind: KnativeKafka
+metadata:
+  name: example-knativekafka
+spec:
+  # Add fields here
+  size: 3
```
aliok, line 7: Default stuff generated by the operator-sdk cli, gonna change it later or remove this example CR file.

knative-operator/pkg/apis/operator/v1alpha1/knativekafka_types.go
```diff
@@ -0,0 +1,87 @@
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	duckv1 "knative.dev/pkg/apis/duck/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
```
markusthoemmes, line 8: Drop all of the notes please.

knative-operator/pkg/apis/operator/v1alpha1/zz_generated.openapi.go
```diff
@@ -0,0 +1,148 @@
+// +build !ignore_autogenerated
```
markusthoemmes, line 1: Let's ditch this file. It'll only confuse people.

knative-operator/pkg/apis/addtoscheme_operator_v1alpha1.go
```diff
@@ -1,11 +1,10 @@
 package apis
```
aliok, line 1: This file was adding the upstream operator schemas.
That happens in another file now: [knative-operator/pkg/apis/addtoscheme_knative_operator_v1alpha1.go](https://github.com/openshift-knative/serverless-operator/pull/472/files#diff-00904b3d74ed5c8bf905a77433451d24)

This file is now adding the stuff under `operator.serverless.openshift.io`.

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -0,0 +1,153 @@
...
+
+		// Pod created successfully - don't requeue
+		return reconcile.Result{}, nil
+	} else if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// Pod already exists - don't requeue
+	reqLogger.Info("Skip reconcile: Pod already exists", "Pod.Namespace", found.Namespace, "Pod.Name", found.Name)
+	return reconcile.Result{}, nil
```
markusthoemmes, line 129: Hollow this out to make clear it's not implemented yet?

knative-operator/pkg/controller/add_knativekafka.go
```diff
@@ -0,0 +1,11 @@
...
+
+//import (
+//	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/controller/knativekafka"
+//)
+
+func init() {
+	// AddToManagerFuncs is a list of functions to create controllers and add them to a manager.
+	// TODO: temp change to disable KnativeKafka reconciliation
+	// AddToManagerFuncs = append(AddToManagerFuncs, knativekafka.Add)
+}
```
aliok, line 11: Disabled for now since the CRD is not installed with the CSV yet. I will revert later once the CSV is adjusted and the controller actually does something with the CRs.

knative-operator/deploy/crds/operator_v1alpha1_knativekafka_cr.yaml
markusthoemmes: Let's just ditch it now anyway. It only confuses peeps and isn't even functional.

markusthoemmes (commented):
lgtm modulo the few comments I left.

knative-operator/deploy/crds/operator_v1alpha1_knativekafka_cr.yaml
matzew: agreed on removing

knative-operator/deploy/crds/operator_v1alpha1_knativekafka_crd.yaml
```diff
@@ -0,0 +1,120 @@
...
+    kind: KnativeKafka
+    listKind: KnativeKafkaList
+    plural: knativekafkas
+    singular: knativekafka
+  scope: Namespaced
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      properties:
```
matzew, line 17: above the `properties`, wanna add something like:

https://github.com/openshift-knative/serverless-operator/blob/master/olm-catalog/serverless-operator/manifests/operator_v1alpha1_knativeeventing_crd.yaml#L28-L29

knative-operator/deploy/crds/operator_v1alpha1_knativekafka_crd.yaml
```diff
@@ -0,0 +1,120 @@
...
+              properties:
+                bootstrapServers:
+                  description: BootstrapServers is comma separated string of bootstrapservers
+                    that the KafkaChannels will use
+                  type: string
+                enabled:
+                  description: Enabled defines if the KafkaChannel installation is
+                    enabled
+                  type: boolean
+                setAsDefaultChannelProvisioner:
```
matzew, line 43: should we add some filed to define what the actual shape of a default channel is ? 

e.g. for:
```
  numPartitions: 3
  replicationFactor: 1
```

Which than would translate to this CM value on the server:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: default-ch-webhook
  namespace: knative-eventing
data:
  # Configuration for defaulting channels that do not specify CRD implementations.
  default-ch-config: |
    clusterDefault:
      apiVersion: messaging.knative.dev/v1beta1
      kind: KafkaChannel
      spec:
        numPartitions: 3
        replicationFactor: 1
```

knative-operator/pkg/apis/operator/v1alpha1/knativekafka_types.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// BootstrapServers is comma separated string of bootstrapservers that the
+	// KafkaChannels will use
+	// +optional
+	BootstrapServers string `json:"bootstrapServers"`
+
+	// SetAsDefaultChannelProvisioner defines if KafkaChannel should be used
+	// as the default channel implementation when using generic Channels.
+	// +optional
+	SetAsDefaultChannelProvisioner bool `json:"setAsDefaultChannelProvisioner,omitempty"`
```
matzew, line 82: as said above, perhaps some option to layou tthe actual values of the channel defaults ?

knative-operator/deploy/crds/operator_v1alpha1_knativekafka_crd.yaml
aliok: Ok, this comment of yours made me remember something: https://issues.redhat.com/browse/SRVKE-397 and https://issues.redhat.com/browse/SRVKE-487

We should actually not have this field anymore, remember the discussion @matzew?

aliok: Our operator shouldn't go and touch that configmap because then we have an operator war over that configmap.

Let me know what you think and I will delete that afterwards.

matzew: I thought we agreed on some shape like:

```
Spec:
  source: 
    enable: true
  channel:
    enable: true
    default: true
    bootstrapServers: foo.bar:9092, baz.bla:9092
```
since the `KnativeEventing` operator CR owns the `default-ch-webhook` CM, we should translate the values from the `KnativeKafka` spec to it. Or do we expect users to touch the "ConfigMap" syntax on the `KnativeEventing` CR, after they _enabled_ the channel in the `KnativeKafka` CR ?

markusthoemmes: fwiw: `KnativeKafka` probably should not mutate the `KnativeEventing` CR since it has no ownership over it.

matzew: yeah, we had the issue in the past - I thought it's easier on a combined Operator.

OK... let's not do the `default` here ...

markusthoemmes: Well the `KnativeEventing` controller could read `KnativeKafka` and use it as "configuration" for its own reconcilation. That way, it'd totally work.

matzew: OK, that would be good.

That way we set the "relevant" info on the `KnativeKafka` CR, and do _NOT_ ask the user to use some funky ConfigMap sytanx to set the previous installed channel as the default.  

OK, that would be better UX, I guess

aliok: > yeah, we had the issue in the past - I thought it's easier on a combined Operator.

The upstream operator we use is the owner of that configmap, not our operator.

> Well the KnativeEventing controller could read KnativeKafka and use it as "configuration" for its own reconcilation. That way, it'd totally work.

I am not clear on this one. @markusthoemmes can you explain this a bit more please?

Downstream KnativeEventing controller doesn't own the configmap, so it also cannot change; same problem. 
It can change the KnativeEventing CR's config field (https://github.com/knative/operator/blob/master/docs/configuration.md#specconfig) so that the upstream operator can adjust the configmap. But, not sure if this is a good idea to change the CR except in a webhok.

markusthoemmes: that's right, good point @aliok. The operator extension design will make this possible though. I'd punt on it for now.

aliok: Ok, let's try to merge this PR for now and we can still think about it later.
I will remove that field now

knative-operator/deploy/crds/operator_v1alpha1_knativekafka_crd.yaml
aliok: Done, thanks

markusthoemmes:
@aliok you prolly won't get the upgrade test to pass as there is a slim chance currently that it actually does pass. 0.16 got new job images that the old version didn't have... so we race.

aliok:
@markusthoemmes all green, it passed :shrug: :D

markusthoemmes (approved):
/lgtm
/approve

PR #475, opened 2020-09-24
Fixing stupid wrong value name...should be image

matzew:

as per title


warrenvw (approved):
/lgtm
/approve

PR #476, opened 2020-09-25
Fail if yq is not on PATH

mgencur:
Otherwise this might leave to error which are not obvious. Such as 
```
./hack/lib/serverless.bash: line 104: $1: unbound variable
```


hack/lib/vars.bash
```diff
@@ -15,6 +15,8 @@ source "$(dirname "${BASH_SOURCE[0]}")/../../test/vendor/knative.dev/test-infra/
 export KNATIVE_SERVING_VERSION="${KNATIVE_SERVING_VERSION:-v0.16.0}"
 export KNATIVE_EVENTING_VERSION="${KNATIVE_EVENTING_VERSION:-v0.16.0}"
 
+# Make sure yq is on PATH.
+yq > /dev/null || exit 127
```
matzew, line 19: what person does NOT have `yq` installed ? 

/cc @skonto 

/lgtm
/approve

PR #477, opened 2020-09-25
Add a readiness probe that waits for the webhook to be up

markusthoemmes:
The webhook in the current implementation sets itself up. It applies Mutating/ValidationWebhookConfig objects and creates a service to point at itself. It does all of that before it starts the webserver actually serving the webhook webserver so by the time it does that, we know that the system knows about the webhook.
This should improve scripted installs in that the operator being ready now actually means ready and KnativeServing/Eventing can be created.

For posterity: This is where the webhook resources get installed. Not how it happens before `run` is called

https://github.com/openshift-knative/serverless-operator/blob/f72d9d1a08efeefdc752821a2bb340c8bf3dab06/knative-operator/vendor/sigs.k8s.io/controller-runtime/pkg/webhook/server.go#L206-L214

### DIFF
--- a/olm-catalog/serverless-operator/csv.template.yaml
+++ b/olm-catalog/serverless-operator/csv.template.yaml
@@ -290,6 +290,11 @@ spec:
                   command:
                   - knative-openshift
                   imagePullPolicy: Always
+                  readinessProbe:
+                    httpGet:
+                      scheme: HTTPS
+                      port: 9876
+                      path: "/mutate-knativeservings"
                   env:
                     - name: WATCH_NAMESPACE
                       value: ""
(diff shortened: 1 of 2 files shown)


aliok:
Looks good. Haven't tried it myself though. Does the mutation endpoint return 200 when called without proper data by the probe?

markusthoemmes:
@aliok I'd assume the pod wouldn't come up if it wouldn't return 200. But yeah, I doublechecked that before coding it up.

skonto:
I need to fix one of the integration tests the waiting condition is not correct. Will update in a minute.

markusthoemmes:
/retest

Known infra issues.

markusthoemmes:
/retest

IAM Infra stuff

matzew (approved):
/lgtm
/approve

PR #484, opened 2020-09-29
Disable connection keep-alive in networkpolicy test

markusthoemmes:
Per the SDN team, keep-alive connections would not be forcefully killed after being established through a NetworkPolicy. This makes sure we always establish a new connection for the test.


nak3:
/lgtm

Thank you!

PR #481, opened 2020-09-25
Fully qualify golang dep in build image

markusthoemmes:
Per discussion with Steve, this should alleviate some of our build failures.

PR #489, opened 2020-09-30
Using fqn image name

matzew:

Original attempt was in #481 

On slack `vrutkovs` suggested using this image instead


matzew:
/retest

failures seem to be not really realted to this

PR #479, opened 2020-09-25
Generate image lists in CSV

markusthoemmes:
As per title. It's oh so annoying to update these lists and we wanted to go towards generating a CSV anyway so this is a first step that automates the needed image list generation.

PR #490, opened 2020-09-30
Generating files that contain release bound variables

cardil:
To make working easier with our project, let's generate more files that changed with release by build tools.

That gives us:

- an easier way to manage values in `project.yaml`
- single source of truth for those variables
- reuse of those metadata values on other levels (productization, QE)

Follows up on https://github.com/openshift-knative/serverless-operator/pull/479


hack/generate/annotations.sh
```diff
@@ -0,0 +1,24 @@
...
+
+set -Eeuo pipefail
+
+template="${1:?Provide template file as arg[1]}"
+target="${2:?Provide a target annotations file as arg[2]}"
+
+# shellcheck disable=SC1091,SC1090
+source "$(dirname "${BASH_SOURCE[0]}")/../lib/metadata.bash"
+
+channel_list=$(yq read metadata.yaml 'olm.channels.list.*' | paste -sd ',' -)
```
cardil, line 11: This is overlooked. I need to parametrize the `metadata.yaml` file here.

metadata.yaml
```diff
@@ -0,0 +1,26 @@
...
+requirements:
+  kube:
+    minVersion: 1.15.0
+  golang: 1.14
+
+components:
+  serving: 0.16.0
+  eventing: 0.16.0
+  cli: 0.16.1
+  kourier: 1.1.0
```
markusthoemmes, line 23: This needs to be `maistra` or something like that. `kourier` needs to be 0.16.0 in this case.

cardil:
This PR supersedes https://github.com/openshift-knative/serverless-operator/pull/357

metadata.yaml
cardil: This is used in https://github.com/cardil/serverless-operator/blob/5caa536/hack/generate/csv.sh#L24 

I'm ok to changing that key to `maistra`, but I don't understand the second part of your comment: 

> kourier needs to be 0.16.0 in this case.

markusthoemmes: Kourier is a separate component that is not necessarily at serving's version.

cardil: Are you talking about this repo: https://github.com/knative-sandbox/net-kourier? Serverless Operator doesn't refer to it, isn't? Changing this to maistra will remove the confusion.

hack/generate/csv.sh
```diff
@@ -0,0 +1,97 @@
...
+images["queue-proxy"]="${serving}-queue"
+images["activator"]="${serving}-activator"
+images["autoscaler"]="${serving}-autoscaler"
+images["autoscaler-hpa"]="${serving}-autoscaler-hpa"
+images["controller"]="${serving}-controller"
+images["webhook"]="${serving}-webhook"
+images["storage-version-migration-serving-$(metadata.get components.serving)__migrate"]="${serving}-storage-version-migration"
+
+images["3scale-kourier-gateway"]="docker.io/maistra/proxyv2-ubi8:$(metadata.get components.kourier)"
+images["3scale-kourier-control"]="${registry}/knative-v$(metadata.get components.serving):kourier"
```
markusthoemmes, line 25: Regarding the other comment: This is wrong. The serving version does not necessarily equal the kourier version, hence we need a kourier component in the metadata too.

hack/lib/vars.bash
```diff
@@ -51,15 +49,8 @@ export UPGRADE_OCP_IMAGE="${UPGRADE_OCP_IMAGE:-}"
...
-function latest_channel_from_metadata {
-  local annotations_file channels
-  annotations_file="$(dirname "${BASH_SOURCE[0]}")/../../olm-catalog/serverless-operator/metadata/annotations.yaml"
-  channels="$(yq r "$annotations_file" 'annotations."operators.operatorframework.io.bundle.channels.v1"')"
-  # Return the last channel.
-  echo "$channels" | awk -F"," '{ print $NF }'
-}
 
-export OLM_CHANNEL="${OLM_CHANNEL:-"$(latest_channel_from_metadata)"}"
+export OLM_CHANNEL="${OLM_CHANNEL:-$(metadata.get 'olm.channels.alternatives[]')}"
```
mgencur, line 53: I didn't see "alternatives" anywhere else. Should this be "list" ?

templates/main.Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+LABEL \
+      com.redhat.component="openshift-serverless-1-serverless-rhel8-operator-container" \
+      name="openshift-serverless-1/serverless-rhel8-operator" \
+      version="__VERSION__" \
+      summary="Red Hat OpenShift Serverless 1 Serverless Operator" \
+      maintainer="serverless-support@redhat.com" \
+      description="Red Hat OpenShift Serverless 1 Serverless Operator" \
+      io.k8s.display-name="Red Hat OpenShift Serverless 1 Serverless Operator" \
+      com.redhat.openshift.versions="v4.5" \
```
mgencur, line 21: This "v4.5" could be in the template as well.

templates/test-source-image.Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM src
+
+COPY oc /usr/bin/oc
+COPY --from=registry.svc.ci.openshift.org/openshift/knative-v__SERVING_VERSION__:knative-serving-src
```
mgencur, line 4: Would it make sense to use `${SERVING_VERSION}` instead of `__SERVING_VERSION__ ` this way it would be clear that it's going to be replaced. And same approach for other places in this PR.

hack/lib/vars.bash
cardil: Yup. It was named like that. It slipped.

metadata.yaml
markusthoemmes: Yes we do reference it, via https://github.com/openshift-knative/net-kourier/

templates/main.Dockerfile
cardil: Yeah. But, I didn't know what it is. How to name this?! @warrenvw ?

templates/test-source-image.Dockerfile
cardil: I didn't use `${SERVING_VERSION}` in Dockerfiles, because it's a legal Docker expression. If it slips in future someone could overlook it, theoretically. It may also be confusing.

We could use `#{SERVING_VERSION}` or stick with `__SERVING_VERSION__`.

What's your opinion?

hack/generate/csv.sh
cardil: Done.

metadata.yaml
cardil: Done.

templates/test-source-image.Dockerfile
mgencur: OK. I get it. Maybe it should be something that Docker doesn't parse. Either `#{}` or `{{ }}`. But it probably doesn't matter.

templates/main.Dockerfile
mgencur: @cardil AFAIK, this means the target OCP version we're releasing for. This should also ensure "stage" index images are built for both 4.5 and 4.6. (when the other Warren's PR is merged)

templates/test-source-image.Dockerfile
cardil: I opt to stay with `__VARNAME__` style. I tried other styles, and they break syntax coloring for Dockerfiles.

templates/main.Dockerfile
cardil: Done

jcrossley3:
I recognize the need for the extra tooling -- to package our operator for OLM -- but it only reinforces my belief that they should not live in the same repo. I wouldn't want any of these scripts to impact the daily development cycle of the operator developers.

cardil:
> I recognize the need for the extra tooling -- to package our operator for OLM -- but it only reinforces my belief that they should not live in the same repo. I wouldn't want any of these scripts to impact the daily development cycle of the operator developers.

I agree. Extracting that to OLM related repo would be good idea. We need a single source of truth about our product. All those parameters are required to be "known" when testing for CVP, INTEROP, or new upcoming serverless release.

cardil:
@warrenvw @mgencur @markusthoemmes can I get review of this?

templates/build-image.Dockerfile
```diff
@@ -0,0 +1,13 @@
+# Dockerfile to bootstrap build and test in openshift-ci
+
+FROM openshift/origin-release:golang-__GOLANG_VERSION__
+
+# Add kubernetes repository
+ADD openshift/ci-operator/build-image/kubernetes.repo /etc/yum.repos.d/
+
+RUN yum install -y kubectl ansible httpd-tools
+
+RUN GO111MODULE=on go get github.com/mikefarah/yq/v3
```
markusthoemmes, line 10: Can we install `yq` the same way for the action above? We had issues with different formats in the past sadly.

olm-catalog/serverless-operator/project.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+dependencies:
+  serving: 0.16.0
+  eventing: 0.16.0
+  kourier: 0.16.0
+  cli: 0.16.1
+  maistra: 1.1.0
+
+deprecated:
+  eventing-images: 0.15.2
```
markusthoemmes, line 31: I believe this can go away. It was a one off AFAIK.

markusthoemmes (commented):
Love it! Do you mind a rebase to the 0.17 stuff?

olm-catalog/serverless-operator/project.yaml
```diff
@@ -0,0 +1,28 @@
...
+    minVersion: 1.15.0
+  golang: '1.14'
+  ocp:
+    - '4.5'
+    - '4.6'
+
+dependencies:
+  serving: 0.17.3
+  eventing: 0.17.2
+  kourier: 0.16.0
```
markusthoemmes, line 26: This is out of whack, I noticed while reviewing this PR :joy:. Fix in #576

olm-catalog/serverless-operator/project.yaml
```diff
@@ -0,0 +1,28 @@
...
+  cli: 0.17.2
+  maistra: 1.1.0
```
markusthoemmes, line 28: ditto.

markusthoemmes (commented):
LGTM modulo the comments

/assign @mgencur

olm-catalog/serverless-operator/project.yaml
cardil: Yes. I thought it was missed, but I didn't change it to make it consistent.

olm-catalog/serverless-operator/project.yaml
cardil: If #576 is merged first, I'll update it here as well.

templates/build-image.Dockerfile
cardil: I don't get you here. What are you proposing. This file is just a template taken from source that was just like this. https://github.com/openshift-knative/serverless-operator/blob/fb56d4c/openshift/ci-operator/build-image/Dockerfile#L10

.github/workflows/validate.yaml
```diff
@@ -0,0 +1,33 @@
+---
```
cardil, line 1: Running workflows can be seen at my fork: https://github.com/cardil/serverless-operator/actions

After merge they will be in this repo

cardil:
The latest commit aa2fb41 is passing [:heavy_check_mark: GH Action : Generated files are commited : 1260827343](https://github.com/cardil/serverless-operator/runs/1260827343)

olm-catalog/serverless-operator/project.yaml
```diff
@@ -21,6 +21,7 @@ requirements:
     - '4.6'
 
 dependencies:
+  knative-release: 0.17.0
```
mgencur, line 24: How did you determine 0.17.0? Just by looking at the CSV? It's confusing to have knative-release and also `eventing` and `serving` separately. Maybe there would be a better name for this variable.

olm-catalog/serverless-operator/project.yaml
```diff
@@ -0,0 +1,29 @@
+---
```
mgencur, line 1: I'm wondering if this `project.yaml` should be moved to the top-level directory. I think it affects the whole repo.

olm-catalog/serverless-operator/project.yaml
```diff
@@ -0,0 +1,29 @@
...
+  skipRange: '>=1.10.0 <1.11.0'
+  channels:
+    default: '4.3'
+    list:
+      - '4.5'
+      - '4.6'
+
+requirements:
+  kube:
+    minVersion: 1.15.0
```
mgencur, line 17: Nit: Some versions are quoted, some are note. E.g. `replaces: 1.10.0` vs `default: '4.3'` Is that intentional?

templates/build-image.Dockerfile
mgencur: I think Markus is referring to using `go get` vs `wget` as you do in the validate.yaml.

mgencur (commented):
Hey, just a few nits. Otherwise looks good.

olm-catalog/serverless-operator/project.yaml
cardil: This is required in one of the names of our images. A bit weird, yes.

olm-catalog/serverless-operator/project.yaml
cardil: It was at top level. I moved it into `olm-catalog/serverless-operator` so it will be packaged inside the image container. Then we get access to those variables via image.

olm-catalog/serverless-operator/project.yaml
cardil: Yup. Quotes are values that need to be. Otherwise, they will be treated as numbers. `1.10.0` is a string, `'4.3'` need to be quoted to not be a number.

mgencur: I see. OK

templates/build-image.Dockerfile
cardil: A yes. But, I think that's better to stick with wget as that will result in predictive version being used. `go get` can change version, and break something unpredictably.

mgencur: I think it's possible to use `go get` with a version specified but I couldn't download 3.4.0 this way.

cardil: Also, using wget/curl is faster and simpler. It doesn't require to have Golang setup & configured.

olm-catalog/serverless-operator/project.yaml
cardil: As of Slack discussion: Removing this as this is used only in single, one time clean job specific to 0.17 release.

cardil:
The latest commit 9b8b8ec is passing [:heavy_check_mark: GH Action : Generated files are commited : 1274900043](https://github.com/cardil/serverless-operator/runs/1274900043)

mgencur:
/lgtm

This is a good step forward towards automated releases. It concentrates the version definitions to one place which could save some confusion in the future, hopefully. Moving this stuff to a separate repo might be done in the future. I don't think this PR should be blocked by that.

PR #492, opened 2020-09-30
Revert "Add a readiness probe that waits for the webhook to be up"

markusthoemmes:
This reverts commit 6e5e35332d996270c80e5bc1d78720f71ef1fcf4.

This causes a lot of noise in the logs and isn't as important anymore given our bandaid solution in the upstream operator.


matzew (approved):
/lgtm
/approve

Issue #486, opened 2020-09-29
Umbrella issue for having Kafka components


aliok:
- [x] Create Kafka types: https://github.com/openshift-knative/serverless-operator/pull/472
- [x] Install Kafka components blindly (no namespace transforms, no owner refs, etc): https://github.com/openshift-knative/serverless-operator/pull/487
- [x] Set ownerRefs on channel resources which are in the same namespace, but not for KafkaSource resources which are in different namespace: https://github.com/openshift-knative/serverless-operator/pull/491
- [x] Set finalizer to be able to manually delete resources in other namespaces (knative-sources) as well as cluster scoped resources. Delete everything from manifest manually.: https://github.com/openshift-knative/serverless-operator/pull/493
- [x] Implement watching secondary resources for proper reconciliation: https://github.com/openshift-knative/serverless-operator/pull/496 AND https://github.com/openshift-knative/serverless-operator/pull/513
- [x] DeploymentsAvailable condition: https://github.com/openshift-knative/serverless-operator/pull/500
- [x] Remove the components when `spec.channel.enabled=false` or `spec.source.enabled=false`: https://github.com/openshift-knative/serverless-operator/pull/502
- [x] Use the `spec.channel.bootstrapservers` field in the spec to configure the relevant configmap: https://github.com/openshift-knative/serverless-operator/pull/503
- [x] Defaulting / requiredness check: https://github.com/openshift-knative/serverless-operator/pull/510
  - IMO it is better if `spec.channel.enabled` and `spec.source.enabled` are required.
  - Defaulting those to `true` doesn't make sense that much because need to specify the Kafka bootstrapServers for KafkaChannel.
  - Defaulting those to `false` doesn't bring any value
  - `spec.channel.bootstrapServers` needs to be required only when `spec.channel.enabled` is `true`
- [x] Use midstream manifest (probe changes, our images, etc.) --> probably needs a new patch release
  1. Need new src image for eventing-contrib (used in s-o e2e tests)
  2. Need new release yaml --> gonna copy & paste over to s-o
  3. Another thing: we need to separate the release YAML into 2 pieces for channel and source (becauce we now install these separately in s-o) --> created https://issues.redhat.com/browse/SRVKE-595
[...]

PR #493, opened 2020-09-30
Delete KnativeKafka resources when CR is deleted

aliok:
Bases https://github.com/openshift-knative/serverless-operator/pull/487

Similar to other PRs in the umbrella ticket, https://github.com/openshift-knative/serverless-operator/issues/486, I didn't want to wait until the CI is green and that's merged. This is a WIP PR to see in advance if things fail and if possibly get some reviews.

This PR deletes the resources when the KnativeKafka CR is deleted.
Specifically,
- KafkaChannel resources --> garbage collection also works for these since they're in the same namespace and we set ownerRefs for them
- KafkaSource resources that are put in another namespace, `knative-sources` -->  no GC, manual deletion for these
- Deletion of cluster-scoped resources

Deletion of the resources when a component gets disabled is to be handled in another PR (e.g. `spec.channel.enabled=true` is changed to `spec.channel.enabled=false`)

Verification:

1. Enable the controller by uncommenting the stuff in `knative-operator/pkg/controller/add_knativekafka.go`

2. Install operator and eventing by `make install`

```
```

4. Install:
```
cat <<EOS |kubectl apply -f -
---
kind: KnativeKafka
metadata:
  name: example
  namespace: knative-eventing
spec:
  source:
    enabled: true
  channel:
    enabled: true
...
EOS
```

5. Delete the CR and check if created resources are deleted:
```
```

6. Following commands should say `Error from server (NotFound): error when deleting "knative-operator/deploy/resources/knativekafka/kafka....yaml": foo "bar" not found` for each of the resource defined in them
```
k delete -f knative-operator/deploy/resources/knativekafka/kafkasource-latest.yaml
k delete -f knative-operator/deploy/resources/knativekafka/kafkachannel-latest.yaml
```

### DIFF
--- a/knative-operator/deploy/resources/knativekafka/README.md
+++ b/knative-operator/deploy/resources/knativekafka/README.md
@@ -1,0 +1,15 @@
+- Download the files from Github releases (no auto download at the moment)
+- Remove comments and blank lines with
+```
+sed -i \
+        -e '/^[ \t]*#/d' \
+        -e '/^[ \t]*$/d' \
+        kafkachannel-vX.XX.X.yaml
+
+sed -i \
+        -e '/^[ \t]*#/d' \
+        -e '/^[ \t]*$/d' \
+        kafkasource-vX.XX.X.yaml
+```
+- Create `kafkachannel-latest.yaml` and `kafkasource-latest.yaml` symlinks for the new files
+- Dump the old files
--- a/knative-operator/deploy/resources/knativekafka/kafkachannel-v0.17.4.yaml
+++ b/knative-operator/deploy/resources/knativekafka/kafkachannel-v0.17.4.yaml
@@ -1,17 +1,3 @@
-# Copyright 2019 The Knative Authors
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#     http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
@@ -19,7 +5,6 @@ metadata:
   labels:
     contrib.eventing.knative.dev/release: "v0.17.4"
     duck.knative.dev/addressable: "true"
-# Do not use this role directly. These rules will be added to the "addressable-resolver" role.
 rules:
 - apiGroups:
   - messaging.knative.dev
@@ -30,22 +15,7 @@ rules:
   - get
   - list
   - watch
-
 ---
-# Copyright 2019 The Knative Authors
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#     http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
@@ -53,7 +23,6 @@ metadata:
   labels:
     contrib.eventing.knative.dev/release: "v0.17.4"
     duck.knative.dev/channelable: "true"
-# Do not use this role directly. These rules will be added to the "channelable-manipulator" role.
 rules:
 - apiGroups:
   - messaging.knative.dev
@@ -67,22 +36,7 @@ rules:
   - watch
   - update
   - patch
-
 ---
-# Copyright 2019 The Knative Authors
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#     http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
@@ -181,22 +135,7 @@ rules:
   resources:
   - leases
   verbs: *everything
-
 ---
-# Copyright 2020 The Knative Authors
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#     http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
 apiVersion: v1
 kind: ServiceAccount
 metadata:
@@ -204,22 +143,7 @@ metadata:
   namespace: knative-eventing
   labels:
     contrib.eventing.knative.dev/release: "v0.17.4"
-
 ---
-# Copyright 2019 The Knative Authors
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#     http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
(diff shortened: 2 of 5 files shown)


knative-operator/deploy/resources/knativekafka/README.md
```diff
@@ -0,0 +1,15 @@
+- Download the files from Github releases (no auto download at the moment)
```
aliok, line 1: Simple instructions to download the KafkaChannel and KafkaSource manifests. Not really relevant to this PR, but let's merge this too.

knative-operator/deploy/resources/knativekafka/kafkachannel-v0.17.4.yaml
```diff
@@ -1,25 +1,10 @@
-# Copyright 2019 The Knative Authors
```
aliok, line 1: Removed all the comments and blank lines in the manifests

knative-operator/deploy/resources/knativekafka/kafkasource-v0.17.4.yaml
```diff
@@ -1,62 +1,18 @@
-# Copyright 2019 The Knative Authors
```
aliok, line 1: Removed all the comments and blank lines in the manifests

knative-operator/pkg/apis/operator/v1alpha1/knativekafka_lifecycle_test.go
```diff
@@ -1,19 +1,3 @@
-/*
```
aliok, line 1: As requested in the previous PR, removed the upstream license headers

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -125,6 +131,18 @@ func (r *ReconcileKnativeKafka) reconcileKnativeKafka(instance *operatorv1alpha1
 	return nil
 }
 
+// set a finalizer to clean up cluster-scoped resources and resources from other namespaces
+func (r *ReconcileKnativeKafka) ensureFinalizers(instance *operatorv1alpha1.KnativeKafka) error {
```
aliok, line 135: Now we're actually setting the finalizer on the CR.
This is necessary to be able to remove resources that can't be GC'ed because they're in a different namespace (`knative-sources` namespace)

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -155,27 +173,45 @@ func applyKnativeKafka(instance *operatorv1alpha1.KnativeKafka, api client.Clien
...
 }
 
+func kafkaChannelManifest(instance *operatorv1alpha1.KnativeKafka, apiClient client.Client) (*mf.Manifest, error) {
+	manifest, err := mfc.NewManifest(kafkaChannelManifestPath(), apiClient, mf.UseLogger(log.WithName("mf")))
+	if err != nil {
+		return nil, fmt.Errorf("failed to load KafkaChannel manifest: %w", err)
+	}
+
+	transformers := []mf.Transformer{
+		mf.InjectOwner(instance),
```
aliok, line 200: setting the ownerRef for the resources here for the KafkaChannel resources (which are in the same namespace with the CR)

knative-operator/deploy/resources/knativekafka/README.md
matzew: Not sure I get the value here, tbh

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -23,6 +24,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
+const (
+	// This needs to remain "knative-kafka-openshift" to be compatible with earlier versions.
+	finalizerName = "knative-kafka-openshift"
```
matzew, line 29: what earlier version? you mean the `knative-kafka` operator?

knative-operator/deploy/resources/knativekafka/README.md
aliok: Well, there's a `sed` command to delete comments and blank lines from manifests.
That's all :)

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
aliok: no, that's copy paste :)

but not a bad one actually. it means, don't change it to something else in the future!

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
matzew: IMO it would be nicer, if we just stash them all to the `knative-eventing` namespace.

I will propose the same upstream - insted of the `knative-sources` namespace ,....

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
matzew: OK, I'd perhaps do that explicit

matzew:
so. this looks good, and similar to related code here

/lgtm

I think @markusthoemmes is approver for final blessing :shrug:

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
aliok: well, we still need the finalizer, so, this code won't change.
There are also cluster scoped resources which can't be GC'ed and we still need the finalizer to prevent the deletion of KnativeKafka CR before them.

Happy to discuss the get rid of `knative-sources` to `knative-eventing` for KafkaSource.

IMO we should do that change separately once everything is working. It might require some upstream changes perhaps, I don't know yet.

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
aliok: Updated now

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -155,27 +174,45 @@ func applyKnativeKafka(instance *operatorv1alpha1.KnativeKafka, api client.Clien
...
 	return nil
 }
 
+func kafkaChannelManifest(instance *operatorv1alpha1.KnativeKafka, apiClient client.Client) (*mf.Manifest, error) {
+	manifest, err := mfc.NewManifest(kafkaChannelManifestPath(), apiClient, mf.UseLogger(log.WithName("mf")))
+	if err != nil {
+		return nil, fmt.Errorf("failed to load KafkaChannel manifest: %w", err)
+	}
+
+	transformers := []mf.Transformer{
```
markusthoemmes, line 200: Do we need the intermediary slice? Can't we just put the `mf.InjectOwner(instance)` call into the Transform call below?

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -216,3 +261,79 @@ func checkDeployments(manifest *mf.Manifest, api client.Client) error {
...
+func (r *ReconcileKnativeKafka) delete(instance *operatorv1alpha1.KnativeKafka) error {
+	finalizers := sets.NewString(instance.GetFinalizers()...)
+
+	if !finalizers.Has(finalizerName) {
+		log.Info("Finalizer has already been removed, nothing to do")
+		return nil
+	}
+
+	log.Info("Running cleanup logic")
+	log.Info("Deleting kourier")
```
markusthoemmes, line 275: copy and paste issue.

knative-operator/deploy/resources/knativekafka/README.md
markusthoemmes: Seems odd to dump on this PR but :man_shrugging:

markusthoemmes (commented):
Small nits, then LGTM

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
aliok: nope, we don't need it. let me get rid of that.

aliok: Done now

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
aliok: fixed, thx

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -155,27 +174,41 @@ func applyKnativeKafka(instance *operatorv1alpha1.KnativeKafka, api client.Clien
...
 	return nil
 }
 
+func kafkaChannelManifest(instance *operatorv1alpha1.KnativeKafka, apiClient client.Client) (*mf.Manifest, error) {
+	manifest, err := mfc.NewManifest(kafkaChannelManifestPath(), apiClient, mf.UseLogger(log.WithName("mf")))
+	if err != nil {
+		return nil, fmt.Errorf("failed to load KafkaChannel manifest: %w", err)
+	}
+
+	manifest, err = manifest.Transform(mf.InjectOwner(instance))
```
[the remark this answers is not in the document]
aliok: That's why the owner refs are only set for the channel resources, which live in `knative-eventing` namespace (same namespace as the CR).
It is not set for the KafkaSource resources, which live in `knative-sources` namespace.

> Otherwise if I create this in namespace foo, how are the owner refs going to work?

You can't create the KnativeKafka CR in a namespace other than `knative-eventing`. This is similar to existing CRs for serving and eventing. That check is done in another PR (https://github.com/openshift-knative/serverless-operator/pull/510/commits/d2895c8729793c98c9bfcd78606ea1f3d6b1865a#diff-186644545bff629e539219c7f924fc96R111) already but there's a lot of stuff to merge until that check is enabled.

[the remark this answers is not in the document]
markusthoemmes: Hmm, ok.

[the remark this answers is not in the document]
aliok: @matzew has some thoughts about using `knative-eventing` for KafkaSource as well, but I don't want to do this until I merge my existing work.

Also, not sure if it requires any changes in the upstream code.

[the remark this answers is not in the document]
matzew: @aliok let's track it

matzew:
@markusthoemmes wanna do the final blessing ?

markusthoemmes (approved):
/lgtm
/approve

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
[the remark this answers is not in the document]
aliok: @matzew already added in the umbrella ticket: https://github.com/openshift-knative/serverless-operator/issues/486

PR #496, opened 2020-09-30
Watch secondary resources of Kafka components

aliok:
Bases https://github.com/openshift-knative/serverless-operator/pull/493

Similar to other PRs in the umbrella ticket, https://github.com/openshift-knative/serverless-operator/issues/486, I didn't want to wait until the CI is green and that's merged. This is a WIP PR to see in advance if things fail and if possibly get some reviews.

-------------------------
## Changes
This PR creates watches for the secondary resources that are created.

Also, I noticed there were a lot of duplicate code, so I created reusable functions following in a separate PR (https://github.com/openshift-knative/serverless-operator/pull/513):
- injecting owner annotations
- creating reconciliation enqueue request for watched resources
- finding the resource types to watch
-------------------------

Verification:

1. Install eventing

```
```

4. Install:
```
cat <<EOS |kubectl apply -f -
---
kind: KnativeKafka
metadata:
  name: example
  namespace: knative-eventing
spec:
  source:
    enabled: true
  channel:
    enabled: true
...
EOS
```

5. Delete any resource created by the operator for Kafka and see if it is recreated.

6. Clean up:
```
```

### DIFF
--- a/knative-operator/pkg/common/constants.go
+++ b/knative-operator/pkg/common/constants.go
@@ -6,6 +6,6 @@ const (
 	ServingOwnerNamespace            = "serving.knative.openshift.io/ownerNamespace"
 	ServerlessOperatorOwnerName      = "operator.knative.openshift.io/ownerName"
 	ServerlessOperatorOwnerNamespace = "operator.knative.openshift.io/ownerNamespace"
-	KafkaOwnerName                   = "kafka.knative.openshift.io/ownerName"
-	KafkaOwnerNamespace              = "kafka.knative.openshift.io/ownerNamespace"
+	KafkaOwnerName                   = "knativekafkas.operator.serverless.openshift.io/ownerName"
+	KafkaOwnerNamespace              = "knativekafkas.operator.serverless.openshift.io/ownerNamespace"
 )
--- a/knative-operator/pkg/common/constants.go
+++ b/knative-operator/pkg/common/constants.go
@@ -6,4 +6,6 @@ const (
 	ServingOwnerNamespace            = "serving.knative.openshift.io/ownerNamespace"
 	ServerlessOperatorOwnerName      = "operator.knative.openshift.io/ownerName"
 	ServerlessOperatorOwnerNamespace = "operator.knative.openshift.io/ownerNamespace"
+	KafkaOwnerName                   = "kafka.knative.openshift.io/ownerName"
+	KafkaOwnerNamespace              = "kafka.knative.openshift.io/ownerNamespace"
 )
--- a/knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
+++ b/knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
@@ -203,7 +203,7 @@ func (r *ReconcileKnativeKafka) ensureFinalizers(instance *operatorv1alpha1.Knat
 
 // Install Knative Kafka components
 func (r *ReconcileKnativeKafka) installKnativeKafka(instance *operatorv1alpha1.KnativeKafka) error {
-	if err := r.applyKnativeKafka(instance, r.client); err != nil {
+	if err := r.applyKnativeKafka(instance); err != nil {
 		instance.Status.MarkInstallFailed(err.Error())
 		return err
 	}
@@ -211,9 +211,9 @@ func (r *ReconcileKnativeKafka) installKnativeKafka(instance *operatorv1alpha1.K
 	return nil
 }
 
-func (r *ReconcileKnativeKafka) applyKnativeKafka(instance *operatorv1alpha1.KnativeKafka, api client.Client) error {
+func (r *ReconcileKnativeKafka) applyKnativeKafka(instance *operatorv1alpha1.KnativeKafka) error {
 	if instance.Spec.Channel.Enabled {
-		if err := r.installKnativeKafkaChannel(instance, api); err != nil {
+		if err := r.installKnativeKafkaChannel(instance); err != nil {
 			return fmt.Errorf("unable to install Knative KafkaChannel: %w", err)
 		}
 	} else {
@@ -221,7 +221,7 @@ func (r *ReconcileKnativeKafka) applyKnativeKafka(instance *operatorv1alpha1.Kna
 	}
 
 	if instance.Spec.Source.Enabled {
-		if err := r.installKnativeKafkaSource(instance, api); err != nil {
+		if err := r.installKnativeKafkaSource(instance); err != nil {
 			return fmt.Errorf("unable to install Knative KafkaSource: %w", err)
 		}
 	} else {
@@ -231,7 +231,7 @@ func (r *ReconcileKnativeKafka) applyKnativeKafka(instance *operatorv1alpha1.Kna
 	return nil
 }
 
-func (r *ReconcileKnativeKafka) installKnativeKafkaChannel(instance *operatorv1alpha1.KnativeKafka, apiclient client.Client) error {
+func (r *ReconcileKnativeKafka) installKnativeKafkaChannel(instance *operatorv1alpha1.KnativeKafka) error {
 	manifest, err := r.kafkaChannelManifest(instance)
 	if err != nil {
 		return err
@@ -241,7 +241,7 @@ func (r *ReconcileKnativeKafka) installKnativeKafkaChannel(instance *operatorv1a
 	if err := manifest.Apply(); err != nil {
 		return fmt.Errorf("failed to apply KafkaChannel manifest: %w", err)
 	}
-	if err := checkDeployments(manifest, apiclient); err != nil {
+	if err := r.checkDeployments(manifest); err != nil {
 		return fmt.Errorf("failed to check deployments: %w", err)
 	}
 	log.Info("Knative KafkaChannel installation is ready")
@@ -265,7 +265,7 @@ func (r *ReconcileKnativeKafka) kafkaChannelManifest(instance *operatorv1alpha1.
 	return &manifest, nil
 }
 
-func (r *ReconcileKnativeKafka) installKnativeKafkaSource(instance *operatorv1alpha1.KnativeKafka, apiclient client.Client) error {
+func (r *ReconcileKnativeKafka) installKnativeKafkaSource(instance *operatorv1alpha1.KnativeKafka) error {
 	manifest, err := r.kafkaSourceManifest(instance)
 	if err != nil {
 		return err
@@ -275,7 +275,7 @@ func (r *ReconcileKnativeKafka) installKnativeKafkaSource(instance *operatorv1al
 	if err := manifest.Apply(); err != nil {
 		return fmt.Errorf("failed to apply KafkaSource manifest: %w", err)
 	}
-	if err := checkDeployments(manifest, apiclient); err != nil {
+	if err := r.checkDeployments(manifest); err != nil {
 		return fmt.Errorf("failed to check deployments: %w", err)
 	}
 	log.Info("Knative KafkaSource installation is ready")
@@ -307,11 +307,11 @@ func kafkaSourceManifestPath() string {
 // TODO: move to a common place. copied from kourier.go
 // Check for deployments
 // This function is copied from knativeserving_controller.go in serving-operator
-func checkDeployments(manifest *mf.Manifest, api client.Client) error {
+func (r *ReconcileKnativeKafka) checkDeployments(manifest *mf.Manifest) error {
 	log.Info("Checking deployments")
 	for _, u := range manifest.Filter(mf.ByKind("Deployment")).Resources() {
 		deployment := &appsv1.Deployment{}
-		err := api.Get(context.TODO(), client.ObjectKey{Namespace: u.GetNamespace(), Name: u.GetName()}, deployment)
+		err := r.client.Get(context.TODO(), client.ObjectKey{Namespace: u.GetNamespace(), Name: u.GetName()}, deployment)
 		if err != nil {
 			return err
 		}
--- a/knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
+++ b/knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
@@ -38,16 +38,36 @@ var log = logf.Log.WithName("controller_knativekafka")
 // Add creates a new KnativeKafka Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
 func Add(mgr manager.Manager) error {
-	return add(mgr, newReconciler(mgr))
+	reconciler, err := newReconciler(mgr)
+	if err != nil {
+		return err
+	}
+	return add(mgr, reconciler)
 }
 
 // newReconciler returns a new reconcile.Reconciler
-func newReconciler(mgr manager.Manager) reconcile.Reconciler {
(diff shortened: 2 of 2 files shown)


matzew:
@aliok I think this needs a rebase ;-)

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -191,22 +246,32 @@ func installKnativeKafkaChannel(instance *operatorv1alpha1.KnativeKafka, apiclie
...
-	manifest, err := mfc.NewManifest(kafkaChannelManifestPath(), apiClient, mf.UseLogger(log.WithName("mf")))
+	manifest, err := rawKafkaChannelManifest(apiClient)
 	if err != nil {
 		return nil, fmt.Errorf("failed to load KafkaChannel manifest: %w", err)
 	}
 
-	manifest, err = manifest.Transform(mf.InjectOwner(instance))
+	transformers := []mf.Transformer{
+		mf.InjectOwner(instance),
+		common.SetOwnerAnnotations(instance.ObjectMeta, common.KafkaOwnerName, common.KafkaOwnerNamespace),
```
aliok, line 237: Impl for `common.SetOwnerAnnitations` exists in a later commit (https://github.com/openshift-knative/serverless-operator/pull/496/commits/359a39b0817b8ccc09c134fa2766efa5c48b447f#diff-052673a614af4880bcfbd6c264d9f495R73). Rebasing messed up stuff.

aliok: But it is a simple function. Instead of setting the owner annotation with hardcoded keys, it now sets it using the keys passed as arguments.

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -57,7 +60,59 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+				NamespacedName: types.NamespacedName{Namespace: ownerNamespace, Name: ownerName},
+			}}
+		}
+		return nil
+	})
+
+	gvkToResource := make(map[schema.GroupVersionKind]runtime.Object)
+
+	// Watch for Knative KafkaChannel resources.
+	kafkaChannelManifest, err := rawKafkaChannelManifest(mgr.GetClient())
```
aliok, line 70: Manifest was also needed in this function. But it was needed as untransformed, so I created a new `rawKafkaChannelManifest` function that returns the raw manifest. Other places that need the transformed manifest calls another function, which eventually calls `rawKafkaChannelManifest`

aliok:
This is ready for review @matzew @markusthoemmes 
I tried it locally, works fine.

There were some refactorings in this PR, moved them to a follow up PR: https://github.com/openshift-knative/serverless-operator/pull/513

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -57,7 +60,59 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
 	}
 
-	// NOT IMPLEMENTED YET
+	err = c.Watch(&source.Kind{Type: &corev1.ConfigMap{}}, &handler.EnqueueRequestForOwner{
+		OwnerType:    &operatorv1alpha1.KnativeKafka{},
+		IsController: true,
+	})
+	if err != nil {
+		return err
+	}
```
markusthoemmes, line 69: Why is this necessary? Are we creating any configmaps specifically that wouldn't be caught by the generic loop below?

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -57,7 +60,59 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+	// Watch for Knative KafkaSource resources.
+	kafkaSourceManifest, err := rawKafkaSourceManifest(mgr.GetClient())
+	if err != nil {
+		return err
+	}
+	kafkaSourceResources := kafkaSourceManifest.Resources()
+
+	for i := range kafkaSourceResources {
+		resource := &kafkaSourceResources[i]
+		gvkToResource[resource.GroupVersionKind()] = resource
```
markusthoemmes, line 107:
```suggestion
		gvkToResource[resource.GroupVersionKind()] = &kafkaSourceResources[i]
```

Same above.

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -215,19 +280,34 @@ func installKnativeKafkaSource(apiclient client.Client) error {
...
+	manifest, err := rawKafkaSourceManifest(apiclient)
 	if err != nil {
-		return mf.Manifest{}, fmt.Errorf("failed to load KafkaSource manifest: %w", err)
+		return nil, fmt.Errorf("failed to load KafkaSource manifest: %w", err)
 	}
-	return manifest, nil
+
+	transformers := []mf.Transformer{
+		setOwnerAnnotations(instance),
+	}
```
markusthoemmes, line 303: Unnecessary intermediary slice.

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -191,22 +246,32 @@ func installKnativeKafkaChannel(instance *operatorv1alpha1.KnativeKafka, apiclie
...
+}
+
 func kafkaChannelManifest(instance *operatorv1alpha1.KnativeKafka, apiClient client.Client) (*mf.Manifest, error) {
-	manifest, err := mfc.NewManifest(kafkaChannelManifestPath(), apiClient, mf.UseLogger(log.WithName("mf")))
+	manifest, err := rawKafkaChannelManifest(apiClient)
 	if err != nil {
 		return nil, fmt.Errorf("failed to load KafkaChannel manifest: %w", err)
 	}
 
-	manifest, err = manifest.Transform(mf.InjectOwner(instance))
```
markusthoemmes, line 200: Why not keep this format and just keep adding the transformers directly? It takes a variadic argument anyway.

knative-operator/pkg/common/constants.go
```diff
@@ -6,4 +6,6 @@ const (
 	ServingOwnerNamespace            = "serving.knative.openshift.io/ownerNamespace"
 	ServerlessOperatorOwnerName      = "operator.knative.openshift.io/ownerName"
 	ServerlessOperatorOwnerNamespace = "operator.knative.openshift.io/ownerNamespace"
+	KafkaOwnerName                   = "kafka.knative.openshift.io/ownerName"
+	KafkaOwnerNamespace              = "kafka.knative.openshift.io/ownerNamespace"
```
markusthoemmes, line 10: Why this group? I'd think this should be `knativekafkas.operator.serverless.openshift.io` to play nice with the group used for the API.

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
markusthoemmes, line 87: We're reading these a few times in the process of the reconciler. Should we maybe create the manifests just once at reconciler initialization and then always use the instance on the reconciler itself to avoid all the duplicated error handling. Plus it'd be a lot quicker.

knative-operator/pkg/common/constants.go
matzew: +1

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
matzew: The config map for the kafka bootstrap server (backed by the CRD)?

aliok: Nah, that will be caught by the generic loop below.

This is a copy paste error

knative-operator/pkg/common/constants.go
aliok: Done

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
aliok: Yep

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
aliok: Removed

knative-operator/pkg/common/constants.go
```diff
@@ -6,4 +6,6 @@ const (
 	ServingOwnerNamespace            = "serving.knative.openshift.io/ownerNamespace"
 	ServerlessOperatorOwnerName      = "operator.knative.openshift.io/ownerName"
 	ServerlessOperatorOwnerNamespace = "operator.knative.openshift.io/ownerNamespace"
+	KafkaOwnerName                   = "knativekafkas.operator.serverless.openshift.io/ownerName"
+	KafkaOwnerNamespace              = "knativekafkas.operator.serverless.openshift.io/ownerNamespace"
```
markusthoemmes, line 10: Thinking about this, this is actually kinda annoying isn't it?

Maybe we should normalize all of them and add a third one that denotes the owner type so can distinguish in the respective operators.

For this specifically that could mean we do these:

```
operator.serverless.openshift.io/ownerName: "foo"
operator.serverless.openshift.io/ownerNamespace: "knative-eventing"
operator.serverless.openshift.io/ownerType: "KnativeKafka"
```

matzew: Do we move all artifacts than to `knative-eventing` ?

markusthoemmes: Nah, that was only an example to make clear what I think this could look like. The namespace would of course be dynamic to the resource.

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
aliok: @markusthoemmes  please have a look at this comment: https://github.com/openshift-knative/serverless-operator/pull/496/commits/41ff1ae8b369f62c883775c8cb51a0d9be135f2c

I did it, but kindaa turned ugly

knative-operator/pkg/common/constants.go
aliok: I can do this for KnativeKafka controller only.

For consistency, I would like to change others too, but I am worried a bit about what happens during an upgrade

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -241,27 +253,22 @@ func rawKafkaChannelManifest(apiclient client.Client) (mf.Manifest, error) {
...
 }
 
-func kafkaChannelManifest(instance *operatorv1alpha1.KnativeKafka, apiClient client.Client) (*mf.Manifest, error) {
-	manifest, err := rawKafkaChannelManifest(apiClient)
-	if err != nil {
-		return nil, fmt.Errorf("failed to load KafkaChannel manifest: %w", err)
-	}
-
-	manifest, err = manifest.Transform(
+func (r *ReconcileKnativeKafka) kafkaChannelManifest(instance *operatorv1alpha1.KnativeKafka, apiClient client.Client) (*mf.Manifest, error) {
```
markusthoemmes, line 256: The `apiClient` is now unused here (and in the other function probably too)

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -354,26 +356,26 @@ func (r *ReconcileKnativeKafka) delete(instance *operatorv1alpha1.KnativeKafka)
...
 			return fmt.Errorf("unable to delete Knative KafkaSource: %w", err)
 		}
 	}
 
 	return nil
 }
 
-func deleteKnativeKafkaChannel(instance *operatorv1alpha1.KnativeKafka, apiclient client.Client) error {
-	manifest, err := kafkaChannelManifest(instance, apiclient)
+func (r *ReconcileKnativeKafka) deleteKnativeKafkaChannel(instance *operatorv1alpha1.KnativeKafka, apiclient client.Client) error {
```
markusthoemmes, line 375: Client will be unused here too I believe.

knative-operator/pkg/common/constants.go
markusthoemmes: Yeah we have to do the changes to the others graduall (add them in one release, only start depending on them in the next one).

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
matzew: yep - I just checked the `kafkaSourceManifest`, IDEA reports unused

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -146,25 +203,25 @@ func (r *ReconcileKnativeKafka) ensureFinalizers(instance *operatorv1alpha1.Knat
 
 // Install Knative Kafka components
 func (r *ReconcileKnativeKafka) installKnativeKafka(instance *operatorv1alpha1.KnativeKafka) error {
-	if err := applyKnativeKafka(instance, r.client); err != nil {
+	if err := r.applyKnativeKafka(instance, r.client); err != nil {
```
markusthoemmes, line 206: I think now the only remaining client usage is in the checkDeployments functions. Should we just pass r.client down in `installKnativeKafkaChannel/Source` and drop the passing through from everywhere?

knative-operator/pkg/common/constants.go
aliok: Let me try doing this later ok? Not sure how this would work because we don't have the CR instance at the creation time of the watches

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
aliok: Fixed

knative-operator/pkg/common/constants.go
markusthoemmes: Yes, let's just make sure we do this before we enable this controller. I can send a patch with how I think it should look like too,  you don't have to do everything :).

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
markusthoemmes: How so? I like it!

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
aliok: Done.

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
aliok: haha ok, everybody has a different taste :D

knative-operator/pkg/common/constants.go
aliok: Cool, I will ping you just before I send a PR to enable the controller.

aliok:
Pushed the latest commit that addresses some comments. Please take another look @markusthoemmes  @matzew

markusthoemmes (approved):
/lgtm
/approve

Beauitful! :tada:

PR #502, opened 2020-10-02
Delete Kafka resources when spec.channel.enabled or spec.source.enabled is changed to false

aliok:
## Changes
With this PR, when the components are disabled in spec their resources are deleted.

I also did some refactoring to reuse stage execution.

------------

Verification:

1. Install eventing

```
```

4. Install:
```
cat <<EOS |kubectl apply -f -
---
kind: KnativeKafka
metadata:
  name: example
  namespace: knative-eventing
spec:
  source:
    enabled: true
  channel:
    enabled: true
...
EOS
```

```
```

6.Once status is ready, disable channel/source:
```
cat <<EOS |kubectl apply -f -
---
kind: KnativeKafka
metadata:
  name: example
  namespace: knative-eventing
spec:
  source:
    enabled: false
  channel:
    enabled: true
...
EOS
```

7. See the resources gone and also check the status of the CR

8. Clean up:
```
```

### DIFF
--- a/knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
+++ b/knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
@@ -151,19 +151,48 @@ func (r *ReconcileKnativeKafka) Reconcile(request reconcile.Request) (reconcile.
 func (r *ReconcileKnativeKafka) reconcileKnativeKafka(instance *operatorv1alpha1.KnativeKafka) error {
 	instance.Status.InitializeConditions()
 
-	manifest, err := r.buildManifest(instance)
+	// install the components that are enabled
+	if err := r.executeInstallStages(instance); err != nil {
+		return err
+	}
+	// delete the components that are disabled
+	if err := r.executeDeleteStages(instance); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (r *ReconcileKnativeKafka) executeInstallStages(instance *operatorv1alpha1.KnativeKafka) error {
+	manifest, err := r.buildManifest(instance, ManifestBuildEnabledOnly)
 	if err != nil {
-		return fmt.Errorf("failed to build manifest: %w", err)
+		return fmt.Errorf("failed to load and build manifest: %w", err)
 	}
 
-	// TODO: ensure components get removed when enabled=true is changed to enabled=false
-
 	stages := []func(*mf.Manifest, *operatorv1alpha1.KnativeKafka) error{
-		// TODO r.configure,
 		r.ensureFinalizers,
 		r.transform,
 		r.apply,
 		r.checkDeployments,
+	}
+
+	// Execute each stage in sequence until one returns an error
+	for _, stage := range stages {
+		if err := stage(manifest, instance); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (r *ReconcileKnativeKafka) executeDeleteStages(instance *operatorv1alpha1.KnativeKafka) error {
+	manifest, err := r.buildManifest(instance, ManifestBuildDisabledOnly)
+	if err != nil {
+		return fmt.Errorf("failed to load and build manifest: %w", err)
+	}
+
+	stages := []func(*mf.Manifest, *operatorv1alpha1.KnativeKafka) error{
+		r.transform,
+		r.deleteResources,
 	}
 
 	// Execute each stage in sequence until one returns an error
@@ -238,6 +267,17 @@ func (r *ReconcileKnativeKafka) checkDeployments(manifest *mf.Manifest, instance
 	return nil
 }
 
+// Delete Knative Kafka resources
+func (r *ReconcileKnativeKafka) deleteResources(manifest *mf.Manifest, instance *operatorv1alpha1.KnativeKafka) error {
+	log.Info("Deleting resources in manifest")
+	if err := manifest.Delete(); err != nil {
+		// TODO: any conditions?
+		return fmt.Errorf("failed to apply manifest: %w", err)
+	}
+	// TODO: any conditions?
+	return nil
+}
+
 func isDeploymentAvailable(d *appsv1.Deployment) bool {
 	for _, c := range d.Status.Conditions {
 		if c.Type == appsv1.DeploymentAvailable && c.Status == corev1.ConditionTrue {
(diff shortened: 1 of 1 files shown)


aliok:
The logic is like this:
- In case of CR delete, build the manifest with `BuildAll` option, and delete everyting in both KafkaChannel and KafkaSource manifests
- In other case (CR not deleted)
  - Step1: Collect resources to apply by building the manifest with `BuildEnabledOnly`. This will include resources for example for KafkaChannel if `spec.channel.enabled=true`. Then apply these resources.
  - Step2: Collect resources to delete by building the manifest with `BuildDisabledOnly`. This will include the resources for the disabled components. Later, we ensure that these resources don't exist.

I am open to any better ideas which provide these:
- In case of CR deletion, we should delete all resources in both manifests
- In case of `spec.channel.enabled` changed from `true` to `false`, we need to remove the KafkaChannel resources
- In case of `spec.channel.enabled` changed from `false` to `true`, we need to create the KafkaChannel resources

Also did a refactoring in the 2nd commit to reuse stage execution.

cc @markusthoemmes

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -280,29 +320,41 @@ func (r *ReconcileKnativeKafka) delete(instance *operatorv1alpha1.KnativeKafka)
...
 	return nil
 }
 
-func (r *ReconcileKnativeKafka) buildManifest(instance *operatorv1alpha1.KnativeKafka) (*mf.Manifest, error) {
+type manifestBuild int
+
+const (
+	ManifestBuildEnabledOnly manifestBuild = iota
+	ManifestBuildDisabledOnly
+	ManifestBuildAll
```
markusthoemmes, line 331: No need to export these.

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -280,29 +320,41 @@ func (r *ReconcileKnativeKafka) delete(instance *operatorv1alpha1.KnativeKafka)
...
+)
+
+func (r *ReconcileKnativeKafka) buildManifest(instance *operatorv1alpha1.KnativeKafka, build manifestBuild) (*mf.Manifest, error) {
 	var resources []unstructured.Unstructured
 
-	if instance.Spec.Channel.Enabled {
+	if build == ManifestBuildAll || (build == ManifestBuildEnabledOnly && instance.Spec.Channel.Enabled) || (build == ManifestBuildDisabledOnly && !instance.Spec.Channel.Enabled) {
```
markusthoemmes, line 337: At this point, I think we should generalize this like so:

```golang
comps := map[mf.Manifest]bool{
  r.rawKafkaChannelManifest: instance.Spec.Channel.Enabled,
  r.rawKafkaSourceManifest: instance.Spec.Source.Enabled,
}

for manifest, enabled := range comps {
  	if build == ManifestBuildAll || (build == ManifestBuildEnabledOnly && enabled) || (build == ManifestBuildDisabledOnly && !enabled) {
		resources = append(resources, manifestResources()...)
	}
}
```

aliok: yeah, why not. not a huge gain for 2 components, but sounds good.

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
markusthoemmes: Not a huge fan of having such a switch on a function btw, that's usually a big old cold-smell (just like boolean switches are).

I can't think of an easy way to not do that without introducing quite a bit of boilerplate, so maybe that's just fine.

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
aliok: hmm, can't use `mf.Manifest` as map key. It doesn't have `==` and `!=`.

can convert the map into a slice of tuples...

any alternative thoughts?

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
aliok: yeah, I don't like this too. I feel you. 

> I can't think of an easy way to not do that without introducing quite a bit of boilerplate, so maybe that's just fine.

Same here

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
markusthoemmes:
```golang
appendIfNecessary := func(enabled bool, manifest mf.Manifest) {
	if build == manifestBuildAll || (build == manifestBuildEnabledOnly && enabled) || (build == manifestBuildDisabledOnly && !enabled) {
		resources = append(resources, manifest.Resources()...)
	}
}

appendIfNecessary(instance.Spec.Channel.Enabled, r.rawKafkaChannelManifest)
appendIfNecessary(instance.Spec.Source.Enabled, r.rawKafkaSourceManifest)
```

Up to you if you like that better though. I mostly wouldn't want to repeat the big `if` condition.

aliok: IMO, 2 explicit `if`s are easier to read than this. if we have a 3rd component  in the future we can convert it.

thanks for the reviews. if we're good, now let's wait until the other PR is merged.

aliok:
@markusthoemmes @matzew 
This is rebased now and ready for your `lgtm`

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -280,29 +310,35 @@ func (r *ReconcileKnativeKafka) delete(instance *operatorv1alpha1.KnativeKafka)
...
+func (r *ReconcileKnativeKafka) buildManifest(instance *operatorv1alpha1.KnativeKafka, build manifestBuild) (*mf.Manifest, error) {
 	var resources []unstructured.Unstructured
 
-	if instance.Spec.Channel.Enabled {
+	if build == manifestBuildAll || (build == manifestBuildEnabledOnly && instance.Spec.Channel.Enabled) || (build == manifestBuildDisabledOnly && !instance.Spec.Channel.Enabled) {
 		resources = append(resources, r.rawKafkaChannelManifest.Resources()...)
 	}
 
-	if instance.Spec.Source.Enabled {
+	if build == manifestBuildAll || (build == manifestBuildEnabledOnly && instance.Spec.Source.Enabled) || (build == manifestBuildDisabledOnly && !instance.Spec.Source.Enabled) {
```
matzew, line 341: that reads like WOW :-)   and is very similar to the above... :grimacing:

matzew (approved):
/lgtm

markusthoemmes (approved):
/lgtm
/approve

PR #503, opened 2020-10-02
Actually set the bootstrapServers in config-kafka configmap from spec

aliok:
## Changes

With this PR, `bootstrapServers` field in `config-kafka` configmap is set with the value from spec.
------------------

Verification:

1. Install eventing

```
```

4. Install:
```
cat <<EOS |kubectl apply -f -
---
kind: KnativeKafka
metadata:
  name: example
  namespace: knative-eventing
spec:
  source:
    enabled: true
  channel:
    enabled: true
...
EOS
```

```
```

6.Once status is ready, check configmap `config-kafka` in `knative-eventing` namespace. The `bootstrapServers` field should be set accordingly.

7. Clean up:
```
```

### DIFF
--- a/knative-operator/pkg/controller/knativekafka/knativekafka_controller_test.go
+++ b/knative-operator/pkg/controller/knativekafka/knativekafka_controller_test.go
@@ -376,3 +376,125 @@ func TestInjectOwner(t *testing.T) {
 		})
 	}
 }
+
+func TestSetBootstrapServers(t *testing.T) {
+	logf.SetLogger(logf.ZapLogger(true))
+
+	tests := []struct {
+		name             string
+		obj              *unstructured.Unstructured
+		bootstrapServers string
+		expect           *unstructured.Unstructured
+	}{
+		{
+			name: "Update config-kafka",
+			obj: &unstructured.Unstructured{
+				Object: map[string]interface{}{
+					"apiVersion": "v1",
+					"kind":       "ConfigMap",
+					"metadata": map[string]interface{}{
+						"name": "config-kafka",
+					},
+				},
+			},
+			bootstrapServers: "example.com:1234",
+			expect: &unstructured.Unstructured{
+				Object: map[string]interface{}{
+					"apiVersion": "v1",
+					"kind":       "ConfigMap",
+					"metadata": map[string]interface{}{
+						"name": "config-kafka",
+					},
+					"data": map[string]interface{}{
+						"bootstrapServers": "example.com:1234",
+					},
+				},
+			},
+		},
+		{
+			name: "Update config-kafka - overwrite",
+			obj: &unstructured.Unstructured{
+				Object: map[string]interface{}{
+					"apiVersion": "v1",
+					"kind":       "ConfigMap",
+					"metadata": map[string]interface{}{
+						"name": "config-kafka",
+					},
+					"data": map[string]interface{}{
+						"bootstrapServers": "TO_BE_OVERWRITTEN",
+					},
+				},
+			},
+			bootstrapServers: "example.com:1234",
+			expect: &unstructured.Unstructured{
+				Object: map[string]interface{}{
+					"apiVersion": "v1",
+					"kind":       "ConfigMap",
+					"metadata": map[string]interface{}{
+						"name": "config-kafka",
+					},
+					"data": map[string]interface{}{
+						"bootstrapServers": "example.com:1234",
+					},
+				},
+			},
+		},
+		{
+			name: "Do not update other configmaps",
+			obj: &unstructured.Unstructured{
+				Object: map[string]interface{}{
+					"apiVersion": "v1",
+					"kind":       "ConfigMap",
+					"metadata": map[string]interface{}{
+						"name": "config-foo",
+					},
+				},
+			},
+			bootstrapServers: "example.com:1234",
+			expect: &unstructured.Unstructured{
+				Object: map[string]interface{}{
+					"apiVersion": "v1",
+					"kind":       "ConfigMap",
+					"metadata": map[string]interface{}{
+						"name": "config-foo",
+					},
+				},
+			},
+		},
+		{
+			name: "Do not update other resources",
+			obj: &unstructured.Unstructured{
+				Object: map[string]interface{}{
+					"apiVersion": "v1",
+					"kind":       "Secret",
+					"metadata": map[string]interface{}{
+						"name": "config-kafka",
+					},
+				},
+			},
+			bootstrapServers: "example.com:1234",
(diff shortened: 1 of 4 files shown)


aliok:
cc @matzew @markusthoemmes

markusthoemmes (commented):
Can we get tests on the behavior of the main reconciler please? We could also test individual methods, like `transform` in this case.

I'd propose a test that at least tests the `transform` function and makes sure stuff gets transformed properly. Ultimately, we'll need tests for the other behavior too (enabled/disabled etc.)

aliok:
Added unit tests for:
- Reconciler flow
- Custom transformations (`injectOwner`, `setBootstrapServers`)

Didn't add unit test for `transform` method as you asked because it basically consists of well-tested units.
`transform` basically consists of sequential execution of these units and I don't see much value in testing their sequential execution.

@markusthoemmes

knative-operator/pkg/controller/knativekafka/knativekafka_controller_test.go
```diff
@@ -0,0 +1,500 @@
...
+	return nil
+}
+
+func (*fakeManager) GetRecorder(name string) record.EventRecorder {
+	return nil
+}
+
+func (*fakeManager) GetRESTMapper() meta.RESTMapper {
+	return nil
+}
```
aliok, line 84: I just pushed `fakeManager` within this file. If you do not know an existing fake manager like this one, I will extract this into a separate file. If you do know, please let me know and I will get rid of this.

markusthoemmes: I would refactor the `newReconciler` function to take a scheme and a client rather than the manager. That obviates this completely, I think.

knative-operator/pkg/controller/knativekafka/knativekafka_controller_test.go
```diff
@@ -0,0 +1,500 @@
...
+func TestKnativeKafkaReconcile(t *testing.T) {
+	logf.SetLogger(logf.ZapLogger(true))
+
+	tests := []struct {
+		name         string
+		instance     *v1alpha1.KnativeKafka
+		exists       []types.NamespacedName
+		doesNotExist []types.NamespacedName
+	}{
+		{
```
markusthoemmes, line 103: nit: You can collapse the tests a bit to get rid of a level of indentation, you can do:

```golang
tests := []struct{
  ...
}{{
  // your first test
}, {
  // your second test
}}
```

nothing blocking obviously, I personally prefer writing (and more importantly reading) that way though. Can be applied to all tests.

knative-operator/pkg/controller/knativekafka/knativekafka_controller_test.go
```diff
@@ -0,0 +1,500 @@
...
+				err = cl.Get(context.TODO(), d, deployment)
+				if err != nil {
+					t.Fatalf("get: (%v)", err)
+				}
+			}
+		})
+	}
+}
+
+func makeCr(channel bool, source bool, deleted bool) *v1alpha1.KnativeKafka {
```
markusthoemmes, line 220: nit: Instead of using booleans here, function options might be a great way to make the tests a bit more expressive.

```golang
func kk(mods ...func(*v1alpha1.KnativeKafka)) *v1alpha1.KnativeKafka
  base := // create a minimal CR
  for _, mod := range mods {
    mod(base)
  }
  return base
}

func withSourceEnabled(kk *v1alpha1.KnativeKafka) {
  kk.Spec.Source.Enabled = true
}

// etc. withChannelEnabled, withDeletionTimestamp

// Usage then goes like
kk()
kk(withChannelEnabled)
kk(withDeletionTimestamp)
kk(withChannelEnabled, withSourceEnabled)

**Review by markusthoemmes — COMMENTED** [review]
Thanks a MILLION! Very nice tests :tada:.

Left a few suggestions to cut down on the boilerplate (as you asked about that) and how to get rid of the amount of boolean switches.

File: knative-operator/pkg/controller/knativekafka/knativekafka_controller_test.go (hunk 1 above)
Comment (aliok) [reply knative-operator/pkg/controller/knativekafka/knativekafka_controller_test.go:84]: Yeah, that's an option, I actually tried it.
But I ended up loading the manifest more than once from the file, so I gave up.

I am open to suggestions for that.

File: knative-operator/pkg/controller/knativekafka/knativekafka_controller_test.go (hunk 3 above)
Comment (aliok) [reply knative-operator/pkg/controller/knativekafka/knativekafka_controller_test.go:220]: I think your suggestion is good, but I don't like the fact that it is not obvious as the boolean arguments.
```
kk() // --> what does this mean? all disabled? all enabled? superstate?
kk(withChannelEnabled, withSourceEnabled) // --> easy to understand, both are enabled
kk(withChannelEnabled) // --> what does this mean? source is enabled or disabled?
```

Comment (markusthoemmes) [reply knative-operator/pkg/controller/knativekafka/knativekafka_controller_test.go:220]: Well you have to define the default state, which is all disabled (because the structs default to disabled).

You can of course also provide the respective `withChannelDisabled` etc. for clarity.

Comment (markusthoemmes) [reply knative-operator/pkg/controller/knativekafka/knativekafka_controller_test.go:220]: But I marked it as a nit, so feel free to ignore :)

File: knative-operator/pkg/controller/knativekafka/knativekafka_controller_test.go (hunk 1 above)
Comment (aliok) [reply knative-operator/pkg/controller/knativekafka/knativekafka_controller_test.go]: @markusthoemmes 
Got rid of `fakeManager` now by loading the manifests within the unit test.

Mind having another look?

File: knative-operator/pkg/controller/knativekafka/knativekafka_controller_test.go [hunk 4]
```diff
@@ -152,11 +94,23 @@ func TestKnativeKafkaReconcile(t *testing.T) {
 			initObjs := []runtime.Object{test.instance}
 
 			cl := fake.NewFakeClient(initObjs...)
-			mgr := &fakeManager{
-				client: cl,
-				scheme: s,
+
+			kafkaChannelManifest, err := mf.ManifestFrom(mf.Path(os.Getenv("KAFKACHANNEL_MANIFEST_PATH")))
```
markusthoemmes, line 98: With this in place, you can also drop the os.Setenv commands on top for good measure, I believe.

aliok: :facepalm: 
that's copy paste w/o thinking :facepalm:

markusthoemmes (approved):
/lgtm
/approve

Thanks for adding the tests again!

PR #510, opened 2020-10-05
Validation for the KnativeKafka CRs

aliok:
With this PR, KnativeKafka spec is validated.

Validations:
- Check if `spec.channel.bootstrapServers` is not empty, when `spec.channel.enabled` is `true`. The requiredness check for `spec.channel`, `spec.source`, `spec.channel.enabled` and `spec.source.enabled` is done by Kubernetes via the OpenAPI validaton rules defined in the CRD.
- CR namespace must be in the predefined one
- There must be only 1 CRs

Verification instructions:

1. Enable the controller by uncommenting the stuff in `knative-operator/pkg/controller/add_knativekafka.go`

2. `make images` && `make install` --> deploys S-O and installs KnativeServing and KnativeEventing

```
```

4. Install KnativeKafka:
```
cat <<EOS |kubectl apply -f -
---
kind: KnativeKafka
metadata:
  name: example
  namespace: knative-eventing
spec:
  source:
    enabled: true
  channel:
    enabled: true
...
EOS
```

```
```

7. Clean up:
```
```

Try with different combinations of the spec:
- Empty spec
- No `source` field in spec, no `source.enabled` in spec, etc.
- No `channel` field in spec, ..., 
- No `channel.bootstrapServers` with `channel.enabled=true`
...

Try loneliness (create 2 instances of the CR)

Try creating in another namespace than `knative-eventing`

### DIFF
--- a/knative-operator/pkg/webhook/knativekafka/webhook_validating.go
+++ b/knative-operator/pkg/webhook/knativekafka/webhook_validating.go
@@ -1,0 +1,139 @@
+package knativekafka
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"os"
+
+	operatorv1alpha1 "github.com/openshift-knative/serverless-operator/knative-operator/pkg/apis/operator/v1alpha1"
+	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/common"
+	webhookutil "github.com/openshift-knative/serverless-operator/knative-operator/pkg/webhook/util"
+	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission/builder"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission/types"
+)
+
+// Creates a new validating KnativeKafka Webhook
+func ValidatingWebhook(mgr manager.Manager) (webhook.Webhook, error) {
+	common.Log.Info("Setting up validating webhook for KnativeKafka")
+	return builder.NewWebhookBuilder().
+		Name("validating.knativekafka.openshift.io").
+		Validating().
+		Operations(admissionregistrationv1beta1.Create, admissionregistrationv1beta1.Update).
+		WithManager(mgr).
+		ForType(&operatorv1alpha1.KnativeKafka{}).
+		Handlers(&KnativeKafkaValidator{}).
+		Build()
+}
+
+// KnativeKafkaValidator validates KnativeKafka CR's
+type KnativeKafkaValidator struct {
+	client  client.Client
+	decoder types.Decoder
+}
+
+// Implement admission.Handler so the controller can handle admission request.
+var _ admission.Handler = (*KnativeKafkaValidator)(nil)
+
+// What makes us a webhook
+func (v *KnativeKafkaValidator) Handle(ctx context.Context, req types.Request) types.Response {
+	ke := &operatorv1alpha1.KnativeKafka{}
+
+	err := v.decoder.Decode(req, ke)
+	if err != nil {
+		return admission.ErrorResponse(http.StatusBadRequest, err)
+	}
+
+	allowed, reason, err := v.validate(ctx, ke)
+	if err != nil {
+		return admission.ErrorResponse(http.StatusInternalServerError, err)
+	}
+	return admission.ValidationResponse(allowed, reason)
+}
+
+// KnativeKafkaValidator checks for a minimum OpenShift version
+func (v *KnativeKafkaValidator) validate(ctx context.Context, ke *operatorv1alpha1.KnativeKafka) (allowed bool, reason string, err error) {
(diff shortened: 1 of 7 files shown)


matzew:
Is it possible to break this down into a few smaller PRs ?

aliok:
/hold

It should work fine as it was working before the rebase, but I want to retest anyway

knative-operator/pkg/webhook/knativekafka/webhook_validating_test.go
```diff
@@ -0,0 +1,120 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission/types"
+)
+
+func init() {
+	configv1.AddToScheme(scheme.Scheme)
+	operatorv1alpha1.SchemeBuilder.AddToScheme(scheme.Scheme)
+}
+
+var ke1 = &operatorv1alpha1.KnativeKafka{
```
markusthoemmes, line 24: These three KK's all look the same. Can we deduplicate? Maybe we can generalize the helpers from the other unit tests in a `testing` package in API and reuse here? 🤔

knative-operator/pkg/webhook/knativekafka/webhook_validating.go
```diff
@@ -0,0 +1,139 @@
...
+var _ inject.Decoder = (*KnativeKafkaValidator)(nil)
+
+// InjectDecoder injects the decoder.
+func (v *KnativeKafkaValidator) InjectDecoder(d types.Decoder) error {
+	v.decoder = d
+	return nil
+}
+
+// validate minimum openshift version
+func (v *KnativeKafkaValidator) validateVersion(ctx context.Context, _ *operatorv1alpha1.KnativeKafka) (bool, string, error) {
```
markusthoemmes, line 106: Do we even need to do this in this webhook? Seems unnecessary

aliok: Do you mean that because `KnativeEventing` is a prerequisite and we already check the OpenShift version there?

knative-operator/pkg/webhook/knativekafka/webhook_validating_test.go
aliok: yeah, sounds good. I will do this

knative-operator/pkg/webhook/knativekafka/webhook_validating.go
markusthoemmes: Meh, seems like we'd do this in a different place anyway. I'd actually expect Serverless to never become ready and I think we should revisit this check at all since OLM now has mechanisms for this too.

Do you mind filing a JIRA to revisit the version check in webhooks and drop it from this specific webhook?

aliok: Sounds good, I will do it

knative-operator/pkg/webhook/knativekafka/webhook_validating_test.go
aliok: @markusthoemmes Do you mind if we leave these as is?

The function options approach doesn't scale well when things to mod are too many.

markusthoemmes: Can we at least name them appropriately here then? I'm fine leaving as-is and potentially refactor later if necessary assuming that the names make it a bit more identifiable what each of them is supposed to represent.

I disagree on the scalability aspect as you can always pass a custom modifier function if you want to change multiple fields at once (upstream does that a lot). Not necessary to block the PR for that though, the naming would be enough imo.

knative-operator/pkg/webhook/knativekafka/webhook_validating.go
aliok: Created https://issues.redhat.com/browse/SRVCOM-1071

knative-operator/pkg/webhook/knativekafka/webhook_validating_test.go
aliok: Ok, let's do that later. 

Renamed the CRs now. Actually added a "happy" test case too

markusthoemmes (approved):
/lgtm
/approve

aliok:
/retest 

Getting this:
```
    --- FAIL: TestKnativeEventing/create_subscription_and_wait_for_CSV_to_succeed (330.32s)
... Error from server (NotFound): namespaces "knative-serving-ingress" not found

```

markusthoemmes:
Retests won't help.

```
{"level":"error","ts":"2020-10-12T08:38:04.051Z","logger":"webhook","msg":"Unable to setup webhook","error":"no matches for kind \"KnativeKafka\" in version \"operator.serverless.openshift.io/v1alpha1\"","stacktrace":"github.com/openshift-knative/serverless-operator/knative-operator/vendor/github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/src/github.com/openshift-knative/serverless-operator/knative-operator/vendor/github.com/go-logr/zapr/zapr.go:128\ngithub.com/openshift-knative/serverless-operator/knative-operator/pkg/webhook.AddToManager\n\t/go/src/github.com/openshift-knative/serverless-operator/knative-operator/pkg/webhook/webhook.go:34\nmain.main\n\t/go/src/github.com/openshift-knative/serverless-operator/knative-operator/cmd/manager/main.go:117\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:203"}
```

I think we're unwillingly activating the webhook already. Maybe comment out the respective file?

aliok:
@markusthoemmes Thanks.

markusthoemmes (approved):
/lgtm
/approve

PR #513, opened 2020-10-06
Refactor secondary resource watching related code

aliok:
Bases https://github.com/openshift-knative/serverless-operator/pull/496

Does not change any logic, simply does some refactoring. 

Created reusable functions for following:
- injecting owner annotations
- creating reconciliation enqueue request for watched resources
- finding the resource types to watch

EACH in different commits

### DIFF
--- a/knative-operator/pkg/common/util.go
+++ b/knative-operator/pkg/common/util.go
@@ -6,6 +6,9 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	mf "github.com/manifestival/manifestival"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -103,4 +106,19 @@ func EnsureContainerMemoryLimit(s *operatorv1alpha1.CommonSpec, containerName st
 		},
 	})
 	return
+}
+
+// common function to enqueue reconcile requests for resources
+func EnqueueRequestByOwnerAnnotations(ownerNameAnnotationKey, ownerNamespaceAnnotationKey string) handler.ToRequestsFunc {
+	return func(obj handler.MapObject) []reconcile.Request {
+		annotations := obj.Meta.GetAnnotations()
+		ownerNamespace := annotations[ownerNamespaceAnnotationKey]
+		ownerName := annotations[ownerNameAnnotationKey]
+		if ownerNamespace != "" && ownerName != "" {
+			return []reconcile.Request{{
+				NamespacedName: types.NamespacedName{Namespace: ownerNamespace, Name: ownerName},
+			}}
+		}
+		return nil
+	}
 }
--- a/knative-operator/pkg/common/util.go
+++ b/knative-operator/pkg/common/util.go
@@ -6,6 +6,8 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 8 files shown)


aliok:
@markusthoemmes @matzew please check each commit individually.

None of the code changes actually change any logic. They're just moving common stuff into reusable functions and eliminating unnecessary loops.

knative-operator/pkg/common/util.go
```diff
@@ -69,11 +75,11 @@ func BuildImageOverrideMapFromEnviron(environ []string) map[string]string {
 }
 
 // SetOwnerAnnotations is a transformer to set owner annotations on given object
-func SetOwnerAnnotations(instance *operatorv1alpha1.KnativeServing) mf.Transformer {
+func SetOwnerAnnotations(meta metav1.ObjectMeta, ownerNameAnnotationKey, ownerNamespaceAnnotationKey string) mf.Transformer {
```
markusthoemmes, line 78: I see why you did the ObjectMeta thing here, but I think it'd be somewhat clearer if we just passed name, namespace as a string here.

markusthoemmes (commented):
One nit, I like this a lot!

knative-operator/pkg/common/util.go
aliok: sgtm.

let me update!

markusthoemmes: Actually thinking about it: We might as well pass a `map[string]string` in to make the mapping VERY clear :joy:. We don't win anything by passing it separately anyway.

markusthoemmes: As such, we might as well rename the function to `SetAnnotations`

aliok: Pushed another commit now that does that.
However, I changed the logic slightly:
- The old function existed in the repo since beginning was clearing the existing annotations and setting our own stuff
- New function keeps the existing ones, adds new ones and if there's already with the same key, it overwrites that

aliok:
ok, this should be good to go now?

knative-operator/pkg/common/util.go
```diff
@@ -103,3 +115,34 @@ func EnsureContainerMemoryLimit(s *operatorv1alpha1.CommonSpec, containerName st
 	})
 	return
 }
+
+// common function to enqueue reconcile requests for resources
+func EnqueueRequestByOwnerAnnotations(ownerNameAnnotationKey, ownerNamespaceAnnotationKey string) handler.ToRequestsFunc {
```
markusthoemmes, line 120: If we're refactoring anyway: Maybe this could return a `handler.EventHandler` and do the wrap into the `EnqueueRequestsFromMapFunc` directly to allow us to write:

```golang
c.Watch(&source.Kind{Type: t}, common.EnqueueRequestByOwnerAnnotations(common.ServingOwnerNamespace, common.ServingOwnerName))
```

To very closely align to osdk's syntax?

aliok: @markusthoemmes check out the latest commit. 

I didn't like the function in function in function, but my dislike is not too strong to revert the change :) I am ok with this change.

knative-operator/pkg/common/util.go
```diff
@@ -103,3 +115,37 @@ func EnsureContainerMemoryLimit(s *operatorv1alpha1.CommonSpec, containerName st
 	})
 	return
 }
+
+// common function to enqueue reconcile requests for resources
+func EnqueueRequestByOwnerAnnotations(ownerNameAnnotationKey, ownerNamespaceAnnotationKey string) handler.EventHandler {
+	enqueueRequests := func() handler.ToRequestsFunc {
+		return func(obj handler.MapObject) []reconcile.Request {
```
markusthoemmes, line 122: The intermediary function seems unnecessary.

Just

```golang
enqueueRequests := func(obj handler.MapObject) []reconcile.Request { ...}
return &handler.EnqueueRequestsFromMapFunc{ToRequests: enqueueRequests}
```

Should work.

markusthoemmes (commented):
Love it!

knative-operator/pkg/common/util.go
aliok: Couldn't make that work. Some golang magic is required.

Let's merge this one and can you send a small PR for that?

markusthoemmes: ack!

markusthoemmes (approved):
/lgtm
/approve

PR #516, opened 2020-10-07
[SRVCOM-1060] Seed the new vendored version of the upstream operator

markusthoemmes:
The seeds our way of pulling in the upstream operator and eventually extend it via the Extension mechanism. For now, this is onlt a straight up copy of the upstream operator with no extensions yet.

Not hooked up anywhere and just here so I can start getting the image built and to create a foundation.


aliok:
/lgtm

Happy to see you starting with this so soon!

PR #521, opened 2020-10-08
Add eventing smoke test in e2e, add eventing dep

lberk:
Please use discretion with my go mod changes to make sure we don't trip over a release boundary unintended.  dep no longer worked with one of eventing's dependencies (which had already switched to go mod) and the upstream bug for dep isn't going to be fixed.

Add three different tests to eventinge2e where we test
source -> ksvc
source -> channel -> subscription -> ksvc
source -> broker -> trigger -> ksvc

@matzew @aliok


matzew:
Is it possible to split this in two PRs: ? 

* go mod
* e2e tests

hack/lib/catalogsource.bash
```diff
@@ -129,7 +129,7 @@ function add_user {
   kubectl create secret generic htpass-secret \
     --from-file=htpasswd="$(pwd)/users.htpasswd" \
     -n openshift-config \
-    --dry-run -o yaml | kubectl apply -f -
+    --dry-run=client -o yaml | kubectl apply -f -
```
markusthoemmes, line 132: Can we back out those changes into their own PR please? I've done them once but we had issues with different `oc` versions for peeps.

Putting them in their own PR at least makes them easily revertable.

test/eventinge2e/source_broker_ksvc_test.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		t.Fatal("Knative PingSource not created: %+V", err)
+	}
+	waitForRouteServingText(t, client, ksvc.Status.URL.URL(), helloWorldText)
+
+	// Delete resources
+	client.Clients.Eventing.EventingV1beta1().Brokers(testNamespace).Delete(broker.Name, &metav1.DeleteOptions{})
+	client.Clients.Eventing.EventingV1beta1().Triggers(testNamespace).Delete(trigger.Name, &metav1.DeleteOptions{})
+	client.Clients.Eventing.SourcesV1alpha2().PingSources(testNamespace).Delete(ps.Name, &metav1.DeleteOptions{})
+	client.Clients.Kube.CoreV1().ConfigMaps(testNamespace).Delete(configMap.Name, &metav1.DeleteOptions{})
```
markusthoemmes, line 117: Schedule a cleanup after every create call above to make sure stuff gets cleaned up even if the test is failed/aborted.

Or, as you know all the names beforehand anyway and we don't care about errors: Do this at the beginning of the test:

```golang
cleanup := func() {
	client.Clients.Eventing.EventingV1beta1().Brokers(testNamespace).Delete(broker.Name, &metav1.DeleteOptions{})
	client.Clients.Eventing.EventingV1beta1().Triggers(testNamespace).Delete(trigger.Name, &metav1.DeleteOptions{})
	client.Clients.Eventing.SourcesV1alpha2().PingSources(testNamespace).Delete(ps.Name, &metav1.DeleteOptions{})
	client.Clients.Kube.CoreV1().ConfigMaps(testNamespace).Delete(configMap.Name, &metav1.DeleteOptions{})
}
test.CleanupOnInterrupt(t, cleanup)
defer cleanup()
```

lberk:
@matzew I've cherry-picked the go.mod changes to their own PR: #523

hack/lib/catalogsource.bash
lberk: thanks for catching that -- backed out this change in 23233e0 and will create a separate PR

test/eventinge2e/source_broker_ksvc_test.go
lberk: done in `b4623e5` thanks for the suggestion

matzew:
/lgtm

thanks for adding these three smoke tests, @lberk

PR #522, opened 2020-10-08
[SRVKS-596] Propagate OpenShift Route labels from Knative Ingress's

nak3:
This patch changes:
- to propagate OpenShift Route labels from Knative Ingress
- to trigger reconcile if OpenShift Route's labels or annotations are different from desired.

/cc @markusthoemmes @bbrowning

### DIFF
--- a/serving/ingress/pkg/controller/ingress/resources/route.go
+++ b/serving/ingress/pkg/controller/ingress/resources/route.go
@@ -13,7 +13,6 @@ import (
 	"knative.dev/pkg/ptr"
 	"knative.dev/serving/pkg/apis/networking"
 	networkingv1alpha1 "knative.dev/serving/pkg/apis/networking/v1alpha1"
-	"knative.dev/serving/pkg/apis/serving"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


serving/ingress/pkg/controller/ingress/resources/route.go
```diff
@@ -91,12 +91,9 @@ func makeRoute(ci *networkingv1alpha1.Ingress, host string, rule networkingv1alp
 		}
 	}
 
-	labels := make(map[string]string)
-	labels[networking.IngressLabelKey] = ci.GetName()
-
-	ingressLabels := ci.GetLabels()
-	labels[serving.RouteLabelKey] = ingressLabels[serving.RouteLabelKey]
```
nak3, line 98: Just a note about this deletion, Ingress has the label https://github.com/knative/serving/blob/b1a021dfcf522d5c7e2d7513501696a07e7a25d0/pkg/reconciler/route/resources/ingress.go#L72-L75
so we don't need to copy it one by one.

markusthoemmes (commented):
Want to add an additional test maybe?

markusthoemmes:
Memo to self: Land the knative/pkg rewrite finally :joy:

markusthoemmes (approved):
/lgtm
/approve

PR #528, opened 2020-10-08
Change test go dep to go mod

lberk:
As requested in #521 -- making a separate PR for just the go.mod changes


test/go.mod
```diff
@@ -0,0 +1,100 @@
+module github.com/openshift-knative/serverless-operator/test
+
+go 1.14
+
+require (
```
markusthoemmes, line 5: Please run `go mod tidy` on this. There's way too much stuff in here.

test/go.mod
```diff
@@ -0,0 +1,100 @@
...
+	knative.dev/test-infra v0.0.0-20200519161858-554a95a37986
+	sigs.k8s.io/yaml v1.2.0 // indirect
+)
+
+replace (
+	github.com/openshift/api => github.com/openshift/api v0.0.0-20200527184302-a843dc3262a0
+	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20191125132246-f6563a70e19a
+
+	// Pin to kube 1.16
+	k8s.io/api => k8s.io/api v0.0.0-20190918155943-95b840bb6a1f
```
markusthoemmes, line 76: Can we use actual tags here, that are readable? Do we need to pin all of those?

lberk: I couldn't get it working without the pins (which came from olm itself)

https://github.com/operator-framework/operator-lifecycle-manager/blob/release-4.3/go.mod#L54

lberk: keep in mind -- it's not serving that made this a royal pain, it was adding eventing that brought in a dependency that wasn't compatible with go mod

test/go.mod
lberk: So, just to make sure I understand what you're asking for (and the implications).

You'd like a reduced subset of deps in go.{mod,sum} via running `go mod tidy`

However, that'll remove all the eventing bits from #521, which is fine, I'll update the title of *this* PR accordingly, but it'll make it less of a cherry pick and I'll have to rebase 521.

I understand this makes the individual PRs cleaner, just want to make sure we're on the same page and you'd still like that done here (vs in #521 where I can run the same `go mod tidy`)

either way, let me know what you'd prefer

markusthoemmes: Just seen this comment now. Let's invert the order then. Let's merge this one as-is (if CI is happy) and then the tests.

lberk: was asked to do this change here -- I'll rebased #521 as needed.

lberk:
@markusthoemmes changes applied ask, please take a look

test/go.mod
lberk: @markusthoemmes just saw our comments collided -- could we please get this change in first? I've already started the work to rebase on this PR, as you suggested earlier/elsewhere, it's cleaner to make these changes here, then the eventing PR on top

markusthoemmes (approved):
/lgtm
/approve

PR #529, opened 2020-10-08
Implement static parts of the serving and eventing extensions

markusthoemmes:
This has gotten somewhat big but the business logic itself should be fairly small in surface. The tests are fairly large. Please let me know if this should be cut smaller, I didn't want to get too spammy with PRs either.

### DIFF
--- a/openshift-knative-operator/pkg/common/images.go
+++ b/openshift-knative-operator/pkg/common/images.go
@@ -1,0 +1,34 @@
+package common
+
+import (
+	"strings"
+)
+
+const imagePrefix = "IMAGE_"
+
+// ImageMapFromEnvironment generates a map of deployment/container to image from
+// the passed environment variables.
+func ImageMapFromEnvironment(env []string) map[string]string {
+	overrideMap := map[string]string{}
+
+	for _, e := range env {
+		pair := strings.SplitN(e, "=", 2)
+		if strings.HasPrefix(pair[0], imagePrefix) {
+			if pair[1] == "" {
+				continue
+			}
+
+			/*
+				converts:
+				IMAGE_container=foo             -> container: foo
+				IMAGE_deployment__container=foo -> deployment/container: foo
+				IMAGE_env_var=foo               -> env_var: foo
+				IMAGE_deployment__env_var=foo   -> deployment/env_var: foo
+			*/
+			name := strings.TrimPrefix(pair[0], imagePrefix)
+			name = strings.Replace(name, "__", "/", 1)
+			overrideMap[name] = pair[1]
+		}
+	}
+	return overrideMap
+}
--- a/openshift-knative-operator/pkg/eventing/extension.go
+++ b/openshift-knative-operator/pkg/eventing/extension.go
@@ -2,12 +2,16 @@ package eventing
 
 import (
 	"context"
+	"os"
 
 	mf "github.com/manifestival/manifestival"
+	"github.com/openshift-knative/serverless-operator/openshift-knative-operator/pkg/common"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"knative.dev/operator/pkg/apis/operator/v1alpha1"
 	operator "knative.dev/operator/pkg/reconciler/common"
 )
 
+// NewExtension creates a new extension for a Knative Eventing controller.
 func NewExtension(ctx context.Context) operator.Extension {
 	return &extension{}
 }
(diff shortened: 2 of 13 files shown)


openshift-knative-operator/pkg/eventing/extension.go
```diff
@@ -0,0 +1,41 @@
...
+	return nil
+}
+
+func (e *extension) Reconcile(ctx context.Context, comp v1alpha1.KComponent) error {
+	ke := comp.(*v1alpha1.KnativeEventing)
+
+	// Override images.
+	images := common.ImageMapFromEnvironment(os.Environ())
+	ke.Spec.Registry.Override = images
+	ke.Spec.Registry.Default = images["default"]
```
jcrossley3, line 31: This might violate "The Principle of Least Surprise". If we want to make the deliberate decision to ignore any values the user might put in their CR, for whatever reason, I don't think we should do it silently.

My personal preference would be to let their spec fields, if present, win, but at the very least I think we should toss an error if they've set any field[s] we won't honor.

Otherwise, SURPRISE!

openshift-knative-operator/pkg/serving/extension.go
```diff
@@ -0,0 +1,64 @@
...
+	return nil
+}
+
+func (e *extension) Reconcile(ctx context.Context, comp v1alpha1.KComponent) error {
+	ks := comp.(*v1alpha1.KnativeServing)
+
+	// Override images.
+	images := common.ImageMapFromEnvironment(os.Environ())
+	ks.Spec.Registry.Override = common.ImageMapFromEnvironment(os.Environ())
+	ks.Spec.Registry.Default = images["default"]
```
jcrossley3, line 31: ibid

jcrossley3 (changes requested):
Much cleaner, thanks!

openshift-knative-operator/pkg/eventing/extension.go
markusthoemmes: I agree in principle, though I was trying to preserve behavior that we are exhibiting currently. We're currently hard-overriding the `Override` section and are "override-if-present" the default and queue-proxy section.

I thought it'd be clearer if we just override the entire registry as those small exceptions didn't seem worth it at all.

I'd rather revisit our behavior separately once the new logic is engaged to keep the amount of moving parts small. This allows us to replace the components while not expecting any significant behavioral changes if that makes sense?

openshift-knative-operator/pkg/common/images.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"strings"
+)
+
+const imagePrefix = "IMAGE_"
+
+// ImageMapFromEnvironment generates a map of deployment/container to image from
+// the passed environment variables.
+func ImageMapFromEnvironment(env []string) map[string]string {
```
matzew, line 11: wasn't there some "logic" like this before, somewhere?

markusthoemmes: yeah, in the OSDK based operator and in the bandaid PRs I sent towards the operator forks. Mostly 1:1 copies with little touchups here and there.

openshift-knative-operator/pkg/eventing/extension.go
markusthoemmes: For the registry specifically we don't really have a mechanism not to override, since on upgrade this will contain outdated images and **has** to be overridden.

openshift-knative-operator/pkg/eventing/extension.go
```diff
@@ -0,0 +1,41 @@
...
+
+	// Ensure webhook has 1G of memory.
+	common.EnsureContainerMemoryLimit(&ke.Spec.CommonSpec, "eventing-webhook", resource.MustParse("1024Mi"))
```
matzew, line 34: I like it /cc @aliok

openshift-knative-operator/pkg/eventing/extension.go
matzew: yeah, good point :-)

matzew:
nice

openshift-knative-operator/pkg/common/images.go
matzew: that's cool with me, @markusthoemmes

matzew:
LGTM

openshift-knative-operator/pkg/eventing/extension.go
markusthoemmes: Added TODOs and a JIRA to capture the very valid sentiment here. I believe we should (and can!) revisit after we're through with the transition and can actually drop the webhook defaulting.

jcrossley3 (approved):

markusthoemmes:
/retest

CI Registry was too slow to ship all the images in time seemingly.

markusthoemmes:
/retest

Namespace not found 🧐 we have to dig out why that stuff is happening. Have seen it on a couple of PRs now.

openshift-knative-operator/pkg/eventing/extension.go
aliok: > I was trying to preserve behavior that we are exhibiting currently

+1 on this approach! Let's do not try to change the architecture and improve units at the same time!

aliok:
/lgtm
/approve

I like the approach. We shouldn't change the logic and copy them over at the same time.
Use the tests as [characterization tests](https://en.wikipedia.org/wiki/Characterization_test)

PR #530, opened 2020-10-08
Eventing upgrade tests

cardil:
Introducing an Eventing upgrade tests to be executed withing a CI.


hack/lib/common.bash
```diff
@@ -56,3 +56,26 @@ function wait_until_hostname_resolves() {
   logger.error "Timeout waiting for hostname $1 to resolve via DNS"
   return 1
 }
+
+function wait_for_file {
```
mgencur, line 60: I suppose we could use something like this (similar to other places where we wait for something)
```
timeout 300 "[[ ! -f "${file} ]]" || return 1
```

test/lib.bash
```diff
@@ -115,12 +115,131 @@ function downstream_eventing_e2e_tests {
...
+  fi
+
+  (( serving_in_scope )) && wait_for_serving_test_services_settle
+
+  logger.info "Running postupgrade tests"
+
+  (( serving_in_scope )) && run_serving_postupgrade_test
+  (( eventing_in_scope )) && run_eventing_postupgrade_test
+
+  (( serving_in_scope )) && end_serving_prober "${serving_prober_pid}"
```
mgencur, line 191: I'm curious if this will work if we already called this function before upgrading cluster. I.e. is it idempotent?

test/serving.bash
```diff
@@ -77,107 +79,121 @@ function upstream_knative_serving_e2e_and_conformance_tests {
...
-    --resolvabledomain || return 1
+    --resolvabledomain || return $?
 
-  oc delete --ignore-not-found=true ksvc pizzaplanet-upgrade-service scale-to-zero-upgrade-service upgrade-probe -n serving-tests
+  logger.success 'Serving post upgrade tests passed'
+}
 
-  return 0
-  )
+function cleanup_serving_test_servinces {
```
mgencur, line 194: Nit: cleanup_serving_test_servinces -> cleanup_serving_test_services

test/upstream-e2e-tests.sh
```diff
@@ -22,7 +22,7 @@ teardown_serverless || failed=1
 # Run upgrade tests
 if [[ $TEST_KNATIVE_UPGRADE == true ]]; then
   (( !failed )) && install_serverless_previous || failed=3
-  (( !failed )) && run_knative_serving_rolling_upgrade_tests || failed=4
+  (( !failed )) && run_rolling_upgrade_tests "${UPGRADE_TEST_SCOPE:-serving}" || failed=4
```
mgencur, line 25: I dont' see this variable to be set anywhere, so we're running just Serving upgrade tests by default?

mgencur (commented):
Overall, looks good! Just a few minor things.
I'd also like to see the tests in action. AFAIK, this PR now runs only the Serving upgrade tests, right?

test/lib.bash
cardil: It should be.

test/upstream-e2e-tests.sh
cardil: Yes. As we talked about this. When I finish this I could set it to `serving,eventing`, in order to see how much that will prolong the test wall time.

mgencur: Ok. We got one run with just Serving. Perhaps you could enable eventing in the next run and we'll see.

test/lib.bash
```diff
@@ -115,12 +115,131 @@ function downstream_eventing_e2e_tests {
...
+  serving_in_scope="$(echo "${scope}" | grep -vq serving ; echo "$?")"
+  eventing_in_scope="$(echo "${scope}" | grep -vq eventing ; echo "$?")"
+
+  prev_serving_version="$(actual_serving_version)"
+  prev_eventing_version="$(actual_eventing_version)"
+
+  # Save the rootdir before changing dir
+  rootdir="$(dirname "$(dirname "$(realpath "${BASH_SOURCE[0]}")")")"
+
+  (( serving_in_scope )) && prepare_knative_serving_tests
```
cardil, line 138: Those instruction looks nice and clean, but they are dangerous as they hide errors.
```bash
(( serving_in_scope )) && something
```

cardil: See: https://superuser.com/questions/1169664/bash-if-on-single-line?newreg=48d7979019cd4e1a87d8e7a8b461e7b6#comment1933718_1169669

mgencur:
@cardil The PR looks good to me but the upgrade tests have a failure in the logs:
```
panic: test timed out after 30m0s
```
But they still report success.

cardil:
Yeah. I know. The error reporting is still not working well for some reason.

test/lib.bash
```diff
@@ -115,12 +115,161 @@ function downstream_eventing_e2e_tests {
...
+    logger.info "Updating Serverless to ${CURRENT_CSV}"
+    logger.debug "Serving version: ${prev_serving_version} -> ${latest_serving_version}"
+    logger.debug "Eventing version: ${prev_eventing_version} -> ${latest_eventing_version}"
+
+    approve_csv "$CURRENT_CSV" "$OLM_UPGRADE_CHANNEL"
+    if (( serving_in_scope )); then
+      check_serving_upgraded "${latest_serving_version}"
+    fi
+    if (( eventing_in_scope )); then
+      check_eventing_upgraded "${latest_eventing_version}"
```
mgencur, line 185: Where is this `check_eventing_upgraded` function defined? I can't find it anywhere.

cardil: It wasn't. That's why I expected that those tests will fail.

cardil:
Two issues are left to solve:

- [ ] wathola configmap is deployed into `/home/nonroot/.config/wathola` directory, but pods looks for it in `/root/.config/wathola` dir (confirmed)
- [ ] prober's PID looks to be different than the background process that is running (unconfirmed)

PR #533, opened 2020-10-09
Pin versions of Openshift related dependencies to 4.5

markusthoemmes:
For more consistency in our dependencies, this makes sure that the Openshift dependencies are somewhat recent. We can go to 4.5 here (unlike in tests) because we only rely on API definition and generate our own clients.


serving/ingress/go.mod
```diff
@@ -4,10 +4,12 @@ go 1.14
 
 require (
 	github.com/google/go-cmp v0.5.2
-	github.com/openshift/api v0.0.0-20200901182017-7ac89ba6b971
+	github.com/openshift/api v0.0.0-20200917102736-0a191b5b9bb0
```
matzew, line 7: hard to guess that's for 4.5 :-)

matzew:
/lgtm
(assuming test pass all fine)

serving/ingress/go.mod
markusthoemmes: The script is the source of truth here :sweat_smile:. These repos don't cut tags and only cut release branches sadly.

PR #536, opened 2020-10-09
Remove config-blah from manifest, add a test to prevent adding them again in the future

aliok:
Remove config-blah from manifest, add a test to prevent adding them again in the future

I wrote a test in `deploy/resources/knativekafka`. 

If we forget and add `config-blah` to the manifest again in the future, this test will fail.

We've forgotten to remove `config-tracing` in the past and I don't want us to be in that situation again. So, I added simple unit tests.


aliok:
@markusthoemmes Can I get an early review on this PR? Please only check the last commit. Rest is already reviwed by you in https://github.com/openshift-knative/serverless-operator/pull/510

aliok:
The context is:
- `tracing-config` already exists in Eventing's core manifest
- Trying to create it again causes problems, especially because of different ownerRefs

aliok:
This is rebased
cc @matzew @markusthoemmes

knative-operator/deploy/resources/knativekafka/kafkasource-v0.17.4.yaml
```diff
@@ -549,21 +549,3 @@ data:
     metrics.request-metrics-backend-destination: prometheus
     metrics.stackdriver-project-id: "<your stackdriver project id>"
     metrics.allow-stackdriver-custom-metrics: "false"
----
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: config-tracing
-  namespace: knative-eventing
```
matzew, line 557: I think that this is the only one in `knative-eventing` and all other `config-...` ConfigMaps are on the `knative-sources` is a bug:  see here: https://github.com/knative-sandbox/eventing-kafka/pull/95
So, it should be in `knative-sources`.  I think that's the "real fix" here.

However, 
That said, in the in the long run I'd like to see the KafkaSource bits installed into the _same_ namespace (e.g. `knative-eventing`). And I think one goal of the Serverless-operator is to also have all Knative components (serving, eventing, eventing-kafka, ...) all playing nicely together in one namespace....

With this ideal state (all components in one namespace), I think that this means that all `config-` ConfigMaps (e.g. there is a `config-tracing` for serving, eventing, kafkasource) should merge their content, if not containing only example data ...

aliok: > I think that this is the only one in knative-eventing and all other config-... ConfigMaps are on the knative-sources is a bug: see here: knative-sandbox/eventing-kafka#95
> So, it should be in knative-sources. I think that's the "real fix" here.

I thought that's by design.

But, that doesn't mean that we can have `config-tracing` in source manifest. We will already have that configmap created when `KnativeEventing` CR is created and the ownerRef will be `KnativeEventing`.

So, we still need to discard `config-tracing` when installing the source.

Let's hold this PR for a while. First let's backport your fix (only our fork) and have the same manifest here.
Then we can revisit this PR.

/hold

matzew:
Srsly, my recommendation is to change all to artifacts `knative-eventing` namespace

and remove all the "clashing" `config-zzz` ConfigMaps - see the kourier update, they are removed there too, if I recall correctly

aliok:
> Srsly, my recommendation is to change all to artifacts knative-eventing namespace

I want to do this later.

> and remove all the "clashing" config-zzz ConfigMaps - see the kourier update, they are removed there too, if I recall correctly

Now the PR is updated to do that.

Let's merge https://github.com/openshift-knative/serverless-operator/pull/546 first and then gonna rebase this one.

matzew (approved):
/lgtm
/approve

PR #540, opened 2020-10-12
Allow the index image to run in privileged mode

markusthoemmes:
4.6 seems to have restricted access to a few features that podman needs to run the commands in our index image. This widens the privileges even further for it to do its job.

### DIFF
--- a/hack/lib/catalogsource.bash
+++ b/hack/lib/catalogsource.bash
@@ -42,8 +42,9 @@ function install_catalogsource {
   # Undo potential changes to the CSV to not pollute the repository.
   mv "${rootdir}/bkp.yaml" "$csv"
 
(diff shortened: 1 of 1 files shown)


hack/lib/catalogsource.bash
```diff
@@ -64,6 +65,8 @@ spec:
       containers:
       - name: registry
         image: quay.io/openshift-knative/serverless-index:v1.14.3
```
matzew, line 67: what's that vor a version? :thinking:

markusthoemmes: version of the opm tool from OLM.

markusthoemmes:
/hold

To await the 4.6 build

PR #552, opened 2020-10-14
[SRVCOM-1071] Drop explicit version checks from our code

markusthoemmes:
As per title, I don't think we're gaining anything by having these checks at this point. They kinda reinforce our channel shipment model for no apparent reason.

/hold

For discussion.


devguyio:
That lgtm, and makes sense. QQ, should there be docs update cause of that change?

markusthoemmes:
Nope, we don't document our min Openshift version AFAIK other than it being encoded in the channels etc.

PR #559, opened 2020-10-14
Switch to use CRD v1 API for KnativeKafka

matzew:
/hold

We introduce a new CRD - let's start w/ `v1` API, instead of `v1beta1`.

The remove description generally belongs to the CSV, that does describe the different owned CRDs/APIs. I will add that later there

### DIFF
--- a/knative-operator/deploy/crds/operator_v1alpha1_knativekafka_crd.yaml
+++ b/knative-operator/deploy/crds/operator_v1alpha1_knativekafka_crd.yaml
@@ -5,14 +5,13 @@ metadata:
 spec:
   group: operator.serverless.openshift.io
   versions:
-  - &version
-    name: v1alpha1
+  - name: v1alpha1
     served: true
     storage: true
     subresources:
       status: {}
     schema:
-      openAPIV3Schema: &openAPIV3Schema
+      openAPIV3Schema:
         description: KnativeKafka is the Schema for the knativekafkas API
         properties:
           apiVersion:
--- a/knative-operator/deploy/crds/operator_v1alpha1_knativekafka_crd.yaml
+++ b/knative-operator/deploy/crds/operator_v1alpha1_knativekafka_crd.yaml
@@ -12,13 +12,104 @@ spec:
     subresources:
       status: {}
     schema:
-      openAPIV3Schema:
-        type: object
-        # this is a work around so we don't need to flush out the
-        # schema for each version at this time
-        #
-        # see issue: https://github.com/knative/serving/issues/912
-        x-kubernetes-preserve-unknown-fields: true
+      openAPIV3Schema: &openAPIV3Schema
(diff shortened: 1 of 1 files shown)


knative-operator/deploy/crds/operator_v1alpha1_knativekafka_crd.yaml
```diff
@@ -1,118 +1,34 @@
...
+    subresources:
+      status: {}
+    schema:
+      openAPIV3Schema:
+        type: object
+        # this is a work around so we don't need to flush out the
+        # schema for each version at this time
+        #
+        # see issue: https://github.com/knative/serving/issues/912
+        x-kubernetes-preserve-unknown-fields: true
```
markusthoemmes, line 21: Please don't drop the entire schema though.

matzew: :-(

knative-operator/deploy/crds/operator_v1alpha1_knativekafka_crd.yaml
```diff
@@ -1,118 +1,125 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: knativekafkas.operator.serverless.openshift.io
 spec:
   group: operator.serverless.openshift.io
+  versions:
+  - &version
```
markusthoemmes, line 8: Anchor is never reused.

knative-operator/deploy/crds/operator_v1alpha1_knativekafka_crd.yaml
```diff
@@ -1,118 +1,125 @@
...
+    name: v1alpha1
+    served: true
+    storage: true
+    subresources:
+      status: {}
+    schema:
+      openAPIV3Schema: &openAPIV3Schema
```
markusthoemmes, line 15: dito, anchor is never reused.

markusthoemmes (commented):
Is the diff so large by design? It's not necessarily bad but it looks like it's doing much more changes than you actually did (at least on a cursory view).

Have you done more than adding the printer columns and bumping the apiversion?

matzew:
> Have you done more than adding the printer columns and bumping the apiversion?

The diff is `+114` and `-107` lines.  The extra `7` lines are the `additionalPrinterColumns`

With my second commit I added the `schema` back, and while doing that, I align my change here with the structure of upstream Knative (Serving/Eventing) CRDs. As a result, the `names` and `scope` are now at the bottom of the file  

I think that's why the diff may look more than just adding 7 lines - sorry.


NOTE: this PR is still on hold - and I wanna keep it there a little bit longer :-)

markusthoemmes:
Looks good on the surface but can't really judge. Adding @aliok  for final approval.

PR #576, opened 2020-10-15
Update Kourier images to the latest references too

markusthoemmes:
Sadly, this was forgotten when upgrading the Kourier stuff. :crossed_fingers: 

/assign @nak3


markusthoemmes:
^^ Eventing upstream test flakes. I think we're good here wrt. Kourier images.

PR #565, opened 2020-10-15
Adding install flags for serving and eventing

matzew:
Signed-off-by: Matthias Wessendorf <mwessend@redhat.com>

A little bit of more modular installs...

New targets:
* `make install-serving`: Scales the cluster appropriately, deploys serverless-operator and Knative Serving.
* `make install-eventing`: Scales the cluster appropriately, deploys serverless-operator and Knative Eventing.

unchanged behavior for `make install`: continues to install both Serving and Eventing

PR #578, opened 2020-10-15
Separete upstream e2e targets 

matzew:
## Requirements

Some notes... this is based on https://github.com/openshift-knative/serverless-operator/pull/565
and also requires config, like: https://github.com/openshift/release/pull/12868

## Proposal

we introduce two new "targets" to run serving and eventing e2e upstream jobs, separately:

* `test-upstream-e2e-no-upgrade-serving`
* `test-upstream-e2e-no-upgrade-eventing`

They will run more isolated, and in different neutral clusters


matzew:
all green;

rebased

PR #584, opened 2020-10-20
Cleanup vendoring settings of the knative-operator

markusthoemmes:
The settings here have been a bit all over the place. Starting a cleanup to eventually move this to go modules too.

**Note:** The commits are created with some level of care. Stuff should be reviewable by walking through them.


knative-operator/Gopkg.toml
```diff
@@ -39,14 +39,10 @@
   name = "knative.dev/operator"
   version = "=v0.17.2"
 
-[[override]]
+[[constraint]]
```
rhuss, line 42: What's the difference between `override` and `constraint` and why has it changed here ?

knative-operator/Gopkg.toml
```diff
@@ -1,6 +1,11 @@
-[[override]]
-  name = "github.com/go-openapi/spec"
-  branch = "master"
+[prune]
```
rhuss, line 1: Looks like that the file has been restructured as well (i.e moving this section to the top), so it's not easy to me to find out the real difference.

What actually is the reasoning for this reordering ?

knative-operator/Gopkg.toml
```diff
@@ -2,10 +2,6 @@
   name = "github.com/go-openapi/spec"
   branch = "master"
 
-[[override]]
```
rhuss, line 10: Why is this override ineffective ?

knative-operator/Gopkg.toml
```diff
@@ -1,8 +1,3 @@
-# Force dep to vendor the code generators, which aren't imported just used at dev time.
```
rhuss, line 1: What was the original motivation to force this dependency for development and why is this not necessary anymore ?

knative-operator/Gopkg.toml
```diff
@@ -76,8 +76,4 @@ required = [
 [prune]
   go-tests = true
   non-go = true
-
-  [[prune.project]]
-    name = "k8s.io/kube-state-metrics"
-    unused-packages = true
-
+  unused-packages = true
```
rhuss: 👍

knative-operator/Gopkg.lock
```diff
@@ -56,14 +56,6 @@
   revision = "37c8de3658fcb183f997c4e13e8337516ab753e6"
   version = "v1.0.1"
 
-[[projects]]
```
rhuss, line 59: How is HEAD updated ? I can't see any change in Gopkg.toml reflecting the update.

rhuss (approved):
From the comment its not clear to me what settings are cleaned up and the reasoing behind this. 

It looks like its mostly to enable pruning of unused packages for all dependencies.

This looks good to me, removing stuff is always good.

And if the CI doesn't moan, then I'm totally fine to merge this technical cleanup.

Thanks for jumping into this !

/approve
/lgtm

knative-operator/Gopkg.toml
markusthoemmes: Because the thing wasn't vendored at all.

knative-operator/Gopkg.toml
markusthoemmes: To group things mostly so it's clearer why certain deps are pinned (because they're usually in relation to each other)

knative-operator/Gopkg.toml
markusthoemmes: override is technically to deeply override something through other deps, but in this case we're actually constraining what **we** need.

knative-operator/Gopkg.toml
markusthoemmes: No clue, I don't think it was ever needed. There's no code generation here.

knative-operator/Gopkg.lock
markusthoemmes: The commit message was likely bad. I just ran `dep ensure -v` on HEAD since that already had diffs.

PR #586, opened 2020-10-20
Add golangci-lint to our components and fix all existing issues

markusthoemmes:
We're using this upstream to catch a multitude of bugs and issues with the code. The config is checked in, so running it locally should be fairly painless.

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -78,6 +78,30 @@ jobs:
 
           echo "${TEMP_PATH}" >> $GITHUB_PATH
 
+      - name: Go Lint - knative-operator
+        uses: golangci/golangci-lint-action@v2
+        with:
+          version: v1.30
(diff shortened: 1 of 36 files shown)


rhuss (commented):
Looks good, but same concerns/suggestion as in #600 : Do we have a way to easily run the validation locally so that we do not need the CI for reporting things back which should be part of the initial development workflow ?

markusthoemmes:
/hold

Until we decide as a team we want this.

@rhuss yeah, since the linting configuration is checked in, we "just" need to install `golangci-lint` and run it respectively. I can add a README section about that.

markusthoemmes:
Rebased on #600, will rebase again once that merges. Only the last commit is significant.

markusthoemmes:

Infra, failing to create namespaces...

aliok (approved):
/lgtm
/approve
/unhold

PR #588, opened 2020-10-20
Drop dependency on pkg/errors in tests

markusthoemmes:
As per title. This dependency has basically made its way into the standard library, so we shouldn't be pulling in a dependency for it any more.


rhuss (approved):
totall +1

/lgtm
/approve

markusthoemmes:
/hold

For Ali's PR to land.

markusthoemmes:
/unhold

The other PR needs to be fixed anyway, so we might as well rebase against this.

PR #589, opened 2020-10-20
Create an even broader make target for generated files and check them

markusthoemmes:
`make generated-files` calls into all of the subproject's dep and generation scripts to make sure vendor directories and generated code is always up to date. This should avoid having diffs in these scripts in the future when just running on HEAD.

/assign @matzew @cardil


cardil:
Good idea to add this!

cardil (approved):
/lgtm

cardil (approved):
/lgtm

PR #592, opened 2020-10-20
Add the test dependencies to the general check too

markusthoemmes:
Whoops, I forgot these in the PR where I added this. So here we are, and we actually have a diff!

/assign @cardil


Makefile
```diff
@@ -85,4 +85,5 @@ release-files:
 generated-files: release-files
 	(cd openshift-knative-operator; ./hack/update-codegen.sh; ./hack/update-deps.sh; ./hack/update-manifests.sh)
```
cardil, line 86: Hmm, now I'm thinking maybe we should write this as:

```bash
cd openshift-knative-operator && ./hack/update-codegen.sh && ./hack/update-deps.sh && ./hack/update-manifests.sh
```
WDYT @markusthoemmes? If yes, we could do it in follow up.

cardil (approved):
/lgtm

PR #593, opened 2020-10-20
Use k8s svc instead of pod IP in catalogsource.bash

maschmid:
The `wait_until_pods_running` and `indexip` combo doesn't always work, as sometimes the pod isn't even created yet, so `wait_until_pods_running` doesn't wait for anything and `indexip` command fails on no item in `items[0]` (as no such pod exists yet)

Let's change it to just use a k8s svc instead...

### DIFF
--- a/hack/lib/catalogsource.bash
+++ b/hack/lib/catalogsource.bash
@@ -85,9 +85,20 @@ spec:
           /bin/opm registry serve -d index.db -p 50051
 EOF
 
-  # Wait for the index pod to be up to avoid inconsistencies with the catalog source.
-  wait_until_pods_running "$OLM_NAMESPACE"
-  indexip="$(oc -n "$OLM_NAMESPACE" get pods -l app=serverless-index -ojsonpath='{.items[0].status.podIP}')"
+  # Install the k8s svc pointing to the index pod.
+  cat <<EOF | oc apply -n "$OLM_NAMESPACE" -f - || return $?
+apiVersion: v1
+kind: Service
+metadata:
+  name: serverless-index
+spec:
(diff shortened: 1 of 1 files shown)


hack/lib/catalogsource.bash
```diff
@@ -96,7 +107,7 @@ kind: CatalogSource
 metadata:
   name: serverless-operator
 spec:
-  address: $indexip:50051
+  address: serverless-index.openshift-marketplace.svc:50051
```
maschmid, line 110: oops, should not hardcode OLM_NAMESPACE...

cardil (approved):
Much more sane approach

/lgtm

markusthoemmes:
/hold

There was a reason I did this. IIRC, the source had trouble coming up if I didn't wait for the pod to be up first and that never reconciled (or at least not in a usual amount of time). Can we please doublecheck that?

hack/lib/catalogsource.bash
```diff
@@ -85,9 +85,20 @@ spec:
           /bin/opm registry serve -d index.db -p 50051
 EOF
 
-  # Wait for the index pod to be up to avoid inconsistencies with the catalog source.
-  wait_until_pods_running "$OLM_NAMESPACE"
```
markusthoemmes, line 89: Do we still need to wait somewhere for the CatalogSource itself to become ready? The pod takes quite a while to come up. Maybe it'd be a compromise to wait for the Available condition on the deployment.

maschmid:
obsolete by https://github.com/openshift-knative/serverless-operator/pull/622

PR #594, opened 2020-10-20
Detached image override

aliok:
With this PR, the image override is implemented for Kafka components.

To help you follow what is going on:

- 0f78480: Just copy&paste the image override mechanism from upstream **as is**. It would be nice to use it as a library, but it is not possible because `KnativeKafka` doesn't support overrides with CR spec.
- 155b07d: Modify the *as is* image override mechanism to get rid of default registry and image cache. Also use the transform in the controller
- 3f2f2f6: Make the necessary changes in the CSV with actual image overrides. Naming clash in `DISPATCHER_IMAGE` env var override.
- ad558aa: Add scopes to image overrides. "DISPATCHER_IMAGE" is now "KAFKA_DISPATCHER_IMAGE" and "EVENTING_DISPATCHER_IMAGE".

About "scoping" image overrides:
- There is an env var which we should override: `DISPATCHER_IMAGE` of `kafka-ch-controller` deployment. However, we already have an override for that env var in Knative Eventing.
- The copied upstream image override mechanism doesn't check the deployment or container name when it is done with env vars.
- Thus, I created "scopes" for image overrides

Chat with Markus: https://coreos.slack.com/archives/CD87JDUB0/p1601892394404600

The scoping should later be used in serving and eventing. That way, we wouldn't have e.g. eventing related image overrides in serving CR when we mutate the CR.

Verification instructions:

1. Enable the controller by uncommenting the code in `webhook/add_knativekafka.go` and `controller/add_knativekafka.go`

2. `make images` && `make install` --> deploys S-O and installs KnativeServing and KnativeEventing

```
```

4. Install KnativeKafka:
```
cat <<EOS |kubectl apply -f -
---
kind: KnativeKafka
metadata:
  name: example
  namespace: knative-eventing
spec:
  source:
    enabled: true
  channel:
    enabled: true
...
EOS
```

```
```

6. Check the images for deployments in `knative-eventing`, `knative-serving`, `knative-sources` and `knative-serving-ingress` namespaces. No gcr.io should be there

7. Clean up:
```
```


knative-operator/pkg/common/eventing.go
```diff
@@ -16,7 +16,7 @@ func MutateEventing(ke *eventingv1alpha1.KnativeEventing) {
 
 // eventingImagesFromEnviron overrides registry images
 func eventingImagesFromEnviron(ke *eventingv1alpha1.KnativeEventing) {
-	ke.Spec.Registry.Override = BuildImageOverrideMapFromEnviron(os.Environ())
+	ke.Spec.Registry.Override = BuildImageOverrideMapFromEnviron(os.Environ(), "")
```
markusthoemmes, line 19: These should probably take `IMAGE_`

markusthoemmes (commented):
I know this is still WIP, so feel free to ignore:

I'd flip the scope thing on its head to avoid any collisions. Maybe make the kafka specific prefix `KAFKA_IMAGE_`. That way it's always clear where the new images belong and you don't have ambiguity issues because `IMAGE_KAFKA_` matches `IMAGE_` too.

knative-operator/pkg/common/util.go
```diff
@@ -47,13 +48,18 @@ func IngressNamespace(servingNamespace string) string {
 }
 
 // BuildImageOverrideMapFromEnviron creates a map to overrides registry images
-func BuildImageOverrideMapFromEnviron(environ []string) map[string]string {
+func BuildImageOverrideMapFromEnviron(environ []string, scope string) map[string]string {
 	overrideMap := map[string]string{}
 
+	prefix := ImagePrefix
+	if scope != "" {
```
markusthoemmes, line 55: I'd prefer we pass `ImagePrefix` in the callers where appropriate rather than silent defaulting. That's more towards my general suggestion though so... maybe not :sweat_smile:

hack/generate/csv.sh
```diff
@@ -91,7 +118,16 @@ cp "$template" "$target"
 
 for name in "${images[@]}"; do
   echo "Image: ${name} -> ${images_addresses[$name]}"
-  add_image "$target" "$name" "${images_addresses[$name]}"
```
markusthoemmes, line 94: This split might be unnecessary if we handle the prefixes differently as proposed. Sure, the images are not needed on the upstream operator but I don't think they'd hurt either.

knative-operator/pkg/common/util.go
aliok: I like your suggestion

hack/generate/csv.sh
aliok: Still need to track if an image is a Kafka image since they need to have a different prefix.
If that's ok, let's leave this as is. In the future, we can actually add scopes to eventing and serving (so that we don't pass eventing overrides to serving CR) and then we can find a unified solution.

knative-operator/pkg/common/util.go
aliok: Done

knative-operator/pkg/common/eventing.go
aliok: Done

aliok:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-knative_serverless-operator/594/pull-ci-openshift-knative-serverless-operator-master-4.5-upstream-e2e-aws-ocp-45/1318610857730510848/artifacts/upstream-e2e-aws-ocp-45/serverless-e2e/build-jKeEQdi8/pod-logs/test-api-server-source-v1-alpha2-event-ref-match-label-68x84/e2e-api-server-source-logger-pod-event-ref-match-label-recordevents.log
```
2020/10/20 19:29:11 Failed to process env varrequired key POD_NAME missing value
```

I've seen this before. It happened when we used the wrong test images. But, we didn't change anything in this PR for that.

@matzew @slinkydeveloper any idea?

aliok:
/test 4.5-upstream-e2e-aws-ocp-45

```
  * could not run steps: step root failed: could not wait for build: could not create watcher for build root: Get "https://172.30.0.1:443/apis/build.openshift.io/v1/namespaces/ci-op-d6x5kqms/builds?fieldSelector=metadata.name%3Droot&watch=true": dial tcp 172.30.0.1:443: connect: connection refused
  * could not run steps: step [release:latest] failed: unable to import image stream stable: could not resolve imagestream stable: Get "https://172.30.0.1:443/apis/image.openshift.io/v1/namespaces/ci-op-d6x5kqms/imagestreams/stable": dial tcp 172.30.0.1:443: connect: connection refused
```

matzew:
```
2020/10/20 21:44:44 Building root
2020/10/20 21:44:44 Tagging openshift/knative-v0.17.2:knative-eventing-src into pipeline:openshift_knative-v0.17.2_knative-eventing-src
2020/10/20 21:44:44 Tagging openshift/knative-v0.17.3:knative-serving-src into pipeline:openshift_knative-v0.17.3_knative-serving-src
I1020 21:44:45.326095      13 request.go:621] Throttling request took 1.054784956s, request: POST:https://172.30.0.1:443/apis/image.openshift.io/v1/namespaces/ci-op-d6x5kqms/imagestreamtags
2020/10/20 21:44:50 Importing release image latest
2020/10/20 21:44:51 Executing pod "release-images-latest-cli"
2020/10/20 21:44:59 Executing pod "release-images-latest"
2020/10/20 21:45:48 Imported release 4.5.14 created at 2020-09-30 12:51:50 +0000 UTC with 109 images to tag release:latest
2020/10/21 00:17:01 Build root failed, printing logs:
2020/10/21 00:17:01 error: Unable to retrieve logs from failed build: build root is in an error state. No logs are available.
2020/10/21 00:17:11 No custom metadata found and prow metadata already exists. Not updating the metadata.
2020/10/21 00:17:11 Ran for 2h32m29s
error: some steps failed:
  * could not run steps: step root failed: could not wait for build: the build root failed after 2h48m35s with reason BuildPodDeleted: The pod for this build was deleted before the build completed.
time="2020-10-21T00:17:11Z" level=info msg="Reporting job state 'failed' with reason 'executing_graph:step_failed:building_image_from_source'"
```

/retest

knative-operator/pkg/controller/knativekafka/images.go
```diff
@@ -0,0 +1,162 @@
+package knativekafka
```
markusthoemmes, line 1: Does this have any changes compared to the upstream version of it?

knative-operator/vendor/knative.dev/operator/pkg/reconciler/common/testing/util.go
```diff
@@ -0,0 +1,87 @@
...
+	t.Fatalf("expected does not equal actual. \nExpected: %v\nActual: %v", expected, actual)
+}
+
+func AssertDeepEqual(t *testing.T, actual, expected interface{}) {
+	t.Helper()
+	if reflect.DeepEqual(actual, expected) {
+		return
+	}
+	t.Fatalf("expected does not deep equal actual. \nExpected: %T %+v\nActual:   %T %+v", expected, expected, actual, actual)
+}
```
markusthoemmes, line 87: I'd move these to the only test that is using them.

markusthoemmes (commented):
I like it! Had two more clarification questions but I believe this is good to go.

knative-operator/pkg/controller/knativekafka/images.go
markusthoemmes: Nevermind, doublechecked it.

markusthoemmes (approved):
/lgtm
/approve

knative-operator/vendor/knative.dev/operator/pkg/reconciler/common/testing/util.go
markusthoemmes: :facepalm: this is vendored of course... too early, sorry for the noise.

PR #597, opened 2020-10-21
Set default channel to 4.6

warrenvw:
Since 1.11 only targets 4.6, we can now set the default channel to 4.6. Shawn concurs.

### DIFF
--- a/olm-catalog/serverless-operator/Dockerfile
+++ b/olm-catalog/serverless-operator/Dockerfile
@@ -20,4 +20,4 @@ LABEL \
       io.k8s.display-name="Red Hat OpenShift Serverless Bundle" \
       com.redhat.openshift.versions="v4.6" \
       com.redhat.delivery.operator.bundle=true \
-      com.redhat.delivery.backport=true
+      com.redhat.delivery.backport=false
--- a/olm-catalog/serverless-operator/Dockerfile
+++ b/olm-catalog/serverless-operator/Dockerfile
@@ -7,8 +7,8 @@ LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
 LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=serverless-operator
-LABEL operators.operatorframework.io.bundle.channel.default.v1="4.3"
-LABEL operators.operatorframework.io.bundle.channels.v1="4.5,4.6"
+LABEL operators.operatorframework.io.bundle.channel.default.v1="4.6"
+LABEL operators.operatorframework.io.bundle.channels.v1="4.6"
 
 LABEL \
       com.redhat.component="openshift-serverless-1-serverless-operator-bundle-container" \
@@ -18,6 +18,6 @@ LABEL \
       maintainer="serverless-support@redhat.com" \
       description="Red Hat OpenShift Serverless Bundle" \
       io.k8s.display-name="Red Hat OpenShift Serverless Bundle" \
-      com.redhat.openshift.versions="v4.5,v4.6" \
+      com.redhat.openshift.versions="v4.6" \
       com.redhat.delivery.operator.bundle=true \
       com.redhat.delivery.backport=true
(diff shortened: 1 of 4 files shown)


mgencur:
Wow! That's great! :-)
Anyway, this change needs to be applied in this way:
* modify the version at https://github.com/openshift-knative/serverless-operator/blob/master/olm-catalog/serverless-operator/project.yaml#L9 and https://github.com/openshift-knative/serverless-operator/blob/master/olm-catalog/serverless-operator/project.yaml#L19
* re-generate the target files by `make release-files`
This will do the exact changes on the files you have now

olm-catalog/serverless-operator/Dockerfile
```diff
@@ -7,8 +7,8 @@ LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
 LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=serverless-operator
-LABEL operators.operatorframework.io.bundle.channel.default.v1="4.3"
-LABEL operators.operatorframework.io.bundle.channels.v1="4.5,4.6"
+LABEL operators.operatorframework.io.bundle.channel.default.v1="4.6"
+LABEL operators.operatorframework.io.bundle.channels.v1="4.6"
```
matzew, line 11: glad to see this defaulting

warrenvw:
I started on the fix as soon as I saw the failed generate test. Thanks for the review.. :)

olm-catalog/serverless-operator/Dockerfile
warrenvw: Yeah, me too. And I'm sure @cardil as well. :)

templates/main.Dockerfile
```diff
@@ -20,4 +20,4 @@ LABEL \
       io.k8s.display-name="Red Hat OpenShift Serverless Bundle" \
       com.redhat.openshift.versions="__OCP_TARGET_VLIST__" \
       com.redhat.delivery.operator.bundle=true \
-      com.redhat.delivery.backport=true
+      com.redhat.delivery.backport=false
```
mgencur, line 23: Just curious. What does this mean?

warrenvw:
More importantly, looks to me like we need to run CI on 4.6, not 4.5...

templates/main.Dockerfile
warrenvw: We do not need to backport the bundle to the appregistry format since 1.11 is targeting 4.6.

mgencur:
Yeah. We should probably switch it. However, 4.6 is not even GA yet. Also, we have a nightly job on 4.6 (with slack notifications to serverless-dev) which currently fails. I think it's due to https://bugzilla.redhat.com/show_bug.cgi?id=1883589 which is being investigated on OpenShift side.
I can try to apply workarounds from the BZ and see if tests on 4.6 start passing. 
Anyway, it's possible right now to trigger tests on 4.6 manually on each PR against master branch. E.g. like this: https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift-knative/serverless-operator/openshift-knative-serverless-operator-master-presubmits.yaml#L472

templates/main.Dockerfile
cardil: I thought we will try to use separate bundles to our advantage. What I mean is to deploy specific versions of Serverless to specific OCP 4.6+ versions, and having channel called `stable` in all of them. That should be possible with the new bundle format, right?

https://issues.redhat.com/browse/SRVCOM-886

mgencur: @cardil this is not the right place to resolve this. I don't remember discussions about changing the channel strategy in this way from our program calls. Maybe you can bring it up there.

cardil: Sure. I think the same.

markusthoemmes:
/retest

Lovely infra.

templates/main.Dockerfile
warrenvw: Per https://issues.redhat.com/browse/RFE-723 and discussions I've had with Shawn, we'll be able to use a `stable` channel beginning with OCP 4.7.

markusthoemmes (approved):
/lgtm
/approve

PR #600, opened 2020-10-21
Setup woke to keep the codebase free of divisive language

markusthoemmes:
We use this tool upstream to keep the language in check too, so it should be fine here too.

A quick manual check yielded nothing, so we should be mostly clear anyway.

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -43,3 +43,53 @@ jobs:
       - name: Check if everything is consistent
         working-directory: ./src/github.com/${{ github.repository }}
         run: git diff --exit-code
+
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    env:
+      GOPATH: ${{ github.workspace }}
+    steps:
+      - name: Set up Go 1.14.x
+        uses: actions/setup-go@v2
+        with:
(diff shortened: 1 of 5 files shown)


markusthoemmes:
The upstream test doesn't seem worth retrying. This doesn't affect it at all, don't want to waste resources :joy:

rhuss (commented):
Looks good, but I wonder how the feedback looks like when the linter finds some issue. We just started to use it upstream, too, but the error reporting was by far not optimal (a lot of digging in log files is needed).

Also, wouldn't it also make sense to add a local make target for these kind of checks so that a round trip could be saved (which then again would trigger the CI for now reason when fixing a lint issue). I think this would be very important.

rhuss:
BTW is there any way to avoid running the CI when the validation fails ? I'm afraid the more we intermix both (validation and E2E CI) the more we have unneeded test runs that only takes away resources for other tests.

markusthoemmes:
@rhuss no digging through logfiles is needed. The `Files` view contains annotations (in bold red) about the errors. I'm game on adding a make target though.

Not sure there is a way we can avoid running CI if such a check fails. Maybe @mgencur knows.

rhuss (approved):
👍 

/lgtm
/approve

PR #602, opened 2020-10-21
Transform knative-operator to use go modules

markusthoemmes:
Let's see how this goes. This is blocking us from being more consistent with our dependencies.

**Note:** This contains a "manual" patch for manifestival, that I expect to be temporary until we can properly upgrade everything to more recent versions. The transform to go modules is kinda needed as a prereq for that though.


markusthoemmes:

Infra

markusthoemmes:
/hold 

until our KnativeKafka work lands.

rhuss (commented):
Looks good to me in general, and if it passes the tests then we should merge anyway.

However, reviewing this PR is hard. Would be cool if there would be one commit for the changes in vendor only, and another for the 'real change's.

markusthoemmes:
Ack @rhuss, I'll pull that apart for a proper review.

markusthoemmes:
@rhuss done, all real changes are now in commit 1.

knative-operator/hack/update-deps.sh
```diff
@@ -0,0 +1,22 @@
...
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# Prune modules.
+go mod tidy
+go mod vendor
+
+# TODO: Remove this once we bump kubernetes versions.
```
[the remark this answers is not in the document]
rhuss: So iiur, the patch is needed because the upstream manifestival only works with new k8s versions. But why are we need not pinning the manifestival also to an older version that works with k8s 1.13 instead of patching it on the fly ?

knative-operator/hack/update-deps.sh
```diff
@@ -0,0 +1,22 @@
...
+git apply "$ROOT_DIR/hack/manifestival.patch"
+
+rm -rf $(find vendor/ -name 'OWNERS')
+# Remove unit tests & e2e tests.
+rm -rf $(find vendor/ -path '*/pkg/*_test.go')
+rm -rf $(find vendor/ -path '*/e2e/*_test.go')
+
```
rhuss, line 20: Upstream script also removes LICENSE files. 

Other remarks:

* You could combine to a single find 
* This command will break totally if files names contains spaces. So user `xargs` instead.
* It also breaks if there are too many files so that the overall command line becomes too long.
* Why not deleting just all `*_test.go` files ?

rhuss: Recommended form to fix those issues:

```
find "vendor/" \( -name "OWNERS" -o -name "*_test.go" \) -print0 | xargs -0 rm -f
```

rhuss (commented):
Thanks for splitting the commits. Find my comments inline.

knative-operator/go.mod
```diff
@@ -0,0 +1,39 @@
...
+	github.com/gregjones/httpcache v0.0.0-20190611155906-901d90724c79 // indirect
+	github.com/manifestival/controller-runtime-client v0.2.0-0.1.12
+	github.com/manifestival/manifestival v0.6.1
+	github.com/openshift/api v0.0.0-20190927182313-d4a64ec2cbd8
+	github.com/openzipkin/zipkin-go v0.2.5 // indirect
+	github.com/operator-framework/operator-sdk v0.10.1
+	github.com/prometheus/client_golang v1.6.0
+	github.com/spf13/pflag v1.0.5
+	k8s.io/api v0.18.7-rc.0
+	k8s.io/apimachinery v0.19.0
```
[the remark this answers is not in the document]
markusthoemmes: They are both overridden in the replace section to nail them to what we currently have to use.

knative-operator/go.mod
```diff
@@ -0,0 +1,39 @@
...
+	k8s.io/client-go v11.0.1-0.20190805182717-6502b5e7b1b5+incompatible
+	knative.dev/operator v0.17.2
+	knative.dev/pkg v0.0.0-20200831162708-14fb2347fb77
+	knative.dev/serving v0.15.3
```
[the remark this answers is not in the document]
markusthoemmes: 0.17.3 is used, but we cannot update currently because we get clashes with client-go dependencies that we in turn cannot update because we have to use an old controller-runtime because of the way our webhooks are setup.

I plan to untangle all this (this PR is keeping all dependencies roughly the same as they used to be before) to be able to update everything to much more appropriate versions.

knative-operator/go.mod
```diff
@@ -0,0 +1,39 @@
...
+	sigs.k8s.io/controller-runtime v0.6.1
+	sigs.k8s.io/yaml v1.2.0
+)
+
+replace (
+	// Because the bitbucket version is broken
+	bitbucket.org/ww/goautoneg => github.com/adjust/goautoneg v0.0.0-20150426214442-d788f35a0315
```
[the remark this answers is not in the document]
markusthoemmes: I haven't deeply checked, but I guess in some apiserver dependency. I had issues with these upstream too. The dep indeed is super old and didn't get updated in between.

knative-operator/go.mod
```diff
@@ -0,0 +1,39 @@
...
+
+	// Kubernetes v1.13.4
+	k8s.io/api => k8s.io/api v0.0.0-20190222213804-5cb15d344471
+	k8s.io/apimachinery => k8s.io/apimachinery v0.0.0-20190221213512-86fb29eff628
+	k8s.io/client-go => k8s.io/client-go v0.0.0-20190228174230-b40b2a5939e4
+
+	// controller-runtime
+	sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.1.12
```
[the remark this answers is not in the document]
markusthoemmes: Some other dependency we have depends on a newer version of controller-runtime and thus make go mod take the newer one above.

knative-operator/go.mod
```diff
@@ -0,0 +1,39 @@
...
+	knative.dev/serving v0.15.3
+	sigs.k8s.io/controller-runtime v0.6.1
+	sigs.k8s.io/yaml v1.2.0
+)
+
+replace (
+	// Because the bitbucket version is broken
+	bitbucket.org/ww/goautoneg => github.com/adjust/goautoneg v0.0.0-20150426214442-d788f35a0315
+
+	// Kubernetes v1.13.4
```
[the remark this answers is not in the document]
markusthoemmes: See longer comment above.

knative-operator/hack/update-deps.sh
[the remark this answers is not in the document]
markusthoemmes: I fighted with go modules quite a while to make this work with the older manifestival version, but it would throw errors at me. This seemed the most non-intrusive way forward for now.

knative-operator/hack/update-deps.sh
markusthoemmes: Totally copied from upstream, lemme check.

knative-operator/hack/manifestival.patch
```diff
@@ -0,0 +1,45 @@
+diff --git a/knative-operator/vendor/github.com/manifestival/manifestival/client.go b/knative-operator/vendor/github.com/manifestival/manifestival/client.go
+index a8d108ce..85994972 100644
+--- a/knative-operator/vendor/github.com/manifestival/manifestival/client.go
++++ b/knative-operator/vendor/github.com/manifestival/manifestival/client.go
```
matzew, line 4: Is it worth to apply upstream? and eventually use a release containing the patch ?

rhuss: I think as soon as we get the k8s deps straight, we get rid of this patch anyway. It's more a 'backwards' patch (to make manifestival compatible with older k8s), so don't think upstream is keen on downgrading their k8s deps :)

rhuss: (this is how I understoo @markusthoemmes )

knative-operator/hack/update-deps.sh
```diff
@@ -13,10 +13,4 @@ go mod vendor
...
 
-rm -rf $(find vendor/ -name 'OWNERS')
-# Remove unit tests & e2e tests.
-rm -rf $(find vendor/ -path '*/pkg/*_test.go')
-rm -rf $(find vendor/ -path '*/e2e/*_test.go')
-
-# Add permission for shell scripts
-chmod +x $(find vendor -type f -name '*.sh')
\ No newline at end of file
+find vendor/ \( -name "OWNERS" -o -name "OWNERS_ALIASES" -o -name "BUILD" -o -name "BUILD.bazel" -o -name "*_test.go" \) -print0 | xargs -0 rm -f
```
rhuss, line 16: 👍

rhuss: btw, is the file mode change on the scripts intended, so that they are not executable anymore  ? (see the other files in this commit, can't really comment on them directly)

Happens because you have removed the chmod line.

knative-operator/hack/manifestival.patch
markusthoemmes: Exactly, we only need this because of our hilariously old k8s deps.

knative-operator/hack/update-deps.sh
markusthoemmes: For now that seems fine, yes. We can always bring that back should we actually need to execute files.

rhuss (approved):
Looks good for me, please unhold when ready to merge

/hold
/lgtm
/approve

knative-operator/go.mod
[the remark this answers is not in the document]
aliok: Also, IMO, keeping the versions the same with `Gopkg.toml` is a better idea. ...unless we really really need to change them here because of some `go mod` problems.

Actually upgrading the dependency versions should be in a separate PR.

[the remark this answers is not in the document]
markusthoemmes: Agree @aliok, that was the goal of this.

aliok:
/lgtm
/approve

Thanks for doing this.

But, let's keep `/hold` for a while until we're done with big Kafka related PRs

aliok:
/unhold

we're done with big Kafka PRs. We can merge this.

PR #605, opened 2020-10-22
Have KafkaSource resources in knative-eventing namespace

aliok:
- Move KafkaSource reosources in `knative-eventing` namespace
- Get rid of custom `injectOwner` that was skipping resources from other namespaces
- Get rid of other references to `knative-sources` namespace

### DIFF
--- a/knative-operator/deploy/resources/knativekafka/kafkasource-v0.17.1.yaml
+++ b/knative-operator/deploy/resources/knativekafka/kafkasource-v0.17.1.yaml
@@ -1,15 +1,8 @@
 apiVersion: v1
-kind: Namespace
-metadata:
-  name: knative-sources
-  labels:
-    contrib.eventing.knative.dev/release: "v0.17.1"
----
-apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: kafka-controller-manager
-  namespace: knative-sources
+  namespace: knative-eventing
   labels:
     contrib.eventing.knative.dev/release: "v0.17.1"
 ---
@@ -146,7 +139,7 @@ metadata:
 subjects:
 - kind: ServiceAccount
   name: kafka-controller-manager
-  namespace: knative-sources
+  namespace: knative-eventing
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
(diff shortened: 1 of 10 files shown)


knative-operator/deploy/resources/knativekafka/kafkasource-v0.17.1.yaml
```diff
@@ -1,15 +1,8 @@
...
-  name: knative-sources
-  labels:
-    contrib.eventing.knative.dev/release: "v0.17.1"
----
-apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: kafka-controller-manager
-  namespace: knative-sources
+  namespace: knative-eventing
```
markusthoemmes, line 5: How are these manifests obtained and updated this way? Should we enhance the deployment checks in our E2E tests to make sure these land there too so we don't forget in future updates?

Even better: Can we get a download script that does it for us (and is hooked into `make generated-files`)?

aliok: I already took some notes about improvements to do in eventing-contrib here: https://github.com/openshift-knative/serverless-operator/issues/486 . Haven't created a ticket for that yet, but I will.

Thanks for the further improvement ideas.

> How are these manifests obtained and updated this way? 

We get the manifest from **our** eventing-contrib. We currently have a big manifest with source and channel in it. Then we separate it into 2 files manually, replace the namespace for the KnativeSource, do some further adjustments (like deleting config-tracing) from the manifest.

All manual. Automating these on S-O depends on eventing-contrib having separate files.


> Even better: Can we get a download script that does it for us (and is hooked into make generated-files)?

Yes, should be possible.

> Should we enhance the deployment checks in our E2E tests to make sure these land there too so we don't forget in future updates?

I am +1 on this. I think this is what we can do in S-O at the moment.
Let me do this. We already have a test that reports failure when there's config-tracing and similar in the manifests.

markusthoemmes: Yeah just a test works for me for now.

markusthoemmes (approved):
/lgtm
/approve

knative-operator/deploy/resources/knativekafka/kafkasource-v0.17.1.yaml
aliok: Done. Also created https://issues.redhat.com/browse/SRVKE-595 to track this.

matzew: That is a good idea.

I can do some script, folks

PR #607, opened 2020-10-22
KnativeKafka switch to CRD_v1 and adding additionalPrinterColumns

matzew:

Since `KnativeKafka` is _new_, let's land this while using `CRD v1` API shape. While doing the change towards `CRD v1`, I decided align the change with how the structure of upstream Knative (Serving/Eventing) CRDs are looking. As a result, the `names` and `scope` are now at the bottom of the file. This might cause the `DIFF` a bit larger than perhaps expected.

While at the file, I am also adding `additionalPrinterColumns` to the CRD...

### DIFF
--- a/knative-operator/deploy/crds/operator_v1alpha1_knativekafka_crd.yaml
+++ b/knative-operator/deploy/crds/operator_v1alpha1_knativekafka_crd.yaml
@@ -1,115 +1,122 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: knativekafkas.operator.serverless.openshift.io
 spec:
   group: operator.serverless.openshift.io
+  versions:
+  - name: v1alpha1
+    served: true
+    storage: true
+    subresources:
+      status: {}
+    schema:
+      openAPIV3Schema:
+        type: object
+        description: KnativeKafka is the Schema for the knativekafkas API
+        properties:
+          apiVersion:
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#resources'
+            type: string
+          kind:
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          spec:
+            type: object
+            description: 'KnativeKafkaSpec defines the desired state of the KnativeKafka (from the client).'
+            properties:
+              channel:
+                description: Allows configuration for KafkaChannel installation
+                properties:
+                  bootstrapServers:
+                    description: BootstrapServers is comma separated string of bootstrapservers
+                      that the KafkaChannels will use
+                    type: string
+                  enabled:
+                    description: Enabled defines if the KafkaChannel installation
+                      is enabled
+                    type: boolean
+                required:
+                - enabled
+                - bootstrapServers
+                type: object
+              source:
+                description: Allows configuration for KafkaSource installation
+                properties:
+                  enabled:
+                    description: Enabled defines if the KafkaSource installation is
+                      enabled
+                    type: boolean
+                required:
+                - enabled
+                type: object
+          status:
+            type: object
+            description: 'KnativeKafkaStatus defines the observed state of KnativeKafka (from the controller).'
+            properties:
+              annotations:
+                additionalProperties:
+                  type: string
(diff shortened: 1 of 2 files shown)


markusthoemmes (approved):

aliok:
```
error checking provided apis in bundle serverless-operator.v1.11.0: couldn't find operator.serverless.openshift.io/v1alpha1/KnativeKafka (knativekafkas) in bundle. found: map[operator.knative.dev/v1alpha1/KnativeEventing (knativeeventings):{} operator.knative.dev/v1alpha1/KnativeServing (knativeservings):{}]"
```

I got exactly this before and that's why I excluded the CRD v1 work in #587

matzew:
ah... ok - let me look here

matzew:
/hold

will sortout the problem ... wy V1 seems to be a problem

olm-catalog/serverless-operator/manifests/operator_v1alpha1_knativekafka_crd.yaml
```diff
@@ -1,115 +1,121 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
awgreene, line 1: It seems like the CRD's version was updated but the CSV still references the [v1alpha1 version](https://github.com/openshift-knative/serverless-operator/blob/master/olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml#L85), which is likely causing the following error:

```
error checking provided apis in bundle serverless-operator.v1.11.0: couldn't find operator.serverless.openshift.io/v1alpha1/KnativeKafka (knativekafkas) in bundle. found: map[operator.knative.dev/v1alpha1/KnativeEventing (knativeeventings):{} operator.knative.dev/v1alpha1/KnativeServing (knativeservings):{}]"
```

awgreene: As of right now, your bundle provides v1 crds and your CSV expects v1alpha1 CRDs to be provided. This applies for both CRDs that you've changed in this PR.

matzew: thanks for the reply

but...  the actual API version of our bits is `v1alpha1`;

I more here mean about the CRD api itself (using v1 - instead of `v1beta1`). Of the CRD api - not our API

awgreene: My mistake - taking another look.

knative-operator/deploy/crds/operator_v1alpha1_knativekafka_crd.yaml
```diff
@@ -1,115 +1,121 @@
...
+                  enabled:
+                    description: Enabled defines if the KafkaSource installation is
+                      enabled
+                    type: boolean
+                required:
+                - enabled
+                type: object
+          status:
+            type: object
+              description: 'KnativeKafkaStatus defines the observed state of KnativeKafka (from the controller).'
```
markusthoemmes, line 60: This is wrongly indented I believe.

matzew: :facepalm:

olm-catalog/serverless-operator/manifests/operator_v1alpha1_knativekafka_crd.yaml
awgreene: It looks like the V1 yaml is wrong - applying the CRD fails:

```
oc apply -f operator_v1alpha1_knativekafka_crd.yaml 
error: error parsing operator_v1alpha1_knativekafka_crd.yaml: error converting YAML to JSON: yaml: line 60: mapping values are not allowed in this context
```

olm-catalog/serverless-operator/manifests/operator_v1alpha1_knativekafka_crd.yaml
```diff
@@ -1,115 +1,121 @@
...
+                  enabled:
+                    description: Enabled defines if the KafkaSource installation is
+                      enabled
+                    type: boolean
+                required:
+                - enabled
+                type: object
+          status:
+            type: object
+              description: 'KnativeKafkaStatus defines the observed state of KnativeKafka (from the controller).'
```
awgreene: This also seems to be indented incorrectly.

matzew: yep :-)

olm-catalog/serverless-operator/manifests/operator_v1alpha1_knativekafka_crd.yaml
awgreene: After fixing the indent issue that @markusthoemmes [pointed out](https://github.com/openshift-knative/serverless-operator/pull/607#discussion_r510125665) in the OLM-Catalog CRD I was able to build the bundle correctly.

PR #608, opened 2020-10-22
Wait for the Kafka CRD to be ready

markusthoemmes:
There is a rare race where this isn't ready when we apply the Kafka CR so let's wait for it to be established.

This avoids

```
error: unable to recognize "https://raw.githubusercontent.com/strimzi/strimzi-kafka-operator/0.19.0/examples/kafka/kafka-persistent.yaml": no matches for kind "Kafka" in version "kafka.strimzi.io/v1beta1"
```


matzew (approved):
/lgtm
/approve

PR #609, opened 2020-10-22
Removing Version from KnativeKafkaStatus

matzew:

looks like we not really use it


aliok (approved):
/lgtm
/approve

Thanks for catching this!

PR #613, opened 2020-10-22
Don't use conditional Bash script style

cardil:
* Addition of ERR stack trace and dump state for OpenShift CI
 * Clean up of code
 * Optimization of catalog source creation

PR #621, opened 2020-10-23
Bugfix: Making upstream functions errorfree (pipefail mostly)

cardil:
Due to the style of our Bash scripts code that isn't error free creeps in.

This PR fixes couple of instances of such problem. Following changes are required to make our scripts fully respects failures, by ditching our "conditional" style of Bash scripts.

PR has been extracted from #613 to ease review.


hack/lib/tracing.bash
```diff
@@ -62,7 +62,7 @@ EOF
 }
 
 function enable_eventing_tracing {
-  header_text "Configuring tracing for Eventing"
+  logger.info "Configuring tracing for Eventing"
```
markusthoemmes, line 65: Is this related to this PR?

test/eventing-contrib.bash
```diff
@@ -5,15 +5,17 @@ function upstream_knative_eventing_contrib_e2e {
...
+  local randomns
 
   TEST_IMAGE_TEMPLATE="registry.svc.ci.openshift.org/openshift/knative-${KNATIVE_EVENTING_CONTRIB_VERSION}:knative-eventing-test-{{.Name}}"
 
   cd "$KNATIVE_EVENTING_CONTRIB_HOME" || return $?
 
   # This the namespace used to install and test Knative Eventing-Contrib.
+  randomns="knative-eventing-$(LC_ALL=C dd if=/dev/urandom bs=256 count=1 2> /dev/null \
+    | tr -dc 'a-z0-9' | fold -w 10 | head -n 1)"
+  TEST_EVENTING_NAMESPACE="${TEST_EVENTING_NAMESPACE:-"${randomns}"}"
```
markusthoemmes, line 17: This seems unrelated.

markusthoemmes:
```
ERROR 01:13:57.424 Upstream function wait_until_pods_running doesn't work well with strict Bash rules that should be used while scripting, and shouldn't be used in OpenShift Serverless. Please use wait_until_labelled_pods_are_ready instead
```

This is a genuine error from the upstream scripts. I guess  we need to disable bash strictness when calling out upstream for now.

/hold

For visibility.

hack/lib/tracing.bash
cardil: Yup. I can remove that.

test/eventing-contrib.bash
cardil: It is related. It fails with pipefail.

cardil:
> ```
> ERROR 01:13:57.424 Upstream function wait_until_pods_running doesn't work well with strict Bash rules that should be used while scripting, and shouldn't be used in OpenShift Serverless. Please use wait_until_labelled_pods_are_ready instead
> ```
> 
> This is a genuine error from the upstream scripts. I guess we need to disable bash strictness when calling out upstream for now.
> 
> /hold
> 
> For visibility.

Okay. I put this, as I thought we are not using this method anywhere. But that's obviously not the case. 

I think I will remove this warning, so this may continue to be used for now. And while enabling strict bash I'll make exception for this case.

hack/lib/tracing.bash
cardil: In fact this is related. This function `header_text` doesn't exist. 

Scripts should have failed on this, but they don't because of our Bash conditional style.

cardil:
Tests are passing. Can I get unhold @markusthoemmes, now?

markusthoemmes (approved):
/lgtm
/approve
/unhold

PR #623, opened 2020-10-26
Remove a couple of unused helper functions

markusthoemmes:
As per title. These aren't in use anywhere.

### DIFF
--- a/hack/lib/common.bash
+++ b/hack/lib/common.bash
@@ -1,24 +1,9 @@
 #!/usr/bin/env bash
 
-function array.contains {
-  local e match="$1"
-  shift
-  for e; do [[ "$e" == "$match" ]] && return 0; done
-  return 1
-}
-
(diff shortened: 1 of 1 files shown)


matzew (approved):
/lgtm
/approve

hack/lib/common.bash
```diff
@@ -1,26 +1,11 @@
 #!/usr/bin/env bash
 
-function array.contains {
```
cardil, line 3: I'm using that in eventing upgrade tests. Please don't remove it.

cardil: /hold

markusthoemmes: Git would've just given it back to you anyway :shrug:. Pushed.

cardil: Sure. Just need to rebase and wait for tests. Thanks.

PR #626, opened 2020-10-26
[SRVCOM-776] Remove webhook self-registration to unblock c-r bumps

markusthoemmes:
As per title. Apart from fixing an important certificate related issue (https://issues.redhat.com/browse/SRVCOM-1053) this will allow us to move forward with our controller-runtime and operator-sdk versions to something muuuuuch more recent.

This is a first step towards that. No version bumps just yet to keep the number of moving parts lowish.

This changes the way the webhooks get registered for our knative-operator deployment. Instead of it generating a cert and registering itself (which is dropped from newer controller-runtime versions anyway), this defers all that to the CSV. OLM auto-generates and injects (and rolls) certificates to the deployments specified in the webhook configurations.

**Note:** This contains a small "patch" that is required for this to work on such an old version. We can drop this on newer versions of controller-runtime.

### DIFF
--- a/knative-operator/hack/certnames.patch
+++ b/knative-operator/hack/certnames.patch
@@ -1,0 +1,16 @@
+diff --git a/knative-operator/vendor/sigs.k8s.io/controller-runtime/pkg/webhook/internal/cert/writer/certwriter.go b/knative-operator/vendor/sigs.k8s.io/controller-runtime/pkg/webhook/internal/cert/writer/certwriter.go
+index ed6b5110..c0c67e16 100644
+--- a/knative-operator/vendor/sigs.k8s.io/controller-runtime/pkg/webhook/internal/cert/writer/certwriter.go
++++ b/knative-operator/vendor/sigs.k8s.io/controller-runtime/pkg/webhook/internal/cert/writer/certwriter.go
+@@ -33,9 +33,9 @@ const (
+ 	// CACertName is the name of the CA certificate
+ 	CACertName = "ca-cert.pem"
+ 	// ServerKeyName is the name of the server private key
+-	ServerKeyName = "key.pem"
++	ServerKeyName = "apiserver.key"
+ 	// ServerCertName is the name of the serving certificate
+-	ServerCertName = "cert.pem"
++	ServerCertName = "apiserver.crt"
+ )
+ 
+ // CertWriter provides method to handle webhooks.
--- a/knative-operator/hack/update-deps.sh
+++ b/knative-operator/hack/update-deps.sh
@@ -13,6 +13,10 @@ go mod vendor
 # TODO: Remove this once we bump kubernetes versions.
 git apply "$ROOT_DIR/hack/manifestival.patch"
 
+# This patch is required to make sure the cert names read by the webhook setup is equal
+# to the names chosen by OLM. We can drop this once we bump to newer versions of c-r.
+git apply "$ROOT_DIR/hack/certnames.patch"
+
 find vendor/ \( -name "OWNERS" \
   -o -name "OWNERS_ALIASES" \
   -o -name "BUILD" \
--- a/knative-operator/pkg/webhook/webhook.go
+++ b/knative-operator/pkg/webhook/webhook.go
@@ -1,10 +1,6 @@
 package webhook
 
 import (
-	"errors"
-	"os"
-
-	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/common"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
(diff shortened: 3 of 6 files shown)


markusthoemmes:
/hold

This shouldn't go in before the 1.11 cut.

markusthoemmes:
/test 4.6-operator-e2e-aws-ocp-46

Let's see if this is consistent.

markusthoemmes:
/unhold

1.11 has been cut, so we can move forward here.

aliok:
/approve

Looks good to me. I liked the possibility to configure webhooks with CSV.

Leaving `/lgtm` to others

knative-operator/pkg/webhook/webhook.go
```diff
@@ -31,26 +27,12 @@ func AddToManager(m manager.Manager) error {
 	}
 
 	log.Info("Setting up webhook server")
-	operatorNs := os.Getenv(common.NamespaceEnvKey)
-	if operatorNs == "" {
-		return errors.New("NAMESPACE not provided via environment")
-	}
 	// This will be started when the Manager is started
+	truePtr := true
```
skonto, line 31: you can use:
```
import "knative.dev/pkg/ptr"
DisableWebhookConfigInstaller: ptr.Bool(true),
```

markusthoemmes: I wanted to avoid pulling this package in, but we indeed already have it somewhere else. Adding.

knative-operator/pkg/webhook/webhook.go
```diff
@@ -31,26 +27,12 @@ func AddToManager(m manager.Manager) error {
...
-				Namespace: operatorNs,
-				Name:      "admission-server-service",
-				// Selectors should select the pods that runs this webhook server.
-				Selectors: map[string]string{
-					"app": "openshift-admission-server",
-				},
-			},
-		},
+		Port:                          9876,
+		CertDir:                       "/apiserver.local.config/certificates",
```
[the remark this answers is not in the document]
markusthoemmes: Yep, as long as we support 4.6 we need this override. Once we go into 4.7, we can drop stuff.

knative-operator/hack/certnames.patch
```diff
@@ -0,0 +1,16 @@
...
++++ b/knative-operator/vendor/sigs.k8s.io/controller-runtime/pkg/webhook/internal/cert/writer/certwriter.go
+@@ -33,9 +33,9 @@ const (
+ 	// CACertName is the name of the CA certificate
+ 	CACertName = "ca-cert.pem"
+ 	// ServerKeyName is the name of the server private key
+-	ServerKeyName = "key.pem"
++	ServerKeyName = "apiserver.key"
+ 	// ServerCertName is the name of the serving certificate
+-	ServerCertName = "cert.pem"
++	ServerCertName = "apiserver.crt"
```
rhuss, line 13: Just a question: Are there any plans to make the name of the cert's location configurable in the upstream controller runtime ? sure, it's fixed for Kubernetes but other distributions like Rancher might have similar issues. On the other hand, I wonder why OpenShift changed that naming ....

knative-operator/hack/update-deps.sh
```diff
@@ -13,6 +13,10 @@ go mod vendor
 # TODO: Remove this once we bump kubernetes versions.
 git apply "$ROOT_DIR/hack/manifestival.patch"
 
+# This patch is required to make sure the cert names read by the webhook setup is equal
+# to the names chosen by OLM. We can drop this once we bump to newer versions of c-r.
+git apply "$ROOT_DIR/hack/certnames.patch"
+
```
rhuss, line 19: So, that patch existed previously but was not applied ? (because this command is new). Any reason for this ?

rhuss (approved):
Looks good to me (with some questions, see inline ;-)

For a next step, and addressing @jcrossley3 concern about not being able to run the operator then without the help of the OLM, I wonder whether we could add a `hack/setup-webhooks.sh` that would create the webhook configuration via scripting ?

But on the other hand, as most of the value add in this repo is about being installable with the OLM, personally I also woulnd't mind to integrate the OLM in the regular developer workflow.

/lgtm
/hold

(unhold as needed)

knative-operator/hack/certnames.patch
markusthoemmes: Yes, this is fixed in a newer c-r version. The followup PR for this (that'll bump the version) will have this configurable in client code rather than patching vendor.

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
```diff
@@ -559,6 +559,102 @@ spec:
...
+            - v1alpha1
+          operations:
+            - CREATE
+            - UPDATE
+          resources:
+            - knativekafkas
+      sideEffects: None
+      webhookPath: /validate-knativekafkas
+    - generateName: mutating.knativeeventings.operator.serverless.openshift.io
+      type: MutatingAdmissionWebhook
```
matzew, line 621: do we need `knativekafkas`1 for mutatingadmissionwebhook?

knative-operator/hack/update-deps.sh
markusthoemmes: Because we auto-generated the certs inside the component itself, it automatically put it into the right place. Handing that off to OLM is new in this PR.

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
markusthoemmes: There's only a ValidatingWebhook for knativekafkas.

knative-operator/pkg/webhook/webhook.go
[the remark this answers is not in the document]
matzew: @markusthoemmes  wanna add a comment, regarding 4.7 ?

markusthoemmes:
/unhold

Off we go!

PR #627, opened 2020-10-26
Few touchups to the error handling during monitoring setup

markusthoemmes:
As per title. I mostly wanted to avoid an error like this on startup:

```
{"level":"error","ts":"2020-10-26T10:16:15.409Z","logger":"cmd","msg":"Failed to start monitoring","error":"failed to setup the Service monitor: servicemonitors.monitoring.coreos.com \"knative-openshift-metrics\" already exists","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/src/github.com/openshift-knative/serverless-operator/knative-operator/vendor/github.com/go-logr/zapr/zapr.go:128\nmain.main\n\t/go/src/github.com/openshift-knative/serverless-operator/knative-operator/cmd/manager/main.go:123\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:203"}
```

and ended up cleaning up a few more parts.

### DIFF
--- a/knative-operator/pkg/common/service_monitor.go
+++ b/knative-operator/pkg/common/service_monitor.go
@@ -13,6 +13,7 @@ import (
 	"github.com/operator-framework/operator-sdk/pkg/metrics"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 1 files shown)


knative-operator/pkg/common/service_monitor.go
```diff
@@ -53,28 +54,31 @@ func SetupServerlessOperatorServiceMonitor(ctx context.Context, cfg *rest.Config
 	// Create Service object to expose the metrics port(s).
 	service, err := metrics.CreateMetricsService(ctx, cfg, servicePorts)
 	if err != nil {
-		log.Info("Could not create metrics Service", "error", err.Error())
+		return fmt.Errorf("failed to create metrics service: %w", err)
```
markusthoemmes, line 57: @skonto did we ignore this error on purpose?

skonto: The idea was that monitoring will be always there on OCP and we could move on if it was not setup (I used to have code that detected if monitoring is available like the coreos packaged does).

markusthoemmes: Hm, sorry, not parsing that. Why'd we ignore an error during creation of the metrics service (a k8s service) and not all other errors then?

skonto: Wait error is returned a few lines bellow.

markusthoemmes: Well I didn't change anything about the flow to main, I think this purely makes the error handling a bit more "narrow" and ignores one somewhat common error that isn't really an error.

markusthoemmes:
/retest

Infra.

knative-operator/pkg/common/service_monitor.go
skonto: Ok makes sense but afaik there was a discussion that we should never hit this.

PR #629, opened 2020-10-26
Reduce reliance on operator-sdk libraries

markusthoemmes:
This is one more step towards bumping our osdk/controller-runtime versions. By reducing the surface area of osdk we're also reducing the contentious area and thus have less potential for conflict.


matzew:
/lgtm
/approval

I appreciate the removal of the external library. but the new code reads all the same..

```
	namespace := os.Getenv(common.NamespaceEnvKey)
	if namespace == "" {
		return errors.New("NAMESPACE not provided via environment")
	}
```

I guess... not really worth moving that to some function... - just thinking out loud

matzew (approved):
/lgtm
/approve

PR #630, opened 2020-10-26
Drop Openshift detection, we're always on Openshift

markusthoemmes:
As per title. Less code is better code.


matzew (approved):
/lgtm
/approve

PR #632, opened 2020-10-26
Use Schemes in a more principled and less redundant way

markusthoemmes:
While upgrading controller-runtime, I noticed we are very picky with how our schemes are setup in tests. I don't think we need to be that picky, as tests should always assume the same scheme as the "main" scheme used by the operator anyway. This sets all tests up to do that, rather than being picky.

/assign @aliok


aliok (approved):
/lgtm
/approve

PR #633, opened 2020-10-26
Timeout func, less verbose, more user friendly

cardil:
Changes:
 * Do not write a test command for each check, do it once
 * Put a dot on each interval
 * Make a default internal of 2 instead of 5 sec to make checks slightly faster

### DIFF
--- a/hack/helloworld.sh
+++ b/hack/helloworld.sh
@@ -1,14 +1,19 @@
 #!/usr/bin/env bash
 
-NAME=hello
-TARGET=${USER:-world}
+set -Eeuo pipefail
+
+# shellcheck disable=SC1091,SC1090
+source "$(dirname "${BASH_SOURCE[0]}")/lib/__sources__.bash"
+
+NAME="${NAME:-hello}"
+TARGET="${USER:-world}"
 
 # Create a sample Knative Service
 cat <<EOF | kubectl apply -f -
 apiVersion: serving.knative.dev/v1alpha1
 kind: Service
 metadata:
-  name: $NAME
+  name: ${NAME}
 spec:
   template:
     spec:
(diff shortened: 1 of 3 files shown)


hack/helloworld.sh
```diff
@@ -1,33 +1,38 @@
...
     spec:
       containers:
         - image: gcr.io/knative-samples/helloworld-go
           env:
             - name: TARGET
-              value: $TARGET
+              value: ${TARGET}
+          readinessProbe:
+            httpGet:
+              path: /
```
markusthoemmes, line 27: This change is unnecessary. Knative will add a default probe anyway.

hack/helloworld.sh
```diff
@@ -1,33 +1,38 @@
...
 EOF
 
 # Wait for the Knative Service to be ready
-while output=$(kubectl get ksvc $NAME); do
-  echo "$output"
-  echo $output | grep True >/dev/null && break
-  sleep 2
-done
+timeout 100 "[[ \$(kubectl get ksvc ${NAME} -o \
+jsonpath='{.status.conditions[?(@.type==\"Ready\")].status}') != 'True' ]]"
```
markusthoemmes, line 32: Why not

```
kubectl wait --timeout=5m --for=Ready=True
```

? I suppose we could make use of that more often throughout our scripts.

hack/helloworld.sh
cardil: Is that a default probe? Or default one is socket based instead of http get?

hack/helloworld.sh
cardil: I basically changed that to test the new impl of timeout

hack/helloworld.sh
markusthoemmes: The default one is socket based, but it shouldn't matter in this case.

hack/lib/common.bash
```diff
@@ -15,16 +15,21 @@ function array.join {
 
 # Loops until duration (car) is exceeded or command (cdr) returns non-zero
 function timeout {
-  local seconds timeout interval
-  interval=5
+  local timeout
+  timeout="${1:?Pass a timeout as arg[1]}"
+  interval="${interval:-1}"
```
markusthoemmes, line 20: Does this change the behavior in that there is no interval at all anymore? How do I set it? How does this interop with the `seconds=$(( seconds + interval ))` thing?

cardil: It sets a variable to 1 if it isn't set.

You can override it by calling:

```bash
interval=5 timeout 300 "[[ \$(something --to-check) != 'True' ]]"
```

That will wait for 5 seconds instead of 1.

hack/lib/common.bash
```diff
@@ -15,16 +15,21 @@ function array.join {
...
   seconds=0
-  timeout=$1
   shift
-  while eval $*; do
+  ln=' ' logger.debug "${*} : Waiting until non-zero (max ${timeout} sec.)"
+  while (eval "$*" 2>/dev/null); do
```
mgencur, line 24: I'm curious if this will print the condition and the result during each execution. It doesn't look like that. I find it useful to see the condition and the actual values when it's evaluated, possibly when running with `-x`

cardil: This will print the command on stdout just once, but you will see all values in debug log.

PR #637, opened 2020-10-27
Fix namespace usage in eventing test and touchups

markusthoemmes:
I have no clue why this even passes on master, but it will start failing on newer controller-runtime versions. Apparently the namespaces aren't correctly matched up on the current one, so this fixes that.

Also contains a few touchups to the test in general.

### DIFF
--- a/knative-operator/pkg/controller/knativeeventing/knativeeventing_controller_test.go
+++ b/knative-operator/pkg/controller/knativeeventing/knativeeventing_controller_test.go
@@ -5,9 +5,9 @@ import (
 	"os"
 	"testing"
 
-	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/common"
-
 	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
+	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/apis"
+	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/common"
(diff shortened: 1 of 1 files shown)


aliok (commented):
@markusthoemmes why exactly this test would normally fail? I didn't get that part

knative-operator/pkg/controller/knativeeventing/knativeeventing_controller_test.go
```diff
@@ -61,65 +63,51 @@ func TestEventingReconcile(t *testing.T) {
...
 			}
 			sourceCM := &corev1.ConfigMap{}
-			err = cl.Get(context.TODO(), types.NamespacedName{Name: "grafana-dashboard-definition-knative-eventing-source", Namespace: ns.Name}, sourceCM)
+			err = cl.Get(context.TODO(), types.NamespacedName{Name: "grafana-dashboard-definition-knative-eventing-source", Namespace: dashboardNamespace.Name}, sourceCM)
 			if err != nil {
 				t.Fatalf("get: (%v)", err)
 			}
 			// Check if the eventing service monitors are installed
 			smFilter := &monitoringv1.ServiceMonitor{}
-			err = cl.Get(context.TODO(), types.NamespacedName{Name: "knative-eventing-metrics-broker-filter", Namespace: ns.Namespace}, smFilter)
```
markusthoemmes, line 117: @aliok because this is plain wrong. The `Namespace` field of a namespace is empty. And `ns` is also the wrong namespace.

PR #638, opened 2020-10-27
Replace DeepEquals in tests with Equals/Diff combos

markusthoemmes:
This aids debuggability a **ton**. If a test fails because a resource didn't look as expected, this will now only print the difference of the two to make fixing tests a breeze.

### DIFF
--- a/knative-operator/pkg/controller/knativeserving/knativeserving_controller_test.go
+++ b/knative-operator/pkg/controller/knativeserving/knativeserving_controller_test.go
@@ -5,13 +5,13 @@ import (
 	"os"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/apis"
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/controller/dashboard"
 	configv1 "github.com/openshift/api/config/v1"
 	consolev1 "github.com/openshift/api/console/v1"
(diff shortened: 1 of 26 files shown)


knative-operator/pkg/controller/knativeserving/knativeserving_controller_test.go
```diff
@@ -347,8 +347,8 @@ func TestCustomCertsConfigMap(t *testing.T) {
 					t.Fatalf("Failed to fetch controller: %v", err)
 				}
 
-				if !equality.Semantic.DeepEqual(got, test.outCtrl) {
-					t.Fatalf("ConfigMaps %#v not equal to %#v", got, test.outCtrl)
+				if !cmp.Equal(got, test.outCtrl) {
+					t.Errorf("Deployments not equal, diff: %s", cmp.Diff(got, test.outCtrl))
```
skonto, line 351: good catch

skonto:
@markusthoemmes should we update [these calls](https://github.com/openshift-knative/serverless-operator/blob/84142bbe9c25a328c4393e2eaea1dd36901e7d46/knative-operator/pkg/controller/knativekafka/images_test.go#L251) too?

markusthoemmes:
@skonto did the latter one, good catch. I'll leave the former ones as-is, as these are copied tests AFAIK. We might want to look at deduping the entire file instead.

markusthoemmes:
/retest

Infra.

PR #639, opened 2020-10-27
Fix KnativeKafka CRD validation

aliok:
I remember seeing this change in @lberk 's PR. 
He rebased it right but I told him that the rebase was wrong. I was wrong :(

Back to the state at https://github.com/openshift-knative/serverless-operator/pull/510


knative-operator/deploy/crds/operator_v1alpha1_knativekafka_crd.yaml
```diff
@@ -44,7 +47,6 @@ spec:
                     type: boolean
                 required:
                 - enabled
-                - bootstrapServers
```
aliok, line 47: We already have a check in webhook. We reject the CR if `spec.channel.enabled=true` and `bootstrapServers` is empty

matzew (approved):
/lgtm
/approve

@aliok   can you also PR that against the 1.11 branch ??

PR #640, opened 2020-10-27
Rework webhook test method to actually send requests

markusthoemmes:
Currently, the webhook's objects are mocked through the decoder interface, but the webhooks don't actually get a request. This replaces the decoder with a generic encoder based on the scheme and instead passes a "proper" request to the webhooks under test.

**Note:** Newer versions of controller-runtime no longer have a decoder interface, so this is prep work to have a much smaller diff once bumping the version.

### DIFF
--- a/knative-operator/pkg/webhook/knativeeventing/webhook_validating_test.go
+++ b/knative-operator/pkg/webhook/knativeeventing/webhook_validating_test.go
@@ -32,12 +32,7 @@ var (
 
 func init() {
 	apis.AddToScheme(scheme.Scheme)
-
-	dec, err := admission.NewDecoder(scheme.Scheme)
-	if err != nil {
-		panic(err)
-	}
-	decoder = dec
+	decoder, _ = admission.NewDecoder(scheme.Scheme)
 }
 
 func TestInvalidNamespace(t *testing.T) {
--- a/knative-operator/pkg/webhook/knativeeventing/webhook_validating_test.go
+++ b/knative-operator/pkg/webhook/knativeeventing/webhook_validating_test.go
@@ -1,4 +1,4 @@
-package knativeeventing_test
+package knativeeventing
 
 import (
 	"context"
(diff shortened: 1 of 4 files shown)


knative-operator/pkg/webhook/knativeeventing/webhook_validating_test.go
```diff
@@ -1,66 +1,77 @@
...
-var ke1 = &eventingv1alpha1.KnativeEventing{
-	ObjectMeta: metav1.ObjectMeta{
-		Name: "ke1",
-	},
-}
-var ke2 = &eventingv1alpha1.KnativeEventing{
-	ObjectMeta: metav1.ObjectMeta{
-		Name: "ke2",
-	},
+	dec, err := admission.NewDecoder(scheme.Scheme)
```
skonto, line 36: It seems that the call will never produce an error:
```
// NewDecoder creates a Decoder given the runtime.Scheme
func NewDecoder(scheme *runtime.Scheme) (types.Decoder, error) {
	return decoder{codecs: serializer.NewCodecFactory(scheme)}, nil
}
```

markusthoemmes: Heh, great catch. Pushed a version that ditches the error handling. I moved it outside of tests since we never expect it to fail so I didn't want to pollute tests. Now it's even better.

knative-operator/pkg/webhook/testutil/testutil.go
```diff
@@ -1,19 +1,25 @@
...
-	return &configv1.ClusterVersion{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: "version",
-		},
-		Status: configv1.ClusterVersionStatus{
-			Desired: configv1.Update{
-				Version: version,
+// RequestFor generates an admission request for the given object.
+func RequestFor(obj runtime.Object) (types.Request, error) {
+	b, err := json.Marshal(obj)
```
skonto, line 13: s/b/bytes?

knative-operator/pkg/webhook/knativeserving/webhook_validating_test.go
```diff
@@ -1,66 +1,77 @@
...
-var ks1 = &servingv1alpha1.KnativeServing{
-	ObjectMeta: metav1.ObjectMeta{
-		Name: "ks1",
-	},
-}
-var ks2 = &servingv1alpha1.KnativeServing{
-	ObjectMeta: metav1.ObjectMeta{
-		Name: "ks2",
-	},
+	dec, err := admission.NewDecoder(scheme.Scheme)
```
skonto: same as above.

knative-operator/pkg/webhook/testutil/testutil.go
markusthoemmes: bytes is a stdlib package name, so I'd rather not use it.

knative-operator/pkg/webhook/knativeeventing/webhook_validating_test.go
```diff
@@ -1,66 +1,72 @@
...
 }
 
 func TestInvalidNamespace(t *testing.T) {
 	os.Clearenv()
 	os.Setenv("REQUIRED_EVENTING_NAMESPACE", "knative-eventing")
+
 	validator := Validator{}
-	validator.InjectDecoder(&mockDecoder{ke1})
-	result := validator.Handle(context.TODO(), types.Request{})
+	validator.InjectDecoder(decoder)
```
skonto, line 43: This also returns an error but is always nil, weird...

markusthoemmes: Yeah, we usually just don't care in tests :shrug:

knative-operator/pkg/webhook/knativeeventing/webhook_validating_test.go
```diff
@@ -1,66 +1,72 @@
...
+	if err != nil {
+		t.Fatalf("Failed to generate a request for %v: %v", ke1, err)
+	}
 
-func (mock *mockDecoder) Decode(_ types.Request, obj runtime.Object) error {
-	if p, ok := obj.(*eventingv1alpha1.KnativeEventing); ok {
-		*p = *mock.ke
+	result := validator.Handle(context.Background(), req)
+	if result.Response.Allowed {
+		t.Errorf("Too many KnativeEventings: %v", result.Response)
```
skonto, line 70: We could aggregate response reasons and check them. Validation checks the namespace and the uniqueness of the namespace. Here since no required env var is set it will return true for the invalid namespace check and will return `Allowed=false` with reason text: "Only one KnativeEventing allowed per namespace". 
If something changes in the implementation this test may pass for the wrong reason.

markusthoemmes: I agree. Would you be fine if we did that in a separate PR? This refactor is on the blocking path for the version bumping, so I'd rather not conflate the concerns.

skonto: +10

PR #649, opened 2020-10-28
Drop redundant config and client creation

markusthoemmes:
As per title, it seems unnecessary to create clients and configs twice here. Maybe I'm missing something though.


skonto:
@markusthoemmes I had this part set explicitly like that because some api calls were failing at that point. I didnt check deeper back then with the old ctr runtime being used. It works now so let's remote it :)

skonto:
Tests fail. Let's re-test.

PR #650, opened 2020-10-28
[SRVCOM-1084] Update all dependencies of knative-operator to very recent ones

markusthoemmes:
Finally. This updates the knative-operator to have very recent versions of all dependencies, most notably:

- Align with Knative 0.18
- Align with Kubernetes 1.18 and Openshift 4.6
- Bring in the newest manifestival releases
- Bump controller-runtime to **0.6.3** (their latest)
- Bump operator-sdk to **0.19.4** (their last before 1.0, the 1.0 line contains breaking changes to how we handle metrics, I think we should look at bumping further separately)

---

For your reviewing pleasure, I put some effort into cutting the commits in a sensible way, so they can be reviewed one after the other without too much hassle.

1. Updates `go.mod` and brings in a sensible `update-deps.sh` file to update all of our dependencies consistently. Also drops the now unnecessary patches.
2. Runs `update-deps.sh --upgrade`, so it contains all the vendor stuff. You can skip reviewing this.
3. Fixes the webhook implementations to implement the new APIs of the updated controller-runtime. This is fairly small in scope as I already preloaded the necessary fixes to tests.
4. Fixes behavior changes in tests. The newer fake client actually bumps `resourceVersion` in objects, so the tests now need to respect that to some extent.
5. Fixes some deprecated imports and very small API changes between the versions.
6. Adjusts the `main.go` file, most notably hooks up the webhooks in their "new" way and sets the logger up in a similar way as it was before (the default logger prints millis since epoch... ugh).

**Note:** This doesn't yet uses the new APIs from controller-runtime to create webhooks/controllers in an "easier" way. I wanted to explore them afterwards and keep the diff here as small as possible to limit risk and moving parts.


markusthoemmes:
/retest

Infra :)

knative-operator/cmd/manager/main.go
```diff
@@ -75,7 +69,6 @@ func main() {
 	// Create a new Cmd to provide shared dependencies and start components
 	mgr, err := manager.New(cfg, manager.Options{
 		Namespace:          "", // The serverless operator always watches all namespaces.
-		MapperProvider:     restmapper.NewDynamicRESTMapper,
```
markusthoemmes, line 78: This is now controller-runtime's default rest mapper. See https://github.com/kubernetes-sigs/controller-runtime/blob/ac380d61764a160b32946e606b0c9ecd2834e3e8/pkg/manager/manager.go#L411-L415

knative-operator/cmd/manager/main.go
```diff
@@ -7,55 +7,49 @@ import (
...
+	prodConf := zap.NewProductionEncoderConfig()
+	prodConf.EncodeTime = zapcore.ISO8601TimeEncoder
+	logf.SetLogger(zapr.New(zapr.Encoder(zapcore.NewJSONEncoder(prodConf))))
+}
 
+func main() {
 	// Add flags registered by imported packages (e.g. glog and
 	// controller-runtime)
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
-	pflag.Set("zap-time-encoding", "iso8601")
```
markusthoemmes, line 45: The flags have been dropped in the new release, so I manually construct an equivalent log encoder above. I slightly changed how the logger is setup to be sure it's initialized properly.

markusthoemmes:
/test 4.6-upstream-e2e-aws-ocp-46

Eventing flake.

knative-operator/hack/update-deps.sh
```diff
@@ -1,24 +1,65 @@
...
+OCP_VERSION="release-4.6"
+
+# The list of dependencies that we track at HEAD and periodically
+# float forward in this repository.
+FLOATING_DEPS=(
+  "github.com/openshift/api@${OCP_VERSION}"
+
+  "knative.dev/operator@${KN_VERSION}"
+  "knative.dev/pkg@${KN_VERSION}"
+  "knative.dev/serving@${KN_VERSION}"
```
skonto, line 25: How about `knative.dev/eventing-contrib` and `knative.dev/eventing`?

markusthoemmes: Neither are required in `go.mod` (nor vendored)

knative-operator/cmd/manager/main.go
```diff
@@ -7,55 +7,49 @@ import (
...
+	log                       = logf.Log.WithName("cmd")
 )
-var log = logf.Log.WithName("cmd")
 
-func main() {
-	// Add the zap logger flag set to the CLI. The flag set must
-	// be added before calling pflag.Parse().
-	pflag.CommandLine.AddFlagSet(zap.FlagSet())
+func init() {
+	prodConf := zap.NewProductionEncoderConfig()
```
skonto, line 42: Do we need a `NewDevelopmentConfig()` to make timestamps more readable etc for the developer enforcing a rule about timestamp format?
Also in production I think its useful either to enforce `UTC` or use local time + `offset from UTC`.

knative-operator/cmd/manager/main.go
```diff
@@ -97,11 +90,21 @@ func main() {
...
-	if err := webhook.AddToManager(mgr); err != nil {
-		log.Error(err, "")
-		os.Exit(1)
-	}
+	// Setup all Webhooks
+	hookServer := mgr.GetWebhookServer()
+	hookServer.Port = 9876
+	hookServer.CertDir = "/apiserver.local.config/certificates"
+	hookServer.KeyName = "apiserver.key"
+	hookServer.CertName = "apiserver.crt"
```
skonto, line 98: Do we care about `hookServer.ClientCAName` in any case for api server validation?

knative-operator/cmd/manager/main.go
markusthoemmes: Not parsing, didn't I configure the timestamp format below this line? The current config should be equivalent to what it was before I think. If we want additional changes, I'd do them separately and to all of our pods to be the same.

knative-operator/cmd/manager/main.go
markusthoemmes: Nope, don't think so.

knative-operator/cmd/manager/main.go
markusthoemmes: I think we're doing what you're asking already. Here's an example log of this build

```
{"level":"info","ts":"2020-10-28T16:29:05.323Z","logger":"knative.openshift.health dashboard.mf","msg":"Parsing manifest"}
```

We log humanly readable (ISO8601) and locked at UTC (signified by the timestamp ending in a `Z`)

PR #653, opened 2020-10-30
[release-1.11] Align w/ the KnativeServing and KnativeEventing

matzew:

We do it like `KnativeServing` and `KnativeEventing`.

Fix for: https://issues.redhat.com/projects/SRVKE/issues/SRVKE-616

### DIFF
--- a/olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
+++ b/olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
@@ -32,10 +32,10 @@ metadata:
           },
           "spec": {
             "source": {
-              "enabled": "false"
+              "enabled": false
             },
             "channel": {
-              "enabled": "false",
+              "enabled": false,
               "bootstrapServers": "REPLACE_WITH_COMMA_SEPARATED_KAFKA_BOOTSTRAP_SERVERS"
             }
           }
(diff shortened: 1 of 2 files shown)


olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
```diff
@@ -31,11 +31,13 @@ metadata:
             "name": "knative-kafka"
           },
           "spec": {
-            source:
-              enabled: true
-            channel:
-              enabled: true
-              bootstrapServers: REPLACE_WITH_COMMA_SEPARATED_KAFKA_BOOTSTRAP_SERVERS
+            "source": {
+              "enabled": "false"
```
maschmid, line 35: Are you sure it should be a string "false", and not a boolean false?

matzew: @maschmid JSON fu :-) 

Before I sent the PR I looked here, how we used to do that in the past for the `alm-examples`:
https://github.com/openshift-knative/knative-serving-operator/blob/v0.8.0/deploy/olm-catalog/knative-serving-operator/0.8.0/knative-serving-operator.v0.8.0.clusterserviceversion.yaml#L5

that uses `"true"` syntax too :man_shrugging:

matzew: and `"false"` is also twice in there.

So, we have done strings for booleans on this dialect before

PR #657, opened 2020-11-02
[SRVKS-187] Add telemetry support

skonto:
- Monitoring is failing on master. Fixes the following error:
```
{"level":"error","ts":"2020-11-02T10:27:53.126Z","logger":"cmd","msg":"Failed to start monitoring","error":"the cache is not started, can not read objects","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/src/github.com/openshift-knative/serverless-operator/knative-operator/vendor/github.com/go-logr/zapr/zapr.go:132\nmain.main\n\t/go/src/github.com/openshift-knative/serverless-operator/knative-operator/cmd/manager/main.go:110\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:203"}
```
The recommendation at this early stage is to use our own [client](https://github.com/jaegertracing/jaeger-operator/issues/809#issue-532751170).

- Adds telemetry support according to  SRVKS-187. Metrics exposed are shown next:
image
For the above example here are the actual resources:
```
$ oc get revisions -n knative-serving
NAME           CONFIG NAME   K8S SERVICE NAME   GENERATION   READY   REASON
kn-cli-r5vr2   kn-cli        kn-cli-r5vr2       1            True    
$ oc get revisions
NAME                  CONFIG NAME     K8S SERVICE NAME      GENERATION   READY   REASON
event-display-8tdww   event-display   event-display-8tdww   1            True    
 $ oc get revisions -n apisource
NAME                  CONFIG NAME     K8S SERVICE NAME      GENERATION   READY   REASON
event-display-nphxs   event-display   event-display-nphxs   1            True    

```

Note: PR i splitted in different commits for easier review. You can skip vendor.

/cc @markusthoemmes @matzew @s-urbaniak @openshift-team-monitoring


knative-operator/hack/update-deps.sh
```diff
@@ -20,6 +20,7 @@ OCP_VERSION="release-4.6"
 FLOATING_DEPS=(
   "github.com/openshift/api@${OCP_VERSION}"
 
+  "knative.dev/eventing@${KN_VERSION}"
```
markusthoemmes, line 23: please add eventing-contrib too as it's now part of go.mod as well.

markusthoemmes:
> Monitoring is failing on master. Fixes the following error:

Please send a separate PR just for this fix. Can we please add E2E tests of some sorts that would catch something like this there too?

knative-operator/pkg/common/telemetry/telemetry_controller.go
```diff
@@ -0,0 +1,122 @@
...
+
+// add adds a new Controller to mgr with r as the reconcile.Reconciler
+func add(mgr manager.Manager, r reconcile.Reconciler) error {
+	// Create a new controller
+	c, err := controller.New("telemetry-resources-discovery-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+	for _, tp := range typesAndMetrics {
+		if err := c.Watch(&source.Kind{Type: tp}, &handler.EnqueueRequestForObject{}, skipUpdatePredicate{}, updateMetricsDeletePredicate{}, updateMetricsCreatePredicate{}); err != nil {
```
markusthoemmes, line 53: Seems like we only need the watch functionality? Can we add that in a different file that doesn't look like a controller to make that a bit more obvious?

knative-operator/pkg/common/telemetry/telemetry_controller.go
```diff
@@ -0,0 +1,122 @@
...
+func (skipUpdatePredicate) Update(_ event.UpdateEvent) bool {
+	return false
+}
+
+type updateMetricsCreatePredicate struct {
+	predicate.Funcs
+}
+
+func (updateMetricsCreatePredicate) Create(e event.CreateEvent) bool {
+	matchAndUpdateMetric(e.Object, inc)
```
markusthoemmes, line 89: I'd change this a bit to be more pluggable, like:

```
metric := getMetricFor(e.Object)
metric.Inc()
```

This makes the `op` type unnecessary and composes better imo.

skonto:
@markusthoemmes yes that is my plan to add integration tests. We had this but due to lack of readiness probe they were not reliable. I will need to add the probe thingy first.

knative-operator/hack/update-deps.sh
skonto: yeah sure.

knative-operator/pkg/common/telemetry/telemetry_metrics.go
```diff
@@ -0,0 +1,25 @@
...
+		prometheus.GaugeOpts{
+			Name: "serverless_telemetry",
+			Help: "Reports number of serverless resources for telemetry",
+		},
+		[]string{"type"},
+	)
+	ServicesG  = serverlessTelemetryG.WithLabelValues("service")
+	RoutesG    = serverlessTelemetryG.WithLabelValues("route")
+	RevisionsG = serverlessTelemetryG.WithLabelValues("revision")
+	SourcesG   = serverlessTelemetryG.WithLabelValues("source")
```
markusthoemmes, line 19: How do these differ from the metrics that we surface via `GenerateAndServeCRMetrics`?

markusthoemmes:
@skonto you can investigate the new c-r implementations regarding this. I believe they have "native" readiness endpoint support, but I haven't yet looked into that.

knative-operator/pkg/common/telemetry/telemetry_controller.go
skonto: You mean in an existing file?

knative-operator/pkg/common/telemetry/telemetry_metrics.go
skonto: I tried that framework before coming up with this simplified solution. Here are the downsides:
- It requires a cfg object passed by a controller which makes it hard to test.
- it adds a redundant label for the CR type that distinguishes the type of metric [here](https://github.com/operator-framework/operator-sdk/blob/125d0dfcc71fef4f9d7e2a42b1354cb79ffdee03/pkg/kube-metrics/metrics.go#L100), increasing the carnality (max is 10 for telemetry).

knative-operator/pkg/common/telemetry/telemetry_controller.go
markusthoemmes: No necessarily, but if we are not going to make this into a controller, let's not make it look like one.

markusthoemmes, line 53: Can we fold all the predicates into just one handler?

knative-operator/pkg/common/telemetry/telemetry_controller.go
```diff
@@ -0,0 +1,122 @@
...
+
+// blank assignment to verify that ReconcileResourcesForTelemetry implements reconcile.Reconciler
+var _ reconcile.Reconciler = &ReconcileResourcesForTelemetry{}
+
+// ReconcileResourcesForTelemetry reconciles a source deployment object
+type ReconcileResourcesForTelemetry struct {
+	// This client, initialized using mgr.Client() above, is a split client
+	// that reads objects from the cache and writes to the apiserver
+	client client.Client
+	scheme *runtime.Scheme
```
markusthoemmes, line 68: Are these used?

knative-operator/pkg/common/telemetry/telemetry_controller.go
```diff
@@ -0,0 +1,122 @@
...
+
+// newReconciler returns a new reconcile.Reconciler
+func newReconciler(mgr manager.Manager) reconcile.Reconciler {
+	return &ReconcileResourcesForTelemetry{client: mgr.GetClient(), scheme: mgr.GetScheme()}
+}
+
+// add adds a new Controller to mgr with r as the reconcile.Reconciler
+func add(mgr manager.Manager, r reconcile.Reconciler) error {
+	// Create a new controller
+	c, err := controller.New("telemetry-resources-discovery-controller", mgr, controller.Options{Reconciler: r})
```
markusthoemmes, line 48: What happens here if we don't provide a Reconciler in the options?

markusthoemmes: nvm, throws an error.

skonto: Let me check to see how it will behave.

skonto: @markusthoemmes the most limited example I see is [this](https://github.com/kubernetes-sigs/controller-runtime/blob/af24f3b6f46793c3210224bdd00c2593970942f1/pkg/controller/example_test.go#L44-L48) one.

markusthoemmes: Nice, yeah let's make use of that!

knative-operator/pkg/common/telemetry/telemetry_controller.go
markusthoemmes: Most likely you won't need this one.

knative-operator/pkg/common/telemetry/telemetry_controller.go
skonto: Reconciler [needs](https://github.com/kubernetes-sigs/controller-runtime/blob/af24f3b6f46793c3210224bdd00c2593970942f1/pkg/controller/controller.go#L92) to be specified.

knative-operator/pkg/common/telemetry/telemetry_controller.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func (metricsPredicate) Delete(e event.DeleteEvent) bool {
+	if metric := getMetricFor(e.Object); metric != nil {
+		(*metric).Dec()
+	}
+	return true
+}
+
+func getMetricFor(obj runtime.Object) *prometheus.Gauge {
```
markusthoemmes, line 73: Why return a pointer if we dereference it above anyway? :thinking:

knative-operator/pkg/common/telemetry/telemetry_controller.go
```diff
@@ -0,0 +1,85 @@
...
+		(*metric).Inc()
+	}
+	return true
+}
+
+func (metricsPredicate) Delete(e event.DeleteEvent) bool {
+	if metric := getMetricFor(e.Object); metric != nil {
+		(*metric).Dec()
+	}
+	return true
```
markusthoemmes, line 70: I think we can just return false everywhere, so the reconciler is just never called.

knative-operator/pkg/common/telemetry/telemetry_controller.go
skonto: It covers the case where the objectype is not the one we want.

knative-operator/pkg/common/telemetry/telemetry_controller.go
skonto: Ok

skonto:
I created 1000 sources at once in order to see the behavior of the telemetry watcher:
image
I didnt notice any issues. Finally the metric was updated to the right value.

knative-operator/pkg/common/telemetry/telemetry_controller.go
skonto: Can be simplified though its a struct anyway, fixed it.

skonto:
@markusthoemmes I updated the PR but see errors [here](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-knative_serverless-operator/657/pull-ci-openshift-knative-serverless-operator-master-4.6-upstream-e2e-aws-ocp-46/1324020938789883904/artifacts/upstream-e2e-aws-ocp-46/gather-extra/pods/openshift-serverless_knative-openshift-ffb8c95f-6lp87_knative-openshift.log).
```
controller","error":"failed to wait for health-controller caches to sync: cache did not sync","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/src/github.com/openshift-knative/serverless-operator/knative-operator/vendor/github.com/go-logr/zapr/zapr.go:132\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func1\n\t/go/src/github.com/openshift-knative/serverless-operator/knative-operator/vendor/sigs.k8s.io/controller-
```
The resources are added to the scheme. This is something I didnt see locally so probably c-r is not capable to handle CRs not yet installed.

knative-operator/pkg/common/telemetry/telemetry_controller_test.go
```diff
@@ -0,0 +1,108 @@
...
+	}, {
+		name: "sinkbindingsource",
+		obj:  SinkBindingSource.DeepCopyObject(),
+	}, {
+		name: "kafkasource",
+		obj:  kafkaSource.DeepCopyObject(),
+	}}
+	ret = []metricCase{}
+	for _, v := range objects {
+		ret = append(ret, metricCase{
```
markusthoemmes, line 91: nit: You can make this all one `append` call, like

```golang
ret = append(ret, metricCase{
  name: fmt.Sprintf("create a %s", v.name),
  event: event.CreateEvent{Object: v.obj},
  expectedMetricValue: 1,
}, metricCase{
  name: fmt.Sprintf("create a %s", v.name),
  event: event.CreateEvent{Object: v.obj},
  expectedMetricValue: 1,
})

File: knative-operator/pkg/common/telemetry/telemetry_controller_test.go [hunk 10]
```diff
@@ -0,0 +1,108 @@
...
+)
+
+var (
+	service           = servingv1.Service{}
+	route             = servingv1.Route{}
+	revision          = servingv1.Revision{}
+	pingSource        = eventingsourcesv1beta1.PingSource{}
+	apiServerSource   = eventingsourcesv1beta1.ApiServerSource{}
+	SinkBindingSource = eventingsourcesv1beta1.SinkBinding{}
+	kafkaSource       = kafkasourcev1beta1.KafkaSource{}
```
markusthoemmes, line 23: If you make all of those pointers, you won't need the `DeepCopyObject` calls below, as they'll then fit into the `runtime.Object` interface themselves.

knative-operator/pkg/common/telemetry/telemetry_controller_test.go
```diff
@@ -0,0 +1,108 @@
...
+	}, {
+		name: "apiserversource",
+		obj:  apiServerSource.DeepCopyObject(),
+	}, {
+		name: "sinkbindingsource",
+		obj:  SinkBindingSource.DeepCopyObject(),
+	}, {
+		name: "kafkasource",
+		obj:  kafkaSource.DeepCopyObject(),
+	}}
```
markusthoemmes, line 88: nit: You could replace this with globally using a map, i.e:

```golang
objects := map[string]runtime.Object{
  "service": &servingv1.Service{}
  ...
}
```

I'd even move the entire initializiation here if you don't need the objects globally.

knative-operator/pkg/common/telemetry/telemetry_controller_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/event"
+)
+
+var (
+	service           = servingv1.Service{}
+	route             = servingv1.Route{}
+	revision          = servingv1.Revision{}
+	pingSource        = eventingsourcesv1beta1.PingSource{}
+	apiServerSource   = eventingsourcesv1beta1.ApiServerSource{}
+	SinkBindingSource = eventingsourcesv1beta1.SinkBinding{}
```
markusthoemmes, line 22: Why export?

knative-operator/pkg/common/telemetry/telemetry_controller_test.go
```diff
@@ -0,0 +1,108 @@
...
+		case event.CreateEvent:
+			mp.Create(tc.event.(event.CreateEvent))
+			metric = getMetricFor(tc.event.(event.CreateEvent).Object)
+		case event.DeleteEvent:
+			mp.Delete(tc.event.(event.DeleteEvent))
+			metric = getMetricFor(tc.event.(event.DeleteEvent).Object)
+		case event.UpdateEvent:
+			mp.Update(tc.event.(event.UpdateEvent))
+			metric = getMetricFor(tc.event.(event.UpdateEvent).ObjectOld)
+		}
```
markusthoemmes, line 49: nit: Consider rewriting this with type assertions maybe to avoid the "double casting", like:

```golang
if create, ok := tc.event.(event.CreateEvent); ok {
  mp.Create(create)
  metric = getMetricFor(create.Object)
} else if delete, ok := tc.event.(event.DeleteEvent); ok {
 ...
}
```

knative-operator/pkg/common/telemetry/telemetry_controller_test.go
```diff
@@ -0,0 +1,108 @@
...
+		if metric == nil {
+			t.Fatal("Cannot get metric")
+		}
+		err := metric.Write(&dto)
+		if err != nil {
+			t.Fatal("Cannot write metric:", err)
+		}
+		if *dto.Gauge.Value != tc.expectedMetricValue {
+			t.Errorf("Got = %v, want: %v for event: %v", *dto.Gauge.Value, tc.expectedMetricValue, n)
```
markusthoemmes, line 58: Is the `n` really helpful here? Seems like we'd actually want to print `tc` maybe?

markusthoemmes (commented):
Mostly nits to make code a bit less repetitive. Looks good otherwise!

skonto:
I will try to use polling instead to avoid the c-r limitations.

knative-operator/pkg/common/telemetry/telemetry_controller_test.go
skonto: Initiall I had map but iterating on it didnt preserve the order.

skonto:
@markusthoemmes I updated the PR addressing your comments but also covering the failing case where CRD was not found (didnt go for the polling solution because of the cpu overhead it may cause). the new design adds an unmanaged controller per component following the example [here](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/controller/example_test.go).
I cover the lifecycle of each component by starting and stopping the corresponding controller accordingly.
If I remove a specific resource I get in the logs:
```
E1105 15:25:41.372782       1 reflector.go:178] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:224: Failed to list *v1beta1.PingSource: the server could not find the requested resource (get pingsources.sources.knative.dev)
```
which is the same log entry we get in the controller in the eventing namespace. This is acceptable. The user remove the pingsource crd for some reason etc.
In another scenario if the user removes a component eg. Eventing and then re-installs it then he gets in the logs:
```
{"level":"info","ts":"2020-11-05T17:06:38.215Z","logger":"knative.openshift.telemetry","msg":"Starting telemetry for:","component":"Serving"}
{"level":"info","ts":"2020-11-05T17:07:55.466Z","logger":"knative.openshift.telemetry","msg":"Starting telemetry for:","component":"Eventing"}
{"level":"info","ts":"2020-11-05T17:21:22.555Z","logger":"knative.openshift.telemetry","msg":"Stopping telemetry for:","component":"Eventing"}
{"level":"info","ts":"2020-11-05T17:23:50.998Z","logger":"knative.openshift.telemetry","msg":"Starting telemetry for:","component":"Eventing"}
```
No watcher error is emitted in this case.

knative-operator/pkg/common/telemetry/telemetry.go
```diff
@@ -0,0 +1,92 @@
...
+	skipTelemetryStoppingFor[component] = false
+	mu.Unlock()
+	return nil
+}
+
+func TryStopTelemetry(component Component) {
+	if os.Getenv(SkipTelemetryEnvVar) != "" {
+		return
+	}
+	mu.Lock()
```
markusthoemmes, line 78: This should add a `defer mu.Unlock()` underneath.

Point in case: You currently have a deadlock if you return early in the diff below.

skonto: Good catch, yes.

skonto:
Updated the implementation and tested the following scenarios:
a) kill the operator pod and check metrics after the leader comes up. Observed metrics come up consistent with their values before kill.
b) removed eventing or serving crds (uninstalling both) and observed metrics after applying again the corresponding crds and after the components came up.

knative-operator/pkg/common/telemetry/telemetry.go
```diff
@@ -0,0 +1,180 @@
...
+	metricsSnapshot = map[string]*sync.Map{
+		"eventing":     {},
+		"serving":      {},
+		"knativeKafka": {},
+	}
+)
+
+// TryStartTelemetry setups telemetry per component either Eventing, KnativeKafka or Serving.
+// When called it assumes that the component has status ready.
+func TryStartTelemetry(c controller.Controller, mgr manager.Manager, stop chan struct{}, component string, api client.Client) error {
```
markusthoemmes, line 32: I would've expected this to be part of a `Telemetry` type that you could store on the reconciler itself. You'd then just call `Start` and `Stop` accordingly.

This would solve two issues in the current code:
1. The pointer-replacement shenanigans with the stop channel below.
2. The skip env var becomes unnecessary as you can initialize the reconciler with a noop telemetry object in the test.

knative-operator/pkg/common/telemetry/telemetry.go
```diff
@@ -0,0 +1,180 @@
...
+func TryStopTelemetry(stop *chan struct{}, component string) {
+	if os.Getenv(SkipTelemetryEnvVar) != "" {
+		return
+	}
+	log.Info("stopping telemetry for:", "component", component)
+	// Stop the telemetry controller
+	// the lock above makes sure we close once and not panic
+	close(*stop)
+	// Can't use a closed channel
+	*stop = make(chan struct{})
```
markusthoemmes, line 66: This is the shenanigans I refer to above.

knative-operator/pkg/common/telemetry/telemetry.go
```diff
@@ -0,0 +1,180 @@
...
+// InitializeAndTakeMetricsSnapshot is used for taking a global snapshot of metrics
+// before we start any telemetry controller. If the operator is restarted
+// client metrics will not reflect the state in the cluster so there is need to restore current state.
+// It should run each time operator is initialized and for each trial to start a Telemetry controller.
+// Snapshot is used for skipping events already counted but also received later after
+// telemetry controllers are started. Metrics are initialized here and kept around as long as the
+// Serverless Operator is running. No metric label is removed when components are removed so that
+// if any component eg. Eventing is removed and CRDs exist, it will show existing CRs correctly.
+// Telemetry should be used in the background so no error is returned here.
+func InitializeAndTakeMetricsSnapshot(component string, api client.Client) {
```
markusthoemmes, line 84: Is this **really** necessary? I suppose we could have slightly skewed metrics for a reporting period if we don't do this (as the controller does its global resync on startup), but is that slight skew really worth the trouble? :thinking:

knative-operator/pkg/common/telemetry/telemetry_ucontroller.go
```diff
@@ -0,0 +1,113 @@
...
+		if !inSnapshot(e.Meta, getComponentFor(e.Object)) { // skip if we have seen this at controller creation time, so we don't count it twice
+			metric.Inc()
+		}
+	}
+	return false
+}
+
+func (metricsPredicate) Delete(e event.DeleteEvent) bool {
+	if metric := getMetricFor(e.Object); metric != nil {
+		if inSnapshot(e.Meta, getComponentFor(e.Object)) {
```
markusthoemmes, line 81: Why don't we store the snapshot alongside the stopChannel in a `Telemtry` type to not need to do this resolution (`getComponentFor`) at all? This is the cardinality breakup we've talked about.

knative-operator/pkg/controller/knativeeventing/knativeeventing_controller.go
```diff
@@ -26,7 +27,14 @@ import (
...
 const finalizerName = "knative-eventing-openshift"
 
-var log = common.Log.WithName("controller")
+var (
+	log           = common.Log.WithName("controller")
+	stopTelemetry = make(chan struct{})
+	// Protects from processing order, if true we should install telemetry
+	// if it is false we need to uninstall it in the next delete stage.
+	// We start by assuming no telemetry is available.
+	shouldInstall = true
```
markusthoemmes, line 36: I don't quite parse the need for this. Again this would be easier if we have a `Telemetry` type encapsulating everything and potentially keeping internal state.

skonto: If you get a delete event first you need to avoid deleting stuff. If there is no chance I can remove it.

knative-operator/pkg/common/telemetry/telemetry.go
skonto: If I create N objects eg 1000 and then restart the operator I get 0. This is a big difference.

knative-operator/pkg/controller/knativeeventing/knativeeventing_controller.go
markusthoemmes: Sure, but it feels quite odd to carry that state in the reconciler itself, seeing as it belongs to the telemetry controller in question.

knative-operator/pkg/common/telemetry/telemetry.go
markusthoemmes: You get that 0 value until the controller has fully resynced, no?

knative-operator/pkg/common/telemetry/telemetry.go
skonto: @markusthoemmes I can try that to improve compaction but I dont see them as issues. The reason I followed this is because I didnt want to add a lot of noise in the reconciler anyway.

knative-operator/pkg/common/telemetry/telemetry_ucontroller.go
skonto: Ok see my comment above I will try to do that way.

knative-operator/pkg/common/telemetry/telemetry.go
markusthoemmes: True, they are not issues per-se, but they are unnecessary obstructions imo. Types are cheap. I think this currently adds more noise to the reconcilers than encapsulating the telemetry controllers in a type would.

skonto:
@markusthoemmes I removed the feature flag, now using a noop obj, moved to using a telemetry type. I tested again by killing eventing, serving etc. I am planning to add integration tests for the monitoring stuff in a feature PR.

knative-operator/pkg/common/telemetry/telemetry.go
```diff
@@ -0,0 +1,186 @@
...
+
+// TryStartTelemetry setups telemetry per component either Eventing, KnativeKafka or Serving.
+// When called it assumes that the component has status ready.
+func (t *Telemetry) TryStartTelemetry(api client.Client, mgr manager.Manager) error {
+	log.Info("starting telemetry for:", "component", t.name)
+	// Initialize metrics before we start the corresponding controller.
+	// There is a tiny window to miss events here, but should be ok for telemetry purposes.
+	t.InitializeAndTakeMetricsSnapshot(api)
+	// Start our controller in a goroutine so that we do not block.
+	if t.tc != nil { // meant to allow nop objects in tests
```
markusthoemmes, line 48: Shouldn't this be a `if t == nil { return }` at the beginning of the function?

knative-operator/pkg/common/telemetry/telemetry.go
```diff
@@ -0,0 +1,186 @@
...
+func NewTelemetry(name string, mgr manager.Manager, objects []runtime.Object) (*Telemetry, error) {
+	t := &Telemetry{}
+	t.name = name
+	t.stop = make(chan struct{})
+	tc, err := newTelemetryController(name, objects, mgr, t)
+	if err != nil {
+		return nil, err
+	}
+	t.tc = tc
+	return t, nil
```
markusthoemmes, line 37:
```suggestion
	return &Telemetry{
	  name: name,
	  stop: make(chan struct{}),
	  tc: tc
}, nil
```

knative-operator/pkg/common/telemetry/telemetry.go
```diff
@@ -0,0 +1,186 @@
...
+		if err := api.List(context.TODO(), routeList); err == nil {
+			routesG.Set(float64(len(routeList.Items)))
+			for _, obj := range routeList.Items {
+				t.addToSnaphost(obj.GetObjectMeta())
+			}
+		}
+	}
+}
+
+func (t *Telemetry) addToSnaphost(obj metav1.Object) {
```
markusthoemmes, line 157:
```suggestion
func (t *Telemetry) addToSnapshot(obj metav1.Object) {
```

knative-operator/pkg/common/telemetry/telemetry.go
```diff
@@ -0,0 +1,186 @@
...
+	if obj == nil {
+		return
+	}
+	t.metricsSnapshot.Store(types.NamespacedName{
+		Namespace: obj.GetNamespace(),
+		Name:      obj.GetName(),
+	}, true)
+}
+
+func (t *Telemetry) deleteFromSnaphost(obj metav1.Object) {
```
markusthoemmes, line 167:
```suggestion
func (t *Telemetry) deleteFromSnapshot(obj metav1.Object) {
```

knative-operator/pkg/common/telemetry/telemetry.go
```diff
@@ -0,0 +1,186 @@
...
+
+func (t *Telemetry) inSnapshot(obj metav1.Object) (ok bool) {
+	if obj == nil {
+		return false
+	}
+	_, ok = t.metricsSnapshot.Load(types.NamespacedName{
+		Namespace: obj.GetNamespace(),
+		Name:      obj.GetName(),
+	})
+	return
```
markusthoemmes, line 185: please don't use anonymous returns. Adding `return ok` makes this a lot clearer imo.

knative-operator/pkg/common/telemetry/telemetry_ucontroller.go
```diff
@@ -0,0 +1,100 @@
...
+	EventingObjects = []runtime.Object{
+		&eventingsourcesv1beta1.PingSource{},
+		&eventingsourcesv1beta1.ApiServerSource{},
+		&eventingsourcesv1beta1.SinkBinding{},
+	}
+
+	KnativeKafkaObjects = []runtime.Object{
+		&kafkasourcev1beta1.KafkaSource{},
+	}
+)
```
markusthoemmes, line 37: These shouldn't live here but in the respectivecontrollers.

knative-operator/pkg/common/telemetry/telemetry.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	return nil
+}
+
+// TryStopTelemetry stops telemetry per component either Eventing, KnativeKafka or Serving
+// When called it assumes that we are reconciling a deletion event.
+func (t *Telemetry) TryStopTelemetry() {
+	log.Info("stopping telemetry for:", "component", t.name)
+	// Stop the telemetry controller
+	// the lock above makes sure we close once and not panic
```
markusthoemmes, line 69: Which lock?

knative-operator/pkg/common/telemetry/telemetry_ucontroller.go
```diff
@@ -0,0 +1,100 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+var (
+	ServingObjects = []runtime.Object{
+		&servingv1.Service{},
+		&servingv1.Revision{},
+		&servingv1.Route{},
```
markusthoemmes, line 25: Why no `Configuration`?

knative-operator/pkg/common/telemetry/telemetry_ucontroller.go
```diff
@@ -0,0 +1,100 @@
...
+
+	KnativeKafkaObjects = []runtime.Object{
+		&kafkasourcev1beta1.KafkaSource{},
+	}
+)
+
+// newTelemetryController creates an unmanaged controller for watching Telemetry resources
+func newTelemetryController(name string, objects []runtime.Object, mgr manager.Manager, telemetry *Telemetry) (*controller.Controller, error) {
+	// Create a new controller
+	c, err := controller.NewUnmanaged(fmt.Sprintf("telemetry-resources-%s-controller-%s", name, uuid.New().String()), mgr, controller.Options{
```
markusthoemmes, line 42: What do we gain by adding the uuid in here?

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -38,6 +40,10 @@ var (
 	role              = mf.Any(mf.ByKind("ClusterRole"), mf.ByKind("Role"))
 	rolebinding       = mf.Any(mf.ByKind("ClusterRoleBinding"), mf.ByKind("RoleBinding"))
 	roleOrRoleBinding = mf.Any(role, rolebinding)
+	// Protects from processing order, if true we should install telemetry
+	// if it is false we need to uninstall in the next delete stage.
+	// We start by assuming no telemetry is available.
+	shouldInstallTelemetry = true
```
markusthoemmes, line 46: This should move inside the telemetry implementation.

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -96,6 +103,12 @@ func add(mgr manager.Manager, r *ReconcileKnativeKafka) error {
 		}
 	}
 
+	t, err := telemetry.NewTelemetry("knativeKafka", mgr, telemetry.KnativeKafkaObjects)
+	if err != nil {
+		return err
+	}
+	r.telemetry = t
```
markusthoemmes, line 110: This should be part of `newReconciler`

knative-operator/pkg/common/telemetry/telemetry.go
```diff
@@ -0,0 +1,186 @@
...
+// telemetry controllers are started. Metrics are initialized here and kept around as long as the
+// Serverless Operator is running. No metric label is removed when components are removed so that
+// if any component eg. Eventing is removed and CRDs exist, it will show existing CRs correctly.
+// Telemetry should be used in the background so no error is returned here.
+func (t *Telemetry) InitializeAndTakeMetricsSnapshot(api client.Client) {
+	switch t.name {
+	case "eventing":
+		sourcesG = serverlessTelemetryG.WithLabelValues("source")
+		pingSourceList := &eventingsourcesv1beta1.PingSourceList{}
+		if err := api.List(context.TODO(), pingSourceList); err == nil {
```
markusthoemmes, line 95: I wonder if we should just always use a List operation to gather metrics then. AFAIK, they should be coming from cache anyway (with the `Watch` being setup), so what if we don't do an `Add/Dec` operation on individual changes but instead do a `Set` operation on each individual change by listing everything. If everything comes from cache, there's 0 cost to that and it'd make the "snapshot" obsolete.

knative-operator/pkg/common/telemetry/telemetry.go
skonto: It could be but it should be a bug if this happens. At least that is the common practice about whether to check pointers in receiver methods.

knative-operator/pkg/common/telemetry/telemetry_ucontroller.go
skonto: I thought the same, in SRVKS mentions only the above, so from a telemetry perspective may not be useful. I can add it if it useful for telemetry.

knative-operator/pkg/common/telemetry/telemetry_ucontroller.go
skonto: Different names in logs.

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
skonto: It could be but it is also applied to the current reconciler.

knative-operator/pkg/common/telemetry/telemetry.go
skonto: The problem is that for stuff that are not in cache yet you will do that call multiple times for each type of resource you get an event for. The reconciler uses c-r client to list stuff, and it touches the cache as you mention. 
The other problem is that if you have no event coming in to trigger the controller because someone restarted the operator pod you will get a zero value for all metrics. T

knative-operator/pkg/common/telemetry/telemetry.go
skonto: Doc needs update.

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
skonto: ok.

knative-operator/pkg/common/telemetry/telemetry_ucontroller.go
skonto: ok.

knative-operator/pkg/common/telemetry/telemetry.go
skonto: ok.

knative-operator/pkg/common/telemetry/telemetry.go
skonto: What if it does not happen and we get 0 for long?

knative-operator/pkg/common/telemetry/telemetry.go
markusthoemmes: I was actually saying that because we rely on the zero value (i.e. nil) in tests apparently. We use that pattern in other methods on this type already.

skonto: I will change in all functions anyway. I could not parse it above because if I do it for one I should do it for all imho.

markusthoemmes: Yeah I agree, so i didn't parse why we did it different here than in the others.

knative-operator/pkg/common/telemetry/telemetry.go
markusthoemmes: But shouldn't we get a global resync on controller startup?

skonto: The complete solution would be to do a global resync there as well. In the code I have this comment:

```
	// Initialize metrics before we start the corresponding controller.
	// There is a tiny window to miss events here, but should be ok for telemetry purposes.
``
At this point we take the snapshot and then start the controller. I dont expect to lose a considerable number of events in between.
For telemetry I am going to use something like `avg by (type) (avg_over_time(serverless_telemetry[30m]))`, so I dont expect much of a difference in general. I can also do that for the sake of correctness. WDYTH?

Comment (markusthoemmes) [reply knative-operator/pkg/common/telemetry/telemetry.go:95]: My point is: Why doesn't it already do a global resync anyway? AFAIK starting a watcher does that by design.

File: knative-operator/pkg/common/telemetry/telemetry.go (hunk 21 above)
Comment (skonto) [reply knative-operator/pkg/common/telemetry/telemetry.go:48]: Because I read that you should not check the pointer in receiver methods as a practice. Anyway will do it.

Comment (markusthoemmes) [reply knative-operator/pkg/common/telemetry/telemetry.go:48]: But you do it already?

File: knative-operator/pkg/common/telemetry/telemetry.go (hunk 22 above)
Comment (skonto) [reply knative-operator/pkg/common/telemetry/telemetry.go:37]: Here I need to pass a telemetry ref to the controller, it too late.

File: knative-operator/pkg/common/telemetry/telemetry_ucontroller.go (hunk 28 above)
Comment (skonto) [reply knative-operator/pkg/common/telemetry/telemetry_ucontroller.go:25]: I will add it, should be of no harm.

File: knative-operator/pkg/common/telemetry/telemetry.go (hunk 18 above)
Comment (markusthoemmes) [reply knative-operator/pkg/common/telemetry/telemetry.go]: Why would it not happen?

File: knative-operator/pkg/common/telemetry/telemetry.go (hunk 32 above)
Comment (skonto) [reply knative-operator/pkg/common/telemetry/telemetry.go:95]: Ok will do.

**Comment by skonto** [comment]
@markusthoemmes I moved to a better design I update state per metric, also in the controller. No coordination is required as metrics match 1-1 to the events. Metrics are as follows:
image

File: knative-operator/pkg/common/telemetry/telemetry.go [hunk 33]
```diff
@@ -0,0 +1,121 @@
...
+// When called it assumes that we are reconciling a deletion event.
+func (t *Telemetry) TryStopTelemetry() {
+	if t == nil {
+		return
+	}
+	if !t.shouldInstallTelemetry {
+		log.Info("stopping telemetry for:", "component", t.name)
+		// Stop the telemetry controller
+		close(t.stop)
+		// Can'telemetry use a closed channel
```
markusthoemmes, line 88:
```suggestion
		// Can't reuse a closed channel
```

knative-operator/pkg/common/telemetry/telemetry.go
```diff
@@ -0,0 +1,121 @@
...
+			if err := (*t.tc).Start(t.stop); err != nil {
+				log.Error(err, "cannot start telemetry controller for", "component", t.name)
+				errCh <- err
+			}
+		}()
+		select {
+		case err := <-errCh:
+			t.shouldInstallTelemetry = true
+			return err
+		case <-time.After(1 * time.Second): // assume no error, as controller has blocked
```
markusthoemmes, line 71: This will randomly block the reconcile loop for 1 second :thinking:. Should we "just" resort on error logging instead?

knative-operator/pkg/common/telemetry/telemetry.go
```diff
@@ -0,0 +1,121 @@
...
+	case "serving":
+		serviceG = serverlessTelemetryG.WithLabelValues("service")
+		routeG = serverlessTelemetryG.WithLabelValues("route")
+		revisionG = serverlessTelemetryG.WithLabelValues("revision")
+		configurationG = serverlessTelemetryG.WithLabelValues("configuration")
+		updateMetricFor(&servingv1.Service{}, api)
+		updateMetricFor(&servingv1.Route{}, api)
+		updateMetricFor(&servingv1.Revision{}, api)
+		updateMetricFor(&servingv1.Configuration{}, api)
+	}
```
markusthoemmes, line 120: Are these lists redundant with the constructor?

I.e. could we store the objects passed and then do:

```golang
for _, obj := t.objects {
  updateMetricFor(obj, api)
}
```

knative-operator/pkg/common/telemetry/telemetry.go
```diff
@@ -0,0 +1,121 @@
...
+
+// initializeAndUpdateMetrics is used for taking a global snapshot of metrics
+// before we start a telemetry controller. Cost should be low since we are fetching from cache.
+func (t *Telemetry) initializeAndUpdateMetrics(api client.Client) {
+	if t == nil {
+		return
+	}
+	switch t.name {
+	case "eventing":
+		pingSourceG = serverlessTelemetryG.WithLabelValues("source_ping")
```
markusthoemmes, line 102: What are these label values for?

markusthoemmes (commented):
The cardinality you show in the screenshot is a showstopper for Telemetry I believe, as this will be different for each cluster :thinking:. Have we checked with them if this is okay?

knative-operator/pkg/common/telemetry/telemetry.go
skonto: It is a gauge vec where you have one vec and set a label to create a number of different time series based on the labels. It is standard prometheus approach instead of creating N different vecs.

knative-operator/pkg/common/telemetry/telemetry.go
skonto: I am trying to find some way to extract the error. I can remove it.

skonto:
@markusthoemmes for telemetry we will have aggregation over these metrics.

knative-operator/pkg/common/telemetry/telemetry.go
markusthoemmes: Ah I see, and we can't initialize them in the metrics code becauuuuse... that would make them all show 0 and we don't want that?

skonto: Yes at least at the beginning we dont want if some component is never used or the operator is restarted, but it is of no harm to initialize them (I had it that way before). 
However, when we stop the controller we could remove the labels by calling `serverlessTelemetryG.DeleteLabelValues("...")`, I am still thinking about it, as discussed [here](https://godoc.org/github.com/prometheus/client_golang/prometheus#MetricVec.GetMetricWithLabelValues).:
```
Keeping the Metric for later use is possible (and should be considered if performance is critical), but keep in mind that Reset, DeleteLabelValues and Delete can be used to delete the Metric from the MetricVec. In that case, the Metric will still exist, but it will not be exported anymore, even if a Metric with the same label values is created later.
```
Right now between deletes/installs of Serving, Eventing while the serverless operator is running we keep the metrics.

knative-operator/pkg/common/telemetry/telemetry.go
```diff
@@ -0,0 +1,104 @@
...
+		log.Info("starting telemetry for:", "component", t.name)
+		// Initialize metrics before we start the corresponding controller.
+		// There is a tiny window to miss events here, but should be ok for telemetry purposes.
+		t.initializeAndUpdateMetrics(api)
+		// Start our controller in a goroutine so that we do not block.
+		go func() {
+			// Block until our controller manager is elected leader. We presume our
+			// entire process will terminate if we lose leadership, so we don'telemetry need
+			// to handle that.
+			<-mgr.Elected()
```
markusthoemmes, line 54: I think we can safely drop this, can't we? We should be guaranteed to be leaders by the time we call Start in the first place.

knative-operator/pkg/common/telemetry/telemetry.go
```diff
@@ -0,0 +1,104 @@
...
+	if err != nil {
+		return nil, err
+	}
+	t.tc = tc
+	return t, nil
+}
+
+// TryStartTelemetry setups telemetry per component either Eventing, KnativeKafka or Serving.
+// When called it assumes that the component has status ready.
+func (t *Telemetry) TryStartTelemetry(api client.Client, mgr manager.Manager) error {
```
markusthoemmes, line 40:
```suggestion
func (t *Telemetry) TryStart(api client.Client, mgr manager.Manager) error {
```

and same for Stop, to have a bit less stutter in the API.

knative-operator/pkg/common/telemetry/telemetry.go
markusthoemmes: Honestly, I would not want to do that. What's the harm in reporting 0 sources in a cluster where there are... well... 0 sources? Seems like unnecessary complexity to me.

knative-operator/pkg/common/telemetry/telemetry_ucontroller.go
markusthoemmes: Do we do that for our other controllers? I'd actually not want to do that as the strings should actually be stable to allow for easy filtering across pods.

knative-operator/pkg/common/telemetry/telemetry.go
skonto: Yes we can.

knative-operator/pkg/common/telemetry/telemetry.go
skonto: A bit nit but ok.

skonto:
> Honestly, I would not want to do that. What's the harm in reporting 0 sources in a cluster where there are... well... 0 sources? Seems like unnecessary complexity to me.

Ok.

>  Do we do that for our other controllers? I'd actually not want to do that as the strings should actually be stable to allow for easy filtering across pods.
I can remove it.

knative-operator/pkg/common/telemetry/telemetry.go
markusthoemmes: Wouldn't have commented if there wasn't more in the review :P

skonto:
@markusthoemmes I will follow up this with an integration test, let me know if it is now ok.

markusthoemmes (approved):
/lgtm
/approve

PR #662, opened 2020-11-03
Remove note that eventing is technical preview

warrenvw:
Eventing is GA as of OpenShift Serverless 1.11, so remove note that it is a technical preview feature.


markusthoemmes (approved):
/lgtm
/approve

Master too pls.

PR #664, opened 2020-11-03
Wait for given image during cluster upgrades - master

mgencur:
* the condition is different when specifying a concrete image vs. OCP
channel

### DIFF
--- a/test/lib.bash
+++ b/test/lib.bash
@@ -183,8 +183,7 @@ function run_rolling_upgrade_tests {
 
   local latest_cluster_version latest_serving_version latest_eventing_version \
     rootdir scope serving_in_scope eventing_in_scope serving_prober_pid \
-    eventing_prober_pid prev_serving_version prev_eventing_version \
-    ocp_target_version retcode
+    eventing_prober_pid prev_serving_version prev_eventing_version retcode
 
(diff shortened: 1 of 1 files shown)


test/lib.bash
```diff
@@ -325,21 +324,20 @@ function end_prober {
...
 function upgrade_ocp_cluster {
-  local ocp_target_version upgrade_ocp_image latest_cluster_version
+  local upgrade_ocp_image latest_cluster_version
   upgrade_ocp_image="${1:-}"
 
   if [[ -n "$upgrade_ocp_image" ]]; then
-    ocp_target_version="$upgrade_ocp_image"
     oc adm upgrade --to-image="${UPGRADE_OCP_IMAGE}" \
       --force=true --allow-explicit-upgrade || return $?
+    timeout 7200 "[[ \$(oc get clusterversion version -o jsonpath='{.status.history[?(@.image==\"${upgrade_ocp_image}\")].state}') != Completed ]]" || return $?
```
skonto, line 333: is this 2h time? it takes that long?

mgencur: It takes something between half an hour and two hours. The time is always different. Anyway, cluster upgrades don't run in CI. We only re-use this to run cluster upgrade tests in downstream.

cardil (approved):
/lgtm

PR #665, opened 2020-11-03
Wait for given image during cluster upgrades - release-1.11

mgencur:
* the condition is different when specifying a concrete image vs. OCP
channel

### DIFF
--- a/test/lib.bash
+++ b/test/lib.bash
@@ -170,8 +170,7 @@ function run_rolling_upgrade_tests {
 
   local latest_cluster_version latest_serving_version latest_eventing_version \
     rootdir scope serving_in_scope eventing_in_scope serving_prober_pid \
-    eventing_prober_pid prev_serving_version prev_eventing_version \
-    ocp_target_version retcode
+    eventing_prober_pid prev_serving_version prev_eventing_version retcode
 
   scope="${1:?Provide an upgrade scope as arg[1]}"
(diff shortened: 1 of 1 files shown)


test/lib.bash
```diff
@@ -312,21 +311,20 @@ function end_prober {
...
 function upgrade_ocp_cluster {
-  local ocp_target_version upgrade_ocp_image latest_cluster_version
+  local upgrade_ocp_image latest_cluster_version
   upgrade_ocp_image="${1:-}"
 
   if [[ -n "$upgrade_ocp_image" ]]; then
-    ocp_target_version="$upgrade_ocp_image"
     oc adm upgrade --to-image="${UPGRADE_OCP_IMAGE}" \
       --force=true --allow-explicit-upgrade || return $?
+    timeout 7200 "[[ \$(oc get clusterversion version -o jsonpath='{.status.history[?(@.image==\"${upgrade_ocp_image}\")].state}') != Completed ]]" || return $?
```
cardil, line 320: I see. Maybe add additional variable `history_key` to be either `image` or `version`. then it will not be required to copy the entire long command.

mgencur: I'd prefer to keep it as it is. It's not necessary to these micro-optimizations at this level, IMO.

cardil: Ok

PR #667, opened 2020-11-04
Remove auth plugins to get rid of jwt vulnerability

markusthoemmes:
I don't think we even need/use these plugins on the platforms where Openshift runs and if we don't, it's just unnecessary distraction.


markusthoemmes:
I tried this on a GCP cluster without issues. This will go through a more thorough check via our weekly's

PR #669, opened 2020-11-04
Collapse all vendor directories into one central one

markusthoemmes:
This will be enormous :joy:. Mostly tried out to see if this is now possible at all. And it is!

This would mean less work keeping dependencies aligned and also no more trouble with using different versions of the same dep across component and test.

/hold

For discussion.


markusthoemmes:
/unhold

This has been discussed and accepted on the dev list.

matzew:
/lgtm
/hold

(holding to let the tests all pass)

PR #670, opened 2020-11-05
Remove unused build files and align Dockerfiles

markusthoemmes:
As per title, this removes the unused `build` folders and aligns all of our Dockerfiles to look roughly the same. There's no reason for them not to all behave almost identical.

This also removes the unnecessary explicit `command` from the CSV as that's just redundant.


markusthoemmes:
/retest

Eventing upgrade failure.

PR #679, opened 2020-11-09
Drop unnecessary WORKDIR commands and simplify manifest copying

markusthoemmes:
1. It's not necessary to build our components in specific locations anymore, as they are now go module based.
2. It's unnecessary to copy manifests in the builder image only to then copy them again.


mgencur:
@markusthoemmes Did you try to build the images locally? Curious if this works..

markusthoemmes:
@mgencur sure, it does.

PR #681, opened 2020-11-10
[SRVKS-654]Improve health status

skonto:
Followed this [best practice](https://www.robustperception.io/booleans-logic-and-math) to check if all installed components are up. The goal is to make health status reliable and correct since now we require both components to be present to report a healthy installation (Kafka status will be added in another PR, independently of this).

We keep metrics only for the lifecycle of the corresponding CR, will use the avg over time to create an OCP alert in the future.
The 3 signals used are `0` for when we are installing or never reached ready state in general, `no data` (prometheus no data) when a component does not exist and `1` for being up.

 Here is a sample run:
* Installing only eventing, status is 0:
image

* Eventing is up
image
* Uninstalling it, no data:
image

Another scenario:
* Installing both Serving, Eventing:
image
* Both are up:
image
* Uninstalled Eventing:
image
Status is 1 since we only run one component.
* Re-installing Eventing:
image

### DIFF
--- a/knative-operator/deploy/resources/dashboards/grafana-dash-knative-health.yaml
+++ b/knative-operator/deploy/resources/dashboards/grafana-dash-knative-health.yaml
@@ -94,7 +94,7 @@ data:
                }
             ],
             "thresholds": "",
-            "title": "Knative Current Status",
+            "title": "Knative Status",
             "type": "singlestat",
             "valueFontSize": "200%",
             "valueMaps": [
--- a/knative-operator/deploy/resources/dashboards/grafana-dash-knative-health.yaml
+++ b/knative-operator/deploy/resources/dashboards/grafana-dash-knative-health.yaml
@@ -85,7 +85,7 @@ data:
             },
             "targets": [
                {
-                 "expr": "avg_over_time ((sum(knative_up{namespace=\"$namespace\", service=\"knative-openshift-metrics\", type=~\"eventing_status|serving_status\"}) / (sum(knative_up{namespace=\"$namespace\", service=\"knative-openshift-metrics\", type=~\"eventing_status|serving_status\"})))[1m:30s]) >bool 0.95",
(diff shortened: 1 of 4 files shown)


skonto:
@markusthoemmes @s-urbaniak pls review.

skonto:
@bbrowning pls also comment on this.

bbrowning:
Overall, I think the health status metric looks good. The only thing I'd ask is that label values include the namespace and name of the instance (KnativeServing/KnativeEventing/etc) being reconciled. The reason for this is as we move these around (like in #386) having the namespace and name of the operand being reconciled in the metric will be useful as we look at aggregation via telemetry and tracking which namespaces are in use and how many clusters have migrated to the new namespaces. It also safeguards us in case our validation webhook preventing multiple installs ever gets a bug or we make a decision in the future to allow more than one instance of some components per cluster.

skonto:
@bbrowning telemetry gets the metric from prometheus correct?
Right now if metrics are exposed as follows at the pod endpoint:
```
# HELP knative_up Reports if a Knative component is up
# TYPE knative_up gauge
knative_up{type="kafka_status"} 0
knative_up{type="serving_status"} 0
knative_up{type="eventing_status"} 0
```
On prometheus they are enriched with default labels:
```
knative_up{endpoint="http-metrics",instance="10.116.0.75:8383",job="knative-openshift-metrics",namespace="openshift-serverless",pod="knative-openshift-d9c66646c-tt4p9",service="knative-openshift-metrics",type="serving_status"}```
So I think all labels you mentioned are added.

@markusthoemmes if there is no objection let's get this merged.

markusthoemmes (approved):
/lgtm
/approve

bbrowning:
@skonto Those namespace and pod labels you see added to the time series are the namespace and pod of the endpoint that got scraped for metrics. I'm asking that we include the namespace and name of the `KnativeServing`, `KnativeEventing`, and other CRs in the metrics. These would be under some other label - `operand_namespace` and `operand_name` or something like that.

skonto:
@bbrowning ok got it will add it in a separate PR.

PR #684, opened 2020-11-10
Update pkg to latest

skonto:
I need [prometheus host config](https://github.com/knative/pkg/pull/1882) in to move on with securing the metric points of the serving control plane pods. Otherwise will have to wait for next releases.
/cc @markusthoemmes


markusthoemmes:
@skonto Why do you need this config in the serverless-operator though?We usually don't run `pkg` versions that differ from the release versions and I'd be cautious to do so, as there are no compatibility guarantees.

skonto:
I was thinking of protecting the endpoints but probably it is too much. I can close it for now.

PR #660, opened 2020-11-02
Add readiness and liveness probes

skonto:
- Follows the approach [here](https://github.com/kubernetes-sigs/controller-runtime/pull/856). Tested with a deployment. Uses the standard paths: `/livez` & `/readyz`, exposed by c-r. 
- We need this especially for integration tests to make sure everything is setup (monitoring is not covered due to the lack of this).
/cc @markusthoemmes @aliok

PR #685, opened 2020-11-11
Wait for namespaces when installing previous releases

mgencur:
This should resolve issues like these in upgrade tests:
```
Error from server (NotFound): namespaces "knative-serving-ingress" not found
01:49:06.299 INFO:    Testing with pre upgrade tests
01:49:06.301 INFO:    Running Serving pre upgrade tests
```
This error is thrown when the scripts are trying to label the namespace with `system-namespace`. This prevents the ingress from functioning later in upgrade tests.
Readiness checks will be available since 1.12 (https://github.com/openshift-knative/serverless-operator/pull/660) and until that time we need to wait for the namespaces in upgrade tests because they're installing a previous version which doesn't have them yet.

### DIFF
--- a/hack/lib/serverless.bash
+++ b/hack/lib/serverless.bash
@@ -31,15 +31,6 @@ function install_serverless_previous {
 
   deploy_serverless_operator "$PREVIOUS_CSV"  || return $?
 
-  # TODO(ksuszyns): Remove this if block if no longer required
-  if versions.le "$(metadata.get olm.replaces)" 1.11.0; then
-    logger.info "Ensure ${SERVING_NAMESPACE} and ${EVENTING_NAMESPACE} \
-namespaces exists, as ${PREVIOUS_CSV} didn't created them automatically."
-
-    ensure_namespace "${SERVING_NAMESPACE}"
-    ensure_namespace "${EVENTING_NAMESPACE}"
-  fi
-
   deploy_knativeserving_cr || return $?
   deploy_knativeeventing_cr || return $?
   logger.success "Previous version of Serverless is installed: $PREVIOUS_CSV"
@@ -152,6 +143,11 @@ function deploy_knativeserving_cr {
   oc apply -n "${SERVING_NAMESPACE}" -f "${rootdir}/test/v1alpha1/resources/operator.knative.dev_v1alpha1_knativeserving_cr.yaml" || return $?
 
   timeout 900 '[[ $(oc get knativeserving.operator.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.conditions[?(@.type==\"Ready\")].status}") != True ]]'  || return 7
+
+  # TODO: Remove this when the previous release has readiness checks available (since 1.12).
+  if versions.le "$(metadata.get olm.replaces)" 1.11.0; then
+    timeout 120 "[[ -z \$(oc get namespace ${INGRESS_NAMESPACE}) ]]" || return $?
+  fi
 
   logger.success 'Knative Serving has been installed successfully.'
 }
--- a/hack/lib/serverless.bash
+++ b/hack/lib/serverless.bash
@@ -149,11 +149,6 @@ function deploy_knativeserving_cr {
 
   timeout 900 '[[ $(oc get knativeserving.operator.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.conditions[?(@.type==\"Ready\")].status}") != True ]]'  || return 7
 
-  # TODO: Remove this when the previous release has readiness checks available (since 1.12).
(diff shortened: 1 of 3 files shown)


hack/lib/serverless.bash
```diff
@@ -31,13 +31,9 @@ function install_serverless_previous {
...
 
-  # TODO(ksuszyns): Remove this if block if no longer required
+  # TODO: Remove this when the previous release has readiness checks available (since 1.12).
   if versions.le "$(metadata.get olm.replaces)" 1.11.0; then
-    logger.info "Ensure ${SERVING_NAMESPACE} and ${EVENTING_NAMESPACE} \
-namespaces exists, as ${PREVIOUS_CSV} didn't created them automatically."
-
-    ensure_namespace "${SERVING_NAMESPACE}"
-    ensure_namespace "${EVENTING_NAMESPACE}"
+    timeout 120 "[[ \$(oc get namespace ${SERVING_NAMESPACE} ${EVENTING_NAMESPACE} ${INGRESS_NAMESPACE} --no-headers | wc -l) != 3 ]]" || return $?
```
markusthoemmes, line 36: The Ingress Namespace is not created as part of auto-creation but actually when creating Knative Serving so this will never pass.

mgencur: Looks better now? Pushed changes...

hack/lib/serverless.bash
```diff
@@ -153,6 +144,11 @@ function deploy_knativeserving_cr {
 
   timeout 900 '[[ $(oc get knativeserving.operator.knative.dev knative-serving -n $SERVING_NAMESPACE -o=jsonpath="{.status.conditions[?(@.type==\"Ready\")].status}") != True ]]'  || return 7
 
+  # TODO: Remove this when the previous release has readiness checks available (since 1.12).
```
markusthoemmes, line 147: This is not captured by readiness checks at all. If this is the namespace that's failing, we have a different bug.

hack/lib/serverless.bash
mgencur: I'm wondering if we should also wait for knative-serving and knative-eventing namespaces in the original place.

hack/lib/serverless.bash
mgencur: OK. Hmm...

mgencur: The failure was here: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-knative-serverless-operator-master-4.6-e2e-aws-ocp-46-continuous/1325770323794071552

markusthoemmes:
I believe this might be a race between our controllers because Kourier is installed by a different controller than the "normal" system and thus KnativeServing gets called ready before Kourier is ready maybe? :thinking:

mgencur:
I think you're right about that but the real cause of `knative-serving-ingress` NS not being created is different. The knative-openshift pod doesn't start and logs this again and again:
```
{"level":"info","ts":"2020-11-11T13:37:58.018Z","logger":"leader","msg":"Trying to become the leader."}
{"level":"info","ts":"2020-11-11T13:37:58.167Z","logger":"leader","msg":"Not the leader. Waiting."}
```
I can see it here: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-openshift-knative-serverless-operator-master-4.6-e2e-aws-ocp-46-continuous/1326495012841590784/artifacts/e2e-aws-ocp-46-continuous/serverless-e2e/gather-knative/quay-io-openshift-knative-must-gather-sha256-51049ace7b41db7499a225b27bf11492e419245a3c49581020c184d6c816d410/openshift-serverless/pods/
But I'm not sure why that happens yet.

mgencur:
OK. I think the latest commit will fix it. I think this whole thing is caused by not removing Serverless properly before running the upgrade tests. We run s-o tests before them.  Let's see.

hack/lib/serverless.bash
```diff
@@ -49,6 +40,11 @@ function install_serverless_latest {
   logger.info "Installing latest version of Serverless..."
   deploy_serverless_operator_latest || return $?
 
+  # TODO: Remove this when the previous release has readiness checks available (since 1.12).
+  if versions.le "$(metadata.get olm.replaces)" 1.11.0; then
+    timeout 120 "[[ \$(oc get namespace ${SERVING_NAMESPACE} ${EVENTING_NAMESPACE} --no-headers | wc -l) != 2 ]]" || return $?
+  fi
```
markusthoemmes, line 46: Isn't this the block where we effectively are at 1.12? (`install_serverless_latest`)

mgencur: Good catch! And sorry. I'm blind. Fixed.

markusthoemmes (approved):
/lgtm
/approve

PR #686, opened 2020-11-11
[wip]Add telemetry checks in serving downstream integration tests

skonto:
This is a follow up of #657.

### DIFF
--- a/test/servinge2e/helpers.go
+++ b/test/servinge2e/helpers.go
@@ -2,11 +2,20 @@ package servinge2e
 
 import (
 	"context"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
 	"net/url"
+	"strconv"
 	"testing"
 
 	"github.com/openshift-knative/serverless-operator/test"
+	v1 "github.com/openshift/api/route/v1"
+	ioprometheusclient "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/expfmt"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	pkgTest "knative.dev/pkg/test"
 	servingv1 "knative.dev/serving/pkg/apis/serving/v1"
 )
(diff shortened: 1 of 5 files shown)


test/servinge2e/helpers.go
```diff
@@ -44,3 +53,128 @@ func withServiceReadyOrFail(ctx *test.Context, service *servingv1.Service) *serv
...
+	routes         float64
+	configurations float64
+	revisions      float64
+}
+
+func getMetricsEndpointPath() string {
+	portAndPath := strconv.Itoa(8383) + "/metrics"
+	metricsPath := "http://knative-openshift-metrics.openshift-serverless:" + portAndPath
+	return metricsPath
+}
```
markusthoemmes, line 68: This won't work as the hostname is hardcoded. You'll have to fetch it from the route after it's applied and accepted. The port will also be 80.

test/servinge2e/helpers.go
```diff
@@ -44,3 +53,128 @@ func withServiceReadyOrFail(ctx *test.Context, service *servingv1.Service) *serv
...
+			Path: "/metrics",
+			To: v1.RouteTargetReference{
+				Kind: "Service",
+				Name: "knative-openshift-metrics",
+			},
+			TLS: &v1.TLSConfig{
+				Termination:                   v1.TLSTerminationEdge,
+				InsecureEdgeTerminationPolicy: v1.InsecureEdgeTerminationPolicyAllow,
+			},
+			WildcardPolicy: v1.WildcardPolicyNone,
```
markusthoemmes, line 167: Are these required? I'd think no.

test/servinge2e/service_to_service_test.go
```diff
@@ -89,6 +98,53 @@ func testServiceToService(t *testing.T, ctx *test.Context, namespace string, tc
 		t.Errorf("the Route at domain %s didn't serve the expected text %q: %v", service.Status.URL.URL(), helloworldText, err)
 	}
 
+	// Check if service monitor service is available, at this point it should be present
+	_, err := ctx.Clients.Kube.CoreV1().Services("openshift-serverless").Get(context.Background(), "knative-openshift-metrics", meta.GetOptions{})
+	if err != nil {
+		t.Errorf("error getting service monitor service: %v", err)
```
markusthoemmes, line 104: This should be Fatal as the rest of the test won't work (actually true for almost all statements hereafter)

test/servinge2e/servicemesh_test.go
```diff
@@ -37,10 +37,11 @@ import (
...
 type testCase struct {
-	name               string
-	labels             map[string]string // Ksvc labels
-	annotations        map[string]string // Revision template annotations
-	expectIstioSidecar bool              // Whether it is expected for the istio-proxy sidecar to be injected into the pod
+	name                string
+	labels              map[string]string // Ksvc labels
+	annotations         map[string]string // Revision template annotations
+	expectIstioSidecar  bool              // Whether it is expected for the istio-proxy sidecar to be injected into the pod
+	expectedMetricCount float64
```
markusthoemmes, line 44: Let's please create an extra test for this. This has nothing to do with the Service to Service tests and will just make picking apart test results harder as the service-to-service tests might fail due to this.

test/servinge2e/helpers.go
```diff
@@ -44,3 +53,128 @@ func withServiceReadyOrFail(ctx *test.Context, service *servingv1.Service) *serv
...
+}
+
+func getMetricsEndpointPath() string {
+	portAndPath := strconv.Itoa(8383) + "/metrics"
+	metricsPath := "http://knative-openshift-metrics.openshift-serverless:" + portAndPath
+	return metricsPath
+}
+func fetchTelemetryMetrics() (*telemetryStat, error) {
+	metricsPath := getMetricsEndpointPath()
+	req, err := http.NewRequestWithContext(context.Background(), http.MethodGet, metricsPath, nil)
```
markusthoemmes, line 71: nit: NewRequest is just as fine if you use a background context anyway.

test/servinge2e/helpers.go
```diff
@@ -44,3 +53,128 @@ func withServiceReadyOrFail(ctx *test.Context, service *servingv1.Service) *serv
...
+	if err != nil {
+		return nil, err
+	}
+	httpClient := http.Client{}
+	resp, err := httpClient.Do(req)
```
markusthoemmes, line 76: The response's body needs to be closed.

test/servinge2e/servicemesh_test.go
skonto: In the past we discussed that I should not create a separate test to save some time. I guess this kind of test which just sets the admin does not cost anything right?

test/servinge2e/helpers.go
skonto: Yeah I need the domain.

test/servinge2e/servicemesh_test.go
markusthoemmes: Well we just need to create a Knative Service and then check the stuff you're checking. That might cost a second or two more, but that's absolutely worth it here.

We wouldn't want to install Serving separately for it though, yeah. We should be good here with a separate test in the same suite.

test/servinge2e/helpers.go
skonto: I guess they are meant to allow non-secure connections within the cluster and so taht tls is terminated to the edge, I guess not relevant if we are hitting with http only.

test/servinge2e/service_to_service_test.go
skonto: Yeah I was thinking about it guess if I move this in a separate test it makes sense.

skonto:
@markusthoemmes @bbrowning as we discussed this is going to be provided by OCP in future release 4.7+. I will close this one and also revert #657

PR #698, opened 2020-11-18
Use correct label for cluster-local

markusthoemmes:
As per title. 0.18 will start failing validation for `ClusterLocal`.

/assign @nak3


nak3 (approved):
/lgtm

PR #703, opened 2020-11-24
[SRVKS-666] Fix race condition in Serving readiness

markusthoemmes:
Depending on which operator picks the Serving resource up first, there is a potential race where the upstream code defaults the dependencies condition to True before the downstream code even sees it and can mark it as False.
This sets the correct default in upstream code so the API never actually observes the race.

/assign @nak3


nak3 (commented):
LGTM defer to @mgencur 
I wonder that we can drop this workaround https://github.com/openshift-knative/serverless-operator/pull/421 now?

PR #705, opened 2020-11-25
manually adding the KnativeKafka yaml bag

matzew:

Adding the Kafka bits manually here. I used yamls from our fork repo, and added the `grc.io` images there...

Manually b/c of a few reasons:
* 1.12 (0.18) is the last release from eventing-contrib, for 0.19 (1.13) we need to update to `knative-sandbox/eventing-kafka`.
* we do (here) have some tweaks for getting the source to the `knative-eventing` namesapce. Starting with 0.20, the upstream is likely doing that too.... So, even with a download, we'd still have to manually tweak that.

Once the upstream is more aligned with what we want, I see room for an automation, similar like we have for the eventing or serving bits - I really like them (I will PR and automated update for eventing 0.18.6 soon)


skonto:
Unit test needs adaption?

matzew:
@aliok We have the rests for the "not allowed" resources. Those are "duplicated" configmaps we already have in `knative-eventing` - to avoid override, like:
* `config-tracing`
* `config-observability`
* `config-logging`

However, I removed the check for the `config-leader-election-kafka`, which is NOT in `knative-eventing` and is relevant for the Kafka bits.

Please check

aliok:
@matzew I remember it was suggested to delete that configmap in the manifest and I also added a test for that here: https://github.com/openshift-knative/serverless-operator/pull/536/commits/30f2bf9b23b620d91f4c2dcddc6238d8d8a0667d

but, can't find when/where/why it was suggested.

matzew:
Isn't that we only have the duplicates to be deleted ?

aliok:
yeah, but why did we delete that in the first place? was it ever duplicated?

aliok:
I mean `config-leader-election-kafka`

matzew:
the kafka one? I guess be accident? and to be better safe, than sorry  ? :grimacing:

aliok:
I looked again in Slack but couldn't find the message.

Should be fine.

/lgtm
/approve

matzew:
We now have two configmaps, just for `kafka` bits:
* `config-kafka` (where we set auth and bootstrapServers)
* `config-leader-election-kafka` ... is for KAFKA... 

but... than, I think.... perhaps we want it NOT in?  and have folks update the _normal_ CM (`config-leader-election`) for adding the Kafka bits? 

/hold

Perhaps @markusthoemmes has a thought on this topic of `config-leader-election` ?

matzew:
I have more questions on that, will follow up on a different thread
For now, that Config-map is just for kafka... and therefore fine.   Upstream is having thread on defaulting to `config-leader-election`, than we would rethink this

/unhold

aliok:
/test 4.6-operator-e2e-aws-ocp-46

Saw some Serving tests failed.

PR #707, opened 2020-11-26
Enable smoke tests for Kafka auth

aliok:
Couldn't enable them during https://github.com/openshift-knative/serverless-operator/pull/699 because we didn't have the TLS/SASL support backported. Should be good now since these are backported and we use eventing-contrib 0.18 in CSV.


aliok:
/retest

infra

aliok:
/retest

infra

matzew (approved):
/lgtm
/approve

PR #713, opened 2020-12-01
Bump OLM to 1.13

matzew:

### DIFF
--- a/hack/lib/catalogsource.bash
+++ b/hack/lib/catalogsource.bash
@@ -118,7 +118,7 @@ spec:
         - -c
         - |-
           podman login -u $pull_user -p $token image-registry.openshift-image-registry.svc:5000 && \
-          /bin/opm registry add -d index.db --container-tool=podman --mode=replaces -b quay.io/openshift-knative/serverless-bundle:1.7.2,registry.svc.ci.openshift.org/openshift/openshift-serverless-v1.8.0:serverless-bundle,registry.svc.ci.openshift.org/openshift/openshift-serverless-v1.9.0:serverless-bundle,registry.svc.ci.openshift.org/openshift/openshift-serverless-v1.10.0:serverless-bundle,registry.svc.ci.openshift.org/openshift/openshift-serverless-v1.11.0:serverless-bundle,image-registry.openshift-image-registry.svc:5000/$OLM_NAMESPACE/serverless-bundle && \
+          /bin/opm registry add -d index.db --container-tool=podman --mode=replaces -b quay.io/openshift-knative/serverless-bundle:1.7.2,registry.svc.ci.openshift.org/openshift/openshift-serverless-v1.8.0:serverless-bundle,registry.svc.ci.openshift.org/openshift/openshift-serverless-v1.9.0:serverless-bundle,registry.svc.ci.openshift.org/openshift/openshift-serverless-v1.10.0:serverless-bundle,registry.svc.ci.openshift.org/openshift/openshift-serverless-v1.11.0:serverless-bundle,registry.svc.ci.openshift.org/openshift/openshift-serverless-v1.12.0:serverless-bundle,image-registry.openshift-image-registry.svc:5000/$OLM_NAMESPACE/serverless-bundle && \
           /bin/opm registry serve -d index.db -p 50051
 ---
 apiVersion: v1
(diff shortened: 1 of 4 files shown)


olm-catalog/serverless-operator/Dockerfile
```diff
@@ -13,7 +13,7 @@ LABEL operators.operatorframework.io.bundle.channels.v1="4.6"
 LABEL \
```
mgencur, line 13: The only issues I can see is with the channels. I think we'll want 4.7 as well. But that can probably be fixed later

hack/lib/catalogsource.bash
```diff
@@ -118,7 +118,7 @@ spec:
...
+          /bin/opm registry add -d index.db --container-tool=podman --mode=replaces -b quay.io/openshift-knative/serverless-bundle:1.7.2,registry.svc.ci.openshift.org/openshift/openshift-serverless-v1.8.0:serverless-bundle,registry.svc.ci.openshift.org/openshift/openshift-serverless-v1.9.0:serverless-bundle,registry.svc.ci.openshift.org/openshift/openshift-serverless-v1.10.0:serverless-bundle,registry.svc.ci.openshift.org/openshift/openshift-serverless-v1.11.0:serverless-bundle,registry.svc.ci.openshift.org/openshift/openshift-serverless-v1.12.0:serverless-bundle,image-registry.openshift-image-registry.svc:5000/$OLM_NAMESPACE/serverless-bundle && \
```
rhuss, line 121: How long do we need to keep the history ? 1.7.2 is out of support already, isn't it ? Could we just remove it here ?

olm-catalog/serverless-operator/Dockerfile
matzew: Adding 4.7?  or replacing 4.6 with 4.7 ? 

I am fine w/ a follow up PR on this, tbh

rhuss (commented):
The changes look straight forward to me, but I can't say if anything has been missed. Would it be possible to write one, two sentences how you have approached the update so that I could add it to the documentation ?

I have one question around how and when to remove versions from the index, but that might need a wider discussion.

hack/lib/catalogsource.bash
matzew: I have no idea, I just append what we have new :-)

I am fine w/ a clean up PR, coordinated w/ Warren 

I just want the metadata bump now, that we have the 1.12 image, as of yesterday

markusthoemmes: It always have to be the entire history I fear, otherwise the `replaces` parts of the CSVs break.

olm-catalog/serverless-operator/Dockerfile
markusthoemmes: Should definitely be discussed separately.

markusthoemmes (approved):
/lgtm
/approve

hack/lib/catalogsource.bash
cardil: We could take this from image.db from public redhat-operators catalog source image.

rhuss: Hmm, does this really scale ? (not sure how expensive the opm operation is). also there is an upper limit for arguments in system.exec() ;-) [128 kb i  think](https://unix.stackexchange.com/questions/120642/what-defines-the-maximum-size-for-a-command-single-argument)

markusthoemmes: No it doesn't really scale. This is already a problema s of today as it takes longer and longer to pull these images, which sucks.
We should defintely look for a more scalable way of doing this. Maybe we can do what @cardil suggested and take all the released versions from an "official" image and only supplement our unreleased bits (1.12, HEAD in this case) on top of it. I'll open a JIRA.

markusthoemmes: https://issues.redhat.com/browse/SRVCOM-1148

mgencur: There's another way to do this stuff: re-use the support for "optional operators" from ci-operator. The ci-operator would build the index image for use before our tests start. I wrote pros/cons on serverless-dev mailing list some time ago (subject "OpenShift CI supports PR testing for OLM-delivered operators"). This would need some changes in ci-operator but not very complicated.

markusthoemmes: Let's discuss in the JIRA.

PR #722, opened 2020-12-03
[1.12] Create the Secrets for SASL/TLS, needed on upstream KafkaSource tests

matzew:

In mid/upstream we create a secret for TLS and SALS based on the configured KAfka (Strimzi with PLAIN, TLS, SASL listeners) and their `StrimziUsers`.

This test suite lacks the creation of the secrets, doing so here.

/assign @lberk 
/assign @aliok 

The other option would be to create the secrets on the midstream's `run_e2e_tests` function...

### DIFF
--- a/test/eventing-contrib.bash
+++ b/test/eventing-contrib.bash
@@ -32,6 +32,8 @@ function upstream_knative_eventing_contrib_e2e {
 
   # run_e2e_tests defined in eventing-contrib
   logger.info 'Starting eventing-contrib tests'
+  (( !failed )) && create_auth_secrets || failed=$?
+
   (( !failed )) && run_e2e_tests || failed=$?
 
   return $failed
--- a/test/eventing-contrib.bash
+++ b/test/eventing-contrib.bash
@@ -30,7 +30,7 @@ function upstream_knative_eventing_contrib_e2e {
 
   failed=0
 
-  # run_e2e_tests defined in eventing-contrib
+  # create_auth_secrets and run_e2e_tests defined in eventing-contrib
   logger.info 'Starting eventing-contrib tests'
   (( !failed )) && create_auth_secrets || failed=$?
 


matzew:
/retest

ci is drunk?

aliok:
My impression is that this is not necessary.

There are paths that are testing Kafka stuff:

Path 1:
`make test-upstream-e2e`
|-> calls `./hack/strimzi.sh`
|-> calls `INSTALL_KAFKA=true TEST_KNATIVE_KAFKA=true ./test/upstream-e2e-tests.sh`

`./hack/strimzi.sh` already installs Strimzi and creates the certs.

Path 2:
`test-upstream-e2e-no-upgrade`
|-> calls `./hack/strimzi.sh`
|-> calls `INSTALL_KAFKA=true TEST_KNATIVE_KAFKA=true .... ./test/upstream-e2e-tests.sh`

same as 1. 

So, there are only 2 entry points that call the script modified in this PR, but in both paths we already create the secrets in advance.

Maybe have a 2nd look @matzew? Maybe I take something wrong

matzew:
Please look here, at these "upstream" tests failures:

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift-knative_serverless-operator/717/pull-ci-openshift-knative-serverless-operator-release-1.12-4.6-upstream-e2e-aws-ocp-46/1334407926441316352

@aliok we do need these secrets BEFORE we run the upstream contrib e2e test:
https://github.com/openshift/knative-eventing-contrib/blob/release-v0.18.0/test/e2e/kafka_source_test.go#L56-L57

The normal Source e2e test runs w/ SASL and TLS,  hence (upstream), lukas moved the creation of those Secrets up, see:
https://github.com/knative/eventing-contrib/blob/release-0.18/test/e2e-tests.sh#L285-L286

matzew:
/retest

infra

aliok:
> @aliok we do need these secrets BEFORE we run the upstream contrib e2e test:

We just create the secrets and we can always use them later, can't we?

Haven't checked the errors in depth, but I think the secrets created in S-O are not used by the KafkaSource tests.

We should create the secrets either when eventing-contrib is sourced, or separetely here in S-O but not both.

So, I have 2 alternative suggestions.

Alternative#1:
- We don't call the `create_auth_secrets` from eventing-contrib in S-O
- We make sure the KafkaSource tests use the secrets created by S-O in strimzi.sh

Alternative#2:
- We get rid of `strimzi.sh` in S-O and only call the script from eventing-contrib
- We adapt the downstream KafkaChannel tests to use secrets created by that script

...ok, here's another one:
- Instead of sourcing and copying scripts around, extract common stuff like this into a separate repo/lib and use them in eventing-contrib or S-O

matzew:
You know what?  I move the `create_auth_secrets` inside the test, on the contrib repo.....

PR #726, opened 2020-12-04
smcp v1 and v2 mesh tests

### LINKED ISSUE #443: Move test-specific SMCP creation and setup into test code
Currently "make install-mesh" contains test-specific setup  (certificate generation and specific ingress configuration in SMCP).

We should change this so that
* new target "make install-mesh-operator" that installs just the servicemesh operator (and no SMCP)
* "make install-mesh" installs a generic SMCP that would not be used in tests.
* mesh tests will use their own test namespaces for their own test SMCP and a different test namespace as the SMMR member.
* mesh tests will generate their certificates in test code.

This will allow mesh test to each have their own customized SMCP and we won't have tight dependencies between test and "make install-mesh" .
The downside is that mesh test will take a little bit longer, as they will each wait for their SMCP readiness.

maschmid:
drop mesh scripts, move setup logic to tests

fixes #443

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,11 +206,56 @@ To install service mesh operator, run `make install-mesh`
 make install-mesh
 ```
 
-Then, create Knative Service with `sidecar.istio.io/inject: "true"` annotation in `default` namespace,
+and create a `ServiceMeshControlPlane`.
+
+```
+apiVersion: maistra.io/v2
+kind: ServiceMeshControlPlane
+metadata:
+  name: basic
+  namespace: istio-system
+spec:
+  version: v2.0
+```
+
+Then, add your namespace to `ServiceMeshMemberRoll`
+
+```
+apiVersion: maistra.io/v1
+kind: ServiceMeshMemberRoll
+metadata:
+  name: default
+  namespace: istio-system
+spec:
+  members:
+    - $NAMESPACE_YOU_WANT_TO_ADD
+    # Add namespace you want to include mesh.
+```
+
+and add `NetworkPolicy` in your namespace.
+
+```
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-from-serving-system-namespace
+  namespace: $NAMESPACE_YOU_WANT_TO_ADD
+spec:
+  ingress:
+  - from:
+    - namespaceSelector:
+        matchLabels:
+          knative.openshift.io/system-namespace: "true"
+  podSelector: {}
+  policyTypes:
+  - Ingress
+```
+
+Then, create Knative Service with `sidecar.istio.io/inject: "true"` annotation in your namespace,
 which is one of the namespaces in the `ServiceMeshMemberRoll`.
 
 ```sh
-cat <<EOF | oc apply -n default -f -
+cat <<EOF | oc apply -n $NAMESPACE_YOU_WANT_TO_ADD -f -
 apiVersion: serving.knative.dev/v1
 kind: Service
 metadata:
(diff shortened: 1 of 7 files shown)


maschmid:
hm, weird, kiali operator didn't install, from the olm-operator logs
```
2020-12-07T12:30:05.990009064Z time="2020-12-07T12:30:05Z" level=info msg="couldn't ensure RBAC in target namespaces" csv=kiali-operator.v1.24.2 error="no owned roles found" id=5lI/p namespace=openshift-operators phase=Pending
2020-12-07T12:30:05.990037015Z E1207 12:30:05.990026       1 queueinformer_operator.go:290] sync {"update" "openshift-operators/kiali-operator.v1.24.2"} failed: no owned roles found
```

maschmid:
ServiceMesh tests passing in https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-knative_serverless-operator/726/pull-ci-openshift-knative-serverless-operator-master-4.6-operator-e2e-aws-ocp-46/1335961934180126720/artifacts/operator-e2e-aws-ocp-46/operator-e2e/build-WAUJd0wp/junit_JT48ZEaz.xml 

I'll remove the https://github.com/openshift-knative/serverless-operator/pull/726/commits/4eb2c13352dcaec8ed31c7577b27b5a631c4013c , as we don't want to run them in all configurations (install-mesh is enabled only in continuous jobs)

maschmid:
/unhold

(note, Service Mesh is not installed with 4.6-operator-e2e-aws-ocp-46 , so the new tests will be skipped in this run )

nak3:
Thank you for this great work! It looks good to me overall.

There is one discussion I think. I think that we can use Manifestival for the resource manipulation. For example, if we want to create serviceentry:

```
  patch := "test/some_path/serviceentry.yaml"
  manifest, err := mf.NewManifest(path, mf.UseClient(mfclient.NewClient(client)))
  if err != nil {
   return err
  }

  if err := manifest.Apply(); err != nil {
   return err
  }
```

Does it make more readable code? I feel that current resources in the code is a little bit difficult to read. WDYT?
(It will need an entire change so I am fine that we can merge this PR now and consider about it later.)

nak3:
@maschmid Friendly reminder. Any update on this?

maschmid:
@nak3 Thanks for the review! (and sorry for the long delay... )

As for using the files via Manifestival I am not convinced... In some/most cases we need some kind of templating and setting values inside a golang structure seems most readable in such cases (instead of manipulating a structure that comes from a completely different file).. (and mixing different approaches for static vs. templated structures doesn't feel nice either).

test/servicemesh.go
```diff
@@ -4,12 +4,100 @@ import (
...
+	}
+}
+
+func AddServiceMeshControlPlaneV1IngressGatewaySecretVolume(smcp *unstructured.Unstructured, name, secretName, mountPath string) error {
+	secretVolume := make(map[string]interface{})
+	secretVolume["name"] = name
+	secretVolume["secretName"] = secretName
+	secretVolume["mountPath"] = mountPath
+
+	secretVolumes, found, _ := unstructured.NestedSlice(smcp.Object, "spec", "istio", "gateways", "istio-ingressgateway", "secretVolumes")
```
nak3, line 57: The error from `NestedSlice` is ignored by design? If so, could you add the comment?

README.md
```diff
@@ -265,6 +251,31 @@ spec:
   - Ingress
 ```
 
+Then, create Knative Service with `sidecar.istio.io/inject: "true"` annotation in your namespace,
+which is one of the namespaces in the `ServiceMeshMemberRoll`.
+
+```sh
+cat <<EOF | oc apply -n default -f -
```
nak3, line 258: nit:

```suggestion
cat <<EOF | oc apply -n $NAMESPACE_YOU_WANT_TO_ADD -f -
```

nak3 (commented):
Thank you. Commented about a few nit.

test/servicemesh.go
maschmid: Fixed a few more of those as well, thanks!

nak3:
/assign @markusthoemmes @mgencur
for `/approve`

markusthoemmes (approved):

markusthoemmes:
/retest

Infra issues.

PR #728, opened 2020-12-04
[1.12] [SRVKE-654] Hacking the manifest

matzew:
@maschmid some hack... and likely not the real solution...

/hold
as this needs a bit more testing ...  

/cc @aliok

### DIFF
--- a/hack/dependencies.sh
+++ b/hack/dependencies.sh
@@ -14,9 +14,7 @@ source "$(dirname "${BASH_SOURCE[0]}")/lib/__sources__.bash"
 
 set -Eeuo pipefail
 
-scale_up_workers || exit $?
-create_namespaces || exit $?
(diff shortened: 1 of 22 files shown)


maschmid:
I suspect we may also need to  disable the bits in eventing-controller that scale pingsource-mt-adapter (if it ever scales it to something else than 1 )

maschmid:
actually, it looks like it really only ever scales it to 1 when it is 0, so it should work as is

matzew:
w/ that hack I get:

```
2020/12/04 19:24:12 Error processing env var: envconfig.Process: assigning K_SINK_TIMEOUT to EnvSinkTimeout: converting '' to type int. details: strconv.ParseInt: parsing "": invalid syntax
```

More next week...

PR #736, opened 2020-12-08
Print subscription when OPENSHIFT_CI is empty

nak3:
This patch makes a tiny change which outputs Subscription when OPENSHIFT_CI is empty.

Currently `cat` command is not executed when `OPENSHIFT_CI` is empty. But
we always would like to know the output for debug.

* Current log (it skips to cat `subscription.GSJnnX.yaml`):
```
++ mktemp /tmp/subscription.XXXXXX.yaml
+ tmpfile=/tmp/subscription.GSJnnX.yaml
+ cat
+ '[' -n '' ']'
+ oc apply -f /tmp/subscription.GSJnnX.yaml
subscription.operators.coreos.com/serverless-operator created
```

/cc @cardil

### DIFF
--- a/hack/lib/serverless.bash
+++ b/hack/lib/serverless.bash
@@ -93,7 +93,7 @@ spec:
   installPlanApproval: Manual
   startingCSV: "${csv}"
 EOF
-  [ -n "$OPENSHIFT_CI" ] && cat "$tmpfile"
+  cat "$tmpfile"
   oc apply -f "$tmpfile"
 
   # Approve the initial installplan automatically


cardil:
I've put there that check to not see that subscription, each time we execute. I didn't like to have such verbose output as default.

Obviously, you could just set `export OPENSHIFT_CI=anything-here` on your console. Maybe we should introduce some `DEBUG` variable, instead?

nak3:
I wanted to see the output when debugging on https://github.com/openshift/release/pull/14135. It does not allow me to add `export OPENSHIFT_CI=anything-here` so simply. The `DEBUG` option was not possible on https://github.com/openshift/release/  too I think.

But anyway https://github.com/openshift/release/pull/14135 was merged so we can close this without this change if you are hesitated to add :smile_cat:.

PR #738, opened 2020-12-08
[release-1.12] Split Strimzi install function to be reusable downstream

mgencur:
We will be replace strimzi operator install with amq_streams install downstream but re-using the other parts.


aliok (commented):
/lgtm
/approve

Looks good!

PR #739, opened 2020-12-08
Fix regex for the Eventing resources dashboard

skonto:
Need to distinguish between a source with name _kafka1_ and another named _kafka2_.
Tested this also by comparing what dev console shows on the Pods section about resource consumption.
Also compared similar dummy sources if have similar values:
image
image
Ideally we could have an integration test for prometheus metric values but gets harder to do.


skonto:
@matzew this should be quick pls review.

aliok (approved):
/lgtm
/approve

PR #741, opened 2020-12-08
Make health status text precise

skonto:
- Previously we reported `Healthy` and `Unhealthy` if any component was not ready. Healthiness definition
can be more complex and since we lack a detailed one eg. should we include reconciliation latency etc to it, we should avoid it for now. As a matter of fact maybe its not practical to express everything via one metric here.
- This also simplifies a few labels to make things cleaner.
- For now knative_up metric is not renamed as it does not do harm, although its a contract we expose to the user, it may be updated in the future.
Example of the updates:
`$ make install-all`
image
`watch oc delete pods --all -n knative-eventing`
 #Eventing and Kafka go down
image 
 #stop killing pods
image

/cc @markito

### DIFF
--- a/knative-operator/deploy/resources/dashboards/grafana-dash-knative-health.yaml
+++ b/knative-operator/deploy/resources/dashboards/grafana-dash-knative-health.yaml
@@ -105,12 +105,12 @@ data:
                  },
                  {
                    "op": "=",
-                   "text": "Healthy",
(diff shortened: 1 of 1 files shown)


skonto:
@aliok @markusthoemmes pls review.

markusthoemmes (approved):
/lgtm
/approve

PR #744, opened 2020-12-09
Update serving lib to 0.19.0

nak3:
This patch updates serving lib to v0.19.0 by following steps:

- Update to `KN_VERSION="release-0.19"` in `hack/update-deps.sh`.
- Update to `knative.dev/serving v0.19.0` in `go.mod`.
- Execute `make generated-files`.

/cc @markusthoemmes


hack/update-deps.sh
```diff
@@ -10,7 +10,7 @@ set -o pipefail
 cd ${ROOT_DIR}
 
 # This controls the knative release version we track.
-KN_VERSION="release-0.18"
+KN_VERSION="release-0.19"
```
markusthoemmes, line 13: Does this actually work if you run `./hack/update-deps.sh --upgrade`? It shouldn't. I believe we need to invest a little since upstream moved to `buoy` for floating the deps and that seems to break our current flow.

/hold

matzew: Doesn't this than have "impact" to other kn libs?

For instance there is no `release-0.19` on eventing-contrib - that's on `knative-sandbox/eventing-kafka` (I was thinking to update that, when I actually can update that (e.g. we have our image builds for the new repo etc etc.)

matzew:
> Execute `make generate-dockerfiles` 

dockerfiles ?

nak3:
> Execute make generate-dockerfiles

Sorry, it is my typo. `generated-files` is correct. Updated.

hack/update-deps.sh
markusthoemmes: Yes it does, and yes that's one of the failures this causes.

You can theoretically update this right away to unblock this specific issue. No need for 0.19 images to land here usually.

markusthoemmes:
`generated-files` does not run `hack/update-deps.sh --upgrade`. That must be done manually.

matzew:
FWIW: here is how I ran it for eventing:
https://github.com/openshift-knative/serverless-operator/pull/745/commits

(I added some sort of commit msg description so that the rest of the eventing team understands the different steps better)

markusthoemmes:
> FWIW: here is how I ran it for eventing:

Commented, that one ain't correct either.

nak3:
Closing as other PRs handling this update.

PR #750, opened 2020-12-10
Yet another hack for the pingsource blocker

matzew:
/cc @maschmid


matzew:
/hold

This has the risk of:  controller changes the `K_....` values to something else, than the operator does roll them back, and the controller resets them, and so on, and so on .

matzew:
Closing to avoid a lot of hackery.

The operator is about to be updated, to ignore the relicas and the ENV_VARs for the PINGSOURCE Deployment...

PR #753, opened 2020-12-11
 CI was here...:fire:

matzew:


nak3:
I think I found the reason of upstream test failing

The test code fails to find `confg-autoscaler`.

```
    autoscale.go:237: Unable to observe the Deployment named autoscale-up-down-up-zmquuuyd-bnscm-deployment scaling down: failed to get autoscaler configmap: configmaps "config-autoscaler" not found
```

This is because `SYSTEM_NAMESPACE` is set to `knative-eventing` instead of `knative-serving`. We need to use `knative-serving` for upstream serving e2e tests.

```
SYSTEM_NAMESPACE=knative-eventing
```

matzew:
/close

was only a test, anyways

PR #756, opened 2020-12-14
Explicitly set SYSTEM_NAMESPACE=knative-serving before running serving e2e test

nak3:
This patch sets `SYSTEM_NAMESPACE=knative-serving` before running serving e2e test.

Currently e2e test is failing as `SYSTEM_NAMESPACE=knative-eventing` was set.
serving and eventing uses both `SYSTEM_NAMESPACE` so it should be explicitly set beforehand
to avoid corruption.

### DIFF
--- a/test/serving.bash
+++ b/test/serving.bash
@@ -57,14 +57,14 @@ function upstream_knative_serving_e2e_and_conformance_tests {
     parallel=2
   fi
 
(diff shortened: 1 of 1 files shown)


nak3:
Fixes https://github.com/openshift-knative/serverless-operator/pull/753#issuecomment-744192484

matzew:
I will see if I need to tweak eventing‘s too (more fine grain)

PR #759, opened 2020-12-15
Replace test-infra with hack completely

markusthoemmes:
Upstream has moved all scripts into `knative.dev/hack`, which is also where they'll be maintained going forward. This allows us to drop the depenedncy on test-infra completely.

/assign @matzew


hack/update-deps.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 readonly ROOT_DIR=$(dirname $0)/..
-source ${ROOT_DIR}/vendor/knative.dev/test-infra/scripts/library.sh
+source ${ROOT_DIR}/vendor/knative.dev/hack/library.sh
```
navidshaikh, line 4:
```suggestion
# shellcheck disable=SC1091,SC1090
source ${ROOT_DIR}/vendor/knative.dev/hack/library.sh
```

matzew:
Thanks for landing this ! 

/lgtm
/hold (to see that all really wokrs)

markusthoemmes:
/unhold

All green.

PR #763, opened 2020-12-16
[SRVKS-648] Add custom domains support for Knative Services

nak3:
This patch adds custom domain support for Knative Services.

The step is same with upstream's.

```
### 1.  Create kservice as usual
$ kn service create helloworld-go --image=gcr.io/knative-samples/helloworld-go

### 2. Create a domainmapping CR
$ cat <<EOF | oc apply -f -
kind: DomainMapping
metadata:
 name: mydomain.com
 namespace: default
spec:
 ref:
   name: hello-example
   kind: Service
   apiVersion: serving.knative.dev/v1
EOF


### 3. Access to the custom domain
$ curl mydomain.com            (or $ curl -H "Host: mydomain.com" $OCP_INGRESS_IP)
```


serving/ingress/pkg/reconciler/ingress/resources/route.go
```diff
@@ -92,7 +95,9 @@ func makeRoute(ci *networkingv1alpha1.Ingress, host string, rule networkingv1alp
 	}
 
 	labels := kmeta.UnionMaps(ci.Labels, map[string]string{
-		networking.IngressLabelKey: ci.GetName(),
+		networking.IngressLabelKey:      ci.GetName(),
+		OpenShiftRouteLabelKey:          ci.GetName(),
+		OpenShiftRouteNamespaceLabelKey: ci.GetNamespace(),
```
markusthoemmes, line 100: Why are we adding these? :thinking:

markusthoemmes (commented):
Well that looked simple enough :sweat_smile:.

Awesome!

serving/ingress/pkg/reconciler/ingress/resources/route.go
nak3: Ingress generated by DomainMapping does not have any label so `serving.knative.dev/{route,routeNamespace}` labels are not added to OCP Route.
Due to this, route lister here https://github.com/openshift-knative/serverless-operator/blob/master/serving/ingress/pkg/reconciler/ingress/ingress.go#L121-L128 cannot list OCP Route for DomainMapping :disappointed_relieved: 

I had a few other ideas:
#### option-a. Use fixed namespace for the lister like:
```
        return r.routeLister.Routes("knative-serving-ingress").List(
                selector,    // networking.IngressLabelKey=ing.GetName()
        )
```
#### b. option-b. Add label to Ingress created by DomainMapping

https://github.com/knative/serving/pull/10370 (But we need to add `serving.knative.dev/domainmappingNamespace` which is not necessary in upstream.)

But I wonder that adding the new label is simple.

markusthoemmes: Most importantly, we can't easily change labels as that'd break potential upgrade cases as the lister can't find the correct routes anymore.

I feel like if anything, we should invest in our component to be less dependent on upstream labels :thinking:. Should we just add our own `ingressName` and `ingressNamespace` labels and don't care about upstream's labels at all?

markusthoemmes: I guess that's kindof what you're doing with the new labels you add. I'd just rename them to be

```
OpenShiftIngressLabelKey          = "serving.knative.openshift.io/ingressName"
OpenShiftIngressNamespaceLabelKey = "serving.knative.openshift.io/ingressNamespace"
```

nak3: Thank you! As discussed on Slack we write bridging code to adjust routeList to do 2 list operations:
1. the one it already has
2. the new one
3. combine the two

nak3:
/hold

I found one minor issue. We still need https://github.com/knative/serving/pull/10370 for the following reconciler.

https://github.com/openshift-knative/serverless-operator/blob/d036615cde9d284ee092b621704743ff608cfa1f/serving/ingress/pkg/reconciler/ingress/controller.go#L54

markusthoemmes:
@nak3 do you mind filing a JIRA to remove the bridging logic for the 1.15 release please?

nak3:
Alright. I filed https://issues.redhat.com/browse/SRVKS-676.

nak3:
Hmm... TestDomainMapping failed. I will fix it.

nak3:
I see. Test is running with `--resolvable` but DomainMapping's custom domain (`domain-mapping-efkuveuw.mapping.com`) is not resolvable, of course... 

```
    spoof.go:171: Retrying http://domain-mapping-efkuveuw.mapping.com: retrying for DNS error: Get "http://domain-mapping-efkuveuw.mapping.com": dial tcp: lookup domain-mapping-efkuveuw.mapping.com on 172.30.0.10:53: no such host
```

markusthoemmes:
Would it be reasonable to single out the DomainMapping tests and run them without the --resolvable flag?

nak3:
Yes, it makes sense. And `TestDomainMapping` passed now :tada:

serving/ingress/pkg/reconciler/ingress/resources/route.go
```diff
@@ -44,7 +47,7 @@ func MakeRoutes(ci *networkingv1alpha1.Ingress) ([]*routev1.Route, error) {
 			// autogenerated name case which is the biggest pain
 			// point.
 			parts := strings.Split(host, ".")
-			if len(parts) > 2 && parts[2] != "svc" {
+			if len(parts) == 2 || (len(parts) > 2 && parts[2] != "svc") {
```
markusthoemmes, line 50: The comment above is now outdated.

test/serving.bash
```diff
@@ -58,10 +58,18 @@ function upstream_knative_serving_e2e_and_conformance_tests {
   fi
 
   SYSTEM_NAMESPACE=knative-serving go_test_e2e -tags=e2e -timeout=30m -parallel=$parallel \
-    ./test/e2e ./test/conformance/api/... ./test/conformance/runtime/... \
+    ./test/e2e ./test/conformance/api/v1/... ./test/conformance/runtime/... \
     --resolvabledomain --kubeconfig "$KUBECONFIG" \
     --imagetemplate "$image_template"
 
+  # Test new features especially DmainMapping without resolvabledomain.
```
markusthoemmes, line 65:
```suggestion
  # Test new features especially DomainMapping without resolvabledomain.
```

serving/ingress/pkg/reconciler/ingress/controller.go
```diff
@@ -56,5 +58,13 @@ func NewController(
 		)),
 	})
 
+	routeInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
```
markusthoemmes, line 61: Good call!

markusthoemmes (approved):
/lgtm
/approve

PR #764, opened 2020-12-16
Update dependencies to their last versions

markusthoemmes:
As per title, bumps a few of our dependencies to more recent versions, to make sure we don't fall behind.
Also drops now overrides which are no longer necessary to make things more maintainable.

/assign @matzew


matzew (approved):
love it!

/lgtm
/approve

PR #766, opened 2020-12-18
Add 4.7 channel

warrenvw:
As suggested, add the 4.7 channel.

### DIFF
--- a/olm-catalog/serverless-operator/project.yaml
+++ b/olm-catalog/serverless-operator/project.yaml
@@ -9,7 +9,8 @@ olm:
   channels:
     default: '4.6'
     list:
-      - '4.6,4.7'
+      - 4.6
+      - 4.7
 
 requirements:
   kube:
--- a/olm-catalog/serverless-operator/project.yaml
+++ b/olm-catalog/serverless-operator/project.yaml
@@ -9,8 +9,8 @@ olm:
   channels:
     default: '4.6'
     list:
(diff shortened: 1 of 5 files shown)


olm-catalog/serverless-operator/project.yaml
```diff
@@ -9,7 +9,7 @@ olm:
   channels:
     default: '4.6'
     list:
-      - '4.6'
+      - '4.6,4.7'
```
markusthoemmes, line 12: This must actually be an enumerated list. This breaks the code figuring out which channel to use for the subscription, see https://github.com/openshift-knative/serverless-operator/blob/28c6716b70c438a15ae747fd6c2e3ce7ed472350/hack/lib/vars.bash#L57.

olm-catalog/serverless-operator/project.yaml
```diff
@@ -9,7 +9,8 @@ olm:
   channels:
     default: '4.6'
     list:
-      - '4.6'
+      - 4.6
+      - 4.7
```
cardil, line 13:
```suggestion
      - '4.6'
      - '4.7'
```
Without quotes they will be read as floats, but they should be strings. I advised adding quotes.

cardil (changes requested):
Nit.

markusthoemmes (approved):
/lgtm
/approve

PR #768, opened 2020-12-18
[SRVKS-658] Secure metric endpoints for Serving control plane pods

skonto:
- Removes some unused files
- Does a minor refactoring on file names for monitoring resources, adds to a proper dir.
- Secures metric endpoints using [rbac proxy](https://github.com/brancz/kube-rbac-proxy).
- Fixes the old security issue described in SRVKS-227. 
image
- The implementation follows the recommendations from MON-127 to use the prometheus account for the bearer token file in order to be on the safe side when it comes to token rotation. 
- For each endpoint we have a service monitor and a service monitor service that targets https port 8844 exposed by rbac proxy that runs as a side car in each pod. Internally rbac proxy proxies port 9090 on `127.0.0.0`.

Resources created:
```
$ oc get servicemonitors.monitoring.coreos.com -n knative-serving
NAME                       AGE
activator-sm               61m
autoscaler-hpa-sm          61m
autoscaler-sm              61m
controller-sm              61m
domain-mapping-sm          61m
domainmapping-webhook-sm   61m
webhook-sm                 61m

$ oc get svc -n knative-serving | grep sm
activator-sm-service               ClusterIP   172.25.52.64     <none>        8444/TCP                            76m
autoscaler-hpa-sm-service          ClusterIP   172.25.47.56     <none>        8444/TCP                            76m
autoscaler-sm-service              ClusterIP   172.25.108.225   <none>        8444/TCP                            76m
controller-sm-service              ClusterIP   172.25.97.151    <none>        8444/TCP                            76m
domain-mapping-sm-service          ClusterIP   172.25.217.36    <none>        8444/TCP                            76m
domainmapping-webhook-sm-service   ClusterIP   172.25.11.177    <none>        8444/TCP                            76m
webhook-sm-service                 ClusterIP   172.25.95.150    <none>        8444/TCP                            76m

$ oc get pods -n knative-serving
NAME                                                     READY   STATUS      RESTARTS   AGE
activator-7b68656785-lcw77                               2/2     Running     0          87m
activator-7b68656785-xks7w                               2/2     Running     0          86m
autoscaler-58c84f84df-twzt9                              2/2     Running     0          87m
autoscaler-hpa-ffd6486d8-cpfwq                           2/2     Running     0          86m
autoscaler-hpa-ffd6486d8-qp775                           2/2     Running     0          86m
controller-7f4d754bd4-grm2p                              2/2     Running     0          83m
controller-7f4d754bd4-z9s4g                              2/2     Running     0          86m
domain-mapping-76c9898f4f-gkf8r                          2/2     Running     0          86m
domainmapping-webhook-75cf969768-bx5sk                   2/2     Running     0          86m
storage-version-migration-serving-serving-0.19.0-9s879   0/1     Completed   0          86m
webhook-657558f9bc-256cc                                 2/2     Running     0          86m
webhook-657558f9bc-6wz6b                                 2/2     Running     0          86m


```
Targets are up:
image
Metrics exposed as expected here we have autoscaler metrics and one ksvc deployed:
image
- Modifies the serving resource files so validation fails.
/cc @markusthoemmes @nak3


skonto:
@aliok @matzew pls review.

skonto:
Getting:
```
time="2021-01-04T13:18:27Z" level=info msg="Loading kube client config from path \"/go/src/github.com/openshift-knative/serverless-operator/user3.kubeconfig\""
    kafka_channel_ksvc_test.go:100: Unable to create KafkaChannel:  Internal error occurred: failed calling webhook "validation.webhook.kafka.messaging.knative.dev": Post "https://kafka-webhook.knative-eventing.svc:443/resource-validation?timeout=28s": no endpoints available for service "kafka-webhook"
    service.go:68: Cleaning up Knative Service 'serverless-tests/helloworld-go'
--- FAIL: TestSourceToKafkaChanelToKnativeService (17.34s)
FAIL
FAIL	github.com/openshift-knative/serverless-operator/test/extensione2e/kafka	17.459s
=== Failed
=== FAIL: test/extensione2e/kafka TestSourceToKafkaChanelToKnativeService (17.34s)
```

skonto:
@markusthoemmes have you seen this fail before? I discussed with the registry team but dont see any reason for the image not to be there.

openshift-knative-operator/cmd/operator/kodata/knative-serving/0.19.0/5-serving-domainmapping.yaml
```diff
@@ -236,6 +243,13 @@ spec:
               weight: 100
       serviceAccountName: controller
       containers:
+        - name: kube-rbac-proxy
+          image: quay.io/brancz/kube-rbac-proxy:v0.7.0
```
markusthoemmes, line 247: Is this image overridable via the CSV? We have to replace this for productization with a productized image.

knative-operator/deploy/resources/monitoring/rbac-proxy-roles.yaml
```diff
@@ -0,0 +1,21 @@
+---
```
markusthoemmes, line 1: Should we make this part of the serving manifest?

knative-operator/deploy/operator.yaml
```diff
@@ -1,43 +0,0 @@
-apiVersion: apps/v1
```
markusthoemmes, line 1: Why remove this? Seems like an unrelated change.

knative-operator/pkg/common/service_monitor.go
```diff
@@ -122,6 +123,93 @@ func SetupSourceServiceMonitor(client client.Client, instance *appsv1.Deployment
 	return nil
 }
 
+func SetupServingControlPlaneServiceMonitors(api client.Client, instance *v1alpha1.KnativeServing) error {
```
markusthoemmes, line 126: Can all this be part of the serving manifest?

knative-operator/deploy/operator.yaml
skonto: It is part of the refactoring and it seems a relic of the past.

openshift-knative-operator/cmd/operator/kodata/knative-serving/0.19.0/5-serving-domainmapping.yaml
skonto: @markusthoemmes afaik this is the productized version of rbac-proxy, at least this what I was suggested to use? @s-urbaniak  can you verify this?

knative-operator/deploy/resources/monitoring/rbac-proxy-roles.yaml
skonto: Which manifest?

knative-operator/pkg/common/service_monitor.go
skonto: I am not sure I am following. We need to set owner references, install/uninstall stuff based on whether serving cr is present so it cant just be a file applied. If you mean move it to the side where we use the extension controller this can follow as a refactoring.

knative-operator/pkg/common/service_monitor.go
```diff
@@ -122,6 +123,93 @@ func SetupSourceServiceMonitor(client client.Client, instance *appsv1.Deployment
...
+				}},
+			NamespaceSelector: monitoringv1.NamespaceSelector{
+				MatchNames: []string{sm.Namespace},
+			},
+			Selector: metav1.LabelSelector{
+				MatchLabels: map[string]string{"name": sName + "-sm-service"},
+			}}
+
+		if err := api.Create(context.Background(), sm); err != nil {
+			if err == metrics.ErrServiceMonitorNotPresent {
```
markusthoemmes, line 172: I don't think this error will be returned :thinking:. That error might be returned by some osdk helpers, but not by the generic API. You should get a more generic ResourceDoesntExist error here.

openshift-knative-operator/cmd/operator/kodata/knative-serving/0.19.0/2-serving-core.yaml
```diff
@@ -248,7 +248,33 @@ roleRef:
   kind: ClusterRole
   name: knative-serving-admin
   apiGroup: rbac.authorization.k8s.io
-
+---
```
markusthoemmes, line 251: Can we put these in separate files? (Along with the YAML above, that can move into the "manifest" itself too I believe)

skonto: ok

openshift-knative-operator/cmd/operator/kodata/knative-serving/0.19.0/5-serving-domainmapping.yaml
skonto: I will make it configurable.

knative-operator/pkg/common/service_monitor.go
skonto: Yes I will update this.

openshift-knative-operator/cmd/operator/kodata/knative-serving/0.19.0/5-serving-domainmapping.yaml
skonto: I will make it configurable sure.

skonto:
Quota issue due to some updates at the infra side.
```
level=info msg="Creating infrastructure resources..."
level=error
level=error msg="Error: Error creating IAM Role ci-op-grrlpfdk-19129-bl9kl-bootstrap-role: LimitExceeded: Cannot exceed quota for RolesPerAccount: 1000"
```

skonto:
@markusthoemmes besides the productization of the rbax-proxy image I addressed the previous comments.
Not sure why upgrade tests fail. Feel free to review.

openshift-knative-operator/cmd/operator/kodata/knative-serving/0.19.0/7-rbac-proxy.yaml
```diff
@@ -0,0 +1,293 @@
...
+  namespace: knative-serving
+spec:
+  endpoints:
+    - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      bearerTokenSecret:
+        key: ""
+      port: https
+      scheme: https
+      tlsConfig:
+        insecureSkipVerify: true
```
markusthoemmes, line 287: If we're skipping the verification everywhere, should we just serve HTTP from the RBAC proxy?

hack/generate/csv.sh
```diff
@@ -37,12 +38,19 @@ function kafka_image {
 
 image "queue-proxy"    "${serving}-queue"
 image "activator"      "${serving}-activator"
+image "activator__kube-rbac-proxy"  "$rbac_proxy"
```
markusthoemmes, line 41: Potential refactor once this is in: We could write a special transformer that picks up an `IMAGE_kube-rbac-proxy` variable and just replaces that in every container that's named `kube-rbac-proxy` to not have to repeat this a gazilion times.

hack/generate/csv.sh
```diff
@@ -37,12 +38,19 @@ function kafka_image {
...
 image "autoscaler"     "${serving}-autoscaler"
+image "autoscaler__kube-rbac-proxy"  "$rbac_proxy"
 image "autoscaler-hpa" "${serving}-autoscaler-hpa"
+image "autoscaler-hpa__kube-rbac-proxy"  "$rbac_proxy"
 image "controller"     "${serving}-controller"
+image "controler__kube-rbac-proxy"  "$rbac_proxy"
```
markusthoemmes, line 47:
```suggestion
image "controller__kube-rbac-proxy"  "$rbac_proxy"
```

knative-operator/pkg/common/service_monitor.go
```diff
@@ -20,9 +20,9 @@ import (
 )
 
 const (
-	EventingBrokerServiceMonitorPath     = "deploy/resources/broker-service-monitors.yaml"
-	EventingSourceServiceMonitorPath     = "deploy/resources/source-service-monitor.yaml"
-	EventingSourcePath                   = "deploy/resources/source-service.yaml"
+	EventingBrokerServiceMonitorPath     = "deploy/resources/monitoring/broker-service-monitors.yaml"
+	EventingSourceServiceMonitorPath     = "deploy/resources/monitoring/source-service-monitor.yaml"
+	EventingSourcePath                   = "deploy/resources/monitoring/source-service.yaml"
```
markusthoemmes, line 25: Why these changes in here? Isn't this only about serving?

markusthoemmes (commented):
Looks great, thanks!

hack/generate/csv.sh
skonto: Yeah I thought about so I was not sure about where to have that. I saw that the manifest is appended at the operator side at reconcile phase. If a transformer at the extension side works then yes.

knative-operator/pkg/common/service_monitor.go
skonto: its part of the refactoring I did.

openshift-knative-operator/cmd/operator/kodata/knative-serving/0.19.0/7-rbac-proxy.yaml
skonto: This is for the ca verification, we want https to encrypt stuff. I will check if I can remove that.

hack/generate/csv.sh
markusthoemmes: Yeah, transformer on the extension side should work.

knative-operator/pkg/common/service_monitor.go
markusthoemmes: I'll swallow this here, but generally please keep PRs with separate concerns separate. It only makes things more contentious.

openshift-knative-operator/cmd/operator/kodata/knative-serving/0.19.0/7-rbac-proxy.yaml
markusthoemmes: well, skipping CA verification opens the door for MITM attacks, which generally renders the encryption useless.

skonto: @markusthoemmes how do we generate certificates for the rest of our stuff maybe we can re-use that.

skonto: Btw this is not a blocker unless we dont trust the rbac-proxy which is too much I think.

markusthoemmes: Sure it's not a blocker, I was just wondering why we don't just do plain HTTP and spare us the extra config

s-urbaniak: If your certs are generated by the internal service CA operator (generated via annotations) then we have the CA available in the client (prometheus). See https://github.com/openshift/cluster-monitoring-operator/blob/6ba1db4ca73eca18c28acef3b865a892fa26e719/assets/alertmanager/service-monitor.yaml#L14-L16 as an example (just random).

You just have to set the serverName of your TLS kube-rbacy-proxy endpoint.

Putting my SRE hat on: For me (personally) this should be a blocker if it is fixable in here.

s-urbaniak: And this is the annotation I talked about above: https://github.com/openshift/cluster-monitoring-operator/blob/6ba1db4ca73eca18c28acef3b865a892fa26e719/assets/alertmanager/service.yaml#L4-L5

It will create server public/private key in a secret called `alertmanager-main-tls`.

markusthoemmes: For the purpose of this PR, it is fine with me to delay the HTTPS stuff to an immediate followup. We should at least record it in JIRA and make it a blocker for 1.13 though.

skonto: @markusthoemmes plain http does not provide encryption. Better secure it end-to-end imho.

markusthoemmes: With the service CA operator mentioned by @s-urbaniak

skonto: Ok let me try and see if I can make it work and update the PR.

knative-operator/pkg/common/service_monitor.go
skonto: Ok I hear you.

openshift-knative-operator/cmd/operator/kodata/knative-serving/0.19.0/7-rbac-proxy.yaml
skonto: @s-urbaniak I dont think the secret is created when I needed do I have to wait for it? Right now we have some yaml files that we apply in order to create the service monitors, services and the deployments.
Also do I need to set these args https://github.com/openshift/cluster-monitoring-operator/blob/6ba1db4ca73eca18c28acef3b865a892fa26e719/assets/alertmanager/alertmanager.yaml#L66-L67 when using the kube-rbac-proxy so that it uses the certificate that is mounted?

skonto: @s-urbaniak is it ok not to use a serverName in the service monitor tlsconfig, I am getting 
```
Get "https://10.116.0.94:8444/metrics": x509: certificate is valid for autoscaler-hpa-644c88686d-cm8zn, not autoscaler-hpa-sm-service.knative-serving.svc
```
Reason is I dont know the name of the pods behind a deployment and they are not static.

skonto: Figured it out will update the PR shortly, I was getting the default certs instead of the service ones.

skonto:
@markusthoemmes I added certificate support. Move also the yaml files for rbac proxy to be applied first to so that secrets are created first before they are needed to be mounted by the deployments.
One downside of moving all stuff in kodata is that we dont have a check if the monitoring resources exist at all. We do that when we setup the sm in code. At the transformer side we should have a check to avoid load the yaml files in the final manifest before that is applied by the operator.

markusthoemmes (approved):
/lgtm
/approve

PR #776, opened 2021-01-18
Add 4.7 channel

markusthoemmes:
As per title. To make it available on master (and thus upcoming releases) too.

### DIFF
--- a/olm-catalog/serverless-operator/metadata/annotations.yaml
+++ b/olm-catalog/serverless-operator/metadata/annotations.yaml
@@ -1,6 +1,6 @@
 annotations:
   operators.operatorframework.io.bundle.channel.default.v1: "4.6"
-  operators.operatorframework.io.bundle.channels.v1: "4.6"
+  operators.operatorframework.io.bundle.channels.v1: 4.6,4.7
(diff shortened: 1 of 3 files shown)


olm-catalog/serverless-operator/metadata/annotations.yaml
```diff
@@ -1,6 +1,6 @@
 annotations:
   operators.operatorframework.io.bundle.channel.default.v1: "4.6"
-  operators.operatorframework.io.bundle.channels.v1: "4.6"
+  operators.operatorframework.io.bundle.channels.v1: 4.6,4.7
```
rhuss, line 3: why not keep the quotes, so it's not forgotten when moving to a pure 4.7 channel (so that quotes are needed again to distinguish it from a float) ?

rhuss (commented):
little nit, otherwise good to me. However I can not really judge whether you might have missed some part.

olm-catalog/serverless-operator/metadata/annotations.yaml
markusthoemmes: This is generated code. When we move to pure 4.7, they'll be readded automatically.

PR #777, opened 2021-01-19
[SRVKS-680] Serving metrics integration tests

skonto:
- Adds integration test for knative kafka readiness status
- Adds integration tests for the secured metric endpoints of the control plane pods


test/e2e/helpers.go
```diff
@@ -64,3 +72,126 @@ func verifyHealthStatusMetric(caCtx *test.Context, metricsPath string, metricLab
...
+		metricsURL, err := url.Parse(metricsPath)
+		if err != nil {
+			caCtx.T.Fatalf("Error parsing url for metrics: %v", err)
+		}
+		bToken := getBearerTokenForAuthorizedAccount(caCtx)
+		var reqOption pkgTest.RequestOption = func(request *http.Request) {
+			request.Header.Set("Authorization", fmt.Sprintf("Bearer %s", bToken))
+		}
+		var transportOption spoof.TransportOption = func(transport *http.Transport) *http.Transport {
+			transport.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
```
skonto, line 152: Using the `reencrypt` option here plus avoiding validating the crt at the route. The connection from the route to the service is validated from the OCP check this [great source](https://www.openshift.com/blog/self-serviced-end-to-end-encryption-approaches-for-applications-deployed-in-openshift) for more info. Also in the case of a service using a service signing certificate (as in our case) there is no need to have a destination CA for the route.
Passthrough would not work here as we dont have a certifacate with two-SANS (we only generate certs for cluster internal use) and using this option will cause an error:
```
TestKnativeServing: spoof.go:175: NOT Retrying https://activator-sm-service-knative-serving.apps.ci-ln-98rqthb-d5d6b.origin-ci-int-aws.dev.rhcloud.com/metrics: Get "https://activator-sm-service-knative-serving.apps.ci-ln-98rqthb-d5d6b.origin-ci-int-aws.dev.rhcloud.com/metrics": x509: certificate is valid for activator-sm-service.knative-serving.svc, activator-sm-service.knative-serving.svc.cluster.local, not activator-sm-service-knative-serving.apps.ci-ln-98rqthb-d5d6b.origin-ci-int-aws.dev.rhcloud.com```

File: test/e2e/helpers.go [hunk 2]
```diff
@@ -64,3 +72,126 @@ func verifyHealthStatusMetric(caCtx *test.Context, metricsPath string, metricLab
...
+			return transport
+		}
```
markusthoemmes, line 154: Could we test this E2E through prometheus as well? After all, it's important for the customer that our metrics end up in prometheus correctly, so they can be queried by Grafana etc and the setup there is critical as prometheus needs to be given the correct RBAC and whatnot. Does that make sense?

markusthoemmes: Rather than "as well" it might be sufficient to **just** test through prometheus if I think about it.

skonto: There is no Grafana support anymore. I thought about querying Prometheus but that needed some investigation how to access it.

markusthoemmes: we can still pull this in, but I think the test is kind of incomplete since the crucial part (to me as a reviewer) is knowing that the connection between prometheus (the thin scraping the metrics) and the actual metric endpoints is working as intended.

skonto: One drawback with the Prometheus being used (although it needs to be tested) is that it takes some time to discover targets.

skonto:
```
The connection to the server api.ci-op-14ihfimp-19129.origin-ci-int-aws.dev.rhcloud.com:6443 was refused - did you specify the right host or port?
14:38:41.998 ERROR:   🚨 Error (code: 1) occurred at /go/src/github.com/openshift-knative/serverless-operator/hack/lib/namespaces.bash:7, with command: oc create namespace "${ns}"````

**Comment by skonto** [comment]
@markusthoemmes updated the PR using the Prometheus endpoint.

File: test/e2e/helpers.go [hunk 3]
```diff
@@ -2,65 +2,166 @@ package e2e
...
-		return nil, fmt.Errorf("error creating service monitor route: %v", err)
+const servingNamespace = "knative-serving"
+
+type PrometheusInfo struct {
+	options   []interface{}
+	queryPath string
+	token     string
+}
+
+func NewPrometheusInfo(caCtx *test.Context) *PrometheusInfo {
```
markusthoemmes, line 28: Doesn't need to be public I think, so `newPrometheusInfo` should be fine, I believe. The same goes for the `PrometheusInfo` type.

test/e2e/helpers.go
```diff
@@ -2,65 +2,166 @@ package e2e
...
+		if err != nil {
+			caCtx.T.Fatalf("Failed to access the Prometheus API endpoint and get the metric value expected: %v", err)
+		}
+	}
+}
+
+func getBearerTokenForPrometheusAccount(caCtx *test.Context) string {
+	sa, err := caCtx.Clients.Kube.CoreV1().ServiceAccounts("openshift-monitoring").Get(context.Background(), "prometheus-k8s", meta.GetOptions{})
+	if err != nil {
+		caCtx.T.Fatalf("Error getting service account prometheus-k8s: %v", err)
```
markusthoemmes, line 110: It feels a little unexpected to me, that these helper functions take the entire test.Context and fail the test immediately. How'd you feel about making them take `kube.Interface` (the client) and returns `(string, error)` and then fail the test in the `VerifyXXX` functions.

test/e2e/helpers.go
```diff
@@ -2,65 +2,166 @@ package e2e
...
+	if len(tokenContents) == 0 {
+		caCtx.T.Fatalf("Token data is missing for token %s", *tokenSecret)
+	}
+	return string(tokenContents)
+}
+
+func getSecretNameForToken(secrets []corev1.ObjectReference) *string {
+	for _, sec := range secrets {
+		if strings.Contains(sec.Name, "token") {
+			return &sec.Name
```
markusthoemmes, line 130: nit: I'd return an empty string on "not found" and not make this a pointer.

test/e2e/helpers.go
markusthoemmes, line 28: Similar to below, I'd rather pass the K8s client in here and return an error to make the code flow a bit clearer.

markusthoemmes (commented):
This is **awesome**! This will make our test coverage of metrics soooo much better, thanks!

markusthoemmes:
> updated the PR using the Prometheus endpoint. Targets take up to 4 minutes to come up (it should not become flaky due to the spoof timeout). Example run:

We could think about moving these tests to run after 

```
downstream_serving_e2e_tests
downstream_eventing_e2e_tests
```

These tests should give prometheus plenty of time to look up the targets without having to wait a few minutes for them to appear.

test/e2e/helpers.go
skonto: I thought about this but wanted just to fail as soon as I cannot move forward don think it is unexpected in this context.

test/e2e/helpers.go
skonto: It is the old question if we should return a pointer for string, faced it upstream, the truth is k8s code base uses it a lot, and it is not a anti-pattern, I dont think it needs a change.

test/e2e/helpers.go
skonto: Yeah left over from a refactoring.

test/e2e/helpers.go
markusthoemmes: wfm!

test/e2e/helpers.go
markusthoemmes: It was unexpected to me and does hamper the pluggability of the code, but not a strong opinion, it's fine to leave as-is.

skonto:
Tests failed it takes longer than the spoof time out

skonto:
@markusthoemmes I addressed the comments above. I increased the spoof timeout as the test suites remove the operator installation so cant wait after integration tests are run.

skonto:
Not sure why this passes with a random cluster and not on the CI, it seems that it never returns a value for `autoscaler_actual_pods`, I will update the list of the metrics and the error msg to see why it gets stuck.

markusthoemmes:
> I increased the spoof timeout as the test suites remove the operator installation so cant wait after integration tests are run.

I don't think that is the case for the other test suite though. We could remove the current tests from the operator test suite and put them into their own package, which we then run at the end of the downstream e2e suite.

skonto:
> I don't think that is the case for the other test suite though. We could remove the current tests from the operator test suite and put them into their own package, which we then run at the end of the downstream e2e suite.

Semantically the metrics should be checked when each component is started and in that suite we test Serving anyway. If you mean after [here](https://github.com/openshift-knative/serverless-operator/blob/master/test/e2e-tests.sh#L32) yes its doable.

markusthoemmes:
> Moving the tests in a separate test suite may implicitly gives us time but it is error-prone eg. if someone changes the order of calls there (you need to remember these conventions), so I would still set a proper time-out. If we move the tests for separation of concerns then yes its doable but not sure if it is a priority at this point.

We can add a comment stating that these should run last to keep that invariant intact. Yes, we'd still have to set a proper timeout, but we'd at least reduce the impact on the overall runtime of the test suite, especially for local use-cases.

markusthoemmes:
FWIW, i'm fine relaying the test reordering to a followup.

skonto:
@markusthoemmes ok I will put it in another test suite, first need to figure out the diff between CI vs a random testing cluster.

skonto:
This fails again, metric `activator_client_results` is in the [metrics dump](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-knative_serverless-operator/777/pull-ci-openshift-knative-serverless-operator-master-4.6-operator-e2e-aws-ocp-46/1352205560858021888/artifacts/operator-e2e-aws-ocp-46/gather-extra/metrics/prometheus-target-metadata.json) and scraping does work since the test for health status passes.
I suspect it takes longer than 10 minutes (this weird)?

skonto:
Time out needed tuning on CI:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-knative_serverless-operator/777/pull-ci-openshift-knative-serverless-operator-master-4.6-operator-e2e-aws-ocp-46/1352327293590048768/artifacts/operator-e2e-aws-ocp-46/operator-e2e/container-logs/test.log
```
clients.go:160: Failed to clean up: knativeservings.operator.knative.dev "knative-serving" not found
=== CONT  TestKnativeServing
subscription.go:61: Cleaning up CSV 'openshift-serverless/serverless-operator.v1.13.0'
subscription.go:74: Cleaning up Subscription 'openshift-serverless/serverless-operator-subscription'
--- PASS: TestKnativeServing (293.27s)
--- PASS: TestKnativeServing/create_subscription_and_wait_for_CSV_to_succeed (30.68s)
--- PASS: TestKnativeServing/deploy_knativeserving_cr_and_wait_for_it_to_be_ready (70.56s)
--- PASS: TestKnativeServing/verify_health_metrics_work_correctly (5.68s)
--- PASS: TestKnativeServing/verify_control_plane_metrics_work_correctly (163.67s)
--- PASS: TestKnativeServing/verify_correct_deployment_shape (0.90s)
--- PASS: TestKnativeServing/make_sure_no_gcr.io_references_are_there (0.97s)
--- SKIP: TestKnativeServing/update_global_proxy_and_verify_calls_goes_through_proxy_server (0.00s)
--- PASS: TestKnativeServing/remove_knativeserving_cr (20.36s)
--- PASS: TestKnativeServing/undeploy_serverless_operator_and_check_dependent_operators_removed (0.43s)
PASS
ok  	github.com/openshift-knative/serverless-operator/test/e2e	882.464s
```
I will move the tests to a separate suite.

skonto:
@markusthoemmes I think this is ready to merge.

markusthoemmes (approved):
/lgtm
/approve

PR #778, opened 2021-01-20
Generate clients to a central pkg repo and reuse them

markusthoemmes:
As per title, since we've moved to a common vendor directory, we might as well start sharing generated clients (and potentially more helpers in the future). Saves duplication and makes code generation quicker, so... YAY?

/assign @nak3 @rhuss @matzew


nak3:
/test 4.7-images
/lgtm
/hold

I set `hold` just in case. @markusthoemmes please feel free to unhold this.

rhuss:
@markusthoemmes sorry, I need a bit more of context. The 'common vendor directory' is within this repo or shared across repos ? I gues the former, so that this PR is about generating the client once for all sub-builds in this repo instead of having each sub build create the client on his own ?

markusthoemmes:
@rhuss exactly what you said!

hack/update-codegen.sh
```diff
@@ -0,0 +1,31 @@
...
+# If we run with -mod=vendor here, then generate-groups.sh looks for vendor files in the wrong place.
+export GOFLAGS=-mod=
+
+if [ -z "${GOPATH:-}" ]; then
+  export GOPATH
+  GOPATH=$(go env GOPATH)
+fi
+
+REPO_ROOT=$(dirname "${BASH_SOURCE[@]}")/..
+CODEGEN_PKG=${CODEGEN_PKG:-$(cd "${REPO_ROOT}"; ls -d -1 "${REPO_ROOT}/vendor/k8s.io/code-generator" 2>/dev/null || echo ../code-generator)}
```
rhuss, line 17:
```suggestion
CODEGEN_PKG=${CODEGEN_PKG:-$(cd "${REPO_ROOT}"; ls -d -1 "${REPO_ROOT}/vendor/k8s.io/code-generator" 2>/dev/null || echo $(pwd)/../code-generator)}
```

to make it an absolute path that works in any context ? (same for the next line)

Other question: Why `cd`ing into  REPO_ROOT when `ls` is called with an absolute path anyway ?

Maybe something like this is more straight forward (and needs no external shell call, checking the error which can be also something different than just a "not exist"):

```
if [ -z "$COGEDGEN_PKG" ]; then
  CODEGEN="${REPO_ROOT}/vendor/k8s.io/code-generator"
  if [ ! -d $CODEGEN ]; then
     CODEGEN="${REPO_ROOT}/../code-generator"
  fi
fi
```

However, I wonder whether you should be that flexible anyway. Maybe a bit more opinionated, breaking with exit 1 if the directories are not setup properly ? That way you can ensure also that the convention are kept more closely.

markusthoemmes: The script is a 1:1 copy from upstream :shrug:  I can take a look at it.

rhuss: no worries if you are not in it. it works, could just be much simpler for our use case (i.e. we already know where to generate the data so we don't need to evaluate different options.

so don't wast time on this if we want to stay in sync, but maybe you could put a reference to the upstream script as comment to this script ?

markusthoemmes: you're right though, this can be simplified a ton. Will do that, I just didn't think about it before and blindly copied :joy:

markusthoemmes: Done, this should be much simpler. We don't actually need the fancy relative paths since we always want the vendor version to be used.

rhuss (approved):
/lgtm
/approve

Thanks !

PR #782, opened 2021-01-22
 Point images to the 'new' registry

markusthoemmes:
The "svc" based registry seems to be deprecated, images have been promoted to the one in the PR. We probably need to fix this throughout the S-O soonish.


matzew:
I always get confused with those :-) 

Is the _old_ (`registry.svc.ci...`) getting removed anytime soon?  I guess on the image mirroring itself there is no change needed, right ? 

But I guess... a good thing to do ... I should do that on eventing as well

matzew:
what about things like: 
https://github.com/openshift-knative/serverless-operator/blob/release-1.13/olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml#L306 

something like `docker pull registry.ci.openshift.org/openshift/knative-v0.19.0:knative-serving-activator` works fine on my machine

markusthoemmes:
@matzew yeah we should certainly revisit the testplatform's teams here and take an approach that's universally accepted.

For now, this fixes the immediate bug we have in that the images changes aren't even available on the old reigstry.

PR #783, opened 2021-01-25
Remove deprecated timeout setting for OpenShift Route

nak3:
This patch removes timeout setting from DeprecatedTimeout

Timeout in Kingress was deprecated and `DeprecatedTimeout` was not used anymore.
For OpenShift Route, it should be safe to set it from `DefaultMaxRevisionTimeoutSeconds`.

/cc @markusthoemmes


markusthoemmes (approved):
/lgtm
/approve

PR #784, opened 2021-01-25
Bump K8s deps to 1.19.7

markusthoemmes:
As per title. We have a bit of a tension between our target OCP version (which is based on K8s 1.19) and Knative's K8s versions (which will switch to 1.19 in the 0.21 release).

I'm doing this to be able to pull in the ImageStream controller from OCP, which I didn't get to work properly on 1.18 dependencies.

Personally, I think if the APIs are compatible and the tests are happy, there's little risk in using one or the other.

/assign @rhuss @matzew


matzew:
I am generally OK with this update.
/lgtm
/approve
/hold

Just holding to get all (non required) tests a chance to pass as well :smile_cat:

rhuss:
If I understand correctly this is an update on the operator only, not the Knative controllers, true ? So I'm +1 as long as the test are going through. As these are deps on a K8s client level, I'd expect when it compiles, it works (e.g. only breaks when API changes).

markusthoemmes:
/unhold

All passed :)

Right @rhuss, this is just operator.

PR #785, opened 2021-01-26
Fix all shellcheck issues and add it to local lint target

markusthoemmes:
As per title, this

1. Lints all shell files
2. Fixes all existing issues
3. Adds shellcheck to `make lint` so it can be run locally easily

/assign @cardil

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -83,10 +83,7 @@ jobs:
         env:
           REVIEWDOG_GITHUB_API_TOKEN: ${{ github.token }}
         run: |
-          find . -type f \
-            -path './hack/*.*sh' \
-            -o -path './test/*.*.sh' \
-            -not -path '*vendor*' | xargs -r shellcheck --format=checkstyle \
+          find . -type f -path './**/*.*sh' -not -path '*vendor*' | xargs -r shellcheck --format=checkstyle \
             | reviewdog -f=checkstyle \
                 -name="shellcheck" \
                 -reporter="github-pr-review" \
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -83,7 +83,9 @@ jobs:
         env:
           REVIEWDOG_GITHUB_API_TOKEN: ${{ github.token }}
         run: |
-          find . -type f -path './**/*.*sh' -not -path '*vendor*' | xargs -r shellcheck --format=checkstyle \
+          find . -type f \
+            -path './**/*.*sh' \
+            -not -path '*vendor*' | xargs -r shellcheck --format=checkstyle \
             | reviewdog -f=checkstyle \
                 -name="shellcheck" \
                 -reporter="github-pr-review" \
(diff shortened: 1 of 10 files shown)


test/lib.bash
```diff
@@ -389,13 +389,14 @@ function create_htpasswd_users {
   if kubectl get secret htpass-secret -n openshift-config -o jsonpath='{.data.htpasswd}' 2>/dev/null | base64 -d > users.htpasswd; then
     logger.info 'Secret htpass-secret already existed, updating it.'
     # Add a newline to the end of the file if not already present (htpasswd will butcher it otherwise).
+    # shellcheck disable=SC1003
     sed -i -e '$a\' users.htpasswd
```
cardil, line 393: I saw that. I don't know it that a problem or not. It looks weird, like a bug.

Warren wrote that. @warrenvw, maybe you remember what that suppose to achieve?

.github/workflows/validate.yaml
```diff
@@ -83,10 +83,7 @@ jobs:
         env:
           REVIEWDOG_GITHUB_API_TOKEN: ${{ github.token }}
         run: |
-          find . -type f \
-            -path './hack/*.*sh' \
-            -o -path './test/*.*.sh' \
-            -not -path '*vendor*' | xargs -r shellcheck --format=checkstyle \
+          find . -type f -path './**/*.*sh' -not -path '*vendor*' | xargs -r shellcheck --format=checkstyle \
```
cardil, line 86: Why this reformatting here?

hack/lib/catalogsource.bash
```diff
@@ -176,6 +176,7 @@ function add_user {
   logger.info "Creating user $name:***"
   if kubectl get secret htpass-secret -n openshift-config -o jsonpath='{.data.htpasswd}' 2>/dev/null | base64 -d > users.htpasswd; then
     logger.debug 'Secret htpass-secret already existed, updating it.'
+    # shellcheck disable=SC1003
```
cardil, line 179: Same as the other comment.

cardil (commented):
Cool :heart:!

Some doubts/questions...

test/lib.bash
markusthoemmes: Well it achieves what the comment states. A quick google told me that this is indeed the notation for that in `sed` (I doublechecked this too)

.github/workflows/validate.yaml
markusthoemmes: easier to read (imo anyway)

test/lib.bash
cardil: Maybe better would be to rewrite that to simpler and POSIX compliant alternative: https://unix.stackexchange.com/a/263965/145501

.github/workflows/validate.yaml
cardil: For me the opposite is true. The original is easier to read and understand. :rofl:

test/lib.bash
markusthoemmes: Indeed, changing!

.github/workflows/validate.yaml
markusthoemmes: Changed :)

cardil (approved):
/lgtm

PR #795, opened 2021-01-29
Add operator-sdk tool for linting the bundle

markusthoemmes:
This currently catches issues as late as us productizing and pushing the images... much too late! Let's do it in our CI!

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -64,6 +64,7 @@ jobs:
         working-directory: ./src/github.com/${{ github.repository }}
         env:
           WOKE_VERSION: v0.4.1
+          OSDK_VERSION: v1.3.0
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/validate.yaml
```diff
@@ -64,6 +64,7 @@ jobs:
         working-directory: ./src/github.com/${{ github.repository }}
         env:
           WOKE_VERSION: v0.4.1
+          OSDK_VERSION: v1.3.0
```
skonto, line 67: We will need to update this often? Or this part is pretty stable?

markusthoemmes: I'd hope it's stable. 1.0.0 had the very same failures before our fix.

skonto: The only corner case would be current version to ignore some new feature we add from a a future release of OLM. But I guess it would complain.

PR #797, opened 2021-02-01
[SRVKS-679] Disable metrics by default

skonto:
- Reverts rbac proxy related stuff (Otherwise prometheus targets will be down if we disable metrics by default, since Prometheus will not be able scrape the pods although service monitors will be setup. At the end of the day this will break QE tests)
- Set metrics backend to none

If the [PR upstream](https://github.com/knative/pkg/pull/2005) get's merged we can backport that one and avoid reverting stuff here. On master I will fix the issue with targets going down so we can freely disable/enable monitoring if something comes up in the future.

/cc @markusthoemmes @maschmid


openshift-knative-operator/pkg/serving/extension.go
```diff
@@ -17,7 +17,7 @@ import (
 )
 
 const loggingURLTemplate = "https://%s/app/kibana#/discover?_a=(index:.all,query:'kubernetes.labels.serving_knative_dev%%5C%%2FrevisionUID:${REVISION_UID}')"
-const observabilityCMName = "observability"
+const observabilityCMName = "config-observability"
```
markusthoemmes, line 20: This is not correct. While the CM is indeed named `config-observability`, the `config-` prefix is swallowed in the CR.

skonto: Yeah I will update it.

markusthoemmes (approved):
/lgtm
/approve

PR #798, opened 2021-02-02
Add docs how to enable Istio sidecar on Knative Serving system pods

nak3:
As per title, this PR adds docs how to enable Istio sidecar on Knative Serving system pods.


nak3:
I don't strongly demand to merge this documentation. But I wonder that I should publish this kind of unofficial instruction doc somewhere.

PR #799, opened 2021-02-02
Increase parallel to make sure the Prober test actually runs

mgencur:
* without parallel the Autoscale test could run first and block the
Prober test, resulting in all the "probing" tests to fail. 

Fixes an issues like this one: https://master-jenkins-csb-serverless-qe.apps.ocp4.prod.psi.redhat.com/job/functional_tests/job/stream1_13/job/rolling-upgrade-1.13/6/

In the next release (1.14) we'll use the new framework for upgrade tests where this will be fixed in a different way.


markusthoemmes (approved):
/lgtm
/approve

PR #801, opened 2021-02-03
Run custom domain test with coformance test

nak3:
Since serving v0.20+, `ResolvableDomain` flag is separated from custom domain.
Please refer to https://github.com/knative/serving/commit/82b59914fb9727f040b4d58073b2e6049b2d4583.

Hence the conformance test should be able to run with customdomain and
other tests.


nak3:
Still 0.19 in project.yaml :sweat_smile:

PR #791, opened 2021-01-28
Bump Knative Serving to v0.20.0

markusthoemmes:
As per title. Had to adjust the patches but we're working on ways to make that less painful, so let's roll with it for now.

PR #802, opened 2021-02-03
Vendor Serving, Eventing, Eventing-Kafka from midstream forks

mgencur:
We should use our forks where we can have OpenShift-specific patches.
This is e.g. important for upgrade tests where we consume tests from Serving and Eventing and they need a patch for RetryingRouteInconsistency. Instead of applying the patch again we can pull the sources from the forks.

I'm not sure if we want to do the same for the knative.dev/operator dependency? WDYT?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -40,4 +40,7 @@ replace (
 	k8s.io/client-go => k8s.io/client-go v0.19.7
 	k8s.io/code-generator => k8s.io/code-generator v0.19.7
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20200805222855-6aeccd4b50c6
+	knative.dev/eventing => github.com/openshift/knative-eventing v0.99.1-0.20210112194111-24e8ffe5a3a5
+	knative.dev/eventing-kafka => github.com/openshift-knative/eventing-kafka v0.19.1-0.20210202142951-dd8077f58870
+	knative.dev/serving => github.com/openshift/knative-serving v0.10.1-0.20210127081219-6330eae520d2
 )
--- a/hack/update-deps.sh
+++ b/hack/update-deps.sh
@@ -13,6 +13,9 @@ cd "${ROOT_DIR}"
 
 # This controls the knative release version we track.
 KN_VERSION="release-0.20"
+EVENTING_VERSION="release-v0.20.0"
+EVENTING_KAFKA_VERSION="release-v0.20.0"
+SERVING_VERSION="release-v0.20.0"
 
 # Controls the version of OCP related dependencies.
 OCP_VERSION="release-4.7"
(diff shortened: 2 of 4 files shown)


matzew:
> I'm not sure if we want to do the same for the knative.dev/operator dependency? WDYT?

I _think_ with OaaL (Operator as a library) we don't need any fork of that repo ?

markusthoemmes:
yeah the fork is no longer maintained.

go.mod
```diff
@@ -40,4 +40,7 @@ replace (
 	k8s.io/client-go => k8s.io/client-go v0.19.7
 	k8s.io/code-generator => k8s.io/code-generator v0.19.7
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20200805222855-6aeccd4b50c6
+	knative.dev/eventing => github.com/openshift/knative-eventing v0.99.1-0.20210112194111-24e8ffe5a3a5
```
cardil, line 43: Why is this version so wild: `v0.99.1-0.20210112194111-24e8ffe5a3a5`?

hack/update-deps.sh
```diff
@@ -13,6 +13,9 @@ cd "${ROOT_DIR}"
 
 # This controls the knative release version we track.
 KN_VERSION="release-0.20"
+EVENTING_VERSION="release-v0.20.0"
+EVENTING_KAFKA_VERSION="release-v0.20.0"
+SERVING_VERSION="release-v0.20.0"
```
cardil, line 18: Those should be read from `project.yaml`, right?

go.mod
mgencur: This version is automatically resolved/updated by `go get` (or ./hack/update-deps.sh) when we point to the given branch (release-0.20).

hack/update-deps.sh
mgencur: Maybe. As long as the version in project.yaml always matches the given branch in those repositories. This would enforce it. But this would also force the team to always do the upgrade at once. As I heard right now it might be done with at least two PRs - one for Golang dependencies, one for yaml changes etc (like here: https://github.com/openshift-knative/serverless-operator/pull/791/files). I don't want to change something that has been working well so far. Will defer to @markusthoemmes

markusthoemmes: Yeah let's defer that. It's generally useful to break these changes apart imo.

go.mod
markusthoemmes: It's likely because we don't tag that repo.

hack/update-deps.sh
cardil: I just hate having different, conflicting, sources of truth :anger:

markusthoemmes: Everybody does, but the Golang deps aer usually their own can of worms when it comes to upgrading versions and they usually don't stricrtly have to move in lockstep to the component versions themselves.

markusthoemmes (approved):
/lgtm
/approve

nak3:
It seems `./hack/update-deps.sh --upgrade` does not work...

mgencur:
@nak3 what happened in your case?

nak3:
It gets this error.
```
$ bash hack/update-deps.sh --upgrade
go get: knative.dev/eventing-kafka@release-v0.20.0: invalid version: unknown revision release-v0.20.0
go get: knative.dev/eventing@release-v0.20.0: invalid version: unknown revision release-v0.20.0
go get: knative.dev/serving@release-v0.20.0: invalid version: unknown revision release-v0.20.0
```

knative.dev/serving repo does not have `release-v0.20.0`.

mgencur:
@nak3 ok, let me check the script. I remember having problems as well. Maybe I can modify to work properly.

mgencur:
@nak3 sent this PR: https://github.com/openshift-knative/serverless-operator/pull/882

PR #803, opened 2021-02-04
Serving 0.20 upgrade tests

mgencur:
Incorporates commit https://github.com/openshift-knative/serverless-operator/pull/791 which will be removed when this PR is integrated.


hack/lib/common.bash
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+# shellcheck disable=SC2001
```
markusthoemmes, line 3: Can we define this locally to the line that violates it?

test/eventing.bash
```diff
@@ -29,70 +29,6 @@ function prepare_knative_eventing_tests {
   logger.info 'Nothing to prepare for Eventing upgrade tests'
 }
 
-function run_eventing_preupgrade_test {
```
markusthoemmes, line 32: Why all the changes to the eventing scripts :thinking: ?

test/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,82 @@
...
+	suite.Execute(c)
+}
+
+func preUpgradeTests() []pkgupgrade.Operation {
+	// We might want to skip pre-upgrade test if we want to re-use the services
+	// from the previous run. For example, to let them survive both Serverless
+	// and OCP upgrades. This allows for more variants of tests, with different
+	// order of upgrades.
+	if os.Getenv("SKIP_PRE_UPGRADE") == "true" {
+		return []pkgupgrade.Operation{}
```
markusthoemmes, line 71: nit: I'd hope the framework allows for returning `nil` here. Best practice and all, not a blocker.

markusthoemmes (approved):
Looks great as far as I can judge! It'd be awesome to get a second QE review on this, to verify it works as expected.

Thanks a megaton @mgencur !

test/eventing.bash
cardil: Martin, you are removing here the eventing upgrade tests, but you didn't add their new `Operation` to new test.

test/upgrade/upgrade_test.go
cardil: This will be executed by: https://github.com/knative/pkg/blob/master/test/upgrade/suite_execution.go#L26

I think it should accept `nil` just fine: https://play.golang.org/p/jzd9nInNaL4

test/eventing.bash
cardil: /hold

hack/lib/common.bash
mgencur: ok.

test/eventing.bash
mgencur: Yes. I know. I was only working on Serving upgrade tests and preparing the scripts for the new upgrade framework. The Eventing tests don't work anyway so the Eventing stuff in scripts can't be used. I was hoping we could do this: merge this for Serving, then the master branch would be running only Serving upgrade tests for a short while until I add the Eventing tests which also depend on the Eventing update to 0.20 (in a separate PR from Matthias). The other way would be also rebasing this on Matthias' PR and add Eventing tests in one go but it's too many dependencies in one PR. This would only work if you review and "stamp" the PR for eventing and I'd include the commits in this PR. WDYT?

test/upgrade/upgrade_test.go
mgencur: @cardil thanks for checking this. Pushed the change

test/eventing.bash
markusthoemmes: IMO it's okay to ship piecemeal in this case, but regarding our process discussions we really need to avoid such situations in the future.

cardil: I'm okay with your original plan, but please add a comment to the code, something like:

```
TODO(mgencur): Restore eventing upgrade tests shortly
```

mgencur: Thanks. Pushed the comment.

mgencur:
/retest

The Autoscaling test failure in upgrade tests should not be releated to "upgrade tests" themselves :-/

mgencur:
/hold

Superseded by https://github.com/openshift-knative/serverless-operator/pull/819

mgencur:
Superseded by #819

PR #806, opened 2021-02-04
[SRVKS-689] Improve sm rbac setup

skonto:
-  adds a feature toggle for serving monitoring, in the future I will do that for eventing as well (at some point all monitoring will be managed a single flag).
- moves rbac setup for serving monitoring to the extension side.
/cc @markusthoemmes

### DIFF
--- a/openshift-knative-operator/hack/004-rbac-proxy.patch
+++ b/openshift-knative-operator/hack/004-rbac-proxy.patch
@@ -1,9 +1,9 @@
 diff --git a/openshift-knative-operator/cmd/operator/kodata/knative-serving/0.20.0/0-rbac-proxy.yaml b/openshift-knative-operator/cmd/operator/kodata/knative-serving/0.20.0/0-rbac-proxy.yaml
 new file mode 100644
-index 00000000..d9df9652
+index 00000000..83c826d2
 --- /dev/null
 +++ b/openshift-knative-operator/cmd/operator/kodata/knative-serving/0.20.0/0-rbac-proxy.yaml
-@@ -0,0 +1,314 @@
+@@ -0,0 +1,342 @@
 +apiVersion: rbac.authorization.k8s.io/v1
 +kind: Role
 +metadata:
@@ -571,7 +571,7 @@ index a980952f..bcee5d5d 100644
 +        - name: secret-domain-mapping-sm-service-tls
 +          secret:
 +            secretName: domain-mapping-sm-service-tls
- 
+
  ---
  # Copyright 2020 The Knative Authors
 @@ -265,6 +287,8 @@ spec:
--- a/openshift-knative-operator/hack/004-rbac-proxy.patch
+++ b/openshift-knative-operator/hack/004-rbac-proxy.patch
@@ -1,9 +1,37 @@
 diff --git a/openshift-knative-operator/cmd/operator/kodata/knative-serving/0.20.0/0-rbac-proxy.yaml b/openshift-knative-operator/cmd/operator/kodata/knative-serving/0.20.0/0-rbac-proxy.yaml
 new file mode 100644
-index 00000000..37558cd9
+index 00000000..d9df9652
 --- /dev/null
 +++ b/openshift-knative-operator/cmd/operator/kodata/knative-serving/0.20.0/0-rbac-proxy.yaml
 @@ -0,0 +1,314 @@
++apiVersion: rbac.authorization.k8s.io/v1
++kind: Role
++metadata:
++  name: knative-serving-prometheus-k8s
++rules:
++  - apiGroups:
++      - ""
++    resources:
++      - services
++      - endpoints
++      - pods
++    verbs:
++      - get
++      - list
++      - watch
++---
++apiVersion: rbac.authorization.k8s.io/v1
++kind: RoleBinding
++metadata:
++  name: knative-serving-prometheus-k8s
++roleRef:
++  apiGroup: rbac.authorization.k8s.io
++  kind: Role
++  name: knative-serving-prometheus-k8s
++subjects:
++  - kind: ServiceAccount
++    name: prometheus-k8s
++    namespace: openshift-monitoring
 +---
 +apiVersion: rbac.authorization.k8s.io/v1
 +kind: ClusterRole
--- a/openshift-knative-operator/pkg/monitoring/monitoring.go
+++ b/openshift-knative-operator/pkg/monitoring/monitoring.go
@@ -9,62 +9,84 @@ import (
 
 	mf "github.com/manifestival/manifestival"
 	"github.com/openshift-knative/serverless-operator/openshift-knative-operator/pkg/common"
-	"go.uber.org/zap"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/client-go/kubernetes"
 	"knative.dev/operator/pkg/apis/operator/v1alpha1"
+	"knative.dev/pkg/logging"
 )
 
 const (
-	DisableMonitoringEnvVar      = "DISABLE_SERVING_MONITORING_BY_DEFAULT"
+	EnableMonitoringEnvVar       = "ENABLE_SERVING_MONITORING_BY_DEFAULT"
 	EnableMonitoringLabel        = "openshift.io/cluster-monitoring"
 	ObservabilityCMName          = "observability"
 	ObservabilityBackendKey      = "metrics.backend-destination"
 	OpenshiftMonitoringNamespace = "openshift-monitoring"
 )
 
-func SetupServingMonitoring(api kubernetes.Interface, ks *v1alpha1.KnativeServing, log *zap.SugaredLogger) error {
-	if shouldDisableMonitoring(ks.Spec.CommonSpec.Config) {
-		log.Info("Disabling Serving monitoring")
-		if err := disableMonitoring(api, ks); err != nil {
-			return fmt.Errorf("failed to disable monitoring %w ", err)
+const (
+	prometheusRoleName        = "knative-serving-prometheus-k8s"
+	prometheusClusterRoleName = "rbac-proxy-metrics-prom"
+)
+
+func SetupServingMonitoring(ctx context.Context, api kubernetes.Interface, ks *v1alpha1.KnativeServing) error {
+	backend, isSet := ks.Spec.CommonSpec.Config[ObservabilityCMName][ObservabilityBackendKey]
+	if shouldEnableMonitoring(backend) {
+		log := logging.FromContext(ctx)
+		log.Info("Enabling Serving monitoring")
+		if err := setMonitoringLabelToNamespace(ctx, ks.Namespace, api, true); err != nil {
+			return fmt.Errorf("failed to enable monitoring %w ", err)
 		}
 		return nil
 	}
-	if err := setMonitoringLabelToNamespace(ks.Namespace, api, true); err != nil {
+	if err := setMonitoringLabelToNamespace(ctx, ks.Namespace, api, false); err != nil {
 		return err
+	}
+	if !isSet {
+		common.Configure(&ks.Spec.CommonSpec, ObservabilityCMName, ObservabilityBackendKey, "none")
 	}
 	return nil
 }
 
-func shouldDisableMonitoring(cfg v1alpha1.ConfigMapData) bool {
-	_, disableByDefault := os.LookupEnv(DisableMonitoringEnvVar)
-	backend, backendIsSet := cfg[ObservabilityCMName][ObservabilityBackendKey]
-	return (disableByDefault && !backendIsSet) || backend == "none"
-}
-
-func disableMonitoring(api kubernetes.Interface, ks *v1alpha1.KnativeServing) error {
-	// Disable metrics backend by default in case we need to eg. SRVKS-679
-	if _, ok := ks.Spec.CommonSpec.Config[ObservabilityCMName][ObservabilityBackendKey]; !ok {
-		common.Configure(&ks.Spec.CommonSpec, ObservabilityCMName, ObservabilityBackendKey, "none")
+func shouldEnableMonitoring(backend string) bool {
+	enable, present := os.LookupEnv(EnableMonitoringEnvVar)
+	// Skip setup from env if feature toggle is off, use whatever the user defines in the Serving CR.
+	if !present && backend != "none" {
+		return true
 	}
-	return setMonitoringLabelToNamespace(ks.Namespace, api, false)
+	parsedEnable, _ := strconv.ParseBool(enable)
+	// Let the user enable monitoring with a proper backend value even if feature toggle is off.
+	if !parsedEnable && backend != "none" && backend != "" {
+		return true
+	}
+	return parsedEnable
 }
 
-func setMonitoringLabelToNamespace(namespace string, api kubernetes.Interface, enable bool) error {
-	ns, err := api.CoreV1().Namespaces().Get(context.Background(), namespace, metav1.GetOptions{})
+func setMonitoringLabelToNamespace(ctx context.Context, namespace string, api kubernetes.Interface, enable bool) error {
+	ns, err := api.CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
 	if err != nil {
 		return err
 	}
+	if shouldSkipSettingLabel(ns.Labels) {
+		return nil
+	}
 	if ns.Labels == nil {
-		ns.Labels = map[string]string{}
+		ns.Labels = make(map[string]string, 1)
 	}
 	ns.Labels[EnableMonitoringLabel] = strconv.FormatBool(enable)
 	if _, err := api.CoreV1().Namespaces().Update(context.Background(), ns, metav1.UpdateOptions{}); err != nil {
 		return fmt.Errorf("could not add label %q to namespace %q: %w", EnableMonitoringLabel, namespace, err)
 	}
 	return nil
+}
+
+func shouldSkipSettingLabel(labels map[string]string) bool {
+	if labelVal, found := labels[EnableMonitoringLabel]; found {
+		if parsedLabelVal, _ := strconv.ParseBool(labelVal); parsedLabelVal {
+			return parsedLabelVal
+		}
+	}
+	return false
 }
 
 // InjectNamespaceWithSubject uses a custom transformation to avoid operator overriding everything with the current namespace including
@@ -74,9 +96,9 @@ func InjectNamespaceWithSubject(resourceNamespace string, subjectNamespace strin
 	return func(u *unstructured.Unstructured) error {
 		kind := strings.ToLower(u.GetKind())
 		// Only touch the related manifests.
-		if kind == "role" && u.GetName() == "knative-serving-prometheus-k8s" {
+		if kind == "role" && u.GetName() == prometheusRoleName {
 			u.SetNamespace(resourceNamespace)
-		} else if (kind == "clusterrolebinding" && u.GetName() == "rbac-proxy-metrics-prom") || (kind == "rolebinding" && u.GetName() == "knative-serving-prometheus-k8s") {
+		} else if (kind == "clusterrolebinding" && u.GetName() == prometheusClusterRoleName) || (kind == "rolebinding" && u.GetName() == prometheusRoleName) {
 			if kind == "rolebinding" {
 				u.SetNamespace(resourceNamespace)
 			}
--- a/openshift-knative-operator/pkg/monitoring/monitoring.go
+++ b/openshift-knative-operator/pkg/monitoring/monitoring.go
@@ -1,0 +1,93 @@
+package monitoring
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+
+	mf "github.com/manifestival/manifestival"
+	"github.com/openshift-knative/serverless-operator/openshift-knative-operator/pkg/common"
+	"go.uber.org/zap"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/kubernetes"
+	"knative.dev/operator/pkg/apis/operator/v1alpha1"
+)
+
+const (
+	DisableMonitoringEnvVar      = "DISABLE_SERVING_MONITORING_BY_DEFAULT"
+	EnableMonitoringLabel        = "openshift.io/cluster-monitoring"
+	ObservabilityCMName          = "observability"
+	ObservabilityBackendKey      = "metrics.backend-destination"
+	OpenshiftMonitoringNamespace = "openshift-monitoring"
+)
+
+func SetupServingMonitoring(api kubernetes.Interface, ks *v1alpha1.KnativeServing, log *zap.SugaredLogger) error {
+	if shouldDisableMonitoring(ks.Spec.CommonSpec.Config) {
+		log.Info("Disabling Serving monitoring")
+		if err := disableMonitoring(api, ks); err != nil {
+			return fmt.Errorf("failed to disable monitoring %w ", err)
+		}
+		return nil
+	}
+	if err := setMonitoringLabelToNamespace(ks.Namespace, api, true); err != nil {
+		return err
+	}
+	return nil
+}
+
+func shouldDisableMonitoring(cfg v1alpha1.ConfigMapData) bool {
+	_, disableByDefault := os.LookupEnv(DisableMonitoringEnvVar)
+	backend, backendIsSet := cfg[ObservabilityCMName][ObservabilityBackendKey]
+	return (disableByDefault && !backendIsSet) || backend == "none"
+}
+
+func disableMonitoring(api kubernetes.Interface, ks *v1alpha1.KnativeServing) error {
+	// Disable metrics backend by default in case we need to eg. SRVKS-679
+	if _, ok := ks.Spec.CommonSpec.Config[ObservabilityCMName][ObservabilityBackendKey]; !ok {
+		common.Configure(&ks.Spec.CommonSpec, ObservabilityCMName, ObservabilityBackendKey, "none")
+	}
+	return setMonitoringLabelToNamespace(ks.Namespace, api, false)
+}
+
+func setMonitoringLabelToNamespace(namespace string, api kubernetes.Interface, enable bool) error {
+	ns, err := api.CoreV1().Namespaces().Get(context.Background(), namespace, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	if ns.Labels == nil {
+		ns.Labels = map[string]string{}
+	}
+	ns.Labels[EnableMonitoringLabel] = strconv.FormatBool(enable)
+	if _, err := api.CoreV1().Namespaces().Update(context.Background(), ns, metav1.UpdateOptions{}); err != nil {
+		return fmt.Errorf("could not add label %q to namespace %q: %w", EnableMonitoringLabel, namespace, err)
+	}
+	return nil
+}
+
+// InjectNamespaceWithSubject uses a custom transformation to avoid operator overriding everything with the current namespace including
+// subjects ns. Here we break the assumption of the operator about all resources being in the same namespace
+// since we need to setup RBAC for the prometheus-k8s account which resides in openshift-monitoring ns.
+func InjectNamespaceWithSubject(resourceNamespace string, subjectNamespace string) mf.Transformer {
+	return func(u *unstructured.Unstructured) error {
+		kind := strings.ToLower(u.GetKind())
+		// Only touch the related manifests.
+		if kind == "role" && u.GetName() == "knative-serving-prometheus-k8s" {
+			u.SetNamespace(resourceNamespace)
+		} else if (kind == "clusterrolebinding" && u.GetName() == "rbac-proxy-metrics-prom") || (kind == "rolebinding" && u.GetName() == "knative-serving-prometheus-k8s") {
+			if kind == "rolebinding" {
+				u.SetNamespace(resourceNamespace)
+			}
+			subjects, _, _ := unstructured.NestedFieldNoCopy(u.Object, "subjects")
+			for _, subject := range subjects.([]interface{}) {
+				m := subject.(map[string]interface{})
+				if _, ok := m["namespace"]; ok {
+					m["namespace"] = subjectNamespace
+				}
+			}
+		}
+		return nil
+	}
+}
--- a/openshift-knative-operator/pkg/monitoring/monitoring.go
+++ b/openshift-knative-operator/pkg/monitoring/monitoring.go
@@ -30,22 +30,17 @@ const (
 )
 
 func ReconcileServingMonitoring(ctx context.Context, api kubernetes.Interface, ks *v1alpha1.KnativeServing) error {
-	backend, isSet := ks.Spec.CommonSpec.Config[ObservabilityCMName][ObservabilityBackendKey]
-	log := logging.FromContext(ctx)
+	backend := ks.Spec.CommonSpec.Config[ObservabilityCMName][ObservabilityBackendKey]
 	if shouldEnableMonitoring(backend) {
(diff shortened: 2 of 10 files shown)


openshift-knative-operator/hack/004-rbac-proxy.patch
```diff
@@ -1,9 +1,37 @@
...
++metadata:
++  name: knative-serving-prometheus-k8s
++roleRef:
++  apiGroup: rbac.authorization.k8s.io
++  kind: Role
++  name: knative-serving-prometheus-k8s
++subjects:
++  - kind: ServiceAccount
++    name: prometheus-k8s
++    namespace: openshift-monitoring
```
skonto, line 34: The part bellow will be removed (most likely) because prometheus-k8s account already has the required role:
```
$ oc describe clusterroles prometheus-k8s
Name:         prometheus-k8s
Labels:       <none>
Annotations:  <none>
PolicyRule:
  Resources                                  Non-Resource URLs  Resource Names  Verbs
  ---------                                  -----------------  --------------  -----
  tokenreviews.authentication.k8s.io         []                 []              [create]
  subjectaccessreviews.authorization.k8s.io  []                 []              [create]
                                             [/metrics]         []              [get]
  namespaces                                 []                 []              [get]
  nodes/metrics                              []                 []              [get]
```

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -0,0 +1,93 @@
...
+	if _, err := api.CoreV1().Namespaces().Update(context.Background(), ns, metav1.UpdateOptions{}); err != nil {
+		return fmt.Errorf("could not add label %q to namespace %q: %w", EnableMonitoringLabel, namespace, err)
+	}
+	return nil
+}
+
+// InjectNamespaceWithSubject uses a custom transformation to avoid operator overriding everything with the current namespace including
+// subjects ns. Here we break the assumption of the operator about all resources being in the same namespace
+// since we need to setup RBAC for the prometheus-k8s account which resides in openshift-monitoring ns.
+func InjectNamespaceWithSubject(resourceNamespace string, subjectNamespace string) mf.Transformer {
```
skonto, line 73: Avoid custom injection to mess up with subjects.

openshift-knative-operator/pkg/serving/extension_test.go
```diff
@@ -113,27 +125,61 @@ func TestReconcile(t *testing.T) {
...
 			}
-			ctx, _ := ocpfake.With(context.Background(), objs...)
-
 			ks := c.in.DeepCopy()
-			ext := NewExtension(ctx)
+			ks.Namespace = "knative-serving"
+			c.expected.Namespace = ks.Namespace
+			ctx, _ = ocpfake.With(ctx, objs...)
+			ext := NewExtension(ctx).(*extension)
+			ext.kubeclient = fakekubeclientset.NewSimpleClientset([]runtime.Object{&servingNamespace}...)
```
markusthoemmes, line 162: Let's not construct this manually. You can use the same `With` function as is used for the `ocpfake` above. That also avoids the unnecessary cast to `*extension`

openshift-knative-operator/pkg/serving/extension_test.go
```diff
@@ -35,14 +43,16 @@ func TestReconcile(t *testing.T) {
...
 	cases := []struct {
-		name     string
-		in       *v1alpha1.KnativeServing
-		objs     []runtime.Object
-		expected *v1alpha1.KnativeServing
+		name                    string
+		in                      *v1alpha1.KnativeServing
+		objs                    []runtime.Object
+		expected                *v1alpha1.KnativeServing
+		shouldDisableMonitoring bool
```
markusthoemmes, line 50: I'd rather not introduce this specific concept to this test. Can we instead test more generic in that we make `expecetd` a list of `runtime.Object` and make sure that all the objects we expect are correctly reflected?

openshift-knative-operator/pkg/serving/extension.go
```diff
@@ -7,30 +7,38 @@ import (
```
markusthoemmes, line 7: specific unit tests for this would be great.

openshift-knative-operator/pkg/serving/extension.go
```diff
@@ -7,30 +7,38 @@ import (
...
+		kubeclient: kubeclient.Get(ctx),
+		logger:     logging.FromContext(ctx),
 	}
 }
 
 type extension struct {
-	ocpclient versioned.Interface
+	ocpclient  versioned.Interface
+	kubeclient kubernetes.Interface
+	logger     *zap.SugaredLogger
```
markusthoemmes, line 37: This is not required. The context passed to `Reconcile` already has a context that can be obtained using `logging.FromContext`.

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -0,0 +1,93 @@
...
+		return nil
+	}
+	if err := setMonitoringLabelToNamespace(ks.Namespace, api, true); err != nil {
+		return err
+	}
+	return nil
+}
+
+func shouldDisableMonitoring(cfg v1alpha1.ConfigMapData) bool {
+	_, disableByDefault := os.LookupEnv(DisableMonitoringEnvVar)
```
markusthoemmes, line 42: What if I set the env var to `false`? Should this parse the env var to an extend?

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -0,0 +1,93 @@
...
+
+const (
+	DisableMonitoringEnvVar      = "DISABLE_SERVING_MONITORING_BY_DEFAULT"
+	EnableMonitoringLabel        = "openshift.io/cluster-monitoring"
+	ObservabilityCMName          = "observability"
+	ObservabilityBackendKey      = "metrics.backend-destination"
+	OpenshiftMonitoringNamespace = "openshift-monitoring"
+)
+
+func SetupServingMonitoring(api kubernetes.Interface, ks *v1alpha1.KnativeServing, log *zap.SugaredLogger) error {
```
markusthoemmes, line 27: This should take the reconcile context and obtain the logger from there.

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -0,0 +1,93 @@
...
+func disableMonitoring(api kubernetes.Interface, ks *v1alpha1.KnativeServing) error {
+	// Disable metrics backend by default in case we need to eg. SRVKS-679
+	if _, ok := ks.Spec.CommonSpec.Config[ObservabilityCMName][ObservabilityBackendKey]; !ok {
+		common.Configure(&ks.Spec.CommonSpec, ObservabilityCMName, ObservabilityBackendKey, "none")
+	}
+	return setMonitoringLabelToNamespace(ks.Namespace, api, false)
+}
+
+func setMonitoringLabelToNamespace(namespace string, api kubernetes.Interface, enable bool) error {
+	ns, err := api.CoreV1().Namespaces().Get(context.Background(), namespace, metav1.GetOptions{})
```
markusthoemmes, line 56: The reconcile context should likely be passed through to here.

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return err
+	}
+	if ns.Labels == nil {
+		ns.Labels = map[string]string{}
```
markusthoemmes, line 61: nit:

```suggestion
		ns.Labels = make(map[string]string, 1)
```

since we know we'll add one.

openshift-knative-operator/pkg/monitoring/monitoring.go
markusthoemmes, line 73: A unit test for this would be great to see what the intended behavior is.

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -0,0 +1,93 @@
...
+	_, disableByDefault := os.LookupEnv(DisableMonitoringEnvVar)
+	backend, backendIsSet := cfg[ObservabilityCMName][ObservabilityBackendKey]
+	return (disableByDefault && !backendIsSet) || backend == "none"
+}
+
+func disableMonitoring(api kubernetes.Interface, ks *v1alpha1.KnativeServing) error {
+	// Disable metrics backend by default in case we need to eg. SRVKS-679
+	if _, ok := ks.Spec.CommonSpec.Config[ObservabilityCMName][ObservabilityBackendKey]; !ok {
+		common.Configure(&ks.Spec.CommonSpec, ObservabilityCMName, ObservabilityBackendKey, "none")
+	}
```
markusthoemmes, line 51: I wonder if it'd be cleaner to pull this apart a bit, i.e. to determine the defaulting of this using the env var shenanigans above and then, as a separate step, just to reconcile based on what KnativeServing actually has set.

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -0,0 +1,93 @@
...
+func setMonitoringLabelToNamespace(namespace string, api kubernetes.Interface, enable bool) error {
+	ns, err := api.CoreV1().Namespaces().Get(context.Background(), namespace, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	if ns.Labels == nil {
+		ns.Labels = map[string]string{}
+	}
+	ns.Labels[EnableMonitoringLabel] = strconv.FormatBool(enable)
+	if _, err := api.CoreV1().Namespaces().Update(context.Background(), ns, metav1.UpdateOptions{}); err != nil {
```
markusthoemmes, line 64: This should not be done if the label is already set to avoid churn on the API server.

markusthoemmes (commented):
I like it!

openshift-knative-operator/pkg/monitoring/monitoring.go
skonto: This does not check the value but it could

openshift-knative-operator/pkg/monitoring/monitoring.go
skonto: This is trivial.

openshift-knative-operator/pkg/monitoring/monitoring.go
skonto: I dont think I follow here. Could you elaborate a bit?

openshift-knative-operator/pkg/monitoring/monitoring.go
skonto: This does not go through the cache?

openshift-knative-operator/pkg/serving/extension.go
skonto: Good to know there is a context otherwise it seems too verbose.

openshift-knative-operator/pkg/serving/extension_test.go
skonto: I would like to signal that the env var for disabling monitoring should be set, this is external to the test case at hand and happens per test case not before tests are run, what is wrong with that? For the ns check at the end of the test I can add it to the expected objects. Do you mean the latter?

openshift-knative-operator/pkg/serving/extension.go
skonto: For?

openshift-knative-operator/pkg/monitoring/monitoring.go
markusthoemmes: Nope. Updates never go via cache FWIW.

openshift-knative-operator/pkg/serving/extension_test.go
markusthoemmes: I see. I'd check the env var behavior with a separate test as this conflates things quite a bit :thinking:

openshift-knative-operator/pkg/serving/extension.go
markusthoemmes: The entire behavior ideally.

openshift-knative-operator/pkg/monitoring/monitoring.go
markusthoemmes: The type case shenanigans are definitely not trivial IMO.

openshift-knative-operator/pkg/serving/extension_test.go
skonto: ok will do.

openshift-knative-operator/pkg/monitoring/monitoring.go
skonto: Ok I thought it was about the Get call above misread. Yes I think I can skip it.

markusthoemmes: The `Get` call also ain't via cache in this case FWIW.

openshift-knative-operator/pkg/serving/extension_test.go
skonto: @markusthoemmes the returned clientSet from ocpfake.With call, only provides access to:
```
	Discovery() discovery.DiscoveryInterface
	ConfigV1() configv1.ConfigV1Interface
	RouteV1() routev1.RouteV1Interface
```
I cannot use it directly to get the namespace bellow that is why I used this fake client explicitly. What do you suggest exactly?

openshift-knative-operator/pkg/monitoring/monitoring.go
skonto: Oh it uses the RestClient directly too bad.

openshift-knative-operator/pkg/serving/extension.go
skonto: ok.

openshift-knative-operator/pkg/serving/extension_test.go
markusthoemmes: You can import the same fake client for K8s and do `kubefake.With(ctx, &servingNamespace)`

skonto: Not sure I see that method I just followed what I read [here](https://github.com/knative/pkg/blob/448ae657fb5fb0d84d99825d184d706478574023/metrics/reflector_test.go#L62).

openshift-knative-operator/pkg/monitoring/monitoring.go
skonto: This is a copy of https://github.com/manifestival/manifestival/blob/18b895087f7b4ad607f86e00faeaa58b2ea2130e/transform.go#L63

openshift-knative-operator/pkg/monitoring/monitoring.go
markusthoemmes: Nevermind, it's fine as-is.

skonto: I will try to separate and see if that is ok.

openshift-knative-operator/pkg/monitoring/monitoring.go
rhuss: just my 2cents: Typically it's better to formulate flags in a 'positive' way, to avoid double negative (like DISABLE_...=false). IMO its a better user experience to use `ENABLE_SERVICE_MONITORING_BY_DEFAULT` (or maybe even more explicit `ENABLE_SERVICE_MONITORING_AS_FALLBACK`) with a value of `false` for this behaviour.

This would be especially important if there are already ENABLE_ kind of environment variables (don't know if this is the case here, but the usage should be consistent, and introducing DISABLE_ now would imply that all future boolean env vars should use DISABLE_ , too)

I don't insist on a change.

rhuss (commented):
minor comment on naming, can't really judge on the logic (haven't dived into it)

skonto:
@rhuss I thought about this before so my assumption was to have features enabled by default when they are really considered as a must and have a flag to disable them, for critical bugs (like the one we faced lately). I think monitoring should be enabled by default thus the "DISABLE_" prefix for special cases. I agree that this requires some form of unification for the feature toggles also if you need a feature always on [remove the flag](https://featureflags.io/feature-flags-best-practices/). So this was kind of a misuse of a feature flag but with a specific goal in mind of how to be used.
Each feature ideally should be stable and be kept that way even when a migration is done as we probably will face when we move to open telemetry. It is a bit unpleasant from a user point of view to disable such useful features from one release to the other. 
To make it proper beyond monitoring though I agree we should use something more positive like "ENABLE_*".
Each feature should have some state and make a decision about the default value when we do releases.
I will rename the var.

rhuss:
> I think monitoring should be enabled by default thus the "DISABLE_" prefix for special cases.

no question about this, but you can equally say `ENABLE_ .... = false` if you want to disable something that is enabled when no env variable is given. And I think it's a good practice to always evaluate the value of an env var and not only check for their existence (because the value could be contradictive then, as Markus pointed out, someone providing a `DISABLE_...=false` would be surprised if the feature would be still disabled because the env value is just ignored).

But as mentioned, I don't insist on ENABLE vs DISABLE but still would suggest giving the value of an env var some meaning (and so would check for it). E.g. its hard and counter-intuitive to set an env var without value in a ContainerSpec's `env:` section. Actually the value would be not empty but an empty string, which is again a different case. To avoid all this, just give the value a meaning an insist on `true` or `false` (and possibly don't even allow empty values).

skonto:
>  And I think it's a good practice to always evaluate the value of an env var and not only check for their existence (because the value could be contradictive then, as Markus pointed out, someone providing a DISABLE_...=false would be surprised if the feature would be still disabled because the env value is just ignored).
I replied to that comment I will update the PR ;)

openshift-knative-operator/pkg/monitoring/monitoring_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	mf "github.com/manifestival/manifestival"
+	"github.com/manifestival/manifestival/fake"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+const servingNamespace = "knative-serving"
+
+func TestSetupServingRbacTrasfmormation(t *testing.T) {
+	client := fake.New()
```
skonto, line 14: I could check the resources array from the manifest but using a client makes it more elegant.

skonto:
Addressed the comments.
-  added extensive testing.
-  added support for on/off values for the feature toggle. Made the toggle not too strict to allow user preference in CR to take precedence. The current combination of values for the toggle and the monitoring backend set in the CR and the result of whether monitoring is on/off are:

```
    --- PASS: TestMonitoring/enable_monitoring_when_monitoring_toggle_is_not_defined,_backend_is_not_defined (0.00s)
    --- PASS: TestMonitoring/enable_monitoring_when_monitoring_toggle_=_not_defined,_backend_=_defined_and_not_`none` (0.00s)
    --- PASS: TestMonitoring/disable_monitoring_when_monitoring_toggle_is_not_defined,_backend_is_`none` (0.00s)
    --- PASS: TestMonitoring/enable_monitoring_when_monitoring_toggle_is_on,_backend_is_not_defined (0.00s)
    --- PASS: TestMonitoring/enable_monitoring_when_monitoring_toggle_is_on,_backend_is_defined_and_not_`none` (0.00s)
    --- PASS: TestMonitoring/disable_monitoring_when_monitoring_toggle_is_on,_backend_is_`none` (0.00s)
    --- PASS: TestMonitoring/disable_monitoring_when_monitoring_toggle_is_off,_backend_is_not_defined (0.00s)
    --- PASS: TestMonitoring/enable_monitoring_when_monitoring_toggle_=_off,_backend_=_defined_and_not_`none` (0.00s)
    --- PASS: TestMonitoring/disable_monitoring_when_monitoring_toggle_is_off,_backend_is_`none` (0.00s)
```

openshift-knative-operator/pkg/serving/extension_test.go
```diff
@@ -121,10 +142,147 @@ func TestReconcile(t *testing.T) {
...
+}
 
+func TestMonitoring(t *testing.T) {
+	cases := []struct {
+		name     string
+		in       *v1alpha1.KnativeServing
+		objs     []runtime.Object
+		expected *v1alpha1.KnativeServing
+		// Returns the expected status for monitoring
+		setupMonitoringToggle func() (bool, error)
```
markusthoemmes, line 168: I don't quite parse why this is a function :thinking:. You seem to be returning just a boolean everytime anyway. Any reason why an error is used here?

openshift-knative-operator/pkg/monitoring/monitoring_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"testing"
+
+	mf "github.com/manifestival/manifestival"
+	"github.com/manifestival/manifestival/fake"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+const servingNamespace = "knative-serving"
+
+func TestSetupServingRbacTrasfmormation(t *testing.T) {
```
markusthoemmes, line 13:
```suggestion
func TestSetupServingRbacTransformation(t *testing.T) {
```

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -0,0 +1,115 @@
...
+	return nil
+}
+
+func shouldEnableMonitoring(backend string) bool {
+	enable, present := os.LookupEnv(EnableMonitoringEnvVar)
+	// Skip setup from env if feature toggle is off, use whatever the user defines in the Serving CR.
+	if !present && backend != "none" {
+		return true
+	}
+	parsedEnable, _ := strconv.ParseBool(enable)
```
markusthoemmes, line 57: Swallowing the error feels a little odd. What I've seen other people do is to compare with "true", i.e.: `enable := strings.EqualFold("true", enable)`.

I'd actually also generalize the !present case above to that as well, i.e.

```golang
func shouldEnableMonitoring(backend string) bool {
  enableByDefault := strings.EqualFold(os.GetEnv(EnableMonitoringEnvVar), "true")
  if !enableByDefault && backend != "none" && backend != "" {
    return true
  }
  return enableByDefault
}
```

Is imo easier to understand.

openshift-knative-operator/pkg/serving/extension.go
```diff
@@ -88,6 +93,11 @@ func (e *extension) Reconcile(ctx context.Context, comp v1alpha1.KComponent) err
 		}
 	}
 
+	logger := logging.FromContext(ctx)
+	logger.Info("Installing Serving Monitoring Requirements")
```
markusthoemmes, line 97: Should we remove this logline? If monitoring is disabled, we won't do anything here.

openshift-knative-operator/pkg/serving/extension_test.go
```diff
@@ -134,10 +292,24 @@ func TestReconcile(t *testing.T) {
 			if !cmp.Equal(ks, c.expected, opt) {
 				t.Errorf("Got = %v, want: %v, diff:\n%s", ks, c.expected, cmp.Diff(ks, c.expected, opt))
 			}
+			ns, err := kube.CoreV1().Namespaces().Get(context.Background(), ks.Namespace, metav1.GetOptions{})
+			if err != nil {
+				t.Errorf("Namespace %s not found %w", ns, err)
```
markusthoemmes, line 297:
```suggestion
				t.Errorf("Failed to get namespace %s: %v", ns, err)
```

openshift-knative-operator/pkg/serving/extension_test.go
```diff
@@ -134,10 +292,24 @@ func TestReconcile(t *testing.T) {
...
+			}
+			if len(ns.Labels) != 1 && ns.Labels[monitoring.EnableMonitoringLabel] != strconv.FormatBool(shouldEnableMonitoring) {
```
markusthoemmes, line 299: I'd drop the explicit len check here. The explicit check for the name itself is fine and it'll be more extensible in the future if we ever add more labels.

openshift-knative-operator/pkg/serving/extension_test.go
```diff
@@ -121,10 +142,147 @@ func TestReconcile(t *testing.T) {
...
+			}
+		})
+	}
+}
 
+func TestMonitoring(t *testing.T) {
+	cases := []struct {
+		name     string
+		in       *v1alpha1.KnativeServing
+		objs     []runtime.Object
```
markusthoemmes, line 165: This is never set. Should we remove it?

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func shouldSkipSettingLabel(labels map[string]string) bool {
+	if labelVal, found := labels[EnableMonitoringLabel]; found {
+		if parsedLabelVal, _ := strconv.ParseBool(labelVal); parsedLabelVal {
+			return parsedLabelVal
+		}
+	}
+	return false
+}
```
markusthoemmes, line 90: Is this correct? I think we should skip the label setting if our nwe value is equal to the value already in the label, so a more simple

```golang
if ns.Labels[EnableMonitoringLabel] == strconv.FormatBool(enable) {
  return nil
}
...
```

above would suffice I think?

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -0,0 +1,115 @@
...
+	ObservabilityBackendKey      = "metrics.backend-destination"
+	OpenshiftMonitoringNamespace = "openshift-monitoring"
+)
+
+const (
+	prometheusRoleName        = "knative-serving-prometheus-k8s"
+	prometheusClusterRoleName = "rbac-proxy-metrics-prom"
+)
+
+func SetupServingMonitoring(ctx context.Context, api kubernetes.Interface, ks *v1alpha1.KnativeServing) error {
```
markusthoemmes, line 32: It'd be great to name this `ReconcileServingMonitoring` to signify the fact that this runs per reconcile loop. Likewise I'd rename `setMonitoringLabelToNamespace` to `reconcileMonitoringLabelOnNamespace` to signify the same thing.

markusthoemmes (commented):
Almost there from my PoV, very nice!

openshift-knative-operator/pkg/serving/extension_test.go
skonto: Error comes from the the os.SetEnv call.

openshift-knative-operator/hack/004-rbac-proxy.patch
simonpasquier: yes prometheus-k8s SA already has the required permissions

openshift-knative-operator/pkg/serving/extension_test.go
markusthoemmes: But you are, i think?

skonto: Yes because `toggleEnvVar` should return error not false, I will fix that.

openshift-knative-operator/pkg/monitoring/monitoring.go
skonto: Yes but `  enableByDefault := strings.EqualFold(os.GetEnv(EnableMonitoringEnvVar), "true")` does not cover the case where no env var is set and backend is not `none`. If var is not set and backend is not `none` I need to return true.
`EqualFold` does not help with that.

skonto: I changed it a bit.

skonto:
I will remove the prometheus rbac part (as Simon clarified above) in the next PR which will also handle the repetition in the rbac proxy yaml files.

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -0,0 +1,110 @@
...
+		return err
+	}
+	if ns.Labels[EnableMonitoringLabel] == strconv.FormatBool(enable) {
+		return nil
+	}
+	if ns.Labels == nil {
+		ns.Labels = make(map[string]string, 1)
+	}
+	ns.Labels[EnableMonitoringLabel] = strconv.FormatBool(enable)
+	if _, err := api.CoreV1().Namespaces().Update(context.Background(), ns, metav1.UpdateOptions{}); err != nil {
```
markusthoemmes, line 81:
```suggestion
	if _, err := api.CoreV1().Namespaces().Update(ctx, ns, metav1.UpdateOptions{}); err != nil {
```

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+		return nil
+	}
+	log.Info("Disabling Serving monitoring")
+	if err := reconcileMonitoringLabelOnNamespace(ctx, ks.Namespace, api, false); err != nil {
+		return err
+	}
+	if !isSet {
+		common.Configure(&ks.Spec.CommonSpec, ObservabilityCMName, ObservabilityBackendKey, "none")
+	}
```
markusthoemmes, line 48: Is this correct? Feels like it should be done regardless, since we also disable monitoring :thinking:

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -0,0 +1,110 @@
...
+const (
+	prometheusRoleName        = "knative-serving-prometheus-k8s"
+	prometheusClusterRoleName = "rbac-proxy-metrics-prom"
+)
+
+func ReconcileServingMonitoring(ctx context.Context, api kubernetes.Interface, ks *v1alpha1.KnativeServing) error {
+	backend, isSet := ks.Spec.CommonSpec.Config[ObservabilityCMName][ObservabilityBackendKey]
+	log := logging.FromContext(ctx)
+	if shouldEnableMonitoring(backend) {
+		log.Info("Enabling Serving monitoring")
```
markusthoemmes, line 36: nit: The two loglines will be somewhat noisy. Should we move them into `reconcileMonitoringLabelOnNamespace` and log them only **after** we determined we actually want to update the NS labels?

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -0,0 +1,110 @@
...
+		if err := reconcileMonitoringLabelOnNamespace(ctx, ks.Namespace, api, true); err != nil {
+			return fmt.Errorf("failed to enable monitoring %w ", err)
+		}
+		return nil
+	}
+	log.Info("Disabling Serving monitoring")
+	if err := reconcileMonitoringLabelOnNamespace(ctx, ks.Namespace, api, false); err != nil {
+		return err
```
markusthoemmes, line 44: nit:

```suggestion
		return fmt.Errorf("failed to disable monitoring %w ", err)
```

to be consisetnt with above.

markusthoemmes (commented):
A few last questions, feel free to ignore everything marked as **nit**.

openshift-knative-operator/pkg/monitoring/monitoring.go
skonto: I can remove it is a relic from the changes.

simonpasquier:
> I will remove the prometheus rbac part (as Simon clarified above) in the next PR which will also handle the repetition in the rbac proxy yaml files.

oops, sorry I was wrong. You need to explicitly grant permissions on services, pods and endpoints resources to the prometheus-k8s SA. For security reasons, it doesn't have cluster-wide permissions.

skonto:
@simonpasquier when I tested it without these extra permissions it did work so I was wondering why it was the case and found out about the default ones. But I will keep it as you suggest.

openshift-knative-operator/pkg/monitoring/monitoring.go
skonto: ok

openshift-knative-operator/pkg/monitoring/monitoring.go
markusthoemmes: I mean I guess we should still explicitly set this if we want to disable, right?

skonto:
Cleaned it up a bit.

openshift-knative-operator/pkg/monitoring/monitoring.go
skonto: Check the new version,  there is no need to have that there.

markusthoemmes (approved):
/lgtm
/approve

Thanks!

PR #807, opened 2021-02-04
[RELEASE-1.13] Bump Openshift dependencies to 4.7

markusthoemmes:
A bit out of bounds, but I need the updated deps to be able to install Quickstarts (I need to be able to pull the API definition into the Schema)

/assign @mgencur


markusthoemmes:
/retest

Infra

hack/update-deps.sh
```diff
@@ -20,7 +20,7 @@ OCP_VERSION="release-4.6"
 # The list of dependencies that we track at HEAD and periodically
 # float forward in this repository.
 FLOATING_DEPS=(
-  "github.com/openshift/api@${OCP_VERSION}"
+  "github.com/openshift/api@release-4.7"
```
mgencur, line 23: Should we update `OCP_VERSION="release-4.7"` and use it here like in the other PR? Why is it different here?

markusthoemmes: It's different here because we have no K8s version bump in here yet, so the bump of all of the OCP deps becomes much more complex. Hence I went with a more surgical change to also reduce the potential for screwups throughout.

mgencur: OK. Makes sense.

markusthoemmes:
/retest

Infra.

PR #808, opened 2021-02-08
[SRVKS-667] Install Service related quickstart manifests

markusthoemmes:
As per SRVKS-667, this installs Service creation related quickstart manifests once Knative Serving is getting installed. It ignores `NoMatchFound` errors, which indicate that the given CRD is not present in this cluster (which is the case for 4.6 clusters for example).

/assign @skonto @mgencur


markusthoemmes:
Tested this on both a 4.6 and a 4.7 cluster and it worked as intended (ignoring the error on 4.6 and installing the Quickstart on 4.7)

skonto:
If this is a copy of https://github.com/openshift-knative/serverless-operator/pull/810 then see my comments there.

PR #818, opened 2021-02-09
Add ocp 4.7 to 1.13.0

warrenvw:
Add v4.7 to `com.redhat.openshift.versions`.


markusthoemmes (approved):
/lgtm
/approve

Should we fold this list with the channels list above? Will they always be equal?

warrenvw:
No, I don't think they'll be equal all the time. For instance, https://github.com/openshift-knative/serverless-operator/blob/9337e564928c3700d7ac0dcbbbb04514e2aff344/olm-catalog/serverless-operator/Dockerfile. Let's review this decision in a future release.

cardil:
> Should we fold this list with the channels list above? Will they always be equal?

Me, neither. I would like to see us finally adopting more same channel naming, like others teams do. For ex.: `stable`, `1.x`. So that it related to serverless feature set, and not to OCP.

PR #820, opened 2021-02-12
Update manifestival to latest and greatest

markusthoemmes:
Title has it all. I made an effort to align manifestival with newer K8s deps, so here we go.

/assign @matzew @rhuss


rhuss (approved):
Looks good to me as long as tests are passing 

/lgtm
/approve

PR #822, opened 2021-02-14
SRVKS-698 Enable Autoscaling tests during upgrades

mgencur:
The PR https://github.com/openshift-knative/serverless-operator/pull/803 showed 4 runs in a row where Autoscaling tests were passing. All these checks ran after the increase of cpu/mem request for the pod running the tests. The autoscaling tests are more resource-intensive since the previous release as they are using more threads. The assumption is that the client's are not able to send enough load to the autoscaler and the tests fail because of insufficient scaling. Let's run all the tests together (with Eventing tests) in this PR to confirm that.


mgencur:
The upgrade tests passing on this PR (2 times in a row). This is ready to be integrated, IMO.

markusthoemmes (approved):
/lgtm
/approve

PR #827, opened 2021-02-16
Drop installing dep in Github Actions

markusthoemmes:
:shrug: 

/assign @mgencur


mgencur:
I don't know why we were installing it. Anyway, if it works without that then great!

PR #829, opened 2021-02-17
Do not use -race flag in upgrade tests

mgencur:
The test suite occasionally fails with a race condition in triggering
the continual tests. Disabling the -race flag should not have effect on
how individual tests run. The tests are re-used from other repositories.
When the issue with the upgrade framework is fixed this can be enabled
again.

/assign @cardil

### DIFF
--- a/test/lib.bash
+++ b/test/lib.bash
@@ -47,7 +47,7 @@ function go_test_e2e {
     [[ -n "$arg" ]] && go_test_args+=("$arg")
   done
   set +Eeuo pipefail
-  report_go_test -race -count=1 "${go_test_args[@]}"
+  report_go_test -count=1 "${go_test_args[@]}"
   retcode=$?
   set -Eeuo pipefail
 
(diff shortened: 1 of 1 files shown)


markusthoemmes:
Have we filed issues upstream to actually hunt and fix these bugs?

mgencur:
Just filed one: https://github.com/knative/pkg/issues/2026

cardil:
The race flag doesn't make much sense in context of upgrade testing.

markusthoemmes:
I disagree with that. If the framework itself has data races going on, we might have racy outcomes and don't know where they're coming from. That might result in false positives or false negatives which would be quite annoying.

markusthoemmes:
(That said, this is LGTM since we've opened a tracking bug to actually fix it and bothering us with errors ain't useful)

cardil (approved):
/lgtm

cardil:
> I disagree with that. If the framework itself has data races going on, we might have racy outcomes and don't know where they're coming from. That might result in false positives or false negatives which would be quite annoying.

Only false positives, like this. Race recognition makes sense when unit testing. If you deploy prebuild artifacts like all e2e tests do, races will be reported only in testing/orchestration code. That code isn't run on production. That's why I think it makes a little sense to use `-race` for upgrade testing, as it has its downsides of longer operation and different output processing.

markusthoemmes:
@cardil well but in this case the race is actually happening in the test code itself, which means that it has a bug that has to be fixed too. Taken to the extreme, such a bug could lead to a test passing when it actually should fail (I'm not saying that's the case here but still).

PR #830, opened 2021-02-18
Upgrade tests - re-enable -race and increase log level

mgencur:
Updated deps to get the midstream fix for the Prober test which was causing race conditions and re-enabled -race because now it should work without problems.


mgencur:
@markusthoemmes can you take a look, please? This is ready to be reviewed/merged.

markusthoemmes (approved):
/lgtm
/approve

PR #832, opened 2021-03-01
Run custom domain test with coformance test

nak3:
Since serving v0.20+, `ResolvableDomain` flag is separated from custom domain.
Please refer to https://github.com/knative/serving/commit/82b59914fb9727f040b4d58073b2e6049b2d4583.

Hence the conformance test should be able to run with customdomain and
other tests.

### DIFF
--- a/test/serving.bash
+++ b/test/serving.bash
@@ -61,15 +61,9 @@ function upstream_knative_serving_e2e_and_conformance_tests {
   fi
 
   SYSTEM_NAMESPACE=knative-serving go_test_e2e -tags=e2e -timeout=30m -parallel=$parallel \
-    ./test/e2e ./test/conformance/api/v1/... ./test/conformance/runtime/... \
+    ./test/e2e ./test/conformance/api/... ./test/conformance/runtime/... \
     --resolvabledomain --kubeconfig "$KUBECONFIG" \
     --imagetemplate "$image_template"
-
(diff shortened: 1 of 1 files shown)


nak3:
/test 4.6-upgrade-tests-aws-ocp-46

upgreade test is not related to this PR.

nak3:
/cc @mgencur 

Could you please take a look?

test/serving.bash
```diff
@@ -61,15 +61,9 @@ function upstream_knative_serving_e2e_and_conformance_tests {
   fi
 
   SYSTEM_NAMESPACE=knative-serving go_test_e2e -tags=e2e -timeout=30m -parallel=$parallel \
-    ./test/e2e ./test/conformance/api/v1/... ./test/conformance/runtime/... \
+    ./test/e2e ./test/conformance/api/... ./test/conformance/runtime/... \
```
mgencur, line 64: The `--customdomain` is not passed anywhere. Is it supposed to be?

nak3: No, when we specify the `--customdomain`, the specified domain needs to be "resolvable". And it is difficult (=needs extra cost) to set up the second domain for CI :disappointed:

So I think it is OK to use the default hard-corded `example.org` with spoof client rather than setting up the resolable domain.

mgencur: OK. Thanks!

PR #833, opened 2021-03-01
Replace rbac prroxy image with the image built by Red Hat

nak3:
For some reason, rbac container does not start due to  `CreateContainerError`.

This patch tries to use rbac proxy image built by Red Hat.


markusthoemmes (approved):
/lgtm
/approve

PR #836, opened 2021-03-01
[SRVCOM-1237] Remove redundant service monitors during upgrades

skonto:
- fixes issue faced during upgrades
- tested it also via creating a sm "knative-openshift-metrics-2" and restarting the knative-openshift pod

### DIFF
--- a/knative-operator/pkg/common/monitoring.go
+++ b/knative-operator/pkg/common/monitoring.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"strings"
 
 	mfclient "github.com/manifestival/controller-runtime-client"
 	mf "github.com/manifestival/manifestival"
@@ -26,8 +25,6 @@ const (
 	monitoringLabel                   = "openshift.io/cluster-monitoring"
 	rolePath                          = "deploy/resources/monitoring/role-service-monitor.yaml"
 	TestRolePath                      = "TEST_ROLE_PATH"
-	operatorServiceMonitorNameEnvKey  = "OPERATOR_SERVICE_MONITOR_NAME"
-	operatorServiceMonitorDefaultName = "knative-openshift-metrics-3"
 )
 
 func SetupMonitoringRequirements(api client.Client, instance mf.Owner) error {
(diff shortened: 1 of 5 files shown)


knative-operator/pkg/common/monitoring.go
```diff
@@ -36,6 +42,36 @@ func SetupMonitoringRequirements(api client.Client, instance mf.Owner) error {
 	return nil
 }
 
+func RemoveOldServiceMonitorResources(namespace string, api client.Client) error {
+	currentSMName := os.Getenv(operatorServiceMonitorNameEnvKey)
+	if currentSMName == "" {
+		currentSMName = operatorServiceMonitorDefaultName
+	}
```
markusthoemmes, line 49: Is the defaulting necessary? Seems like we use explicit values everywhere, so I'd rather drop this to catch accidental issues.

knative-operator/pkg/common/monitoring.go
```diff
@@ -36,6 +42,36 @@ func SetupMonitoringRequirements(api client.Client, instance mf.Owner) error {
...
+	smList := monitoringv1.ServiceMonitorList{}
+	if err := api.List(context.TODO(), &smList, client.InNamespace(namespace)); err != nil {
+		return err
+	}
+	for _, sm := range smList.Items {
+		// Skip the sm that is being installed with the current operator version
+		if sm.Name != currentSMName && strings.HasPrefix(sm.Name, "knative-openshift-metrics-") {
+			// Delete service monitor and the related service monitor service, skip error if not found
+			if err := api.Delete(context.TODO(), sm); err != nil && !errors.IsNotFound(err) {
```
markusthoemmes, line 58:
```suggestion
			if err := api.Delete(context.Background(), sm); err != nil && !errors.IsNotFound(err) {
```

knative-operator/pkg/common/monitoring.go
```diff
@@ -36,6 +42,36 @@ func SetupMonitoringRequirements(api client.Client, instance mf.Owner) error {
...
+				return err
+			}
+			service := v1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: namespace,
+					Name:      sm.Name,
+				},
+			}
+			if err := api.Delete(context.TODO(), &service); err != nil && !errors.IsNotFound(err) {
```
markusthoemmes, line 67:
```suggestion
			if err := api.Delete(context.Background(), &service); err != nil && !errors.IsNotFound(err) {
```

knative-operator/pkg/common/monitoring.go
```diff
@@ -36,6 +42,36 @@ func SetupMonitoringRequirements(api client.Client, instance mf.Owner) error {
 	return nil
 }
 
+func RemoveOldServiceMonitorResources(namespace string, api client.Client) error {
+	currentSMName := os.Getenv(operatorServiceMonitorNameEnvKey)
+	if currentSMName == "" {
+		currentSMName = operatorServiceMonitorDefaultName
+	}
+	smList := monitoringv1.ServiceMonitorList{}
+	if err := api.List(context.TODO(), &smList, client.InNamespace(namespace)); err != nil {
```
markusthoemmes, line 51: I sort of wondered if it'd be simpler to enumerate the items we want to delete explicitly, but not a strong opinion, this approach looks viable as well.

markusthoemmes (commented):
Do we know if this is "only" an issue with our very first Service/SM and the subsequent ones (managed by OLM) are fine being handled by OLM?

If so, I'd target this towards deleting just this very first set of resources. If OLM doesn't garbage collect the newer stuff, this approach is fine too.

knative-operator/pkg/common/monitoring.go
skonto: I kept this a bit generic not having to define them per release.

knative-operator/pkg/common/monitoring.go
skonto: I could remove it yes.

skonto:
> Do we know if this is "only" an issue with our very first Service/SM and the subsequent ones (managed by OLM) are fine being handled by OLM?

I will double check and adapt the PR if we only need to remove the very first one.

knative-operator/pkg/common/monitoring.go
skonto: I was not sure if I could pass any ctx later on thus TODO but Background probably makes more sense.

markusthoemmes: Yeah the line is somewhat blurry, but TODO shouldn't be in any production code (IMO at least). It's supposed to trigger linter warnings I guess :shrug:

markusthoemmes (approved):
/lgtm
/approve

PR #837, opened 2021-03-02
[SRVKS-590] Don't reconcile KS/KE if placed in the wrong namespace.

markusthoemmes:
Apart from the checks in the webhooks, this aborts reconcilation for Knative Serving/Knative Eventing resources if they don't have been placed into the expected namespace. This can happen if the resources are placed before the webhook is up and similar cases. An error message is surfaced in the status of the resource.

/assign @skonto

### DIFF
--- a/knative-operator/pkg/controller/knativeeventing/knativeeventing_controller.go
+++ b/knative-operator/pkg/controller/knativeeventing/knativeeventing_controller.go
@@ -28,7 +28,7 @@ const (
 	// This needs to remain "knative-eventing-openshift" to be compatible with earlier versions.
 	finalizerName = "knative-eventing-openshift"
 
-	requiredNsKey = "REQUIRED_EVENTING_NAMESPACE"
+	requiredNsEnvName = "REQUIRED_EVENTING_NAMESPACE"
 )
 
 var log = common.Log.WithName("controller")
@@ -44,7 +44,7 @@ func newReconciler(mgr manager.Manager) reconcile.Reconciler {
 	client := mgr.GetClient()
 
 	// Create required namespace first.
-	if ns, required := os.LookupEnv(requiredNsKey); required {
+	if ns, required := os.LookupEnv(requiredNsEnvName); required {
 		client.Create(context.Background(), &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{
 			Name: ns,
 		}})
@@ -65,8 +65,8 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	}
 
 	// Watch for changes to primary resource KnativeEventing
+	requiredNs := os.Getenv(requiredNsEnvName)
 	return c.Watch(&source.Kind{Type: &eventingv1alpha1.KnativeEventing{}}, &handler.EnqueueRequestForObject{}, predicate.NewPredicateFuncs(func(obj client.Object) bool {
-		requiredNs := os.Getenv(requiredNsKey)
 		if requiredNs == "" {
 			return true
 		}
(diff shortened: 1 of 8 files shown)


markusthoemmes:
Unrelated but

/retest

knative-operator/pkg/controller/knativeeventing/knativeeventing_controller.go
```diff
@@ -19,12 +19,17 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
-// This needs to remain "knative-eventing-openshift" to be compatible with earlier versions.
-const finalizerName = "knative-eventing-openshift"
+const (
+	// This needs to remain "knative-eventing-openshift" to be compatible with earlier versions.
+	finalizerName = "knative-eventing-openshift"
+
+	requiredNsKey = "REQUIRED_EVENTING_NAMESPACE"
```
skonto, line 31: Better use the suffix "EnvName", "requiredNsEnvName"  eg. to be aligned with the rest of the code base.

knative-operator/pkg/controller/knativeeventing/knativeeventing_controller.go
```diff
@@ -60,7 +65,13 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	}
 
 	// Watch for changes to primary resource KnativeEventing
-	return c.Watch(&source.Kind{Type: &eventingv1alpha1.KnativeEventing{}}, &handler.EnqueueRequestForObject{})
+	return c.Watch(&source.Kind{Type: &eventingv1alpha1.KnativeEventing{}}, &handler.EnqueueRequestForObject{}, predicate.NewPredicateFuncs(func(obj client.Object) bool {
+		requiredNs := os.Getenv(requiredNsKey)
```
skonto, line 69: Can we move this out of the func?

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
```diff
@@ -80,8 +83,14 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
-	// Watch for changes to primary resource KnativeServing
-	err = c.Watch(&source.Kind{Type: &servingv1alpha1.KnativeServing{}}, &handler.EnqueueRequestForObject{})
+	// Watch for changes to primary resource KnativeServing, only in the expected namespace.
+	err = c.Watch(&source.Kind{Type: &servingv1alpha1.KnativeServing{}}, &handler.EnqueueRequestForObject{}, predicate.NewPredicateFuncs(func(obj client.Object) bool {
+		requiredNs := os.Getenv(requiredNsKey)
```
skonto, line 88: Same as above.

openshift-knative-operator/pkg/serving/extension_test.go
```diff
@@ -134,10 +135,24 @@ func TestReconcile(t *testing.T) {
...
+		expected: ks(func(ks *v1alpha1.KnativeServing) {
+			ks.Namespace = "foo"
+			ks.Status.MarkInstallFailed(`Knative Serving must be installed into the namespace "knative-serving"`)
+		}),
 	}}
 
 	for _, c := range cases {
 		t.Run(c.name, func(t *testing.T) {
+			// Default the namespace to the correct one if not set for brevity.
+			if c.in.Namespace == "" {
```
skonto, line 152: Could we also test the case where ns = ""

markusthoemmes: That's not possible in practice, is it? Since the API is namespace scoped there always has to be a namespace set.

skonto: My bad I was not clear, I mean the `requiredNs` to be "".

PR #842, opened 2021-03-04
Move proxy injection to OaaL

markusthoemmes:
As per title, this moves the super ugly injection of the proxy environment variables from an actual API command to a transformer in the new operator.

---

**EDIT:** Ran this with the (flaky) test enabled and it passed just fine. Ready for review.


markusthoemmes:
/retest

Infra...

openshift-knative-operator/pkg/common/deployment.go
```diff
@@ -0,0 +1,73 @@
...
+	return transformDeployment(deploymentName, func(deploy *appsv1.Deployment) error {
+		containers := deploy.Spec.Template.Spec.Containers
+		for i := range containers {
+			c := &containers[i]
+			if c.Name != containerName {
+				continue
+			}
+
+			for key, value := range env {
+				c.Env = upsert(c.Env, key, value)
```
skonto, line 25: Do we add this to the rbac proxy container too?

markusthoemmes: Currently, we don't. This specifically targets the controller container to enable digest resolution to work properly even in a HTTP_PROXY settings. AFAIK we don't have other components that might call to the outside and thus need those settings.

markusthoemmes:
Upstream leg is hosed on main. Informed the eventing team.

openshift-knative-operator/pkg/common/deployment.go
```diff
@@ -0,0 +1,73 @@
...
+			}
+		}
+
+		return nil
+	})
+}
+
+// upserts updates the env var if the key already exists or inserts it if it didn't
```
skonto, line 33: s/`upserts`/upsert

openshift-knative-operator/pkg/common/deployment.go
```diff
@@ -0,0 +1,73 @@
...
+		Name:  key,
+		Value: value,
+	})
+}
+
+// transformDeployment returns a transformer that transforms a deployment with the given
+// name.
+func transformDeployment(name string, f func(*appsv1.Deployment) error) mf.Transformer {
+	return func(u *unstructured.Unstructured) error {
+		if u.GetKind() != "Deployment" || u.GetName() != name {
```
skonto, line 54: Should not this be an AND condition?

openshift-knative-operator/pkg/common/deployment_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			u := &unstructured.Unstructured{}
+			if err := scheme.Scheme.Convert(test.in, u, nil); err != nil {
+				t.Fatal("Failed to convert deployment to unstructured", err)
+			}
+
+			if err := InjectEnvironmentIntoDeployment(test.deployment, test.container, test.env)(u); err != nil {
+				t.Fatal("Unexpecetd error from transformer", err)
```
skonto, line 123: s/`Unexpecetd`/`Unexpected`

openshift-knative-operator/pkg/common/deployment.go
markusthoemmes: Note that this exits early. So we want to exit early if the Kind is not a deployment OR if it is and the name does not match.

openshift-knative-operator/pkg/common/deployment.go
```diff
@@ -0,0 +1,73 @@
...
+	"k8s.io/client-go/kubernetes/scheme"
+)
+
+// InjectEnvironmentIntoDeployment injects the specified environment variables into the
+// specified deployment/container.
+// Note: We're not deleting empty environment variables and instead set them to empty
+// string. Three-way-merging of the deployment drops the update otherwise.
+func InjectEnvironmentIntoDeployment(deploymentName, containerName string, env map[string]string) mf.Transformer {
+	return transformDeployment(deploymentName, func(deploy *appsv1.Deployment) error {
+		containers := deploy.Spec.Template.Spec.Containers
```
skonto, line 17: Are we planning to modify the deployment in other ways in the future? I am thinking about having this high-order func vs just collapsing this with the transformDeployment body anyway.

openshift-knative-operator/pkg/common/deployment.go
skonto: I understand the short-cut condition but we might accidentally modify the wrong deployment if the kind is right but the name is not. Basically if it is a deployment we ignore the name no?

openshift-knative-operator/pkg/common/deployment.go
skonto: Ok I see that we skip if the container name does not match good.

openshift-knative-operator/pkg/common/deployment.go
markusthoemmes: No we don't. If it's a Deployment the first condition will be false, so the second one will be evaluated which will be true (because the name doesn't match), so it's true overall.

openshift-knative-operator/pkg/common/deployment.go
markusthoemmes: I had the collapsed version before and this is imo nicer to read in that the annoying bits are out of the codepath I want to read.

openshift-knative-operator/pkg/common/deployment.go
skonto: Right. So it is equivelant to ` u.GetKind() == "Deployment" && u.GetName() == name` afaik.

skonto: Right. Since we are transforming a Deployment (name of the method) we should return quickly if the kind is not a deployment anyway, not check the name.

skonto: Right.

skonto:
@markusthoemmes what is the status with the eventing issue?

markusthoemmes:
Nope :(

PR #847, opened 2021-03-11
Rewrite Serverless and OpenShift upgrade to Golang

mgencur:
This is the remaining step to convert upgrade test suites in serverless-operator to Golang. The other parts - tests and test framework have already been migrated.


test/flags.go
```diff
@@ -13,9 +13,15 @@ var Flags = initializeFlags()
...
-	Channel       string // serverless-operator Subscription channel
+	Kubeconfigs     string // Path to .kube/config
+	CatalogSource   string // CatalogSource in the openshift-marketplace namespace for the Serverless operator Subscription
+	Channel         string // Serverless operator Subscription channel
+	Subscription    string // Serverless operator Subscription name
+	UpgradeChannel  string // Target OLM channel for upgrades
+	CSV             string // Target CSV for upgrades
+	ServingVersion  string // Target Serving version for upgrades
+	EventingVersion string // Target Eventing version for upgrades
+	OpenShiftImage  string // Target OpenShift image for upgrades
```
cardil, line 24: It would be much easier to use environmental variables instead of go flags for those. Using https://github.com/kelseyhightower/envconfig

mgencur: Thanks for the suggestion. It's not clear to me how it would things easier though. We need to set the name, set the description, and the default value anyway. What would it actually bring? Curious...
The problem with kelseyhightoewer library is that there's no clear connection between environment variables and the fields of the given struct. Basically, you need to know that the given project is using this library otherwise it's confusing and hard to find where the fields are being set. You can't grep the code. There's no direct assignment.
OTOH, passing them as flags makes it clear where they're coming from.
When it comes to calling the tests, whether you pass the argument as a flag or you define an environment variable, it's more or less the same effort.
I used flags to make it consistent with the other flags that we already have such as `channel`, `catalogsource`. 
It's true that the upgrade tests still use two env variables: SKIP_SERVING_PREUPGRADE and UPGRADE_CLUSTER. However, I see them as something that is orthogonal - you don't have to use them - they're optional. Anyway. I could turn them into flags as well if this is desired.

markusthoemmes: Imo, let's stay consistent and use flags.

mgencur:
The latest failures are related to the recent changes to Kafka. I've just rebased, pushed, and trying to see if it helps.

mgencur:
Alright. Tests passed.

test/installplan.go
```diff
@@ -0,0 +1,57 @@
...
+		}
+		return false, nil
+	}); waitErr != nil {
+		return plan, fmt.Errorf("installplan for CSV %s and OLM source %s not found: %w", csvName, olmSource, waitErr)
+	}
+	return plan, nil
+}
+
+func installsCSVFromSource(installPlan v1alpha1.InstallPlan, csvName, olmSource string) bool {
+	if installPlan.Status.BundleLookups[0].CatalogSourceRef.Name != olmSource {
```
maschmid, line 35: Probably could panic if either

BundleLookups is nil or empty
CatalogSourceRef is nil

test/installplan.go
```diff
@@ -0,0 +1,57 @@
...
+	return false
+}
+
+func ApproveInstallPlan(ctx *Context, name string) error {
+	plan, err := ctx.Clients.OLM.OperatorsV1alpha1().InstallPlans(OperatorsNamespace).Get(context.Background(), name, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	plan.Spec.Approved = true
+	_, err = ctx.Clients.OLM.OperatorsV1alpha1().InstallPlans(OperatorsNamespace).Update(context.Background(), plan, metav1.UpdateOptions{})
```
maschmid, line 52: We should use either use Patch, or use RetryOnConflict   (note we hit issues before we switched to "oc patch" for installplan approval , see https://github.com/openshift-knative/serverless-operator/pull/727

test/subscription.go
```diff
@@ -91,21 +93,34 @@ func WaitForSubscriptionState(ctx *Context, name, namespace string, inState func
...
 }
 
+func UpdateSubscriptionChannelSource(ctx *Context, name, channel, source string) (*v1alpha1.Subscription, error) {
+	subs, err := ctx.Clients.OLM.OperatorsV1alpha1().Subscriptions(OperatorsNamespace).Get(context.Background(), name, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	subs.Spec.Channel = channel
+	subs.Spec.CatalogSource = source
+	return ctx.Clients.OLM.OperatorsV1alpha1().Subscriptions(OperatorsNamespace).Update(context.Background(), subs, metav1.UpdateOptions{})
```
maschmid, line 103: We should either use Patch, or RetryOnConflict

test/upgrade/installation/openshift.go
```diff
@@ -0,0 +1,92 @@
...
+
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
+
+	configv1 "github.com/openshift/api/config/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+func UpgradeOpenShift(ctx *test.Context) error {
+	clusterVersionName := "version"
```
maschmid, line 19: const?

test/upgrade/installation/openshift.go
```diff
@@ -0,0 +1,92 @@
...
+
+	return nil
+}
+
+func WaitForClusterVersionState(ctx *test.Context, name string, inState func(s *configv1.ClusterVersion) bool) (*configv1.ClusterVersion, error) {
+	var lastState *configv1.ClusterVersion
+	var err error
+	waitErr := wait.PollImmediate(30*time.Second, 2*time.Hour, func() (bool, error) {
+		lastState, err = ctx.Clients.ConfigClient.ClusterVersions().Get(context.Background(), name, metav1.GetOptions{})
+		if apierrs.IsUnauthorized(err) {
```
maschmid, line 69: Is this really the only kind of errors during a cluster upgrade? (I'd expect even networking errors when api servers are restarted... maybe we should just log any kind of errors here and patiently wait... we can reconsider if we ever hit a case when a clear error is returned for the 2 hours...

test/upgrade/installation/serverless.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	if err := test.ApproveInstallPlan(ctx, installPlan.Name); err != nil {
+		return err
+	}
+	if _, err := test.WaitForClusterServiceVersionState(ctx, test.Flags.CSV, test.OperatorsNamespace, test.IsCSVSucceeded); err != nil {
+		return err
+	}
+
+	knativeServing := "knative-serving"
```
maschmid, line 27: const?

test/upgrade/installation/openshift.go
mgencur: Yeah. Good point. That's the only error I saw during the cluster upgrade. We're only getting the ClusterVersion object. Maybe there are other errors in the cluster but we're only retrieving the ClusterVersion. What about this: keep this check as strict/narrow as possible and if we hit other errors we can reconsider the approach, after checking if the errors make sense. Rather than opening the doors completely. The errors will be visible in Jenkins because the build will fail. Anyway, either approach is good, IMO.

test/installplan.go
mgencur: Right. Previously I was looking at the other places in the repo and what's used there. It's the Update method but there we use it in the "reconciler" loop so it will be reapplied if it fails. Just pushed a fix using the Patch.

mgencur:
/hold
Waiting for https://github.com/openshift-knative/serverless-operator/pull/854 to be merged first

mgencur:
/hold cancel
The other PR has been merged.

PR #848, opened 2021-03-12
[main] UI serving tests independent execution

cardil:
Serving tests are using the same ksvc for each test case. If one of them fails, it will not clean up the app, so then subsequent tests fails.

This PR should make those independent, by using different names for each test case.


test/e2e-tests.sh
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+# TODO(ksuszyns): Remove this before merging
+exec "$(dirname "${BASH_SOURCE[0]}")/ui-e2e-tests.sh"
```
cardil, line 4: /hold

Remove those lines before merging

cardil:
Here's a [passing build `1382387299718795264`](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift-knative_serverless-operator/848/pull-ci-openshift-knative-serverless-operator-main-4.7-operator-e2e-aws-ocp-47/1382387299718795264) for cbf56a6.

Removing temp commit.

test/e2e-tests.sh
cardil: Commit removed.

/unhold

PR #856, opened 2021-03-24
Disable kafka topic autocreation

maschmid:


matzew:
good catch... we have that in upstream too:

https://github.com/knative-sandbox/eventing-kafka/pull/271

/lgtm
/approve

PR #857, opened 2021-03-24
Remove last remnants of svc.ci registry

maschmid:


maschmid:
registry.svc.ci.openshift.org is now gone, however, some of the old images are probably not on the new one....

hack/lib/catalogsource.bash
```diff
@@ -119,7 +119,7 @@ spec:
...
+          /bin/opm registry add -d index.db --container-tool=podman --mode=replaces -b quay.io/openshift-knative/serverless-bundle:1.7.2,quay.io/openshift-knative/serverless-bundle:1.8.0,quay.io/openshift-knative/serverless-bundle:1.9.0,quay.io/openshift-knative/serverless-bundle:1.10.0,quay.io/openshift-knative/serverless-bundle:1.10.1,quay.io/openshift-knative/serverless-bundle:1.11.0,registry.ci.openshift.org/openshift/openshift-serverless-v1.12.0:serverless-bundle,registry.ci.openshift.org/openshift/openshift-serverless-v1.13.0:serverless-bundle,image-registry.openshift-image-registry.svc:5000/$OLM_NAMESPACE/serverless-bundle && \
```
matzew, line 122: Do we have image mirroring setup for these? or is there a "manual" push/sync to `quay.io` ?

maschmid: manual for now... we need to fix the branches first before we can define mirroring (those images don't exist yet even on registry.svc.ci.openshift.org , because the jobs for old branches are broken, so there is nothing to mirror yet)

matzew:
/lgtm
/approve

:see_no_evil:

PR #863, opened 2021-03-29
Test CI after build image fix

mgencur:


mgencur:
There are test failures which are unrelated. But the image build works. Closing.

PR #866, opened 2021-03-29
Increase cluster upgrade timeout to 3 hours

mgencur:
* some OpenShift cluster upgrades might take longer than 2 hours

The last upgrade I tried from OCP 4.7.2 to 4.7.4 took 2h 12m.

Note: Increasing the timeout shouldn't have any effect on the tests because cluster upgrade tests are only executed downstream (not in this repo).


markusthoemmes (approved):
/lgtm
/approve

PR #870, opened 2021-03-31
[SRVKS-712] Add sidecar.istio.io/rewriteAppHTTPProbers annotation for TestKsvcWithServiceMeshJWTDefaultPolicy

nak3:
This patch adds `sidecar.istio.io/rewriteAppHTTPProbers` annotation to
the Ksvc in TestKsvcWithServiceMeshJWTDefaultPolicy.

When JWT is enabled, HTTP requests is blocked by auth policy. So
rewriteAppHTTPProbers option lets sidecar redirect the request to
application.

/cc @markusthoemmes @maschmid @mgencur

### DIFF
--- a/test/servinge2e/servicemesh_test.go
+++ b/test/servinge2e/servicemesh_test.go
@@ -601,7 +601,8 @@ func TestKsvcWithServiceMeshJWTDefaultPolicy(t *testing.T) {
 
 		// Create a test ksvc, should be accessible only via proper JWT token
 		testKsvc := test.Service("jwt-test", testNamespace, image, map[string]string{
-			"sidecar.istio.io/inject": "true",
+			"sidecar.istio.io/inject":                "true",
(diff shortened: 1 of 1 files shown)


markusthoemmes:
Ohhhh interesting! We do this upstream too right? Do we need to change docs/add release notes for this?

nak3:
Istio enabled the option by default since Istio v1.6 https://github.com/istio/istio/commit/76ec44c27f079f5254463d157fc94f2f69cadf21 so upstream does not hit this issue as long as users do not use old Istio ([Istio 1.6 was EOL since Nov/2020](https://istio.io/latest/about/supported-releases/#support-status-of-istio-releases) so I think we don't need to care about it in upstream :sunglasses:).
For docs change, yes I will file the request once this solution was accepted.

markusthoemmes (approved):
/lgtm
/approve

PR #872, opened 2021-04-01
Fix memory usage unit

skonto:
We don't use rate() for memory usage. [Memory usage (container_memory_usage_bytes)](https://github.com/google/cadvisor/blob/master/docs/storage/prometheus.md#prometheus-container-metrics) is measured via a gauge which cant be used with [rate()](https://prometheus.io/docs/prometheus/latest/querying/functions/#rate) otherwise it would cause wrong results when gauge decreases. Thus the unit is wrong.

### DIFF
--- a/knative-operator/deploy/resources/dashboards/grafana-dash-knative-eventing-resources.yaml
+++ b/knative-operator/deploy/resources/dashboards/grafana-dash-knative-eventing-resources.yaml
@@ -154,7 +154,7 @@ data:
           "thresholds": [],
           "timeFrom": null,
           "timeShift": null,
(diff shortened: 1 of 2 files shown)


markusthoemmes (approved):
/lgtm
/approve

nak3:
/lgtm

Yes, definitely memory doesn't(cannot) call `rate()`:

https://github.com/openshift-knative/serverless-operator/blob/e440d216b39028699cc9ce54f2fbba4534c07783/knative-operator/deploy/resources/dashboards/grafana-dash-knative-serving-resources.yaml#L161

PR #871, opened 2021-03-31
svc.ci -> ci for release-1.11

maschmid:

PR #817, opened 2021-02-09
Update CSV for 1.10.1

warrenvw:
Updated CSV template to reference 1.10.1, and executed `make csv` to render `olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml`.

PR #873, opened 2021-04-05
[1.9] registry.svc.ci -> registry.ci

aliok:
Similar to https://github.com/openshift-knative/serverless-operator/pull/871


aliok:
I guess we need to do something like this : https://github.com/openshift-knative/serverless-operator/pull/817/commits/c4a2ed3cf9fb8ebd8bd7373654058ccbb56219db

matzew:
I think there is more needed... on the related fork branches too...

/cc @maschmid

matzew:
@maschmid @aliok what's the status here?   Do we still need this ?

aliok:
I don't know if this is still needed. Leaving it to @maschmid

maschmid:
Let's close this one. It would require changes in the other repos first, and also the latest supported is 1.10, so it is not a priority to fix the older versions now. If anybody wants to make older work, they can open a new one then...

PR #878, opened 2021-04-08
Delete KnativeKafka before KnativeEventing to workaround SRVKE-667

mgencur:
Workaround for https://issues.redhat.com/browse/SRVKE-667 . The teardown function is used in nightly jobs where multiple test suites run in the same job. The jobs often timeout because of this issue.


aliok (approved):
/lgtm
/approve

PR #880, opened 2021-04-08
Increase timeout for WaitForUnstructuredState

mgencur:
* we might wait for ServiceMesh control plane which might take longer on
some platforms to start

Fix for the weekly job on Azure: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-knative-serverless-operator-main-4.7-azure-e2e-azure-ocp-47-continuous/1378497697421463552

The Grafana pod for ServiceMesh is being started as the last one. And in my tests I got the same failure on Azure for the first time. But any further attempt on the cluster was successful.

### DIFF
--- a/test/unstructured.go
+++ b/test/unstructured.go
@@ -3,6 +3,7 @@ package test
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


test/unstructured.go
```diff
@@ -61,7 +62,7 @@ func WaitForUnstructuredState(ctx *Context, schema schema.GroupVersionResource,
 		lastState *unstructured.Unstructured
 		err       error
 	)
-	waitErr := wait.PollImmediate(Interval, Timeout, func() (bool, error) {
+	waitErr := wait.PollImmediate(Interval, 10*time.Minute, func() (bool, error) {
```
nak3, line 65: Could you leave a comment in the code that we need the longer timeout for SeerviceMesh control plane?

mgencur: I'd prefer the comment in the commit as I did. This function can be used for other resources, not only Service Mesh. And I don't know if anyone will ever wonder why this is 10 minutes. I mean, both 5 and 10 minutes are just fine. That comment might be unnecessary. WDYT?

nak3:
LGTM
/assign @maschmid

maschmid:
/lgtm

I guess we might want to move the timeout to the args eventually...

PR #884, opened 2021-04-09
[main] Set KafkaChannel as default if INSTALL_KAFKA is true

mgencur:
This is required for upgrade tests so that Eventing Brokers use the KafkaChannel.

### DIFF
--- a/hack/lib/serverless.bash
+++ b/hack/lib/serverless.bash
@@ -44,6 +44,7 @@ function install_serverless_previous {
   fi
   if [[ $INSTALL_KAFKA == "true" ]]; then
     deploy_knativekafka_cr
+    ensure_kafka_channel_default
   fi
 
   logger.success "Previous version of Serverless is installed: $PREVIOUS_CSV"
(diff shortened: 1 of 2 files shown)


matzew (commented):
/lgtm
/assign @lberk for final review (he recently had to deal with this)

FWIW: upstream doc for that is now also present: https://github.com/knative/docs/pull/3433/files

matzew:
/hold
for @lberk

matzew:
@mgencur do we want this on the 1.14 branch too ?

mgencur:
@matzew yes, I was waiting for a final review before creating the other PR

lberk:
lgtm in principle.  My only question is if we want to set a default `repliactionFactor` or `numPartitions` for testing in the future?
/unhold

mgencur:
@lberk I'm not sure about setting the `replicationFactor` and `numPartitions`. It would be good if the tests used the most common value, perhaps the default. When do you think we'd need to set the two parameters?

PR #887, opened 2021-04-12
[SRVCOM-1274] Secure Metrics Endpoints for Eventing control plane pods

skonto:
- secures all endpoints for the Eventing control plane pods. Note: it does not cover Eventing sources (will be done in another PR)
- moves rbac and other resource setup from knative-operator to the openshift-knative-operator.
- adds required tests
- first commit secures the endpoints while the second is for testing.


knative-operator/pkg/common/service_monitor.go
```diff
@@ -14,39 +14,21 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
```
skonto, line 14: Eventually `service_monitor.go` will be moved to the sources dir.

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -120,27 +152,77 @@ func InjectNamespaceWithSubject(resourceNamespace string, subjectNamespace strin
...
+			if smManifest != nil {
+				rbacManifest = rbacManifest.Append(*smManifest)
+			}
 		}
 	}
+
 	return []mf.Manifest{rbacManifest}, nil
 }
 
+func GetCompMonitoringPlatformManifests(comp v1alpha1.KComponent) ([]mf.Manifest, error) {
```
matzew, line 207: `GetCompMonitoringPlatformManifests` ?

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -21,45 +22,76 @@ import (
...
-	if backend == "none" {
+func getCommonSpec(comp v1alpha1.KComponent) (*v1alpha1.CommonSpec, KComponentType) {
+	switch c := comp.(type) {
+	case *v1alpha1.KnativeServing:
+		return &c.Spec.CommonSpec, Serving
+	case *v1alpha1.KnativeEventing:
+		return &c.Spec.CommonSpec, Eventing
+	}
+	return nil, ""
+}
```
markusthoemmes, line 77: Is this necessary? Why can't we use `comp.GetSpec()`?

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -200,6 +282,66 @@ func createServiceMonitorService(component string) corev1.Service {
...
+
+// getDefaultMetricsPort returns the expected metrics port under the assumption that this will not change
+// This is static information since observability cm does not allow any changes for the prometheus config
+func getDefaultMetricsPort(name string) string {
+	if servingComponents.Has(name) {
+		return "9090"
+	}
+	if name == "mt-broker-ingress" || name == "mt-broker-filter" {
+		return "9092"
```
matzew, line 323: different b/c upstream is "wrong" ?

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -81,15 +113,15 @@ func reconcileMonitoringLabelOnNamespace(ctx context.Context, namespace string,
 	}
 	log := logging.FromContext(ctx)
 	if enable {
-		log.Info("Enabling Serving monitoring")
+		log.Info(fmt.Sprintf("Enabling %s monitoring", cType))
```
markusthoemmes, line 116:
```suggestion
		log.Infof("Enabling %s monitoring", cType)
```

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -200,6 +282,66 @@ func createServiceMonitorService(component string) corev1.Service {
...
+	if servingComponents.Has(name) {
+		return "9090"
+	}
+	if name == "mt-broker-ingress" || name == "mt-broker-filter" {
+		return "9092"
+	}
+	return "9090"
+}
+
+func getSelectorLabels(component string) map[string]string {
```
matzew, line 328: perhaps a comment on the exceptions/cases?

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/openshift-knative/serverless-operator/openshift-knative-operator/pkg/common"
```
matzew, line 11: would it make sense to have specific `monitorign` files for:
* serving
* eventing
* kafka 
?

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -120,27 +152,77 @@ func InjectNamespaceWithSubject(resourceNamespace string, subjectNamespace strin
...
+			rbacManifest = rbacManifest.Append(*crbM)
+		}
+		for c := range eventingComponents {
+			smManifest, err := constructServiceMonitorResourceManifests(c, ns)
+			if err != nil {
+				return nil, err
+			}
+			if smManifest != nil {
+				rbacManifest = rbacManifest.Append(*smManifest)
+			}
```
matzew, line 200: same block as: https://github.com/openshift-knative/serverless-operator/pull/887/files#diff-9441c1af1419ac92923cabd71411c98c7702acbdad9baec26fddf7a8a4bb221bR173-R179

perhaps move to function?

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -21,45 +22,76 @@ import (
...
 
 func init() {
 	_ = monitoringv1.AddToScheme(scheme.Scheme)
 }
 
-func ReconcileServingMonitoring(ctx context.Context, api kubernetes.Interface, ks *v1alpha1.KnativeServing) error {
-	backend := ks.Spec.CommonSpec.Config[ObservabilityCMName][ObservabilityBackendKey]
-	if shouldEnableMonitoring(backend) {
-		if err := reconcileMonitoringLabelOnNamespace(ctx, ks.Namespace, api, true); err != nil {
+func ReconcileMonitoringForKComponent(ctx context.Context, api kubernetes.Interface, comp v1alpha1.KComponent) error {
```
markusthoemmes, line 52: This is imo unnecessary "code-deduplication". When calling this, you know if you want to deal with Serving and Eventing. You forget that through the boundary of this interface only to then pick it apart in the implementation after the interface again.

Can we come up with a version for Serving and Eventing that composes the "helper" functions herein but doesn't need `KComponentType`?

matzew: yeah. would be good to call into specific code for each "Component", from the monitoring file

openshift-knative-operator/pkg/monitoring/monitoring.go
skonto: @matzew what is the question?

openshift-knative-operator/pkg/monitoring/monitoring.go
skonto: @matzew I dont want to block this until upsteam is fixed.

openshift-knative-operator/pkg/monitoring/monitoring.go
skonto: I dont see any difference at the moment, maybe in the future.

openshift-knative-operator/pkg/monitoring/monitoring.go
skonto: Because we want to the common Spec not the direct Spec for each comp. Common Spec is embedded in the top level spec.

openshift-knative-operator/pkg/monitoring/monitoring.go
skonto: I made it a bit agnostic. If I do it the way you propose (initially it was like this) then it means I need to calculate several parameters before `ReconcileMonitoringForKComponent` is called like the common spec, ns all at the extension side so I can take advantage that I know what instance the comp is and avoid type casting. Also I would have to pass the name of the instance explicitly because it is used in logging. It is a trade-off, either pollute the extension code and use the info or make it less specific and hide the details.

openshift-knative-operator/pkg/monitoring/monitoring.go
skonto: Yes could be done.

openshift-knative-operator/pkg/monitoring/monitoring.go
markusthoemmes: I guess the point is that it doesn't quite hide the details, but just moves them to other places. The monitoring functions "fake" a genericness that doesn't seem to be applicable.

openshift-knative-operator/pkg/monitoring/monitoring.go
matzew: let's do it ?

openshift-knative-operator/pkg/monitoring/monitoring.go
matzew: that's fine. let's have a comment here to clean this up, and file upstream issue, to fix it there too ?

openshift-knative-operator/pkg/monitoring/monitoring.go
skonto: Extension code does not know the details. That is what I mean by hiding.

openshift-knative-operator/pkg/monitoring/monitoring.go
matzew: I see already if lot of if/else, so perhaps make specific file for serving/eventing, with the usage of "common" util?

Adding kafka will benefit from this

skonto: Kafka has no special handling in this context.

skonto:
@markusthoemmes gentle ping :)

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -21,45 +22,73 @@ import (
...
 
 func init() {
 	_ = monitoringv1.AddToScheme(scheme.Scheme)
 }
 
-func ReconcileServingMonitoring(ctx context.Context, api kubernetes.Interface, ks *v1alpha1.KnativeServing) error {
-	backend := ks.Spec.CommonSpec.Config[ObservabilityCMName][ObservabilityBackendKey]
-	if shouldEnableMonitoring(backend) {
-		if err := reconcileMonitoringLabelOnNamespace(ctx, ks.Namespace, api, true); err != nil {
+func ReconcileMonitoringForServing(ctx context.Context, api kubernetes.Interface, config v1alpha1.ConfigMapData, spec *v1alpha1.CommonSpec, ns string) error {
```
markusthoemmes, line 52: Why pass both ConfigMapData and the CommonSpec? Can't the former be taken off the latter easily?

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -21,45 +22,73 @@ import (
...
+	return reconcileMonitoring(ctx, api, config, spec, Serving, ns)
+}
+
+func ReconcileMonitoringForEventing(ctx context.Context, api kubernetes.Interface, config v1alpha1.ConfigMapData, spec *v1alpha1.CommonSpec, ns string) error {
+	return reconcileMonitoring(ctx, api, config, spec, Eventing, ns)
+}
+
+func reconcileMonitoring(ctx context.Context, api kubernetes.Interface, config v1alpha1.ConfigMapData, spec *v1alpha1.CommonSpec, cType KComponentType, ns string) error {
```
markusthoemmes, line 60: Same q as above.

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -120,27 +149,78 @@ func InjectNamespaceWithSubject(resourceNamespace string, subjectNamespace strin
 	}
 }
 
-func LoadServingMonitoringPlatformManifests(ns string) ([]mf.Manifest, error) {
-	rbacPath := os.Getenv(servingSMRbacManifestPath)
+func getMonitoringPlarformNanifests(ns string, cType KComponentType) ([]mf.Manifest, error) {
```
markusthoemmes, line 152:
```suggestion
func getMonitoringPlatformManifests(ns string, cType KComponentType) ([]mf.Manifest, error) {
```

openshift-knative-operator/pkg/monitoring/monitoring.go
```diff
@@ -120,27 +149,78 @@ func InjectNamespaceWithSubject(resourceNamespace string, subjectNamespace strin
...
+	rbacPath := os.Getenv(smRbacManifestPath)
 	if rbacPath == "" {
 		return nil, fmt.Errorf("failed to get the Serving sm rbac manifest path")
```
markusthoemmes, line 155: Not serving specific anymore.

openshift-knative-operator/pkg/monitoring/monitoring.go
skonto: I suppose, this is a confusing struct hierarchy.

skonto:
@matzew could you add your comments here before I do another update round? I have addressed some of your comments but some more @markusthoemmes added need to be addressed.

openshift-knative-operator/pkg/monitoring/monitoring.go
skonto: This is done now

openshift-knative-operator/pkg/monitoring/monitoring.go
skonto: done

openshift-knative-operator/pkg/monitoring/monitoring.go
skonto: done

openshift-knative-operator/pkg/monitoring/monitoring.go
skonto: done

openshift-knative-operator/pkg/monitoring/monitoring.go
skonto: Name has changed.

markusthoemmes (approved):
/lgtm
/approve

Thanks for the changes!

PR #888, opened 2021-04-12
[SRVCOM-1237] Backport 836

skonto:
- Backports #836 to avoid the issue when users upgrade to 4.6(1.13) but they are not on 1.14 yet (faced by users lately).

### DIFF
--- a/knative-operator/cmd/manager/main.go
+++ b/knative-operator/cmd/manager/main.go
@@ -136,6 +136,12 @@ func setupMonitoring(cfg *rest.Config) error {
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


markusthoemmes:
Do we even plan to release 1.13 again? The respective user should be able to just upgrade to 1.14.

skonto:
@markusthoemmes If images are not updated automatically downstream then we need to think about it because not all people go to 1.14 although they should if possible. At least this is what I get as feedback.

maschmid:
/lgtm

not sure if there'll ever be a 1.13.1, but sure, we can have this there if there is...

PR #889, opened 2021-04-12
[release-1.14] Set E2E_UPGRADE_TESTS_FAILONERRORS=true as we're running with Kafka

mgencur:
* https://issues.redhat.com/browse/SRVKE-593 should be fixed now

### DIFF
--- a/test/lib.bash
+++ b/test/lib.bash
@@ -203,6 +203,7 @@ function run_rolling_upgrade_tests {
   E2E_UPGRADE_TESTS_SERVING_USE=true \
   E2E_UPGRADE_TESTS_CONFIGMOUNTPOINT=/.config/wathola \
   E2E_UPGRADE_TESTS_INTERVAL="50ms" \
+  E2E_UPGRADE_TESTS_FAILONERRORS=true \
   GO_TEST_VERBOSITY=standard-verbose \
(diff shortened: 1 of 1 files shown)


test/lib.bash
```diff
@@ -203,6 +203,7 @@ function run_rolling_upgrade_tests {
   E2E_UPGRADE_TESTS_SERVING_USE=true \
   E2E_UPGRADE_TESTS_CONFIGMOUNTPOINT=/.config/wathola \
   E2E_UPGRADE_TESTS_INTERVAL="50ms" \
```
cardil, line 205: What do you think about lowering this interval (between events)? EventingKafka on upstream operates with `2 msec`

https://github.com/knative-sandbox/eventing-kafka/blob/48351009c15aa61fa0837b7fc3ac6aba286c6ad9/test/upgrade/continual.go#L48

mgencur: IMO, 50ms (or 20 times per second) is a good enough setting for upgrade tests. Should we implement a load test we'd need to increase it. But for this case it looks good to me.

mgencur:
@cardil hmm, the upgrade tests fail with `Duplicate events: event #2043 received 2 times, but should be received only once` .

cardil (approved):
/lgtm
/retest

PR #893, opened 2021-04-13
SRVCOM-1282 Fix UI test failures from nightlies

mgencur:
https://issues.redhat.com/browse/SRVCOM-1282

The problem was that the Kafka tests left a pull secret in the default serviceaccount in serverless-tests namespace. The UI tests are running after that, were trying use the pull secret but it didn't exist anymore. The Kafka tests were removing the pull secret but didn't remove it from the service account. This PR removes them from SA as well. 
There were also a few more problems with resource cleanup from previous test suites.

### DIFF
--- a/test/ui/cypress/integration/serving.spec.js
+++ b/test/ui/cypress/integration/serving.spec.js
@@ -101,7 +101,10 @@ describe('OCP UI for Serverless', () => {
 
   const showcaseKsvc = new ShowcaseKservice()
 
-  it('can deploy kservice and scale it', () => {
+  it('can deploy kservice and scale it',
+    // TODO: Remove the increased timeout when https://issues.redhat.com/browse/ODC-5685 is fixed.
+    { defaultCommandTimeout: 300000 },
+    () => {
     describe('with authenticated via Web Console', () => {
       cy.login()
     })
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -51,7 +51,7 @@ test-e2e:
 
 # Run E2E tests from the current repo for serving+eventing+knativeKafka
 test-e2e-with-kafka:
-	INSTALL_KAFKA=true TEST_KNATIVE_KAFKA=true ./test/e2e-tests.sh
+	INSTALL_KAFKA=true TEST_KNATIVE_KAFKA=true ./test/e2e-tests.sh ./test/ui-e2e-tests.sh
```
mgencur, line 54: Will remove this line when the tests pass

test/ui/cypress/integration/serving.spec.js
```diff
@@ -91,7 +91,10 @@ describe('OCP UI for Serverless', () => {
 
   const showcaseKsvc = new ShowcaseKservice()
 
-  it('can deploy kservice and scale it', () => {
+  it('can deploy kservice and scale it',
+    // TODO: Remove the increased timeout when https://issues.redhat.com/browse/ODC-5685 is fixed.
+    { defaultCommandTimeout: 300000 },
```
cardil, line 96: Maybe we can move it to a specific invocation that fails?

mgencur: Pushed.

cardil (approved):
/lgtm

PR #894, opened 2021-04-13
Don't recompute image-from-env map twice

markusthoemmes:
A small nit found while hunting down the most recent upgrade test bug. Let's see if the upgrade test passes here.


markusthoemmes:
/retest

This should work now that 1.14 is fixed.

PR #895, opened 2021-04-13
[1.14] Configuring non trivial KafkaChannel options

cardil:
[1.14] Apart from adding retries, this configures also the number of partitions and replication factor. Both of those is 1 by default.

This PR should address https://github.com/knative-sandbox/eventing-kafka/issues/533 for serverless-operator and productized builds.

### DIFF
--- a/hack/lib/serverless.bash
+++ b/hack/lib/serverless.bash
@@ -220,8 +220,8 @@ function ensure_kafka_channel_default {
   logger.info 'Set KafkaChannel as default'
   local defaultChConfig channelTemplateSpec yamls patchfile
   yamls="$(dirname "$(realpath "${BASH_SOURCE[0]}")")/yamls"
-  read.in defaultChConfig < "${yamls}/kafka-default-ch-config.yaml"
-  read.in channelTemplateSpec < "${yamls}/kafka-channel-templatespec.yaml"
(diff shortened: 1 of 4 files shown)


hack/lib/serverless.bash
```diff
@@ -218,16 +218,19 @@ EOF
 
 function ensure_kafka_channel_default {
   logger.info 'Set KafkaChannel as default'
-
+  local defaultChConfig channelTemplateSpec yamls
+  yamls="$(dirname "$(realpath "${BASH_SOURCE[0]}")")/yamls"
+  read.in defaultChConfig < "${yamls}/kafka-default-ch-config.yaml"
+  read.in channelTemplateSpec < "${yamls}/kafka-channel-templatespec.yaml"
```
markusthoemmes, line 224: I read the comment on the `read.in` function but honestly don't quite understand the reasoning :sweat_smile:. What's the benefit vs. doing `defaultChConfig=$(cat "${yamls}/kafka-default-ch-config.yaml)`?

cardil: You're right. Let's simplify that. I just thought it will be necessary...

cardil:
Flake `TestContainerSourceV1Alpha2`

PR #898, opened 2021-04-13
Like on older releases, go with a dispatcher directly

matzew:

As on 1.13 of the product, we go with the dispatcher to default to `1` replica initially 

See old behavior: https://github.com/openshift-knative/serverless-operator/blob/release-1.13/knative-operator/deploy/resources/knativekafka/kafkachannel-v0.19.1.yaml#L512


matzew:
/hold
to make sure all goes green, before merges

PR #901, opened 2021-04-15
Add knative.openshift.io/system-namespace label in README for servicemesh

nak3:
This patch adds commands to add `knative.openshift.io/system-namespace` in README.

```
oc label namespace knative-serving knative.openshift.io/system-namespace=true
oc label namespace knative-serving-ingress knative.openshift.io/system-namespace=true
```

OSSM's networkpolicy blocks traffic from other namespaces so official
doc has [the guide](https://docs.openshift.com/container-platform/4.7/serverless/networking/serverless-ossm.html) but README misses it even though it has `allow-from-serving-system-namespace` NetworkPolicy.

/cc @maschmid

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,6 +232,13 @@ spec:
     # Add namespace you want to include mesh.
 ```
 
+and add `knative.openshift.io/system-namespace` label to system namespaces.
+
+```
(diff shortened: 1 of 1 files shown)


markusthoemmes (approved):
/lgtm
/approve

PR #907, opened 2021-04-21
[SRVCOM-1284] Update Serving and Eventing CRDs

nak3:
This patch updates `operator_v1alpha1_knativeeventing_crd.yaml` and `operator_v1alpha1_knativeserving_crd.yaml` that are very old.

### DIFF
--- a/olm-catalog/serverless-operator/hack/update-manifests.sh
+++ b/olm-catalog/serverless-operator/hack/update-manifests.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -Eeuo pipefail
+
+root="$(dirname "${BASH_SOURCE[0]}")/../../.."
+
+# Source the main vars file to get the operator version to be used.
+# shellcheck disable=SC1091,SC1090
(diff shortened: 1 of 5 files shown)


nak3:

I think infra issue.

olm-catalog/serverless-operator/hack/update-manifests.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
```
markusthoemmes, line 1: Are we calling this from anywhere? Should we add it to the `Makefile` so it's automatically taken into account?

matzew: Upstream they were already updated, so I am suprised we have old files here.

Perhaps update the script to pull em from upstream?

nak3: Sure, added it to the Makefile.

And yes this script will pull them from upstream :+1:

markusthoemmes (approved):
/lgtm
/approve

PR #909, opened 2021-04-22
Deleting the patch to remove HPA for broker's ingress/filter pods

matzew:

For [SRVKE-629](https://issues.redhat.com/browse/SRVKE-629) we did disable them...
Since then they are not active part of the product.

Wondering... if ... for 1.15 we try to get HPA back in?

/hold


maschmid:
For scaling up, we need to make sure the readiness probe in filter and ingress works OK . 
For scaling down, we need to make sure the in-fligt requests are processed before the filter/ingress terminates. Should there be a prestop hook on those?

matzew:
> Should there be a prestop hook on those?

that's perhaps something we can define in the upstream repo?  

here, we just "enable" or configure bits


Ideally we also have some good automated (load) tests for those?

PR #914, opened 2021-04-28
Add net-istio manifest

nak3:
This patch changes to:
- add net-istio manifest (TODO: remove webhook and Gateway)
- allow openshift-ingress to reconcile ingress by istio ingressclass

This still does not add e2e test because we have an issue OSSM-449.

/cc @markusthoemmes


nak3:
Close and I will send separated PR.

PR #920, opened 2021-04-29
[SRKVS-732] Allow ingress.class to be overridden

markusthoemmes:
1. This stops the webhook from defaulting a value into the KnativeServing's spec. This is necessary for us to be able to tell a system default apart from a user-set value.
2. Only defaults the `ingress.class` to kourier if it's not otherwise set.

/assign @nak3


markusthoemmes:
```
could not upsert output imagestreamtag: imagestreamtag.image.openshift.io "stable:serverless-bundle" already exists
```

/retest

PR #923, opened 2021-04-29
Add install-operator target to only install the serverless-operator

markusthoemmes:
This is quite useful if you want to test specific Knative Serving settings for example, without having the default applied first. This basically does the equivalent of installing the Serverless Operator through Operatorhub, but with the current codebase.

/assign @rhuss

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,9 @@ images:
 install:
 	./hack/install.sh
 
+install-operator:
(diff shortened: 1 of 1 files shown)


markusthoemmes:
/retest 
```
could not upsert output imagestreamtag: imagestreamtag.image.openshift.io "stable:serverless-bundle" already exists
```

matzew (approved):
/lgtm
/approve

:shipit:

PR #924, opened 2021-04-29
[SRVKS-732] Optionally install net-istio

markusthoemmes:
This allows customers to install `net-istio` for a better ServiceMesh integration experience. They need to explicitly enable net-istio by adding

```yaml
spec:
  ingress:
    istio:
      enabled: true
```

to their KnativeServing instance.

This is done purely as a means to unblock certain use-cases for now. Mid-term, we probably want to rework all of our Service Mesh related support to this scheme. Currently the intention is to shadow-launch this capability.


Makefile
```diff
@@ -9,6 +9,9 @@ images:
 install:
 	./hack/install.sh
 
+install-operator:
```
matzew, line 12: `install-operator-only` ?

markusthoemmes: None of the other targets (like `install-serving`) specify the "only" bit though they all mean "xxx only", so I'd personally keep that for consistency. This has been broken out to https://github.com/openshift-knative/serverless-operator/pull/923 anyway since it's not directly related here.

matzew: yeah. It's perhaps just for us anyways - but I was wondering if "install operator" might be misunderstood.

Just a nit comment. I am fine w/ keeping it, though, just was wondering

matzew: > they all mean "xxx only"

that actually is a good point, Markus

hack/generate/csv.sh
```diff
@@ -45,7 +45,7 @@ image "autoscaler-hpa" "${serving}-autoscaler-hpa"
 image "autoscaler-hpa__kube-rbac-proxy"  "$rbac_proxy"
 image "controller"     "${serving}-controller"
 image "controller__kube-rbac-proxy"  "$rbac_proxy"
-image "webhook"        "${serving}-webhook"
+image "webhook__webhook" "${serving}-webhook"
```
markusthoemmes, line 48: Had to fully qualify this because otherwise this matches the istio webhook as well and breaks it.

openshift-knative-operator/hack/update-manifests.sh
```diff
@@ -37,8 +41,32 @@ function download {
...
+    url="https://raw.githubusercontent.com/openshift-knative/${component}/${version}/config/${file}"
+    wget --no-check-certificate "$url" -O "$ingress_target_file"
+  done
+}
+
 download serving "$KNATIVE_SERVING_VERSION" "${serving_files[@]}"
 
+download_ingress net-istio "v$(metadata.get dependencies.net_istio)" "${istio_files[@]}"
+
 # Create an empty ingress directory.
```
nak3, line 70: nit: The empty ingress directory and `.gitkeep` are not necessary anymore.

nak3 (commented):
one nit otherwise LGTM!

olm-catalog/serverless-operator/project.yaml
```diff
@@ -22,9 +22,12 @@ requirements:
...
+  kourier: 0.21.0
+  net_istio: 0.21.0
+  maistra: 2.0.0
+
   eventing: 0.21.0
   eventing_kafka: 0.20.0
-  kourier: 0.21.0
+
   cli: 0.20.0
-  maistra: 2.0.0
```
rhuss, line 29: Why was maistra already a dependency before (re-)adding net_istio again ?

openshift-knative-operator/cmd/operator/kodata/ingress/0.21/1-200-clusterrole.yaml
```diff
@@ -0,0 +1,27 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  # These are the permissions needed by the Istio Ingress implementation.
+  name: knative-serving-istio
+  labels:
+    serving.knative.dev/release: devel
```
rhuss, line 21: Is this a temporary label used to identify these new objects or does is it evaluated in code, too ?

openshift-knative-operator/cmd/operator/kodata/ingress/0.21/2-500-mutating-webhook.yaml
```diff
@@ -0,0 +1,34 @@
...
+  - v1beta1
+  clientConfig:
+    service:
+      name: istio-webhook
+      namespace: knative-serving
+  failurePolicy: Fail
+  sideEffects: None
+  objectSelector:
+    matchExpressions:
+      - {key: "serving.knative.dev/configuration", operator: Exists}
```
rhuss, line 33: TIL: interesting YAML syntax, didn't know that you can collapse objects that way (I know only `{}` for an empty object)

openshift-knative-operator/cmd/operator/kodata/ingress/0.21/2-500-mutating-webhook.yaml
```diff
@@ -0,0 +1,34 @@
...
+apiVersion: admissionregistration.k8s.io/v1
+kind: MutatingWebhookConfiguration
+metadata:
+  name: webhook.istio.networking.internal.knative.dev
+  labels:
+    serving.knative.dev/release: devel
+    networking.knative.dev/ingress-provider: istio
+webhooks:
+- admissionReviewVersions:
+  - v1beta1
```
rhuss, line 24: Is this really valid yaml ?

openshift-knative-operator/cmd/operator/kodata/ingress/0.21/3-500-validating-webhook.yaml
```diff
@@ -0,0 +1,35 @@
...
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  name: config.webhook.istio.networking.internal.knative.dev
+  labels:
+    serving.knative.dev/release: devel
+    networking.knative.dev/ingress-provider: istio
+webhooks:
+- admissionReviewVersions:
+  - v1beta1
```
rhuss, line 24: valid yaml ? (i.e. that's not a key-value pair as part of the arrya's element definition)

openshift-knative-operator/cmd/operator/kodata/ingress/0.21/3-500-validating-webhook.yaml
```diff
@@ -0,0 +1,35 @@
...
+  clientConfig:
+    service:
+      name: istio-webhook
+      namespace: knative-serving
+  failurePolicy: Fail
+  sideEffects: None
+  name: config.webhook.istio.networking.internal.knative.dev
+  namespaceSelector:
+    matchExpressions:
```
rhuss, line 33: Here the expanded version is used for the expressions selector. Maybe for consistency, use the same format for the Validating and MutatingWebhook definitions ?

openshift-knative-operator/pkg/serving/extension.go
```diff
@@ -97,6 +97,15 @@ func (e *extension) Reconcile(ctx context.Context, comp v1alpha1.KComponent) err
 	// Use Kourier by default but allow a manual override.
 	common.ConfigureIfUnset(&ks.Spec.CommonSpec, "network", "ingress.class", "kourier.ingress.networking.knative.dev")
 
+	// Apply an Ingress config with Kourier enabled if nothing else is defined.
+	if ks.Spec.Ingress == nil {
```
rhuss, line 101: How was the behaviour before when no `.spec.ingress` was specified ? 

What happens if `.spec.ingress` is defined but neither `.spec.ingress.kourier` nor `.spec.ingress.istio` is given ?

rhuss (commented):
looks good, mostly questions so far.

olm-catalog/serverless-operator/project.yaml
markusthoemmes: Because we use the productized envoy that maistra uses for our Kourier deployment, so we don't have to be in the business of compiling (and patching) envoy.

openshift-knative-operator/cmd/operator/kodata/ingress/0.21/1-200-clusterrole.yaml
markusthoemmes: I should've noted that everything under `kodata` in this PR is automatically propagated from net-istio. These are literally straight copies from the net-istio repo. We might want to fix this label though, though it prolly doesn't do any harm to leave it like this either.

openshift-knative-operator/cmd/operator/kodata/ingress/0.21/2-500-mutating-webhook.yaml
markusthoemmes: Yep, `admissionReviewVersions` is an array of strings.

openshift-knative-operator/cmd/operator/kodata/ingress/0.21/3-500-validating-webhook.yaml
markusthoemmes: Not quite parsing the key-value comment. `admissionReviewVersions` is an array of strings.

openshift-knative-operator/cmd/operator/kodata/ingress/0.21/3-500-validating-webhook.yaml
markusthoemmes: As noted above, this is automatically fetched from net-istio, so if anything, we should go fix it upstream

openshift-knative-operator/pkg/serving/extension.go
markusthoemmes: This is a little intricate so I'm going to try to answer extensively:

The `Spec.Ingress` API was introduced upstream fairly recently. We **don't** use it to install Kourier today. Instead, we're still installing Kourier through our own component (the knative-openshift component in this repo).

So, the `Spec.Ingress` API was a noop before this PR. Whatever you changed there, nothing would change, except if you set everything to false (which is equivalent to not mentioning the child keys, as they all default to false, since they are not pointers).

It's defaulting is defined here https://github.com/knative/operator/blob/39d7e473b24ae08101d440b43145d2d135a71fae/pkg/reconciler/knativeserving/ingress/ingress.go#L43-L59 and if you squint a little, you might spot a "bug": It says: If `Spec.Ingress` is not defined (note that it's a pointer for that reason), we assume you want to install Istio. Now, we don't want that. At first, I only specified an empty `IngressConfigurations` object. That causes everything to be `Enabled: false`, which should be fine because we don't actually install Kourier here, right? Wrong!

If everything is `Enabled: false`, this piece of code will return `mf.Any()` with an empty list of subfilters. `mf.Any` only returns true, if any of the predicates return true. Since we don't have any predicates, we're effectively going to filter away **everything** and won't install Knative Serving at all.

So that's a bug I'm going to go fix upstream and meanwhile we default to "Kourier" (in quotes because this doesn't actually install Kourier). That doesn't do any harm to the logic as it's going to filter the manifest for all resources that are **not** Kourier, which is exactly what we want in this case.

openshift-knative-operator/hack/update-manifests.sh
markusthoemmes: Ack!

openshift-knative-operator/cmd/operator/kodata/ingress/0.21/3-500-validating-webhook.yaml
rhuss: ah, sorry, thought that the `clientConfig:` is on thee same level as `v1beta1`

damned yaml ....

openshift-knative-operator/cmd/operator/kodata/ingress/0.21/1-200-clusterrole.yaml
rhuss: Just out of curiosity: This label is updated during a release, right ? (so it's e.g. `"0.21"` for the 0.21 release) and then reflected here as well when we are doing our release that depends on a released version of Knative.

If so, it makes totally sense (but I would probably call it `nightly` or so, or even better like in Maven land `0.23-SNAPSHOT` ....

But I agree that this should be fixed upstream, not here.

markusthoemmes: Yeah, usually it's updated during a release. In fact, the other manifests pull from released versions of Knative and we could do that here too. We'd only have to implement a YAML filter then because we don't want the entire manifest of net-istio but leave out the Gateways for example.

openshift-knative-operator/hack/update-manifests.sh
```diff
@@ -37,8 +41,32 @@ function download {
   done
 }
 
+function download_ingress {
```
matzew, line 44: Why is the `net-istio` download inside the `openshift-knatve-operator`, while `kourier` is in the `knative-operator`?

hack/generate/csv.sh
matzew: good container names, FTW :-)

openshift-knative-operator/hack/update-manifests.sh
markusthoemmes: Because we didn't yet do the work to migrate the Kourier installation over. The Ingress API of the operator is quite new and us using Kourier predates it's existence. Moving Kourier over is also slightly non-trivial as we install it in `knative-serving-ingress`, while upstream assumes it's installed into `knative-serving`. We have a plan to move that over, but it's not quite necessary for the purposes of this work.

An additional mention: This means that net-kourier will **always** be installed (even if Istio is enabled), and I think that's okay for now, since we're shadow-launching this. We'll just tell people that use Istio that Kourier is still installed but won't work as expected anymore.

matzew: thanks

openshift-knative-operator/pkg/serving/extension.go
rhuss: I see and iiuc having both Kourier enabled here and in our own component doesn't collide as this config here is complementary and has no issues with our own way to install kourier.

Still my question would be what happens if the user provides an erroneous config like `.spec.ingress.courier.enabled=true`. I would expect this to be either ignore (so the same behaviour as if `.spec.ingress` was not given at all or an error raised.

However, I suspect from the code change above, providing this kind of config would prevent your workaround because `.spec.ingress` is != nil (but should be treated as it was nil).

Maybe a test for this kind of user error could be added, too, to validate any expected behaviour ?

openshift-knative-operator/pkg/serving/extension.go
```diff
@@ -97,6 +97,15 @@ func (e *extension) Reconcile(ctx context.Context, comp v1alpha1.KComponent) err
...
+		ks.Spec.Ingress = &v1alpha1.IngressConfigs{
+			Kourier: v1alpha1.KourierIngressConfiguration{
+				Enabled: true,
```
matzew, line 104: clean override to our default :+1:

openshift-knative-operator/pkg/serving/extension.go
rhuss, line 101:
```suggestion
	if ks.Spec.Ingress == nil || (ks.Spec.Ingress.Kourier == nil && ks.Spec.Ingress.Istio == nil) {
```

oder so ... ?

openshift-knative-operator/hack/update-manifests.sh
```diff
@@ -37,14 +41,31 @@ function download {
...
+  ingress_dir="$root/openshift-knative-operator/cmd/operator/kodata/ingress/$(versions.major_minor "${KNATIVE_SERVING_VERSION}")"
+  rm -r "$ingress_dir"
+  mkdir -p "$ingress_dir"
+
+  for (( i=0; i<${#files[@]}; i++ ));
+  do
+    index=$(( i+1 ))
+    file="${files[$i]}.yaml"
+    ingress_target_file="$ingress_dir/$index-$file"
+    url="https://raw.githubusercontent.com/openshift-knative/${component}/${version}/config/${file}"
```
matzew, line 61: should we download from upstream?

the files on the TAGs there do have the proper `release` labels, e.g. see here: https://github.com/knative-sandbox/net-istio/releases/download/v0.21.0/net-istio.yaml

markusthoemmes: See https://github.com/openshift-knative/serverless-operator/pull/924#discussion_r623650459. We **can**, but then have to deal with filtering this YAML to only the set we actually want.

openshift-knative-operator/pkg/serving/extension.go
markusthoemmes: Currently, we'd indeed ignore it. I'll extend the behavior a little to catch the bug mentioned above for good measure.

markusthoemmes: I don't quite want to surface an error via a webhook here because in theory, "all disabled" is a valid config I don't necessarily want to forbid (just maybe not officially support). So I'll extend the defaulting to Kourier to also be done (for now) if all ingresses are disabled.

rhuss (approved):
thanks !

/lgtm
/approved

markusthoemmes:
/hold

To see everything pass.

markusthoemmes:
/unhold

YAY

PR #925, opened 2021-04-30
[SRVKS-733] Add an annotation to make routes use passthrough

markusthoemmes:
As per title, this adds an annotation to be added to a Knative Service (or a Route) to make us generate a route that has passthrough encryption enabled on it.

/assign @nak3

### DIFF
--- a/serving/ingress/pkg/reconciler/ingress/resources/route.go
+++ b/serving/ingress/pkg/reconciler/ingress/resources/route.go
@@ -17,9 +17,12 @@ import (
 )
 
 const (
-	TimeoutAnnotation      = "haproxy.router.openshift.io/timeout"
-	DisableRouteAnnotation = "serving.knative.openshift.io/disableRoute"
-	KourierHTTPPort        = "http2"
+	TimeoutAnnotation                = "haproxy.router.openshift.io/timeout"
+	DisableRouteAnnotation           = "serving.knative.openshift.io/disableRoute"
+	EnablePassthroughRouteAnnotation = "serving.knative.openshift.io/enablePassthrough"
+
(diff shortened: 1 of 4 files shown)


serving/ingress/pkg/reconciler/ingress/resources/route.go
```diff
@@ -132,6 +135,14 @@ func makeRoute(ci *networkingv1alpha1.Ingress, host string, rule networkingv1alp
 			WildcardPolicy: routev1.WildcardPolicyNone,
 		},
 	}
+
+	// If the passthrough annotation is set, target the HTTPS port and configure passthrough.
+	if _, ok := annotations[EnablePassthroughRouteAnnotation]; ok {
+		route.Spec.Port.TargetPort = intstr.FromString(HTTPSPort)
+		route.Spec.TLS.Termination = routev1.TLSTerminationPassthrough
+		route.Spec.TLS.InsecureEdgeTerminationPolicy = routev1.InsecureEdgeTerminationPolicyNone
```
nak3, line 143: Just in case, `InsecureEdgeTerminationPolicyNone` is better than `InsecureEdgeTerminationPolicyAllow` here?
`oc get ksvc` prints `http://` but actually it is not available.

markusthoemmes: hmm, good point. Maybe Redirect is indeed the safer choice?

nak3: Yes, I think so.

nak3:
/hold
/lgtm

Please feel free to unhold :smile:

PR #931, opened 2021-05-06
Bump to 1.14.1

warrenvw:
See title.

### DIFF
--- a/openshift/ci-operator/build-image/Dockerfile
+++ b/openshift/ci-operator/build-image/Dockerfile
@@ -1,6 +1,6 @@
 # Dockerfile to bootstrap build and test in openshift-ci
 
-FROM docker.io/openshift/origin-release:golang-1.14
+FROM docker.io/openshift/origin-release:golang-1.15
 
 # Add kubernetes repository
 ADD openshift/ci-operator/build-image/kubernetes.repo /etc/yum.repos.d/
--- a/openshift/ci-operator/build-image/Dockerfile
+++ b/openshift/ci-operator/build-image/Dockerfile
@@ -1,6 +1,6 @@
 # Dockerfile to bootstrap build and test in openshift-ci
 
-FROM docker.io/openshift/origin-release:golang-1.15
+FROM registry.ci.openshift.org/openshift/release:golang-1.15
 
 # Add kubernetes repository
 ADD openshift/ci-operator/build-image/kubernetes.repo /etc/yum.repos.d/
--- a/Makefile
+++ b/Makefile
@@ -102,6 +102,15 @@ release-files:
 	./hack/generate/dockerfile.sh \
 		templates/build-image.Dockerfile \
 		openshift/ci-operator/build-image/Dockerfile
+	./hack/generate/dockerfile.sh \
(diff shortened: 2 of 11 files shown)


openshift/ci-operator/build-image/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # Dockerfile to bootstrap build and test in openshift-ci
 
-FROM docker.io/openshift/origin-release:golang-1.14
+FROM docker.io/openshift/origin-release:golang-1.15
```
mgencur, line 3: I think we need to changes these lines too:
https://github.com/openshift-knative/serverless-operator/blob/release-1.14/knative-operator/Dockerfile#L1
https://github.com/openshift-knative/serverless-operator/blob/release-1.14/openshift-knative-operator/Dockerfile#L1
Otherwise this looks good to me.

mgencur: These files are not part of the templating mechanism. Maybe good to add it there.

mgencur: Actually, it's probably the right time to change the base image to the ones from OpenShift CI: registry.ci.openshift.org/openshift
Similar to https://github.com/openshift-knative/serverless-operator/commit/1176cb1fe5d6442f431dea93ed5da42a40f99deb and https://github.com/openshift-knative/serverless-operator/commit/aa7513338efe92c81d2060b7aa52d718cb6d391f

maschmid:
I think you may need to add the 1.14.0 bundle here https://github.com/openshift-knative/serverless-operator/blob/release-1.14/hack/lib/catalogsource.bash#L122

rhuss (commented):
For the version numver update it looks ok to me, but I'm not the clear about the changes in the registries (from registry.redhat.com to quay.io and vice versa). 

What is the reasoning for that switch for a patch level release ?

maschmid:
> For the version numver update it looks ok to me, but I'm not the clear about the changes in the registries (from registry.redhat.com to quay.io and vice versa).
> 
> What is the reasoning for that switch for a patch level release ?

We don't have CI image promotion for micro releases, so when we update release-1.14 branch for 1.14.1 , we don't have any bundle in the CI registry with 1.14.0 , so we have to use one from quay (or elsewhere)

maschmid:
Hm, this approach actually breaks the upgrade tests (or, the install of 1.14.0), The serving storage-version-migration is created with the product image `registry.redhat.io/openshift-serverless-1/serving-storage-version-migration-rhel8@sha256:64c679be96c33703c801e55564eac7e9d782a4d1ba92cd434a819b743cdc4041` , but then the operator tries to replace it with `registry.ci.openshift.org/openshift/knative-v0.20.0:knative-serving-storage-version-migration` , which doesn't work, as the job cannot be modified once created...

maschmid:
We could probably fix this by using a bundle for 1.14.0 that uses the CI images...

maschmid:
FWIW, I have replaced the quay.io/openshift-knative/serverless-bundle:1.14.0 with the CI bundle for 1.14.0, so this works now... (we should use CI bundles in quay anyway)

PR #940, opened 2021-05-11
Update Strimzi API version to v1beta2

scholzj:
Strimzi is moving soon to use the API versions `v1beta2` which is compatible with the `apiextensions/v1` CRDs and will work with Kubernetes 1.22. This API version works with latest versions of Strimzi as well as AMQ Streams. This PR updates the uses of Strimzi in this repository to use the new API version as well.

### DIFF
--- a/hack/lib/strimzi.bash
+++ b/hack/lib/strimzi.bash
@@ -16,7 +16,7 @@ function install_strimzi_operator {
 function install_strimzi_cluster {
   header "Applying Strimzi Cluster file"
   cat <<-EOF | oc apply -f -
-    apiVersion: kafka.strimzi.io/v1beta1
+    apiVersion: kafka.strimzi.io/v1beta2
     kind: Kafka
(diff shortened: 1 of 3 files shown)


matzew:
@scholzj thanks for the PR.

There is still some hackery needed, since some of the vendored bits have the api shape in it:
https://github.com/openshift-knative/serverless-operator/pull/940/checks?check_run_id=2554952276

scholzj:
Ok, I realised that changing the Golang is not so easy so I reverted the changes. The changes to the shell file are important to keep this working with the next Strimzi release. The topics referenced in the Golang files should keep working for the time being

matzew:
/hold

Let's please wait here a bit until we have done our cut for 1.15 

/assign @matzew

matzew:
I've rebased this, and adjusted 

/lgtm
/approve

/hold 
to wait for all checks ......

PR #949, opened 2021-05-11
Add upgrade build tag to golangci-lint config

markusthoemmes:
Just so we also check (and compile) the upgrade tests.


matzew (approved):
/lgtm
/approve

PR #950, opened 2021-05-11
Remove arm64 link from consoleclidownload.go

warrenvw:
The arm64 client is not yet tested, and QE does not want it included in the 1.14.1 release.


matzew (approved):
/lgtm
/approve

PR #955, opened 2021-05-12
WIP: CI was here...:fire:

matzew:
/hold

this is only a test!


markusthoemmes:
Issues should be solved via https://github.com/openshift-knative/serverless-operator/pull/956

PR #961, opened 2021-05-16
Add `make install-full-mesh` to deploy mesh to test secure traffic

nak3:
This patch adds new script `make install-full-mesh`.
It deploys ServiceMesh with secure Gateway and mTLS strict mode.

Please also refer to this doc: https://github.com/nak3/serverless-operator/blob/add-net-istio-script/docs/full-mesh.md


hack/lib/mesh.bash
```diff
@@ -60,3 +34,70 @@ function undeploy_servicemesh_operators {
...
+function deploy_gateways {
+  oc apply -f "${resources_dir}"/smmr.yaml || return $?
+
+  # Generate wildcard certs with cluster's subdomain.
+
+  local out_dir
+  out_dir="$(mktemp -d /tmp/certs-XXX)"
+
+  oc extract -n openshift-apiserver configmap/config --to="${out_dir}"
+  subdomain=$(yq read "${out_dir}"/config.yaml "routingConfig.subdomain")
```
markusthoemmes, line 65: Should this use the `spec.domain` field from `oc get config cluster` instead? You should be able to get away with a `--ojsonpath=".spec.domain"` or thelike I think. That's what our operator is doing as well AFAIK.

markusthoemmes (commented):
LGTM modulo that one comment.

hack/lib/mesh.bash
nak3: Sure! updated.

markusthoemmes (approved):

PR #963, opened 2021-05-18
Fix rbac proxy images for Eventing

skonto:
- adds rbac proxy image productization support for Eventing.


matzew:
@skonto .  I think we need to update the `hack/generate/csv` file ? 

that file is gneerated

markusthoemmes:
It'll actuallty have to be done in `csv.sh` I think (similar to serving)

matzew:
yeah, look here: https://github.com/openshift-knative/serverless-operator/commit/882cdb2444f6adc35d0675bb688c17f1f3b81915

skonto:
Yeah I remember its generated from the cvs.sh :(

matzew:
@skonto do we have a JIRA to track similar work for Knative Kafka components?

skonto:
@matzew SRVCOM-1270

matzew:
> SRVCOM-1270
that jira does not mention anything for Knative-Kafka  - just Eventing (which is what we here update). Perhaps worth to track in there? 

The Kafka images are not yet affected

skonto:
> that jira does not mention anything for Knative-Kafka - just Eventing (which is what we here update). Perhaps worth to track in there?
I will add it if it makes sense to secure these endpoints the same way. Knative Kafka is part of Eventing anyway, but to move forward I will add a ticket under that jira.

markusthoemmes (approved):
/lgtm
/approve

markusthoemmes:
/hold

I think these references are mostly wrong.

hack/generate/csv.sh
```diff
@@ -60,14 +60,22 @@ image "networking-istio" "${registry}/knative-v$(metadata.get dependencies.net_i
...
 image "mt-broker-controller__mt-broker-controller"  "${eventing}-mtchannel-broker"
+image "mt-broker-controller__mt-broker-controller__kube-rbac-proxy"  "$rbac_proxy"
 image "mt-broker-filter__filter"                    "${eventing}-mtbroker-filter"
+image "mt-broker-filter__filter__kube-rbac-proxy"                    "$rbac_proxy"
 image "mt-broker-ingress__ingress"                  "${eventing}-mtbroker-ingress"
+image "mt-broker-ingress__ingress__kube-rbac-proxy"                  "$rbac_proxy"
 image "imc-controller__controller"                  "${eventing}-channel-controller"
+image "imc-controller__controller__kube-rbac-proxy"                 "$rbac_proxy"
 image "imc-dispatcher__dispatcher"                  "${eventing}-channel-dispatcher"
+image "imc-dispatcher__dispatcher__kube-rbac-proxy"                  "$rbac_proxy"
```
markusthoemmes, line 78: `__XXX` already indicates a container name, so I think all of these have to change to 

 ```suggestion
image "imc-dispatche__kube-rbac-proxy"                  "$rbac_proxy"
```

(i.e. remove the first container name)

matzew: can we do that "legacy" clean up in a separate PR ? 

I am happy to clean up all those "dups" - ok ?

matzew: ah, nevermind ...   Ignore my comment ... that's the rbac_proxy ...

matzew: but... they are lke: https://github.com/openshift-knative/serverless-operator/pull/964/files#diff-da13200d4ae862c053922b38b3bf7dfdc1bd6f52518c2f160156fe41c119bf39R62-R79

Hrm....

markusthoemmes: Yes, these are wrong too, that's how we noticed :joy:

matzew: :rofl:  ok...

markusthoemmes:
Since we have to touch these again anyway, what if we dropped all of the `__kube-rbac-proxy` env vars in favor of 1 `KUBE_RBAC_PROXY_IMAGE` var that contains the image (it's the same everywhere anyway). We can then change https://github.com/openshift-knative/serverless-operator/blob/e4fa5d11add625462875db4221f2bee17c761839/openshift-knative-operator/pkg/monitoring/rbac_proxy_injection.go#L75-L81 to get only that single image regardless of the deployment name and won't have to deal with this in the future.

Would that work @skonto ?

skonto:
@markusthoemmes yes this was a pending refactoring from the initial PR that added rbac proxy. I am wondering why tests didnt capture this failure.

markusthoemmes:
Because of the hardcoded fallback. That made wrong overrides be silently ignored.

skonto:
Cleaned up things and removed the default image.

markusthoemmes (approved):
/lgtm
/approve

Very nice, thanks!

matzew:
thanks, @skonto

PR #970, opened 2021-05-19
Apply kourier defaulting a bit more diligently

markusthoemmes:
If the user forgets to set `enabled: true` to the Kourier ingress configuration while also configuring other fields like `service-type`, we'll currently swallow the extra config, which is a weird experience. This fixes that.

/assign @nak3

### DIFF
--- a/openshift-knative-operator/pkg/serving/extension_test.go
+++ b/openshift-knative-operator/pkg/serving/extension_test.go
@@ -189,6 +189,26 @@ func TestReconcile(t *testing.T) {
 			}
 		}),
 	}, {
+		name: "respect kourier settings",
+		in: &v1alpha1.KnativeServing{
+			Spec: v1alpha1.KnativeServingSpec{
+				Ingress: &v1alpha1.IngressConfigs{
+					Kourier: v1alpha1.KourierIngressConfiguration{
+						// Enabled: true omitted explicitly.
+						ServiceType: corev1.ServiceTypeClusterIP,
+					},
+				},
+			},
+		},
+		expected: ks(func(ks *v1alpha1.KnativeServing) {
+			ks.Spec.Ingress = &v1alpha1.IngressConfigs{
+				Kourier: v1alpha1.KourierIngressConfiguration{
+					Enabled:     true,
+					ServiceType: corev1.ServiceTypeClusterIP,
+				},
+			}
+		}),
+	}, {
 		name: "respects different status",
 		in: ks(func(ks *v1alpha1.KnativeServing) {
 			ks.Status.MarkDependenciesInstalled()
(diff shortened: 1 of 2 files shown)


markusthoemmes:
ehrm, okay?

/retest

openshift-knative-operator/pkg/serving/extension.go
```diff
@@ -165,11 +165,16 @@ func (e *extension) fetchLoggingHost(ctx context.Context) string {
...
+	if ks.Spec.Ingress == nil {
 		ks.Spec.Ingress = &v1alpha1.IngressConfigs{
 			Kourier: v1alpha1.KourierIngressConfiguration{
 				Enabled: true,
 			},
 		}
+		return
+	}
+
+	if !ks.Spec.Ingress.Istio.Enabled && !ks.Spec.Ingress.Kourier.Enabled && !ks.Spec.Ingress.Contour.Enabled {
```
nak3, line 177:
```suggestion
	if !ks.Spec.Ingress.Istio.Enabled && !ks.Spec.Ingress.Kourier.Enabled {
```

I wonder that we should enable Kourier regardless of the Contour setting. But current behavior is also fine as users can realize it by the deploy error. What do you think?

markusthoemmes: You mean we should bail out if a user tries to enable contour? We could do that yeah, though I'd like to do that separately as we'd ideally also add webhook validation to that then.

nak3: Sure, webhook validation is ideal. Let's leave it as it is.

nak3:
/lgtm
/hold

Please feel free to unhold.

openshift-knative-operator/pkg/serving/extension_test.go
```diff
@@ -187,6 +187,26 @@ func TestReconcile(t *testing.T) {
...
 			}
 		}),
+	}, {
+		name: "respect kourier settings",
+		in: &v1alpha1.KnativeServing{
+			Spec: v1alpha1.KnativeServingSpec{
+				Ingress: &v1alpha1.IngressConfigs{
+					Kourier: v1alpha1.KourierIngressConfiguration{
+						// Enabled: true omitted explicitly.
+						ServiceType: corev1.ServiceTypeClusterIP,
```
skonto, line 197: Do we cover the case where all ingresses are disabled?

markusthoemmes: This is that case, technically. The `Enabled` field is false on all Ingresses.

skonto: Yeah it is implicit due to default false value. How about the nil case?

markusthoemmes: Tested as the default case in pretty much every other test.

PR #971, opened 2021-05-19
Determine default ingress class based on enabled ingress

markusthoemmes:
Since we now support using the Ingress API (`spec.ingress`) we can actually use that to determine the ingress class so the user doesn't have to configure it "twice" (once via the ingress API and then explicitly setting the ingress class to the same value).

Essentially, a small quality-of-life improvement.

/assign @nak3


nak3:
/lgtm
/hold

Please feel free to unhold.

PR #978, opened 2021-05-19
[RELEASE-1.15] Fix kube-rbac references in the CSV

markusthoemmes:
All of the eventing references contained two `__XXX`, where `__XXX` stands for the container name inside a deployment. So we need to drop the first of these references for each for it to be correct.


markusthoemmes:
If all legs pass at least once (i.e. operator, upgrade and upstream), feel free to force merge @warrenvw

PR #979, opened 2021-05-20
Update yq to v4

### LINKED ISSUE #939: Update yq version
This project expects yq v3 like https://github.com/openshift-knative/serverless-operator/blob/main/.github/workflows/validate.yaml#L28 but it is too old version and I need to switch v3 and v4 in my local env :sweat_smile: 

Is there any blocker to update yq version to v4?

markusthoemmes:
This was much more painful than I anticipated at first. Serves as a showcase mostly since this is still under discussion.

Fixes #939 

/hold


matzew:
Would we also need to change these:
* https://github.com/openshift/knative-serving/blob/main/openshift/ci-operator/build-image/Dockerfile#L10
* https://github.com/openshift/knative-eventing/blob/main/openshift/ci-operator/build-image/Dockerfile#L10
* https://github.com/openshift-knative/eventing-kafka/blob/main/openshift/ci-operator/build-image/Dockerfile#L10

skonto:
They changed the model a lot I remembered I got stuck when downloaded the latest version locally and then found out I need y3.

markusthoemmes:
FWIW, the failures are expeceted as we need to fix the build image first, merge that and then have the new yq available. We likely have to create some dual-installation thing where the script can choose. Sigh.

markusthoemmes:
I think imma close this for now. Seems much too invasive.

PR #980, opened 2021-05-20
Fix a few misspellings

markusthoemmes:
As per title.

Gonna include the respective linter in a subsequent PR as well.


hack/lib/metadata.bash
```diff
@@ -13,8 +13,8 @@ yq > /dev/null || exit 127
 #   Writes metadata value on STDOUT
 #######################################
 function metadata.get {
-  local matadata_file
-  matadata_file="$(dirname "${BASH_SOURCE[0]}")/../../olm-catalog/serverless-operator/project.yaml"
```
matzew, line 17: :joy_cat:

PR #990, opened 2021-05-31
Go with 2.7.0 of Apache Kafka

matzew:

The 0.22.x line that we currently use has support for 2.7.0 as well


skonto:
@matzew should not we use 2.8?

matzew:
2.8.0 is not yet supported in _that_ version of Strimzi

matzew:
For 2.8.0, we should wait until we have https://github.com/knative-sandbox/eventing-kafka/pull/627  in place (and the vendored golang updates)

PR #993, opened 2021-05-31
Bump Golang to 1.15 everywhere

markusthoemmes:
For consistency. We should be using 1.15 in productization already anyway.


matzew (approved):
thanks for the updates!

/lgtm
/approve

PR #994, opened 2021-05-31
[SRVKS-741] Drop namespaces from manifests to prevent reconcilation

markusthoemmes:
Having the namespaces in the manifests doesn't do anything useful as
we're requiring the user to create the namespace beforehand anyway. It
does however actively cause us to drop labels and annotations that users
have added via `kubectl apply` commands (versus `kubectl annotate` or
`kubectl label`), so it's actually harmful.

### DIFF
--- a/openshift-knative-operator/cmd/operator/kodata/knative-eventing/0.22.0/2-eventing-core.yaml
+++ b/openshift-knative-operator/cmd/operator/kodata/knative-eventing/0.22.0/2-eventing-core.yaml
@@ -11,27 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-apiVersion: v1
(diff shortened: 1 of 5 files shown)


matzew:
LGTM.

One question. For https://github.com/openshift-knative/serverless-operator/blob/main/openshift-knative-operator/cmd/operator/kodata/ingress/0.22/kourier.yaml#L15-L21  it's not relevant, b/c that's a "system" namespace, and we not ask the user to create it, right? so here it's good to have it?


Unlike for serving or eventing, where we want users to do:
```yaml
apiVersion: v1
kind: Namespace
metadata:
 name: knative-eventing
---
kind: KnativeEventing
metadata:
  name: knative-eventing
  namespace: knative-eventing
spec: {}
```

Correct?

markusthoemmes:
Yes, the Kourier namespace is even necessary and critical to keep.

matzew (approved):
/lgtm
/approve

PR #995, opened 2021-05-31
[SRVKS-744] Only configure domain template if not otherwise set

markusthoemmes:
Users may have a legitimate reason to override this, so there is no point in us forcing a specific template on them.


matzew (approved):
/lgtm
/approve

PR #998, opened 2021-06-02
[SRVCOM-1331] Secure metric endpoints for Kafka control plane pods

skonto:
- Adds rbac proxy to the right deployments
- Adds tests for the metrics exposed

Note: Kafka sources will be secured in another PR along with the rest of the sources.

### DIFF
--- a/knative-operator/pkg/controller/knativekafka/rbacproxytransforms.go
+++ b/knative-operator/pkg/controller/knativekafka/rbacproxytransforms.go
@@ -1,0 +1,50 @@
+package knativekafka
+
+import (
+	"context"
+	"errors"
+
+	mf "github.com/manifestival/manifestival"
+	operatorv1alpha1 "github.com/openshift-knative/serverless-operator/knative-operator/pkg/apis/operator/v1alpha1"
+	"github.com/openshift-knative/serverless-operator/openshift-knative-operator/pkg/monitoring"
+	"k8s.io/apimachinery/pkg/util/sets"
+	eventingv1alpha1 "knative.dev/operator/pkg/apis/operator/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	kafkaChannelComponents = []string{"kafka-ch-controller", "kafka-ch-dispatcher", "kafka-webhook"}
+	kafkaSourceComponents  = []string{"kafka-controller-manager"}
+)
+
+func addRBACProxySupportToManifest(instance *operatorv1alpha1.KnativeKafka, components []string) (*mf.Manifest, error) {
+	proxyManifest := mf.Manifest{}
+	// Only create the roles needed for the deployment service accounts as Prometheus has already
+	// the rights needed due to eventing that is assumed to be installed.
+	for _, c := range components {
+		crbM, err := monitoring.CreateClusterRoleBindingManifest(c, instance.GetNamespace())
+		if err != nil {
+			return nil, err
+		}
+		proxyManifest = proxyManifest.Append(*crbM)
+		if err = monitoring.AppendManifestsForComponent(c, instance.GetNamespace(), &proxyManifest); err != nil {
+			return nil, err
+		}
+	}
+	return &proxyManifest, nil
+}
+
+func getRBACProxyInjectTransformer(apiClient client.Client) (mf.Transformer, error) {
+	eventingList := &eventingv1alpha1.KnativeEventingList{}
+	err := apiClient.List(context.Background(), eventingList)
+	if err != nil {
+		return nil, err
+	}
+	if len(eventingList.Items) == 0 {
+		return nil, errors.New("eventing instance not found")
+	}
+	if monitoring.ShouldEnableMonitoring(eventingList.Items[0].GetSpec().GetConfig()) {
+		return monitoring.InjectRbacProxyContainerToDeployments(sets.NewString(append(kafkaChannelComponents, kafkaSourceComponents...)...)), nil
+	}
+	return nil, nil
+}
--- a/openshift-knative-operator/pkg/monitoring/monitoring_eventing.go
+++ b/openshift-knative-operator/pkg/monitoring/monitoring_eventing.go
@@ -18,13 +18,12 @@ func ReconcileMonitoringForEventing(ctx context.Context, api kubernetes.Interfac
 }
 
 func GetEventingTransformers(comp v1alpha1.KComponent) []mf.Transformer {
-	if shouldEnableMonitoring(comp.GetSpec().GetConfig()) {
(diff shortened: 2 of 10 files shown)


matzew:
:eyes:

knative-operator/pkg/controller/knativekafka/rbacproxytransforms.go
```diff
@@ -0,0 +1,50 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	kafkaChannelComponents = []string{"kafka-ch-controller", "kafka-ch-dispatcher", "kafka-webhook"}
+	kafkaSourceComponents  = []string{"kafka-controller-manager"}
+)
+
+func addRBACProxySupportToManifest(instance *operatorv1alpha1.KnativeKafka, components []string) (*mf.Manifest, error) {
+	proxyManifest := mf.Manifest{}
```
matzew, line 21: for eventing, I see https://github.com/openshift-knative/serverless-operator/blob/6f9d2ca6d6a8a61b251315d2d326b330094dcdca/openshift-knative-operator/pkg/monitoring/helpers.go#L260  is called ?

knative-operator/pkg/controller/knativekafka/rbacproxytransforms.go
```diff
@@ -0,0 +1,50 @@
...
+	// Only create the roles needed for the deployment service accounts as Prometheus has already
+	// the rights needed due to eventing that is assumed to be installed.
+	for _, c := range components {
```
matzew, line 24: could we use `kafkaDeployments`?  (looking here: https://github.com/openshift-knative/serverless-operator/blob/6f9d2ca6d6a8a61b251315d2d326b330094dcdca/openshift-knative-operator/pkg/monitoring/monitoring_eventing.go#L36)

knative-operator/pkg/controller/knativekafka/rbacproxytransforms.go
```diff
@@ -0,0 +1,50 @@
...
+	eventingv1alpha1 "knative.dev/operator/pkg/apis/operator/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	kafkaChannelComponents = []string{"kafka-ch-controller", "kafka-ch-dispatcher", "kafka-webhook"}
+	kafkaSourceComponents  = []string{"kafka-controller-manager"}
+)
+
+func addRBACProxySupportToManifest(instance *operatorv1alpha1.KnativeKafka, components []string) (*mf.Manifest, error) {
```
matzew, line 20: not sure I get why this is a bit different than the same for eventing: https://github.com/openshift-knative/serverless-operator/blob/6f9d2ca6d6a8a61b251315d2d326b330094dcdca/openshift-knative-operator/pkg/monitoring/monitoring_eventing.go#L30

knative-operator/pkg/controller/knativekafka/rbacproxytransforms.go
```diff
@@ -0,0 +1,50 @@
...
+		proxyManifest = proxyManifest.Append(*crbM)
+		if err = monitoring.AppendManifestsForComponent(c, instance.GetNamespace(), &proxyManifest); err != nil {
+			return nil, err
+		}
+	}
+	return &proxyManifest, nil
+}
+
+func getRBACProxyInjectTransformer(apiClient client.Client) (mf.Transformer, error) {
+	eventingList := &eventingv1alpha1.KnativeEventingList{}
```
matzew, line 38: `KnativeEventingList` ?

openshift-knative-operator/pkg/monitoring/monitoring_eventing.go
```diff
@@ -18,13 +18,12 @@ func ReconcileMonitoringForEventing(ctx context.Context, api kubernetes.Interfac
 }
```
matzew, line 18: Is it possible to move some `monitoring_kafka.go` in here?

matzew:
Do we need something like: https://github.com/openshift-knative/serverless-operator/commit/226efb1967d89fa355596351b23f51be2b98b46f  ?

knative-operator/pkg/controller/knativekafka/rbacproxytransforms.go
skonto: This is different part. The Kafka setup is done in the knative-operator. This needs to remain as is because it deals with the Eventing components like broker-ingress, broker-filter etc.

knative-operator/pkg/controller/knativekafka/rbacproxytransforms.go
skonto: Because for example there is no need to call `proxyManifest := mf.Manifest{}` and there is a specific rule about `"mt-broker-controller"` sa not all components have the sa name be equal to component name. I could separate the common code but does not add much to do so right now.

knative-operator/pkg/controller/knativekafka/rbacproxytransforms.go
skonto: Yes for example `GetEventingMonitoringPlatformManifests` calls that. We dont need to do so for Kafka because the Prometheus account in `openshift-monitoring` ns already gets the proper rbac roles due to the GetEventingMonitoringPlatformManifests call.

knative-operator/pkg/controller/knativekafka/rbacproxytransforms.go
skonto: when requesting a namespaced resource the name may vary, so I didnt want to assume that the name is `knative-eventing`  because the user may change that no?

openshift-knative-operator/pkg/monitoring/monitoring_eventing.go
skonto: Which part you want to be moved here?

skonto:
> Do we need something like: 226efb1 ?
I am re-using the code that sets the rbac proxy image so it should be the same for all components including the Kafka ones.

matzew (approved):
/lgtm
/approve

skonto:
> INFO[2021-06-07T06:39:50Z]     knative_kafka_test.go:121: Deployment kafka-ch-dispatcher is not gone: deployment kafka-ch-dispatcher in namespace knative-eventing not gone in time: timed out waiting for the condition 
INFO[2021-06-07T06:39:50Z]     --- FAIL: TestKnativeKafka/remove_knativekafka_cr (310.14s) 
INFO[2021-06-07T06:39:50Z] FAIL test/e2ekafka.TestKnativeKafka/remove_knativekafka_cr (310.14s) 

This seems irrelevant.

PR #1007, opened 2021-06-09
Dispacther is at replicas: 0, initially

matzew:

The dispatcher is at `0` replicas, hence removing from the smoke test.

We will need to add a more sophisticated smoke test here.  Will create a JIRA for that, once _this_ is in.


matzew:
I am looking....

matzew:
/close
in fav of: #1010

PR #1010, opened 2021-06-09
Skip 0 replica optimization

matzew:
see if we are still better w/ default to `replicas: 1` for the kafka channel deployment

### DIFF
--- a/knative-operator/deploy/resources/knativekafka/1-channel-consolidated.yaml
+++ b/knative-operator/deploy/resources/knativekafka/1-channel-consolidated.yaml
@@ -726,9 +726,7 @@ metadata:
   labels:
     kafka.eventing.knative.dev/release: "v0.22.3"
 spec:
-  # this deployment is going to be scaled up by the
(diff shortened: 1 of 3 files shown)


knative-operator/deploy/resources/knativekafka/1-channel-consolidated.yaml
```diff
@@ -728,7 +728,7 @@ metadata:
 spec:
   # this deployment is going to be scaled up by the
   # controller when the very first KafkaChannel is created
```
skonto, line 730: Should we remove this comment since default is 1?

matzew: yeah, we can patch remove the comments too

matzew:
@skonto I've changed the patch, to also remove the comments!  Thanks for the feedback!

markusthoemmes:
/retest

Infra.

aliok (approved):
/lgtm
/approve

Let's see how it goes.

PR #1011, opened 2021-06-09
Including eventing-kafka upgrade tests for source and channel

cardil:
Port of #983 to main branch with 0.22 eventing

This PR requires also https://github.com/openshift/release/pull/19133 to be merged

PR #1017, opened 2021-06-10
Cleanup monitoring resource files

skonto:
-  removed some dependencies on resource yaml files.
-  cleaned up some old code

Note: This the first of a series of PR to make things a bit more compact. There is still code that can be shared between knative-operator and openshift-knative-operator.


skonto:
@matzew could you approve this one?

cardil:
@skonto Merge main. With #1011 new tests was introduced. That should help the flakes.

skonto:
@markusthoemmes could you review pls?

knative-operator/pkg/common/monitoring.go
```diff
@@ -108,37 +99,69 @@ func addMonitoringLabelToNamespace(namespace string, api client.Client) error {
...
+	}
+	rb := rbacv1.RoleBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "knative-serving-prometheus-k8s",
+			Namespace: ns,
+		},
+		RoleRef: rbacv1.RoleRef{
+			APIGroup: "rbac.authorization.k8s.io",
+			Kind:     "Role",
+			Name:     "knative-serving-prometheus-k8s",
```
markusthoemmes, line 140:
```suggestion
			Name:     role.Name,
```

knative-operator/pkg/common/service_monitor.go
```diff
@@ -13,96 +12,102 @@ import (
...
+		return err
 	}
-	return manifest.Apply()
-}
 
-func getMonitorPath(envVar string, defaultVal string) string {
-	path := os.Getenv(envVar)
-	if path == "" {
-		return defaultVal
+	if *smManifest, err = smManifest.Transform(mf.InjectOwner(srv)); err != nil {
```
markusthoemmes, line 53: Why not use the instance as an owner here too? Both are created from `instance` technically. The ServiceMonitor isn't owned by the `service`.

That way you could prevent the interim fetching and also create just one manifest as with the RBAC stuff.

markusthoemmes (commented):
Some driveby comments. I was wondering if we should even create manifests in this case vs. applying the resources with the client "directly". I guess we safe the fetch-if-not-found-update-else-create type logic.

knative-operator/pkg/common/service_monitor.go
skonto: One way to do this is to have the service depend on the instance deployment and service monitor on the service because when the deployment is gone there is nothing to scrape from thus service is gone and then service monitor which depends on the existence of the service is gone. This is [how](https://github.com/operator-framework/operator-sdk/blob/53b00d125fb12515cd74fb169149913b401c8995/pkg/metrics/service-monitor.go#L72) it was done at the operator sdk side.

skonto:
```
INFO[2021-06-14T08:42:17Z] level=info msg=Creating infrastructure resources... 
INFO[2021-06-14T08:42:17Z] level=error                                  
INFO[2021-06-14T08:42:17Z] level=error msg=Error: [ERR]: Error building changeset: InvalidChangeBatch: [Tried to create resource record set [name='api.ci-op-yv01ixlw-fbf8c.origin-ci-int-aws.dev.rhcloud.com.', type='A'] but it already exists] 
INFO[2021-06-14T08:42:17Z] level=error msg=	status code: 400, request id: 2800ac89-2312-4901-8682-d39aca3d1796 
INFO[2021-06-14T08:42:17Z] level=error                                  
INFO[2021-06-14T08:42:17Z] level=error msg=  on ../tmp/openshift-install-267055916/route53/base.tf line 42, in resource "aws_route53_record" "api_external_alias": 
INFO[2021-06-14T08:42:17Z] level=error msg=  42: resource "aws_route53_record" "api_external_alias" { 
INFO[2021-06-14T08:42:17Z] level=error                                  
INFO[2021-06-14T08:42:17Z] level=error                                  
INFO[2021-06-14T08:42:17Z] level=fatal msg=failed to fetch Cluster: failed to generate asset "Cluster": failed to create cluster: failed to apply Terraform: failed to complete the change 
```
What is happening here?

skonto:
> I was wondering if we should even create manifests in this case vs. applying the resources with the client "directly". I guess we safe the fetch-if-not-found-update-else-create type logic.

I thought about this because I was tempted to remove them completely. So I thought of two advantages:
- as you mentioned not to check for the already exists error
- get owner ref set with one line transform. The code for that part would have to be copied explicitly and I thought it was too much.

knative-operator/pkg/common/service_monitor.go
markusthoemmes: Meh, I guess it's a detail anyway but I personally don't see a reason to build a requirement chain here as both resources are owned by the same thing, technically.

skonto: Ok I can remove it no problem for me.

skonto:
@markusthoemmes comments addressed lets see if all tests pass.

skonto:
infra ```sg=  on ../tmp/openshift-install-513259411/route53/base.tf line 42, in resource "aws_route53_record" "api_external_alias":```

knative-operator/pkg/common/service_monitor.go
```diff
@@ -12,97 +10,76 @@ import (
...
+		Spec: monitoringv1.ServiceMonitorSpec{
+			Endpoints: []monitoringv1.Endpoint{{Port: "http-metrics"}},
+			NamespaceSelector: monitoringv1.NamespaceSelector{
+				MatchNames: []string{ns},
+			},
+			Selector: metav1.LabelSelector{
+				MatchLabels: map[string]string{"name": depName},
+			},
+		}}
+	sm.Labels["name"] = sm.Name
```
markusthoemmes, line 76: Does the ServiceMonitor even need this label?

knative-operator/pkg/common/service_monitor.go
```diff
@@ -12,97 +10,76 @@ import (
...
+			Namespace: ns,
+			Labels:    kmeta.CopyMap(labels),
+		},
+		Spec: monitoringv1.ServiceMonitorSpec{
+			Endpoints: []monitoringv1.Endpoint{{Port: "http-metrics"}},
+			NamespaceSelector: monitoringv1.NamespaceSelector{
+				MatchNames: []string{ns},
+			},
+			Selector: metav1.LabelSelector{
+				MatchLabels: map[string]string{"name": depName},
```
markusthoemmes, line 73: It'd be cool to "declaratively" connect this with the service above, for example like so:

```golang
selector := map[string]string{"name": depName}
sms := v1.Service{
	ObjectMeta: metav1.ObjectMeta{
		Name:      depName,
		Namespace: ns,
		Labels:    kmeta.UnionMaps(labels, selector),
	},
	Spec: v1.ServiceSpec{
		Ports: []v1.ServicePort{{
			Name:       "http-metrics",
			Port:       9090,
			TargetPort: intstr.FromInt(9090),
			Protocol:   "TCP",
		}},
		Selector: kmeta.CopyMap(labels),
	}}
sm := monitoringv1.ServiceMonitor{
	ObjectMeta: metav1.ObjectMeta{
		Name:      depName,
		Namespace: ns,
		Labels:    kmeta.CopyMap(labels),
	},
	Spec: monitoringv1.ServiceMonitorSpec{
		Endpoints: []monitoringv1.Endpoint{{Port: "http-metrics"}},
		NamespaceSelector: monitoringv1.NamespaceSelector{
			MatchNames: []string{ns},
		},
		Selector: metav1.LabelSelector{
			MatchLabels: selector,
		}
	}}

File: knative-operator/pkg/common/service_monitor.go [hunk 5]
```diff
@@ -12,97 +10,76 @@ import (
...
-	if path == "" {
-		return defaultVal
+func createServiceMonitorManifest(labels map[string]string, depName string, ns string, options mf.Option) (*mf.Manifest, error) {
+	var svU = &unstructured.Unstructured{}
+	var smU = &unstructured.Unstructured{}
+	sms := v1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      depName,
+			Namespace: ns,
+			Labels:    kmeta.CopyMap(labels),
```
markusthoemmes, line 46: Any particular reason we apply the same labels here that we match on? Same for the ServiceMonitor. Do we need labels at all (modulo the one to match on for the ServiceMonitor).

markusthoemmes (approved):
/lgtm
/approve

Some flyby comments, that are just nits really though. Feel free to unhold if you don't think we should change stuff.

/hold

knative-operator/pkg/common/service_monitor.go
skonto: Added for consistency with the service labels as done in that operator framework.

knative-operator/pkg/common/service_monitor.go
skonto: These labels come from the deployment so I use them to tag the svc/sm too. It is a way to filter resources also from a cli perspective.

skonto:
@markusthoemmes I will unhold this one and will update anything needed in another PR. I am planning to refactor more stuff anyway.

knative-operator/pkg/common/service_monitor.go
markusthoemmes: They actually are the selector of the deployment though, right? So they'd be the same labels as on the pods.

markusthoemmes:
WFM, thanks! :+1:

PR #1020, opened 2021-06-15
[SRVCOM-1358] Fix getting started and installation links in CSV

markusthoemmes:
As per title, these links have changed around a bit, so let's fix it.

/assign @maschmid


olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
```diff
@@ -105,7 +105,7 @@ spec:
 
     The components provided with the OpenShift Serverless operator require minimum cluster sizes on
     OpenShift Container Platform. For more information, see the documentation on [Getting started
-    with OpenShift Serverless](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.6/html-single/serverless_applications/index#serverless-getting-started).
+    with OpenShift Serverless](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.6/html-single/serverless/index#serverless-getting-started).
```
skonto, line 108: Should we use 4.7 for the link or 4.6 is used as the oldest one?

markusthoemmes: This is generated via the project settings. I thought the same thing but shrugged at it :joy:. In theory, 1.16.0 will even be shipped to 4.8 so I dunno what the mechanics will be there wrt. shipping the docs first etc etc.

PR #1023, opened 2021-06-15
[SRVCOM-1285] Drop unsupported fields from the Serving CRD

markusthoemmes:
This improves our UX by dropping all the fields we don't expect users to set from the `explain` output of the Serving CRD. We'll still allow users to make use of those fields generally (by adding the `preserve-unknown-fields` clause) but we'll no longer actively advertise them.

/hold

For feedback.


skonto:
I guess nothing changes when it comes to validation as illegal arbitrary fields will be detected correct?

markusthoemmes:
Not sure I follow: What kinda validation are you referring to?

skonto:
CR validation.

markusthoemmes:
So in this case, it'd allow arbitrary keys on `spec` and `spec.ingress` (to allow users to theoretically still use the stuff we've hidden). Without this, unknown keys would just be swallowed. AFAIK, validation itself doesn't complain.

skonto:
Ok I am asking because there were some bugs (fixed now) https://github.com/kubernetes/kubernetes/pull/96369 and not sure what is the status of this.

markusthoemmes:
/retest

Infraaaa.

skonto:
@markusthoemmes should we unhold now that we got the feedback needed?

PR #1024, opened 2021-06-16
[SRVCOM-1148] Automatically use stop-image from previous version

markusthoemmes:
This approach is being adopted for now. The 1.15 image has already been produced, so this just puts the pieces together. "Automated" fetching the "last" image via a suggestion by @skonto.

### DIFF
--- a/hack/lib/catalogsource.bash
+++ b/hack/lib/catalogsource.bash
@@ -119,7 +119,7 @@ spec:
         - -c
         - |-
           podman login -u $pull_user -p $token image-registry.openshift-image-registry.svc:5000 && \
-          /bin/opm registry add -d index.db --container-tool=podman --mode=replaces -b quay.io/openshift-knative/serverless-bundle:1.7.2,quay.io/openshift-knative/serverless-bundle:1.8.0,quay.io/openshift-knative/serverless-bundle:1.9.0,quay.io/openshift-knative/serverless-bundle:1.10.0,quay.io/openshift-knative/serverless-bundle:1.10.1,quay.io/openshift-knative/serverless-bundle:1.11.0,registry.ci.openshift.org/openshift/openshift-serverless-v1.12.0:serverless-bundle,registry.ci.openshift.org/openshift/openshift-serverless-v1.13.0:serverless-bundle,quay.io/openshift-knative/serverless-bundle:1.14.0,registry.ci.openshift.org/openshift/openshift-serverless-v1.15.0:serverless-bundle,image-registry.openshift-image-registry.svc:5000/$OLM_NAMESPACE/serverless-bundle && \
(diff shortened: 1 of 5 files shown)


markusthoemmes:
/retest

Infra.

markusthoemmes:
/retest

Infra...

markusthoemmes:
/retest

Infra.

hack/lib/catalogsource.bash
```diff
@@ -119,7 +119,7 @@ spec:
         - -c
         - |-
           podman login -u $pull_user -p $token image-registry.openshift-image-registry.svc:5000 && \
-          /bin/opm registry add -d index.db --container-tool=podman --mode=replaces -b quay.io/openshift-knative/serverless-bundle:1.7.2,quay.io/openshift-knative/serverless-bundle:1.8.0,quay.io/openshift-knative/serverless-bundle:1.9.0,quay.io/openshift-knative/serverless-bundle:1.10.0,quay.io/openshift-knative/serverless-bundle:1.10.1,quay.io/openshift-knative/serverless-bundle:1.11.0,registry.ci.openshift.org/openshift/openshift-serverless-v1.12.0:serverless-bundle,registry.ci.openshift.org/openshift/openshift-serverless-v1.13.0:serverless-bundle,quay.io/openshift-knative/serverless-bundle:1.14.0,registry.ci.openshift.org/openshift/openshift-serverless-v1.15.0:serverless-bundle,image-registry.openshift-image-registry.svc:5000/$OLM_NAMESPACE/serverless-bundle && \
+          /bin/opm registry add -d index.db --container-tool=podman --mode=replaces -b docker.io/markusthoemmes/serverless-bundle:1.15-stop,image-registry.openshift-image-registry.svc:5000/$OLM_NAMESPACE/serverless-bundle && \
```
skonto, line 122: project yaml should have the previous and the current versions and we can use them here, correct?

markusthoemmes: Good idea! Let's do that as a followup. I'll need to discuss this "stop" image stuff with others first, but if we land that it'd be great to get this "bump" automated!

mgencur: Should that be `1.16-stop` instead of `1.15-stop`? (Looking at the stopbundle.Dockerfile below). 
Otherwise this approach looks good to me. And +1 for @skonto 's suggestion.

markusthoemmes: No, in this case you need 1.15 (the "old" version) and the current one (which is 1.16, which is the internally built one).

markusthoemmes: Done.

olm-catalog/serverless-operator/stopbundle.Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM alpine
```
matzew, line 1: is that "OK" ?

markusthoemmes: Why would it not be? It's only used in our CI and the only reason this is not `scratch` is us needing `sed`.

matzew: Just asking, in case some "policies" ... :smile: 

thanks for the PR!

/lgtm

PR #1026, opened 2021-06-16
Add patch to change kafka-webhook pdb to 1 (absolute value)

lberk:
related: SRVKE-817

PR #1032, opened 2021-06-17
SRVCOM-1375 Check Serverless alerts after tests

mgencur:
Midstream impl for https://issues.redhat.com/browse/SRVCOM-1375

### DIFF
--- a/test/lib.bash
+++ b/test/lib.bash
@@ -270,6 +270,22 @@ function gather_knative_state {
     --dest-dir "$gather_dir" > "${gather_dir}/gather-knative.log"
 }
 
+function check_serverless_alerts {
+  logger.info 'Checking Serverless alerts'
+  local alerts_file monitoring_route num_alerts
+  alerts_file="${ARTIFACT_DIR:-/tmp}/alerts.json"
+  monitoring_route=$(oc -n openshift-monitoring get routes alertmanager-main -oyaml -ojsonpath='{.spec.host}')
+  curl -k -H "Authorization: Bearer $(oc -n openshift-monitoring sa get-token prometheus-k8s)" \
+    "https://${monitoring_route}/api/v1/alerts" | \
+    jq -c '.data | map(select(.labels.namespace == "'"${OPERATORS_NAMESPACE}"'" or .labels.namespace == "'"${EVENTING_NAMESPACE}"'" or .labels.namespace == "'"${SERVING_NAMESPACE}"'" or .labels.namespace == "'"${INGRESS_NAMESPACE}"'"))' > "${alerts_file}"
+
+  num_alerts=$(jq 'length' "${alerts_file}")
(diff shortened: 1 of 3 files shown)


test/upstream-e2e-tests.sh
```diff
@@ -48,4 +48,6 @@ if [[ $TEST_KNATIVE_E2E == true ]]; then
   upstream_knative_eventing_e2e
 fi
 
+check_serverless_alerts
```
mgencur, line 51: Maybe I could call this function also here in operator-e2e: https://github.com/openshift-knative/serverless-operator/blob/main/test/e2e-tests.sh#L42 . WDYT?

test/lib.bash
```diff
@@ -270,6 +270,22 @@ function gather_knative_state {
     --dest-dir "$gather_dir" > "${gather_dir}/gather-knative.log"
 }
 
+function check_serverless_alerts {
+  logger.info 'Checking Serverless alerts'
+  local alerts_file monitoring_route num_alerts
+  alerts_file="${ARTIFACT_DIR:-/tmp}/alerts.json"
+  monitoring_route=$(oc -n openshift-monitoring get routes alertmanager-main -oyaml -ojsonpath='{.spec.host}')
```
markusthoemmes, line 277: Will this work in CRC too? If you don't know, I can help doublecheck.

mgencur: Yeah. I don't know. ...if you could check that would be great.

mgencur:
So, the alert check is already in place and the "upstream-e2e" check failed because there was one alert: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-knative_serverless-operator/1032/pull-ci-openshift-knative-serverless-operator-main-4.7-upstream-e2e-aws-ocp-47/1405770657081331712/artifacts/upstream-e2e-aws-ocp-47/serverless-e2e/artifacts/alerts.json
This was caused by https://github.com/openshift-knative/serverless-operator/pull/1026 not being merged yet. I'll rebase and try again. The error should be fixed then.

test/lib.bash
markusthoemmes: The route doesn't exist in a CRC cluster (In fact: It seems like the entire alert management system doesn't exist). Can we make room for that here? I.e. we can key off of `$OPENSHIFT_CI` being set or key off of the route not being there. I lean towards the former as that wouldn't "hide" issues if the route is renamed or whatever. Thoughts?

mgencur: Thanks. OK. Something like this?
```
[ -n "$OPENSHIFT_CI" ] && check_serverless_alerts
```

markusthoemmes: yep

markusthoemmes (approved):
/lgtm
/approve

PR #1033, opened 2021-06-17
[RELEASE-1.15] [SRVCOM-1148] Add a 'stop' bundle image that can act as the first in the chain

markusthoemmes:
As per title, this adds a new bundle image which drops the `replaces` stanza so it can act as the beginning of the graph for OLM. This allows us to start an index image from 1.15 (and every version from now on) to limit the amount of stuff that has to be downloaded.

/assign @mgencur


markusthoemmes:
/retest

Infra.

PR #1034, opened 2021-06-18
Use DomainMapping test with `--customdomain` option

nak3:
DomainMapping test needs two domain like:

- `original-ksvc.serverless.devcluster.openshift.com`
- `mydomain.example.com`

We have a wildcard DNS for `*.serverless.devcluster.openshift.com` but cannot prepare for DNS the second domain `mydomain.example.com` so we used spoof client for the `mydomain.example.com`.
But if we changed the second domain to `mydomain.serverless.devcluster.openshift.com`, it should work.

Hence this patch changes to use `--customdomain` option.

### DIFF
--- a/test/serving.bash
+++ b/test/serving.bash
@@ -63,7 +63,7 @@ function upstream_knative_serving_e2e_and_conformance_tests {
     ./test/e2e ./test/conformance/api/... ./test/conformance/runtime/... \
     --resolvabledomain --kubeconfig "$KUBECONFIG" \
     --imagetemplate "$image_template" \
-    --customdomain="$subdomain"
+    --customdomain="$subdomain" \
     --enable-beta \
     --enable-alpha
 
--- a/test/serving.bash
+++ b/test/serving.bash
@@ -49,6 +49,7 @@ function upstream_knative_serving_e2e_and_conformance_tests {
     --patch='{"spec": {"config": { "autoscaler": {"allow-zero-initial-scale": "true"}}}}'
 
   image_template="registry.ci.openshift.org/openshift/knative-${KNATIVE_SERVING_VERSION}:knative-serving-test-{{.Name}}"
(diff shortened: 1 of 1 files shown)


nak3:
Ah, I think this PR also needs https://github.com/knative/serving/commit/26c93b9c4af345ef4231659c4c0ac75ef35bec71 to be backported

test/serving.bash
```diff
@@ -62,6 +63,7 @@ function upstream_knative_serving_e2e_and_conformance_tests {
     ./test/e2e ./test/conformance/api/... ./test/conformance/runtime/... \
     --resolvabledomain --kubeconfig "$KUBECONFIG" \
     --imagetemplate "$image_template" \
+    --customdomain="$subdomain"
```
mgencur, line 66: I guess this line is missing the backslash.

nak3: Ah! Thank you!

nak3:
It still does not work but I am guessing that the DomainMapping test does not use `pkgtest.WaitForEndpointState` which has `RetryingRouteInconsistency`. (And we add our patch to `RetryingRouteInconsistency`).
So `CheckDistribution()` accesses to the OpenShift Route before it is actually ready.

nak3:
I realized that there are two issues in current test.

1. `RetryingRouteInconsistency` is necessary.
2. `--customdomain` with cluster subdomain [duplicated with the original ksvc's url](https://github.com/knative/serving/blob/1c042a1a8c977e570d29642aaef94ca426579704/test/conformance/api/v1alpha1/domain_mapping_test.go#L65-L66 ).  we need `host = svc.Service.Name + "-sumrandom." + test.ServingFlags.CustomDomain` for the customdomain.

After 1 & 2, running the test locally and CI on 4.6 passed but [4.7 does not](https://github.com/openshift/knative-serving/pull/751#issuecomment-864558368).

nak3:
Closing for now.

PR #1036, opened 2021-06-18
[SRVKS-743] Disable monitoring if istio is enabled

skonto:
This is intended for 1.16 and should be fixed in 1.17.

### DIFF
--- a/openshift-knative-operator/pkg/serving/extension.go
+++ b/openshift-knative-operator/pkg/serving/extension.go
@@ -119,7 +119,7 @@ func (e *extension) Reconcile(ctx context.Context, comp v1alpha1.KComponent) err
 
 	// Temporary fix for SRVKS-743
 	if ks.Spec.Ingress.Istio.Enabled {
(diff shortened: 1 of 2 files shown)


openshift-knative-operator/pkg/serving/extension.go
```diff
@@ -116,6 +116,11 @@ func (e *extension) Reconcile(ctx context.Context, comp v1alpha1.KComponent) err
 			Type: "ConfigMap",
 		}
 	}
+
+	// Temporary fix for SRVKS-743
+	if ks.Spec.Ingress.Istio.Enabled {
+		common.Configure(&ks.Spec.CommonSpec, monitoring.ObservabilityCMName, monitoring.ObservabilityBackendKey, "none")
```
markusthoemmes, line 122: let's make this `ConfigureIfUnset` just to leave the gate open for people who **really** want to play with it enabled explicitly.

skonto: Ok good to know updating in a sec.

markusthoemmes (approved):
/lgtm
/approve

PR #1043, opened 2021-06-22
[SRVKE-799] Channel visualization for the admin

skonto:
- adds a dashboard to visualize imc and kafka channel metrics. Uses `rate` to help identify [spikes/trends](https://www.metricfire.com/blog/understanding-the-prometheus-rate-function), same as with the existing dashboards. Background info of what we can use with a counter [here](https://www.robustperception.io/how-does-a-prometheus-counter-work) and [here](https://www.robustperception.io/rate-then-sum-never-sum-then-rate).
- adds dashboard lifecycle management

image

image

To test kafka metrics just use:
```
cat <<-EOF | kubectl apply -f -
---
apiVersion: messaging.knative.dev/v1beta1
kind: KafkaChannel
metadata:
  name: my-kafka-channel
spec:
  numPartitions: 3
  replicationFactor: 1
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: events-sa
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: event-watcher
rules:
  - apiGroups:
      - ""
    resources:
      - events
    verbs:
      - get
      - list
      - watch
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: k8s-ra-event-watcher
roleRef:
  kind: ClusterRole
  name: event-watcher
subjects:
  - kind: ServiceAccount
    name: events-sa
    namespace: default
---
apiVersion: sources.knative.dev/v1beta1
kind: ApiServerSource
metadata:
  name: testevents
spec:
  serviceAccountName: events-sa
  mode: Resource
  resources:
    - apiVersion: v1
      kind: Event
  sink:
    ref:
      apiVersion: messaging.knative.dev/v1beta1
      kind: KafkaChannel
      name: my-kafka-channel
---
apiVersion: messaging.knative.dev/v1beta1
kind: Subscription
metadata:
  name: my-subscription
  namespace: default
spec:
  channel:
    apiVersion: messaging.knative.dev/v1beta1
    kind: KafkaChannel
    name: my-kafka-channel
  delivery:
    deadLetterSink:
      ref:
        apiVersion: serving.knative.dev/v1
        kind: Service
        name: event-display
  subscriber:
    ref:
      apiVersion: serving.knative.dev/v1
      kind: Service
      name: event-display
---
apiVersion: serving.knative.dev/v1
kind: Service
metadata:
  name: event-display
spec:
  template:
    spec:
      containers:
        - # This corresponds to
          # https://github.com/knative/eventing-contrib/tree/master/cmd/event_display/main.go
          image: quay.io/openshift-knative/knative-eventing-sources-event-display

EOF
```

### DIFF
--- a/knative-operator/pkg/controller/dashboard/testdata/grafana-dash-knative-eventing-channel.yaml
+++ b/knative-operator/pkg/controller/dashboard/testdata/grafana-dash-knative-eventing-channel.yaml
@@ -1,0 +1,1065 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: grafana-dashboard-definition-knative-eventing-channel
+  namespace: openshift-config-managed
+  labels:
+    console.openshift.io/dashboard: "true"
+data:
+  eventing-source-dashboard.json: |+
+    {
+      "__inputs": [
+        {
+          "description": "",
+          "label": "prometheus",
+          "name": "prometheus",
+          "pluginId": "prometheus",
+          "pluginName": "Prometheus",
+          "type": "datasource"
+        }
+      ],
+      "annotations": {
+        "list": [
+          {
+            "builtIn": 1,
+            "datasource": "-- Grafana --",
+            "enable": true,
+            "hide": true,
+            "iconColor": "rgba(0, 211, 255, 1)",
+            "name": "Annotations & Alerts",
+            "type": "dashboard"
+          }
+        ]
+      },
+      "editable": false,
+      "gnetId": null,
+      "graphTooltip": 0,
+      "id": 16,
+      "links": [],
+      "panels": [
+        {
+          "collapsed": false,
+          "gridPos": {
+            "h": 1,
+            "w": 24,
+            "x": 0,
+            "y": 0
+          },
+          "id": 2,
+          "panels": [],
+          "repeat": null,
+          "title": "In-Memory Channel Aggregated Metrics",
+          "type": "row"
+        },
+        {
+          "cacheTimeout": null,
+          "colorBackground": false,
+          "colorValue": false,
+          "colors": [
+            "#299c46",
+            "rgba(237, 129, 40, 0.89)",
+            "#d44a3a"
+          ],
+          "datasource": "prometheus",
+          "decimals": 3,
+          "description": "",
+          "format": "ops",
+          "gauge": {
+            "maxValue": 100,
+            "minValue": 0,
+            "show": false,
+            "thresholdLabels": false,
+            "thresholdMarkers": true
+          },
+          "gridPos": {
+            "h": 8,
+            "w": 4,
+            "x": 0,
+            "y": 1
+          },
+          "id": 29,
+          "interval": "",
+          "links": [],
+          "mappingType": 1,
+          "mappingTypes": [
+            {
+              "name": "value to text",
+              "value": 1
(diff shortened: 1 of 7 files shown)


skonto:
```rate(v range-vector) calculates the per-second average rate of increase of the time series in the range vector.```
In the dashboards `rate per minute` means that the average was calculated over a minute of time-series data.

**Comment by aliok** [comment]
@skonto 
> Might be better to switch to `avg per sec, over 1m window'.

Yeah, depends a lot on the use case, but I think showing `avg per sec, over 1m window' would be more useful for most of the use cases.

**Comment by aliok** [comment]
LGTM

Reconciliation code looks good and the screenshots are awesome ❤️  ...though they're hard to review

**Comment by skonto** [comment]
@aliok thanks for the review. This is already the case the rate is defined per minute anyway in the dashboard and this cannot be configured dynamically with native dashboards unless user edits them on the fly. What I am trying to say is the label text could be changed to be more precise because users might think that we show counts per minute while we show avg counts per sec on data gathered per minute ;)

File: knative-operator/pkg/controller/dashboard/testdata/grafana-dash-knative-eventing-channel.yaml [hunk 1]
```diff
@@ -0,0 +1,1065 @@
...
+          "5m",
+          "15m",
+          "30m",
+          "1h",
+          "2h",
+          "1d"
+        ]
+      },
+      "timezone": "",
+      "title": "Knative Eventing - Channel",
```
matzew, line 1062: since these has IMC and KC channels.  Should the title say `Channels` ?

skonto: I didnt use plural for the other dashboards so I kept this convention here too.

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
```diff
@@ -426,6 +426,8 @@ spec:
                         value: "deploy/resources/dashboards/grafana-dash-knative-eventing-source.yaml"
                       - name: EVENTING_BROKER_DASHBOARD_MANIFEST_PATH
                         value: "deploy/resources/dashboards/grafana-dash-knative-eventing-broker.yaml"
+                      - name: EVENTING_CHANNEL_DASHBOARD_MANIFEST_PATH
```
markusthoemmes, line 429: driveby comment: Is there a reason we don't just read and apply the entire `dashboards` folder? If we need scoping between serving and eventing etc, we could add subfolders and thelike. Maybe food for a followup.

skonto: Yes no reason to install an eventin dashboard if eventing is never installed. The lifecycle of a dashboard depends on the instance of the corresponding CR (serving, eventing etc).

aliok:
> I am not sure I can do much for the dashboards to be easier to review :( I am open to suggestions.

Sorry, I don't have any suggestions. But it is on me/us and not you.

aliok (approved):
/lgtm
/approve

PR #1045, opened 2021-06-22
Pull all manifests from upstream

markusthoemmes:
We pull all manifests via upstream except net-istio, so this fixes that. Also use the correct Kourier version for manifest pulling.


matzew (approved):
/lgtm
/approve

PR #1051, opened 2021-06-23
[RELEASE-1.15][SRVKE-841] Fix source resources dashboard for long source names

skonto:
Right now long names are truncated by Eventing and dashboards dont work as expected. More details in SRVKE-841.
Apiserversource:
image
image

Ping source:
image
image

Kafka source:
image

PR #1052, opened 2021-06-23
[Main][SRVKE-841] Fix source resources dashboard for long source names

skonto:
As in #1051.


maschmid:
I don't quite like that we're now showing the ugly pod names in what should be a list of sources, because the dashboard is supposed to be a list of sources, not pods. (and that pods are internal details of the sources which should be hidden in this dashboard). 

I mean even if the metric internally is just a metric of the pods, IMHO it should not present itself as such... 

But I am OK with this fix if we're sure this is the way for now (but IMHO we should aim for a better solution for the future)

skonto:
@maschmid it is not bad if you think about sources. I dont see pods as details this is not a dashboard for funcs. So imagine you have a source which has HA and some pod is misbehaving then you can find more about it with this dashboard. Anyway we can improve it in the future by having a dashboard per source to make things more clear. If eventing team @matzew @aliok thinks changing source name implementation as an easy way to go, then I am happy to do it.

matzew:
I am also OK w/ it.  I leave the approval to @maschmid since he found the issue, and is therefore better for judging here.

maschmid:
OK.

/lgtm

matzew:
Not sure, but worth to think about doing a cherry pick to the 1.16 branch. In case we need a respin ?

aliok (approved):
/lgtm
/approve

Thanks!

skonto:
I can create a PR against 1.16. Any clarifications on:
a) do all sources support HA (not dispatcher)?
b) how easy it is to change the source name upstream (consequences?).

aliok:
@skonto 

> a) do sources support HA (not dispatcher)?

No, they don't. 

> b) how easy it is to change the source name upstream (consequences?).

You mean the controller pod name?

skonto:
> You mean the controller pod name?

I mean changing the expression [here](https://github.com/knative/eventing/blob/main/pkg/reconciler/apiserversource/resources/receive_adapter.go#L63) to keep the last dash even if the suffix is over than 31 chars.

aliok:
@skonto Technical part is very easy, we need to just update it there and in the tests.

Not sure if there are other vendors who already has that deployment name used somewhere.

skonto:
@maschmid given @aliok feedback I updated the PR removing the ugly pod name:

image

maschmid:
Hm, that's weird, I don't see the "Source" selector at all here: 

knative_eventing_source_cpu_and_memory_usage_no_source_selector

(OCP 4.8.0-fc.9 , serverless-operator from `main` (eventing 0.22.0)

maschmid:
It seems to work fine on 4.7.9 . I suppose this could be some change in the metrics in 4.8?  (I don't see the "Total CPU Usage (rate per minute)" graph at all there)

skonto:
@maschmid check last commit it should fix the issue.
image

maschmid:
@skonto The "Source" selector is still missing on OCP 4.8 though.

skonto:
@maschmid changing to `            "query": "label_values(kube_pod_labels{label_eventing_knative_dev_source_name=~\".+\", namespace=\"$namespace\", label_eventing_knative_dev_source=\"$scontroller\"}, label_eventing_knative_dev_source_name)",`
makes this work. Kube-state-metrics have changed. I guess it is due to this: https://github.com/kubernetes/kube-state-metrics/pull/1165

skonto:
@maschmid I fixed the missing selector. Unfortunately kube-state-metrics broke compatibility. In order to avoid detecting ocp version I used each label in a mutually exclusive manner via [this approach](https://stackoverflow.com/questions/51293895/how-to-display-all-metrics-that-dont-have-a-specific-label) which defines a non existing label via the empty string (to be on the safe side).
4.8:
image

4.7:
image
I had to change the name of the selector from `source` to `nameOfSource` because native dashboards cannot distinguish between names with the same prefix (obsiously a bug imho).

knative-operator/deploy/resources/dashboards/grafana-dash-knative-eventing-resources.yaml
```diff
@@ -488,9 +488,29 @@ data:
             "includeAll": false,
             "label": "SourceName",
             "multi": false,
-            "name": "source",
+            "name": "nameofSource",
             "options": [],
-            "query": "label_values(kube_pod_labels{label_eventing_knative_dev_sourceName=~\".+\", namespace=\"$namespace\", label_eventing_knative_dev_source=\"$scontroller\"}, label_eventing_knative_dev_sourceName)",
+            "query": "label_values(kube_pod_labels{label_eventing_knative_dev_sourceName=~\".+\", label_eventing_knative_dev_source_name=\"\", namespace=\"$namespace\", label_eventing_knative_dev_source=\"$scontroller\"}, label_eventing_knative_dev_sourceName)",
```
skonto, line 493: I am using an [empty string](https://stackoverflow.com/questions/51293895/how-to-display-all-metrics-that-dont-have-a-specific-label) to indicate that the metric shouldnt exist.

aliok (approved):
/approve
/lgtm

maschmid:
/unhold
looks good to me

PR #1072, opened 2021-06-30
[1.16] Setting Kafka Webhook HPA to minReplica of two

matzew:
As per title, done for SRVKE-884

### DIFF
--- a/knative-operator/deploy/resources/knativekafka/1-channel-consolidated.yaml
+++ b/knative-operator/deploy/resources/knativekafka/1-channel-consolidated.yaml
@@ -957,7 +957,7 @@ spec:
     apiVersion: apps/v1
     kind: Deployment
     name: kafka-webhook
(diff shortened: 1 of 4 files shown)


knative-operator/deploy/resources/knativekafka/1-channel-consolidated.yaml
```diff
@@ -957,7 +957,7 @@ spec:
     apiVersion: apps/v1
     kind: Deployment
     name: kafka-webhook
-  minReplicas: 1
+  minReplicas: 2
```
markusthoemmes, line 960: Do we really want this? It'll mean users won't be able to tune this down to 1 :thinking:.

matzew: I think I will anyways tackle this differently

PR #1073, opened 2021-06-30
[1.16] Modify the minReplica for HPA components based on HA

matzew:

### DIFF
--- a/knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
+++ b/knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
@@ -34,11 +34,12 @@ const (
 )
 
 var (
-	log               = logf.Log.WithName("controller_knativekafka")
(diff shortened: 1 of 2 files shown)


maschmid:
Hm...even with this change, it still seems the deployment and the HPA is being updated several times per minute by knative-openshift, and there is still only 1 instance of kafka-webhook running. (a second one appears only for a brief moment before it's immediately terminated)

There are also a lot of these (for various resources), so I suspect it may be an additional issue with the manifestival or the manifests we use... 

```
{"level":"info","ts":"2021-06-30T17:55:07.267Z","logger":"controller_knativekafka.mf","msg":"Merging","diff":"{\"metadata\":{\"creationTimestamp\":null}}"}
```

https://gist.github.com/maschmid/5dfd8c9de544b4e7b66989a2422d1697

PR #1078, opened 2021-07-01
Update serverless-operator with secure traffic doc

nak3:
This patch makes two updates:
- remove `"ingress.class": "istio.ingress.networking.knative.dev"`.
  - We don't need it as https://github.com/openshift-knative/serverless-operator/commit/70631c9c22d4a4b198d0c385d4b04faa43951546 improved.
- remove `default` namespace.
  - The default namespace messes up the NetworkPolicy so it was removed from default script.

/cc @maschmid @markusthoemmes

### DIFF
--- a/docs/full-mesh.md
+++ b/docs/full-mesh.md
@@ -58,12 +58,11 @@ spec:
     - knative-serving
     - serving-tests
     - serving-tests-alt
-    - default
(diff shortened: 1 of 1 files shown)


markusthoemmes (approved):
/lgtm
/approve

PR #1080, opened 2021-07-01
Stop auto-upgrading Openshift related golang deps

markusthoemmes:
These are just causing trouble we didn't quite ask for and it's ultra rarely needed anyway --> more trouble than it's worth.

This also includes a run of `./hack/update-deps.sh --upgrade` for consistency.


matzew (approved):
/lgtm
/approve

PR #1091, opened 2021-07-07
SRVCOM-1420 Test Serverless teardown after upgrade

mgencur:
* including CRD deletion

Related to https://issues.redhat.com/browse/SRVCOM-1420

### DIFF
--- a/hack/lib/serverless.bash
+++ b/hack/lib/serverless.bash
@@ -358,6 +358,8 @@ function teardown_serverless {
   done
   oc delete namespace openshift-serverless --ignore-not-found=true
 
+  oc get crd -oname | grep 'knative.dev' | xargs oc delete --timeout=60s
+
   logger.success 'Serverless has been uninstalled.'
 }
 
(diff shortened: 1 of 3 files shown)


hack/lib/serverless.bash
```diff
@@ -358,6 +358,10 @@ function teardown_serverless {
   done
   oc delete namespace openshift-serverless --ignore-not-found=true
 
+  if [[ ! $(oc get crd -oname | grep -c 'knative.dev') -eq 0 ]]; then
+    oc get crd -oname | grep 'knative.dev' | xargs oc delete --timeout=60s
+  fi
```
markusthoemmes, line 363: Are we expecting stuff to hang/fail if it doesn't work as we expect?

markusthoemmes (commented):
Looks sound, are we already testing shutdown in non-upgrade paths?

hack/lib/serverless.bash
mgencur: If the deletion of CRDs hangs it means there's a problem, but hanging is limited to 60 seconds. I didn't want the whole job to be timing out after 2hrs.

Regarding running shutdown in non-upgrade paths. The serverless-operator Golang tests deploy/undeploy KnativeServing and the operator but they don't delete the CRDs. I'm not aware of other places where we'd be testing that.

markusthoemmes (approved):
/lgtm
/approve

PR #1092, opened 2021-07-07
Serve CLI artifacts from the operator deployment rather than a ksvc

markusthoemmes:
As per title, this changes the CLI artifacts to no longer be served through a Knative Service but rather launches the respective container as part of the serverless-operator deployment itself. To minimize resource consumption, the container is left without any resource requirements and/or limits, so the Openshift scheduler pretty much ignores it.

**Reasons:**
- Interdependence with Knative Serving: We can only serve the CLI binaries if Knative Serving is installed.
- Interdependence networking wise: When deploying with istio and Service Mesh support, we have to make sure that the CLI artifact service is available "generally", regardless of the implications that customers might want to put on their networking config. Having it as a "hidden" service (and thus naturally part of the mesh etc) makes this difficult to assume.
- Interdependence installation wise: When uninstalling Knative Serving, we also trigger the clean up of the artifact service. However, this is racy: Knative can be deleted before the Service finishes deleting, causing hanging resources because of unhandled finalizers etc.

**Out of scope:**
- Making the ConsoleCLIDownload available regardless of installing Knative Serving or not. This can (and should eventually) be done in a subsequent step.

### DIFF
--- a/knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
+++ b/knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
@@ -6,91 +6,83 @@ import (
 	"os"
 	"strings"
 
-	v1 "github.com/openshift/api/route/v1"
-	appsv1 "k8s.io/api/apps/v1"
-
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/common"
-
 	consolev1 "github.com/openshift/api/console/v1"
+	routev1 "github.com/openshift/api/route/v1"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	servingv1alpha1 "knative.dev/operator/pkg/apis/operator/v1alpha1"
 	servingv1 "knative.dev/serving/pkg/apis/serving/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
-	knCLIDownload               = "kn"
-	knConsoleCLIDownloadService = "kn-cli"
-	deprecatedResourceName      = "kn-cli-downloads"
+	knCLIDownload          = "kn"
+	deprecatedResourceName = "kn-cli-downloads"
 )
+
+var operatorNamespace = os.Getenv(common.NamespaceEnvKey)
 
 var log = common.Log.WithName("consoleclidownload")
 
 // Apply installs kn ConsoleCLIDownload and its required resources
 func Apply(instance *servingv1alpha1.KnativeServing, apiclient client.Client, scheme *runtime.Scheme) error {
-	if !instance.Status.IsReady() {
-		// Don't return error, wait silently until Serving instance is ready
-		return nil
-	}
 	// Remove deprecated resources from previous version
 	if err := deleteDeprecatedResources(instance, apiclient); err != nil {
 		return err
 	}
-	service := &servingv1.Service{}
-	if err := reconcileKnCCDResources(instance, apiclient, service); err != nil {
+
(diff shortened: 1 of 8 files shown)


knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
```diff
@@ -158,87 +154,62 @@ func deleteDeprecatedResources(instance *servingv1alpha1.KnativeServing, apiclie
...
-									},
-								},
-							},
-						},
-					},
-				},
+		Spec: routev1.RouteSpec{
+			To: routev1.RouteTargetReference{
+				Kind: "Service",
+				Name: "knative-openshift-metrics-3",
```
markusthoemmes, line 170: non straightforward name :thinking: we might want to consider renaming the service or adding a new one.

skonto: @markusthoemmes I would just add a new one to separate things because these resources are usually tight to the corresponding deployment (onwener refs). Btw we had that name because it is used by the `knative-openshift` deployment plus the suffix for the migration issues we faced with OLM.  Why is this route to the service monitor service being created?

markusthoemmes: FWIW, the lifecycle of this is managed by OLM, not by the deployment. This is not one of the autoimatically injected services.

The route is created to make the CLI artifact server available for download via the ConsoleCLIDownload resource.

dsimansk:
Thanks @markusthoemmes! Looks good and clean. :)
/lgtm

Adding hold for @rhuss to take a look. Feel free to cancel if needed.
/hold

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
skonto: Yes it is not I am saying when I would add a new one. Ok from what I see you take advantage of that service to get access to the container for the cli. It seems to should work, cant thing of a problem right now. The service monitor access only the metrics port so no collision there.

rhuss:
> To minimize resource consumption,

why does this reduce resource consumption, when the very same container is now running all the time instead of being scaled down to 0 if not used ? The other benefits are probably worth the extra memory spend (@dsimansk do you know how much heap is used by the download HTTP server), but I don't see how resource consumption is minimized (at least not when it comes to the download HTTP server).

markusthoemmes:
You're right, that was poorly phrased. I should've said: To minimize the impact on scheduling decisions.

A next step (and we've already discussed this) can be to ditch the artifact image's server and serve the binaries from the "operator" container, as an HTTP server part of that process. However, I'm quite dubious that that is worth anything, at least at this moment in time. I'd think we have simpler (and more powerful) ways of reducing our overall overhead vs. doing everything we can to avoid the overhead of this, as you mention, low traffic http server.

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
```diff
@@ -385,6 +383,12 @@ spec:
               spec:
                 serviceAccountName: knative-operator
                 containers:
+                  - name: cli-downloads
+                    image: registry.ci.openshift.org/openshift/knative-v0.22.0:kn-cli-artifacts
+                    imagePullPolicy: Always
+                    ports:
+                      - name: http-cli
+                        containerPort: 8080
```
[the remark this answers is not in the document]
rhuss: @markusthoemmes just for my understanding, the serverless-operator runs without any resource requirements (i.e. no container has requests/limits), so being in QoS class "Best Effort" ? (i.e. being the first ones killed when coming under overcommit pressure). Is this what we want to have ?

rhuss:
@dsimansk do we have CI tests in the client that check the CLI download ? I would be positive for this PR and the let the client CI find out whether it works as before. Does this make sense ?

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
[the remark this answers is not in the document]
markusthoemmes: I don't think anybody has ever thought about that, no. We likely want to revisit that though, to my knowledge, this hasn't surfaced itself as an issue for now.

markusthoemmes:
@rhuss we even have tests here, I had to adjust them slightly, see `test/servinge2e/verify_deploy_resources_test.go`

PR #1096, opened 2021-07-08
Use of ocp range feature

rudyredhat1:
- Replaced the OCP key from the project.yaml and used ocpVersion with min and label sub-keys. One can be used for the csv.sh file for the document link and another can help in the generation of a range of ocp versions.
- I have not run the make release-files command, because its changing the Dockerfile present in olm-catalog.
- But tried in local setup, the values are now coming in range, but the "LABELoperators.operatorframework.io.bundle.channels.v1="- stable" , value is being changing to "- stable" , not able to get why the '-' is being added in the beginning.

### DIFF
--- a/olm-catalog/serverless-operator/project.yaml
+++ b/olm-catalog/serverless-operator/project.yaml
@@ -16,10 +16,9 @@ requirements:
     minVersion: 1.19.0
   golang: '1.15'
   nodejs: 14.x
-  ocp:
-    - '4.6'
-    - '4.7'
-    - '4.8'
+  ocpVersion:
+    min: '4.6'
+    label: 'v4.6-v4.8'
 
 dependencies:
   serving: 0.23.1
--- a/hack/generate/csv.sh
+++ b/hack/generate/csv.sh
@@ -84,7 +84,7 @@ yaml_keys[spec.minKubeVersion]="$(metadata.get requirements.kube.minVersion)"
 yaml_keys[spec.replaces]="$(metadata.get project.name).v$(metadata.get olm.replaces)"
 
 declare -A vars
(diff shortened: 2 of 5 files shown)


markusthoemmes:
The changes look sound to me, thanks! Asking for a review from Warren though since he has the context regarding the functionality of these image labels.

mgencur:
@rudyredhat1 Just curious, is it a requirement to use ranges in `com.redhat.openshift.versions=` ? Or it's just another option that can be used aside from listing individual versions? Asking because the previous way was more intuitive to me.

warrenvw:
Yes, per https://docs.google.com/document/d/1X4xwBK4ECIXjaA_0DuVNuWieY9Qjbl42BMRAyIYc_HM/edit#heading=h.4o2snu2vbv1v, `com.redhat.openshift.versions` no longer supports a comma separated list of OCP versions.

mgencur:
@rudyredhat1 @warrenvw OK. Then I guess this looks good to me:) Thanks

rudyredhat1:
Thanks @mgencur . Should I push the recent commits from the main branch? Like this is 4 commits behind as of now? Apart from this working for yq-4 version changes also. If those are done, then will be easy for next release.

olm-catalog/serverless-operator/project.yaml
```diff
@@ -16,10 +16,9 @@ requirements:
     minVersion: 1.19.0
   golang: '1.15'
   nodejs: 14.x
-  ocp:
-    - '4.6'
-    - '4.7'
-    - '4.8'
+  ocpVersion:
+    min: '4.6'
+    label: 'v4.6-v4.8'
```
cardil, line 21: That might be a breaking change for upgrade test triggering, right @mgencur?

PR #1097, opened 2021-07-08
[QE] Retry the flaky UI tests

cardil:
There's an option in Cypress to retry tests that fail: https://docs.cypress.io/guides/guides/test-retries

This change has pros and cons. 

It should make the UI tests more stable. On the other hand, it will effectively hide the flakes, as the tests would pass - so it would be harder to notice them.

The UI tests are running only nightly, so maybe merging this PR isn't a good idea. 

WDYT? @mgencur

### DIFF
--- a/test/ui/cypress.json
+++ b/test/ui/cypress.json
@@ -3,5 +3,9 @@
   "reporter": "cypress-multi-reporters",
   "reporterOptions": {
     "configFile": "reporter-config.json"
+  },
+  "retries": {
+    "runMode": 2,
+    "openMode": 0
   }
 }


cardil:
/hold

I'm now thinking that the retries will make it harder to spot those flakes. And we are raising a lot of valid ODC issues because we are able to find those flaky tests.

mgencur:
/lgtm
Let's enable this temporarily until the ODC issues causing failures are fixed.

cardil:
Yes. When we enable retries, we should still be able to monitor known problem using search.ci.openshift.org.

After flakes got fixes, we can remove retries and start looking for new flakes (if there are any).

/unhold

PR #1098, opened 2021-07-09
[SRVKE-897] Improve monitoring reconciliation for sources

skonto:
- Fixes SRVKE-897
- Makes monitoring setup for sources easier to configure. Right now by default we setup cluster monitoring for sources per namespace they reside in, for better get started UX. We do this by marking the corresponding ns with a special flag: `openshift.io/cluster-monitoring=true`. This causes any services running side by side with a source (same namespace) being scraped by the cluster monitoring Prometheus instance. This should be avoided in general because user services should have a different metrics budget and not impose load to the cluster instance. As a general rule of thumb though we should not restrict operations options.
Since OCP Monitoring (discussed with the team there) allows only to use either cluster monitoring or user-workload monitoring per ns (no mixed cases) we should allow users to setup things according to their needs. By default cluster monitoring is enabled to ship all Source metrics to the cluster Prometheus instance and feed admin dashboards with data (for better UX). 
With this PR there is more control by introducing two vars that affect how much automation we offer:
If `SOURCES_USE_CLUSTER_MONITORING` is set to true cluster monitoring is setup automatically.
If `SOURCES_GENERATE_SERVICE_MONITORS` is set to true service monitors are created for each source.
If user sets cluster monitoring to false and allows the generation of service monitors then automatically the sources are ready to be scraped by the user-workload monitoring Prometheus instance without effort. Enabling user-workload monitoring is a [one step process](https://docs.openshift.com/container-platform/4.7/monitoring/enabling-monitoring-for-user-defined-projects.html).
The user-workload monitoring Prometheus instance will not scrape ns targeted by the cluster one and vice versa, for more check [here](https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/user-workload-monitoring.md) and [here](https://www.openshift.com/blog/thanos-querier-versus-thanos-querier).
These env vars provide the maximum flexibility for the end-user.
- Each time an env update is done at the S-O side, the S-O deployment is restarted and a full resync of source deployments is being done (ctr runtime cache is synched). The same happens if the config for metrics backend is changed at the Eventing CR level. This guarantees that we dont have some sources being updated and some others not. Should be equivalent to the global resync in knative controllers.

### DIFF
--- a/knative-operator/cmd/manager/main.go
+++ b/knative-operator/cmd/manager/main.go
@@ -119,7 +119,7 @@ func main() {
 	hookServer.Register("/mutate-knativekafkas", &webhook.Admission{Handler: knativekafka.NewConfigurator(decoder)})
 	hookServer.Register("/validate-knativekafkas", &webhook.Admission{Handler: knativekafka.NewValidator(mgr.GetClient(), decoder)})
 
-	if err := setupMonitoring(cfg); err != nil {
+	if err := setupServerlesOperatorMonitoring(cfg); err != nil {
 		log.Error(err, "Failed to start monitoring")
 	}
 
@@ -142,7 +142,7 @@ func main() {
 	}
 }
 
-func setupMonitoring(cfg *rest.Config) error {
+func setupServerlesOperatorMonitoring(cfg *rest.Config) error {
 	cl, err := client.New(cfg, client.Options{})
 	if err != nil {
 		return fmt.Errorf("failed to create a client: %w", err)
@@ -163,7 +163,7 @@ func setupMonitoring(cfg *rest.Config) error {
 		return err
 	}
 
-	if err = monitoring.SetupMonitoringRequirements(cl, operatorDeployment); err != nil {
+	if err = monitoring.SetupClusterMonitoringRequirements(cl, operatorDeployment); err != nil {
 		return fmt.Errorf("failed to setup monitoring resources: %w", err)
 	}
 
--- a/knative-operator/cmd/manager/main.go
+++ b/knative-operator/cmd/manager/main.go
@@ -163,7 +163,7 @@ func setupServerlesOperatorMonitoring(cfg *rest.Config) error {
 		return err
 	}
 
-	if err = monitoring.SetupClusterMonitoringRequirements(cl, operatorDeployment); err != nil {
+	if err = monitoring.SetupClusterMonitoringRequirements(cl, operatorDeployment, namespace); err != nil {
 		return fmt.Errorf("failed to setup monitoring resources: %w", err)
 	}
 
--- a/knative-operator/cmd/manager/main.go
+++ b/knative-operator/cmd/manager/main.go
@@ -163,7 +163,7 @@ func setupServerlesOperatorMonitoring(cfg *rest.Config) error {
 		return err
 	}
 
-	if err = monitoring.SetupClusterMonitoringRequirements(cl, operatorDeployment, namespace); err != nil {
+	if err = monitoring.SetupClusterMonitoringRequirements(cl, operatorDeployment, namespace, nil); err != nil {
 		return fmt.Errorf("failed to setup monitoring resources: %w", err)
 	}
 
--- a/knative-operator/pkg/monitoring/sources/source_deployment_discovery_controller.go
+++ b/knative-operator/pkg/monitoring/sources/source_deployment_discovery_controller.go
@@ -3,8 +3,6 @@ package sources
 import (
 	"context"
 	"os"
-	"sigs.k8s.io/controller-runtime/pkg/event"
-	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"strconv"
 
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/common"
@@ -17,8 +15,10 @@ import (
 	"knative.dev/operator/pkg/apis/operator/v1alpha1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
--- a/knative-operator/pkg/monitoring/sources/source_deployment_discovery_controller.go
+++ b/knative-operator/pkg/monitoring/sources/source_deployment_discovery_controller.go
@@ -9,6 +9,7 @@ import (
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/monitoring"
 	okomon "github.com/openshift-knative/serverless-operator/openshift-knative-operator/pkg/monitoring"
 	v1 "k8s.io/api/apps/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -26,6 +27,8 @@ import (
 var (
 	generateSourceServiceMonitorsEnvVar = "SOURCES_GENERATE_SERVICE_MONITORS"
 	useClusterMonitoringEnvVar          = "SOURCES_USE_CLUSTER_MONITORING"
+	rbacLabelKey                        = "serverless.monitoring"
+	sourceRbacLabels                    = map[string]string{rbacLabelKey: "true"}
 
 	log = common.Log.WithName("source-deployment-discovery-controller")
 )
@@ -62,7 +65,25 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		}
 		return nil
 	})
-	return c.Watch(&source.Kind{Type: &v1.Deployment{}}, handler.EnqueueRequestsFromMapFunc(enqueueRequests), skipDeletePredicate{})
+
+	enqueueRequestsRBAC := handler.MapFunc(func(obj client.Object) []reconcile.Request {
+		oLabels := obj.GetLabels()
+		if _, ok := oLabels[rbacLabelKey]; ok {
+			// Use a special request to trigger ns level reconcilation related to sources
+			return []reconcile.Request{{
+				NamespacedName: types.NamespacedName{Namespace: obj.GetNamespace(), Name: ""},
+			}}
+		}
+		return nil
+	})
+
+	if err = c.Watch(&source.Kind{Type: &v1.Deployment{}}, handler.EnqueueRequestsFromMapFunc(enqueueRequests), skipDeletePredicate{}); err != nil {
+		return err
+	}
+	if err = c.Watch(&source.Kind{Type: &rbacv1.RoleBinding{}}, handler.EnqueueRequestsFromMapFunc(enqueueRequestsRBAC), skipCreatePredicate{}); err != nil {
+		return err
+	}
+	return c.Watch(&source.Kind{Type: &rbacv1.Role{}}, handler.EnqueueRequestsFromMapFunc(enqueueRequestsRBAC), skipCreatePredicate{})
 }
 
 // blank assignment to verify that ReconcileSourceDeployment implements reconcile.Reconciler
@@ -78,20 +99,27 @@ type ReconcileSourceDeployment struct {
 
 // Reconcile reads that state of the cluster for an eventing source deployment
 func (r *ReconcileSourceDeployment) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	reconcileForSource := true
+	if request.Name == "" {
+		// We will handle only rbac resources this time
+		reconcileForSource = false
+	}
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.Info("Reconciling the source deployment, setting up a service/service monitor if required")
 	dep := &v1.Deployment{}
 	inDeletion := false
-	err := r.client.Get(context.TODO(), request.NamespacedName, dep)
-	if apierrors.IsNotFound(err) {
-		// The deployment does not exist anymore, deletions are shown as failing reads
-		log.Info("Source in deletion phase")
-		inDeletion = true
-	} else if err != nil {
-		return reconcile.Result{}, err
+	if reconcileForSource {
+		err := r.client.Get(context.TODO(), request.NamespacedName, dep)
+		if apierrors.IsNotFound(err) {
+			// The deployment does not exist anymore, deletions are shown as failing reads
+			log.Info("Source in deletion phase")
+			inDeletion = true
+		} else if err != nil {
+			return reconcile.Result{}, err
+		}
 	}
 	eventing := &v1alpha1.KnativeEventing{}
-	if err = r.client.Get(context.TODO(), types.NamespacedName{Namespace: "knative-eventing", Name: "knative-eventing"}, eventing); err != nil {
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: "knative-eventing", Name: "knative-eventing"}, eventing); err != nil {
 		return reconcile.Result{}, err
 	}
 	// If monitoring is set to on/off this triggers a global resync to source adapters.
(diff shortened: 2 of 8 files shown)


skonto:
@aliok @markusthoemmes  pls review.

/assign @aliok

knative-operator/pkg/monitoring/sources/source_deployment_discovery_controller.go
```diff
@@ -69,38 +77,102 @@ type ReconcileSourceDeployment struct {
...
+	enable := os.Getenv(generateSourceServiceMonitorsEnvVar)
+	if enable != "" {
+		parsed, err := strconv.ParseBool(enable)
+		if err != nil {
+			// ignore value if garbage
+			return true
+		}
+		return parsed
+	}
+	return true
```
skonto, line 177: TODO: remove defaults

skonto:
@markusthoemmes gentle ping.

skonto:
CSV fields are meant to be idempotent (meant to be metadata shipped together with the operator pod), however we can configure env vars via the subscription as defined [here](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/subscription-config.md#configuring-operators-deployed-by-olm). I already tried this via the ODC and it works as expected and we can override any env var defined in the S-O csv.

knative-operator/pkg/monitoring/sources/source_deployment_discovery_controller.go
```diff
@@ -69,38 +77,94 @@ type ReconcileSourceDeployment struct {
...
-type skipDeletePredicate struct {
-	predicate.Funcs
+	shouldEnableMonitoring := okomon.ShouldEnableMonitoring(eventing.Spec.GetConfig())
+	return reconcile.Result{}, r.reconcileSourceMonitoring(r.client, dep, shouldEnableMonitoring, inDeletion)
 }
 
-func (skipDeletePredicate) Delete(e event.DeleteEvent) bool {
-	return false
+func (r *ReconcileSourceDeployment) reconcileSourceMonitoring(rclient client.Client, dep *v1.Deployment, shouldEnableMonitoring bool, inDeletion bool) error {
+	// Setup monitoring resources only if monitoring should be on and source deployment is not being deleted
```
devguyio, line 109: I'm wondering if we should extract some logic out of this method into other methods to explain the logic more and not rely on the detailed comments. Specially the the two branches of the big if statements? don't have a very strong opinion, just thinking out loud

devguyio (commented):
left a few small comments, other than that it's lgtm for me. I didn't do a deep of the tests because of the time and limited capacity but that should be ok.
/lgtm

knative-operator/pkg/monitoring/sources/source_deployment_discovery_controller.go
skonto: Let me think about it.

knative-operator/pkg/monitoring/sources/source_deployment_discovery_controller.go
```diff
@@ -69,38 +77,94 @@ type ReconcileSourceDeployment struct {
...
-	if err := r.client.Get(context.TODO(), request.NamespacedName, dep); err != nil {
-		// the deployment does not exist anymore do nothing
+	err := r.client.Get(context.TODO(), request.NamespacedName, dep)
+	inDeletion := false
+	if apierrors.IsNotFound(err) {
+		// The deployment does not exist anymore, deletions are shown as failing reads
+		log.Info("Source in deletion phase")
+		inDeletion = true
+	} else if err != nil {
+		// Do nothing if can't get the source for some other error
```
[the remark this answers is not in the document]
skonto: I think we should catch the error. I will update.

skonto:
@devguyio thanks for the review, I addressed the comments. If you have time to test manually then you can set env vars like `USE_CLUSTER_MONITORING_FOR_SOURCES_BY_DEFAULT` at the subscritpion level. 
@rhuss pls review.

skonto:
@devguyio gentle ping

knative-operator/cmd/manager/main.go
```diff
@@ -118,7 +118,7 @@ func main() {
 	hookServer.Register("/mutate-knativekafkas", &webhook.Admission{Handler: knativekafka.NewConfigurator(decoder)})
 	hookServer.Register("/validate-knativekafkas", &webhook.Admission{Handler: knativekafka.NewValidator(mgr.GetClient(), decoder)})
 
-	if err := setupMonitoring(cfg); err != nil {
+	if err := setupServerlesOperatorMonitoring(cfg); err != nil {
```
rhuss, line 121:
```suggestion
	if err := setupServerlessOperatorMonitoring(cfg); err != nil {
```

and the corresponding definition

rhuss: Why is "ServerlessOperator" needed here as extra context information ? Is it not clear that this is about the setting up monitoring for Knative/Serverless ?

knative-operator/pkg/monitoring/sources/source_deployment_discovery_controller.go
```diff
@@ -2,23 +2,31 @@ package sources
...
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
-	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
-var log = common.Log.WithName("source-deployment-discovery-controller")
+var (
+	generateSourceServiceMonitorsEnvVar = "GENERATE_SERVICE_MONITORS_FOR_SOURCES_BY_DEFAULT"
```
rhuss, line 25: Why the extension `BY_DEFAULT` ? How can this be overridden ? (e.g. a user sets this variable, but for certain sources it does not want a ServiceMonitor to be created ?

If this variable triggers to _always_ create a ServiceMonitor for a Source, then I would ommit the `BY_DEFAULT` suffix.

knative-operator/pkg/monitoring/sources/source_deployment_discovery_controller.go
```diff
@@ -2,23 +2,31 @@ package sources
...
+	useClusterMonitoringEnvVar          = "USE_CLUSTER_MONITORING_FOR_SOURCES_BY_DEFAULT"
```
rhuss, line 26: Similar comment as above. If not needed, ommit the `BY_DEFAULT` suffix.

knative-operator/pkg/monitoring/sources/source_deployment_discovery_controller.go
```diff
@@ -53,7 +61,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		}
 		return nil
 	})
-	err = c.Watch(&source.Kind{Type: &v1.Deployment{}}, handler.EnqueueRequestsFromMapFunc(enqueueRequests), skipDeletePredicate{}, skipUpdatePredicate{})
+	err = c.Watch(&source.Kind{Type: &v1.Deployment{}}, handler.EnqueueRequestsFromMapFunc(enqueueRequests))
```
rhuss, line 64: not really deep into it, but why are the predicates not needed anymore ?

knative-operator/pkg/monitoring/sources/source_deployment_discovery_controller.go
```diff
@@ -69,38 +77,119 @@ type ReconcileSourceDeployment struct {
...
 	reqLogger.Info("Reconciling the source deployment, setting up a service/service monitor if required")
 	dep := &v1.Deployment{}
-	if err := r.client.Get(context.TODO(), request.NamespacedName, dep); err != nil {
-		// the deployment does not exist anymore do nothing
-		return reconcile.Result{}, nil
-	}
-	if err := monitoring.SetupMonitoringRequirements(r.client, dep); err != nil {
+	err := r.client.Get(context.TODO(), request.NamespacedName, dep)
+	inDeletion := false
+	if apierrors.IsNotFound(err) {
```
rhuss, line 92: Is 'in deletion' the only state that can cause a 404 ?

knative-operator/pkg/monitoring/sources/source_deployment_discovery_controller.go
rhuss: Another recommendation. If you start the env var name with "SOURCES" then its much clearer that those env vars belong together. Also, I wonder whether you do need two env vars.

E.g. you could differentiate on the value

```
SOURCES_SERVICE_MONITOR=cluster
SOURCES_SERVICE_MONITOR=generate
```

That makes the configurations surface smaller and less to remember.

rhuss (commented):
@skonto quick review on the UI surface, not deep enough into the weeds to understand the full thing (and the original requirement). Sorry, still time constrained right now.

knative-operator/pkg/monitoring/sources/source_deployment_discovery_controller.go
skonto: These are different settings. They affect different functionality pls take a look at the description.

knative-operator/pkg/monitoring/sources/source_deployment_discovery_controller.go
skonto: This can be overridden via a subscription update check comments above. I can ommit BY_DEFAULT.

knative-operator/cmd/manager/main.go
skonto: I added this for clarity because there are other funcs with similar functionality.

knative-operator/pkg/monitoring/sources/source_deployment_discovery_controller.go
skonto: Afaik this us typical, the first step in reconciliation is to retrieve the object. Either the call fails for some error, object does not exist (due to deletion) or it exists and then you can update its state etc.

knative-operator/pkg/monitoring/sources/source_deployment_discovery_controller.go
skonto: Update is needed when deployments are updated due to changes in the observability config in the eventing CR for example which maps to changes in the deployments env vars. Delete can be safely ignored.

skonto:
I will rename the env vars.

skonto:
@rhuss ready for another round.

skonto:
Validation passes locally not sure why yq fails.

skonto:
@rhuss gentle ping. Pls take a look.

skonto:
@devguyio @rhuss gentle ping

rhuss:
yes, sorry. quite busy and many folks are/were on PTO. Hope to continue it still before the weekend, but don't consider me to be a blocker.

skonto:
@rhuss we ce can continue on Monday @markusthoemmes will be back, but I think you are the only one who can approve it right now.

skonto:
@markusthoemmes gentle ping.

knative-operator/pkg/monitoring/sources/source_deployment_discovery_controller.go
```diff
@@ -69,38 +79,133 @@ type ReconcileSourceDeployment struct {
...
 
-type skipUpdatePredicate struct {
+// Remove any Source service monitor resources in any user namespace
+func (r *ReconcileSourceDeployment) cleanUpSourceMonitoringResources(dep *v1.Deployment, inDeletion bool) error {
+	if dep.Namespace != "knative-eventing" {
+		if err := monitoring.RemoveClusterMonitoringRequirements(r.client, dep); err != nil {
+			return err
+		}
+		// No need to do anything if in deletion phase as owner references will do the cleanup
+		if !inDeletion {
```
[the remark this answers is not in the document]
markusthoemmes: Ahhhh, you distinguish between "disable monitoring" and "removing the deployment". Is that distinction necessary? I think we can safely drop the `inDeletion` state and handle both cases the same. It's not harmful to actively delete things even if the ownerRef would technically take care of it.

templates/csv.yaml
```diff
@@ -467,8 +467,10 @@ spec:
                       value: "deploy/resources/quickstart/serverless-application-quickstart.yaml"
                     - name: DASHBOARDS_ROOT_MANIFEST_PATH
                       value: "deploy/resources/dashboards"
-      
-      # Openshift Ingress adapter for Knative's Ingress implementation.
```
markusthoemmes, line 471: Please don't remove this.

knative-operator/pkg/monitoring/sources/source_deployment_discovery_controller.go
markusthoemmes: From below: I think I follow this now. I feel like it's unnecessary to distinguish between "inDeletion" and "disabling monitoring" though. The code paths would likely benefit from handling both cases the same, I believe.

knative-operator/pkg/monitoring/sources/source_deployment_discovery_controller.go
```diff
@@ -69,38 +79,133 @@ type ReconcileSourceDeployment struct {
...
 		return reconcile.Result{}, err
 	}
-	if err := SetupSourceServiceMonitorResources(r.client, dep); err != nil {
+	eventing := &v1alpha1.KnativeEventing{}
+	if err = r.client.Get(context.TODO(), types.NamespacedName{Namespace: "knative-eventing", Name: "knative-eventing"}, eventing); err != nil {
 		return reconcile.Result{}, err
 	}
-	return reconcile.Result{}, nil
+	shouldEnableMonitoring := okomon.ShouldEnableMonitoring(eventing.Spec.GetConfig())
+	return reconcile.Result{}, r.reconcileSourceMonitoring(dep, shouldEnableMonitoring, inDeletion)
```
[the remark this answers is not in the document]
skonto: If someone disables monitoring eg. set backend to `none` this triggers an update to all source adapter deployments.

knative-operator/pkg/monitoring/sources/source_deployment_discovery_controller.go
```diff
@@ -69,38 +79,133 @@ type ReconcileSourceDeployment struct {
...
 }
 
-type skipDeletePredicate struct {
-	predicate.Funcs
+func (r *ReconcileSourceDeployment) reconcileSourceMonitoring(dep *v1.Deployment, shouldEnableMonitoring bool, inDeletion bool) error {
+	// If monitoring is set to on/off this triggers a global resync to source adapters.
```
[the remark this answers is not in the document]
skonto: Yes.

knative-operator/pkg/monitoring/sources/source_deployment_discovery_controller.go
```diff
@@ -69,38 +79,133 @@ type ReconcileSourceDeployment struct {
...
 		return reconcile.Result{}, err
 	}
-	return reconcile.Result{}, nil
+	shouldEnableMonitoring := okomon.ShouldEnableMonitoring(eventing.Spec.GetConfig())
+	return reconcile.Result{}, r.reconcileSourceMonitoring(dep, shouldEnableMonitoring, inDeletion)
 }
 
-type skipDeletePredicate struct {
-	predicate.Funcs
+func (r *ReconcileSourceDeployment) reconcileSourceMonitoring(dep *v1.Deployment, shouldEnableMonitoring bool, inDeletion bool) error {
```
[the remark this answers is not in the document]
skonto: I changed it because previous review suggested to separate.

templates/csv.yaml
skonto: ok probably removed by accident will bring it back.

knative-operator/pkg/monitoring/sources/source_deployment_discovery_controller.go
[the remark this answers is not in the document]
skonto: If I get an update event for a source deployment and monitoring is enabled then I want to avoid setting up the source resources (if in the meanwhile someone deleted the source). Note that when I am saying source resources it is not just the service monitors that have owner refs set to the source but also global rbac rules for the source namespace to be used by Prometheus sa that lives in the openshift-monitoring ns. If I setup source monitoring resources and the source is in deletion phase rbac rules will be left undeleted if that is the only source in that namespace. 
That is why I keep the state there because I have the delete predicate that ignores deletes. I can change that to handle deletions but the logic so far was only to handle updates.

knative-operator/pkg/monitoring/sources/source_deployment_discovery_controller.go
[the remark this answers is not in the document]
skonto: > Shouldn't inDeletion be triggered by a finalizer being present? If we need to cleanup arbitrary (non ownerRef) stuff from the Deployment, we should be adding a finalizer to it. That would cause this to no longer work though, as it'd stick around until the finalizer has been removed by us.

If we were to manage the non ownerRefresources like Prometheus rbac rules yes. Currently if someone accidentally deletes them then I dont restore them unless some source update happens. Otherwise will have to setup a watcher/reconciler for any different kind involved.

knative-operator/pkg/monitoring/sources/source_deployment_discovery_controller.go
```diff
@@ -69,38 +79,133 @@ type ReconcileSourceDeployment struct {
 	// that reads objects from the cache and writes to the apiserver
 	client client.Client
 	scheme *runtime.Scheme
+
+	testShouldUseClusterMonitoring          *bool
+	testShouldGenerateSourceServiceMonitors *bool
```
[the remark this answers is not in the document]
skonto: ok I can set/unset the env vars in tests as an alternative.

skonto:
@markusthoemmes I refactored all major issues. I will address the last pending issue which is rbac resource reconciling (right now they are not attached to some owner) in a separate PR because it requires some thought. However, this PR is an improvement in many ways.

matzew (approved):
/lgtm
/approve

skonto:
Checking if we constantly trigger a reconciliation.

skonto:
@matzew or @rhuss  could you approve once more?

skonto:
@matzew gentle ping.

matzew (approved):
/lgtm
/approve

PR #1103, opened 2021-07-14
Serve CLI artifacts from operator

markusthoemmes:
Gets rid of the "unnecessary" (very very very seldomly used) CLI artifact container in favor of serving the respective artifacts from a server launched by the operator itself. The artifacts are copied in via using the existing artifact image as an init container so no new images or builds necessary.

**Note:** Please doublecheck this for security. We don't want an attacker stealing files from the operator's filesystem (i.e. serviceaccount tokens) through a path based attack. We might decide that the additional safety is actually worth the extra container, maybe. Sending this anyway, as I wanted to know how much work it is.

/assign @rhuss @dsimansk

### DIFF
--- a/knative-operator/cmd/manager/main.go
+++ b/knative-operator/cmd/manager/main.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"net/http"
 	"os"
 
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/apis"
@@ -123,6 +124,16 @@ func main() {
 	}
 
 	log.Info("Starting the Cmd.")
+
+	go func() {
+		// This web server is unimportant enough to not bother connecting its lifecycle to
+		// signal handling so the process can just tear it down with it.
+		log.Info("Serving CLI artifacts on :8080")
+		http.Handle("/", http.FileServer(http.Dir("/cli-artifacts")))
+		if err := http.ListenAndServe(":8080", nil); err != nil && err != http.ErrServerClosed {
+			log.Error(err, "Failed to launch CLI artifact server")
+		}
+	}()
 
 	// Start the Cmd
 	if err := mgr.Start(signals.SetupSignalHandler()); err != nil {
--- a/knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
+++ b/knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
@@ -194,19 +194,19 @@ func populateKnConsoleCLIDownload(baseURL string, instance *servingv1alpha1.Knat
 			Description: "The OpenShift Serverless client `kn` is a CLI tool that allows you to fully manage OpenShift Serverless Serving and Eventing resources without writing a single line of YAML.",
 			Links: []consolev1.CLIDownloadLink{{
 				Text: "Download kn for Linux for x86_64",
-				Href: baseURL + "/amd64/linux/kn-linux-amd64.tar.gz",
+				Href: baseURL + "/kn-linux-amd64.tar.gz",
 			}, {
 				Text: "Download kn for Linux for IBM Power little endian",
-				Href: baseURL + "/ppc64le/linux/kn-linux-ppc64le.tar.gz",
+				Href: baseURL + "/kn-linux-ppc64le.tar.gz",
 			}, {
 				Text: "Download kn for Linux for IBM Z",
-				Href: baseURL + "/s390x/linux/kn-linux-s390x.tar.gz",
+				Href: baseURL + "/kn-linux-s390x.tar.gz",
 			}, {
 				Text: "Download kn for macOS",
-				Href: baseURL + "/amd64/macos/kn-macos-amd64.tar.gz",
+				Href: baseURL + "/kn-macos-amd64.tar.gz",
 			}, {
 				Text: "Download kn for Windows",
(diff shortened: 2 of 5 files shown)


knative-operator/cmd/manager/main.go
```diff
@@ -124,6 +125,16 @@ func main() {
 
 	log.Info("Starting the Cmd.")
 
+	go func() {
+		// This web server is unimportant enough to not bother connecting its lifecycle to
+		// signal handling so the process can just tear it down with it.
+		log.Info("Serving CLI artifacts on :8080")
+		http.Handle("/", http.FileServer(http.Dir("/cli-artifacts")))
```
skonto, line 132: Should we use https?

knative-operator/cmd/manager/main.go
```diff
@@ -124,6 +125,16 @@ func main() {
...
+		if err := http.ListenAndServe(":8080", nil); err != nil && err != http.ErrServerClosed {
```
skonto, line 133: Should we protect this with rbac-proxy?

knative-operator/cmd/manager/main.go
markusthoemmes: Why would we? The external route is secured appropriately and the traffic is not worth to secure any further, imo. We can doublecheck what the other hosted CLIs do though.

knative-operator/cmd/manager/main.go
markusthoemmes: See above, I don't think any auth is necessary here.

rhuss:
For me that looks good and simple enough for being included, even when sharing the service account (but I'm not a security expert).

My minor concerns are:

* The init-container copy would increase the operator's start-up time. It's about copying  ~ 100-200 MB (more could come when we support more archs, like darwin-arm64). Is this something that concerns us ? Are we in danger to hit any quotas on local filesystems ? (eg `ephemeral-storage` for a namespace's resource quotas)
* Can this be an entrypoint for a DoS attack that brings down the controller ? Could this be prevented with the alternate setup when running in a separate container ?

For me since it looks like that both approaches are relatively easy to implement, I wonder whether we should go for this one (which also looks simpler), and, when an issue arises (e.g because of a security audit), might switch over to the separate side-car container ?

rhuss:
Another question to understand the data path: The CLI Download CR is used to allow and find the OCP Console the CLI to download, but the endpoints mentioned in there are not routed to the outside of the cluster but are proxied by the OCP Console, so that only the OCP Console accessed the URL that are connected, right ? Would it make sense (and is this even possible ?) to only allow the service acount of the OCP Console deployment to access this endpoint?

markusthoemmes:
> The init-container copy would increase the operator's start-up time. It's about copying ~ 100-200 MB (more could come when we support more archs, like darwin-arm64). Is this something that concerns us ?

Not at all. We don't care about a second or two in startup latency here.

> Are we in danger to hit any quotas on local filesystems ? (eg ephemeral-storage for a namespace's resource quotas)

I don't actually know. Seeing as this is in the `openshift-serverless` namespace though, I would hope that we're fine here.

>Can this be an entrypoint for a DoS attack that brings down the controller ? Could this be prevented with the alternate setup when running in a separate container ?

Potentially, yes. The alternate setup is safer as the processes are strictly divided and thus CPU usage of the server does not directly affect usage of the controller.

> The CLI Download CR is used to allow and find the OCP Console the CLI to download, but the endpoints mentioned in there are not routed to the outside of the cluster but are proxied by the OCP Console, so that only the OCP Console accessed the URL that are connected, right ?

No, there's no proxying going on. The routing is done via an OCP route so the endpoints are accessible externally by default (as it was before).

> Would it make sense (and is this even possible ?) to only allow the service acount of the OCP Console deployment to access this endpoint?

We'd have to check. In theory this **is** possible, provided that we know which ServiceAccounts to allow in the deployment and that the console is actively adding the respective Bearer tokens to the requests. I dunno if either is true.

markusthoemmes:
(the upgrade test failure is due to docker rate limiting. Cherry picking my fixes to 1.16)

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
```diff
@@ -194,19 +194,19 @@ func populateKnConsoleCLIDownload(baseURL string, instance *servingv1alpha1.Knat
 			Description: "The OpenShift Serverless client `kn` is a CLI tool that allows you to fully manage OpenShift Serverless Serving and Eventing resources without writing a single line of YAML.",
 			Links: []consolev1.CLIDownloadLink{{
 				Text: "Download kn for Linux for x86_64",
-				Href: baseURL + "/amd64/linux/kn-linux-amd64.tar.gz",
+				Href: baseURL + "/kn-linux-amd64.tar.gz",
```
dsimansk, line 197: For the backward compatibility, e.g. scripts relying on the URL format. Should we try to replicate the former format. Or it doesn't matter much since base is changing as well?

markusthoemmes: We certainly could. You think it's worth it?

dsimansk: Well, the URL was changed previously as well moving from `Deployment` to Ksvc. If I'm not mistaken there weren't any serious objections. I suppose it's fine as now as well.

markusthoemmes:
Decided against this for the potential security implications for now.

PR #1105, opened 2021-07-14
Explicitly set controller deployment name to distinguish Serving and Kourier controller image

nak3:
Upstream changed Kourier controller's container name to `controller`.

It is duplicated with Serving controller and so we need to explicitly
set the `<deploy_name>`/`<container_name>` to override the image.

As the net-kourier deployment name is not changed in the downstream yet,
this PR changes only serving controller.
(It works with nigthly test in https://github.com/openshift/knative-serving)

/cc @markusthoemmes @mgencur


markusthoemmes (approved):
/lgtm
/approve

PR #1108, opened 2021-07-15
Use prettier to consistently format the CSV

markusthoemmes:
As per title. This makes it slightly less annoying to work on the CSV as it's clear how stuff should be indented etc.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -131,10 +131,11 @@ generated-files: release-files
 	(cd olm-catalog/serverless-operator && ./hack/update-manifests.sh)
 	./hack/update-deps.sh
 
-# Runs the lints Github Actions do too.
+# Runs the lints Github Actions do too, applies fixes where possible.
 lint:
 	woke
 	golangci-lint run
 	find . -type f -path './**/*.*sh' -not -path '*vendor*' | xargs -r shellcheck
 	operator-sdk bundle validate ./olm-catalog/serverless-operator
 	git ls-files | grep -Ev '^(vendor/|.git)' | xargs misspell -error
+	prettier --write templates/*.yaml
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -131,10 +131,11 @@ generated-files: release-files
...
 
-# Runs the lints Github Actions do too.
+# Runs the lints Github Actions do too, applies fixes where possible.
 lint:
 	woke
 	golangci-lint run
 	find . -type f -path './**/*.*sh' -not -path '*vendor*' | xargs -r shellcheck
 	operator-sdk bundle validate ./olm-catalog/serverless-operator
 	git ls-files | grep -Ev '^(vendor/|.git)' | xargs misspell -error
+	prettier --write templates/*.yaml
```
rhuss, line 141: This shifts the semantics of the `lint:` target a bit, right ? Instead of only validating it also tries to fix it in the YAML case.

My concern here is: I don't think `prettier` will fail if the YAML is not is correct format but just fixes it. Which means that afterwards you have to check whether a file was modified so that it can be committed (and so that the GitHub action does not fail). This can be easily forgotten IMO.

What's about either moving the formatting targets to a new target or, alternatively, at least write out here a warning that formatting changed a file (checking with `git status`) and that this changes need to be committed & pushed ?

markusthoemmes: Fair enough, yeah I did adjust the comment above wrt. that. I'll change accordingly.

rhuss: thanks !

markusthoemmes: Done.

markusthoemmes:

infra.

rhuss (commented):
/approve
/lgtm

PR #1110, opened 2021-07-15
Reorder fields in the CSV to group better

markusthoemmes:
This is supposed to be a noop change (modulo removing the stray label from knative-openshift) that reorders fields in the CSV and adds a few comments to make it smoother to work on the CSV.

Specifically, this reorders the fields of the containers to all be in the same order to make it very easy to glance over them to find out where exactly one is looking right now.


matzew (approved):
/lgtm
/approve

PR #1111, opened 2021-07-15
[SRVCOM-1432] Drop unnecessary permissions

markusthoemmes:
This is the "easy" part of SRVCOM-1432:

1. Drops the completely unnecessary `permissions` block. It specifies a Role (vs. a ClusterRole) and since our operators operate cluster-wide, we don't benefit from that in any way right now. We might as well just completely drop it.
2. Reduces the permissions for the ingress pod to the bare minimum. I have no clue how all of the stuff got in there in the first place (especially the KnativeServing and KnativeKafka stuff?) but :shrug:.

Next step will be actually breaking up the big `*/*/*` clusterpermission the rest of the operator deployments have.

### DIFF
--- a/olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
+++ b/olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
@@ -224,84 +224,28 @@ spec:
           serviceAccountName: knative-operator
         - rules:
             - apiGroups:
-                - ""
+                - ''
               resources:
-                - pods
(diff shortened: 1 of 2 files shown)


skonto:
Does this affect monitoring stuff? Do we still have access to openshift-config-managed ns for example, do dashboards get installed?

markusthoemmes:
@skonto nope, the changes are not supposed to impact any functionality. The aim is to be more explicit with the required permissions.

markusthoemmes:

Infra

PR #1113, opened 2021-07-21
Add yq to local bin

skonto:
Since a few days ago (6), yq is being installed by default on [ubuntu latest vm](https://github.com/actions/virtual-environments/blob/main/images/linux/ubuntu2004.json). Discussion [here](https://github.com/actions/virtual-environments/pull/3646/files#r673836709). This causes S-O actions to [fail](https://github.com/openshift-knative/serverless-operator/pull/1098/checks?check_run_id=3121938547). Reason is yq is installed under `/usr/local/bin` in vm and under `/usr/bin` in our setup. The former takes precedence in PATH, thus the error.


nak3 (approved):
/lgtm

PR #1117, opened 2021-07-26
[SRVCOM-1432] Drop namespace local permissions

markusthoemmes:
Drops the completely unnecessary permissions block. It specifies a Role (vs. a ClusterRole) and since our operators operate cluster-wide, we don't benefit from that in any way right now. We might as well just completely drop it.

(I fumbled somewhere as I thought I had this in #1111 already)


markusthoemmes:
This error makes no sense, let's see if it's actually related

matzew (approved):
/lgtm
/approve

PR #1121, opened 2021-07-28
Serve CLI artifacts from operator

markusthoemmes:
Reviving this as it turns out that the python based server in the "other" way of doing this currently unnecessarily requires read access to the file system, which I'd like to ditch from all of our deployments.

We've discussed the security implications of this in Slack quite a bit and @rhuss and @skonto made good points regarding this actually being "more secure" (or rather less surface) in that we use the same technology (Golang) to serve the CLI and to develop all of our other components. That means, that we wouldn't have to watch for potential python CVEs to get fixed too, which is nice!

/assign @rhuss

### DIFF
--- a/knative-operator/cmd/manager/main.go
+++ b/knative-operator/cmd/manager/main.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"net/http"
 	"os"
 
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/apis"
@@ -123,6 +124,16 @@ func main() {
 	}
 
 	log.Info("Starting the Cmd.")
+
+	go func() {
+		// This web server is unimportant enough to not bother connecting its lifecycle to
+		// signal handling so the process can just tear it down with it.
+		log.Info("Serving CLI artifacts on :8080")
+		http.Handle("/", http.FileServer(http.Dir("/cli-artifacts")))
+		if err := http.ListenAndServe(":8080", nil); err != nil && err != http.ErrServerClosed {
+			log.Error(err, "Failed to launch CLI artifact server")
+		}
+	}()
 
 	// Start the Cmd
 	if err := mgr.Start(signals.SetupSignalHandler()); err != nil {
--- a/olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
+++ b/olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
@@ -490,13 +490,15 @@ spec:
                   name: knative-openshift
               spec:
                 serviceAccountName: knative-operator
-                containers:
-                  - name: cli-downloads
+                initContainers:
+                  - name: cli-artifacts
                     image: registry.ci.openshift.org/openshift/knative-v0.22.0:kn-cli-artifacts
                     imagePullPolicy: Always
-                    ports:
-                      - name: http-cli
-                        containerPort: 8080
+                    command: ["sh", "-c", "cp /usr/share/kn/**/* /cli-artifacts && chmod 444 /cli-artifacts/*"]
+                    volumeMounts:
+                      - mountPath: /cli-artifacts
+                        name: cli-artifacts
+                containers:
                   - name: knative-openshift
                     # This reference will be replaced in local builds and CI via hack/lib/catalogsource.bash.
                     image: registry.ci.openshift.org/openshift/openshift-serverless-nightly:knative-operator
@@ -509,6 +511,12 @@ spec:
                       httpGet:
                         path: /healthz
                         port: 8687
+                    ports:
+                      - name: http-cli
+                        containerPort: 8080
+                    volumeMounts:
+                      - mountPath: /cli-artifacts
+                        name: cli-artifacts
                     env:
                       - name: WATCH_NAMESPACE
                         value: ""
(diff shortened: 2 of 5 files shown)


olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
```diff
@@ -490,13 +490,15 @@ spec:
...
-                containers:
-                  - name: cli-downloads
+                initContainers:
+                  - name: cli-artifacts
                     image: registry.ci.openshift.org/openshift/knative-v0.22.0:kn-cli-artifacts
                     imagePullPolicy: Always
-                    ports:
-                      - name: http-cli
-                        containerPort: 8080
+                    command: ['sh', '-c', 'cp /usr/share/kn/**/* /cli-artifacts']
```
rhuss, line 497: Do we want to mangle with permissions here? So we could make it a-x to avoid being executable on the server (which would work for the Linux binary anyway only, but when you download it, I guess it will end up with 0644 on the downloaders side anyway). But I'm overcautious here probably.

rhuss, line 497:
```suggestion
                    command: ['sh', '-c', 'cp /usr/share/kn/**/* /cli-artifacts && chmod -R a-x /cli-artifacts']
```

knative-operator/cmd/manager/main.go
```diff
@@ -124,6 +125,16 @@ func main() {
...
+	go func() {
+		// This web server is unimportant enough to not bother connecting its lifecycle to
+		// signal handling so the process can just tear it down with it.
+		log.Info("Serving CLI artifacts on :8080")
+		http.Handle("/", http.FileServer(http.Dir("/cli-artifacts")))
+		if err := http.ListenAndServe(":8080", nil); err != nil && err != http.ErrServerClosed {
+			log.Error(err, "Failed to launch CLI artifact server")
+		}
+	}()
+
```
rhuss, line 137: Do we have an easy way to E2E test the download ? I think it should not be that hard to add this to the test, just download some or all of the binaries.

rhuss (commented):
Looks good, however I would add a simple E2E test to check the download via the route.

knative-operator/cmd/manager/main.go
markusthoemmes: We have that (though just doing HEAD requests and making sure the binary size is plausible)

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
markusthoemmes: Done

markusthoemmes: Fair enough, done!

knative-operator/cmd/manager/main.go
markusthoemmes: See https://github.com/openshift-knative/serverless-operator/blob/main/test/servinge2e/verify_deploy_resources_test.go for posterity

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
rhuss: While we are at, I wonder whether we should also do a `a-w`, just to avoid that someone can inject a trojan binary by some exploit. If someone can jump directly into the container, that is not much helpful, but more like leveraging some other potential loophole that just tries to overwrite such file (without being able to do a +w on it) ?

templates/csv.yaml
```diff
@@ -447,13 +447,15 @@ spec:
...
-              containers:
-                - name: cli-downloads
+              initContainers:
+                - name: cli-artifacts
                   image: TO_BE_REPLACED
                   imagePullPolicy: Always
-                  ports:
-                    - name: http-cli
-                      containerPort: 8080
+                  command: ['sh', '-c', 'cp /usr/share/kn/**/* /cli-artifacts && chmod -R a-x /cli-artifacts']
```
rhuss, line 454:
```suggestion
                  command: ['sh', '-c', 'cp /usr/share/kn/**/* /cli-artifacts && chmod -R a-xw /cli-artifacts']
```

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
```diff
@@ -490,13 +490,15 @@ spec:
...
-                containers:
-                  - name: cli-downloads
+                initContainers:
+                  - name: cli-artifacts
                     image: registry.ci.openshift.org/openshift/knative-v0.22.0:kn-cli-artifacts
                     imagePullPolicy: Always
-                    ports:
-                      - name: http-cli
-                        containerPort: 8080
+                    command: ['sh', '-c', 'cp /usr/share/kn/**/* /cli-artifacts && chmod -R a-x /cli-artifacts']
```
rhuss, line 497:
```suggestion
                    command: ['sh', '-c', 'cp /usr/share/kn/**/* /cli-artifacts && chmod -R a-wx /cli-artifacts']
```

markusthoemmes: Changed to 444 per discussion in Slack.

rhuss (commented):
/approve
/lgtm

Thanks !

markusthoemmes:
Rebased to pick up Kenjiro's fix.

markusthoemmes:

Unrelated.

rhuss (commented):
/approve
/lgtm

PR #1115, opened 2021-07-22
[WIP] mgencur: Add eventings 0.23 2nd try extras

mgencur:
Trying with increased logging.

PR #1127, opened 2021-08-02
Enable DEBUG log by default for Serving and Eventing components in tests

mgencur:
We were enabling debug logs to get extra logging in https://github.com/openshift-knative/serverless-operator/pull/1115 
However, it would be good to have the DEBUG logs enabled by default as it helps with debugging and doesn't need re-running CI. This approach is used upstream as well.


matzew (approved):
/lgtm
/approve

PR #1144, opened 2021-08-11
Remove kafka-ch-disptacher from knativekafka manifests.

devguyio:
Now the KafkaChannel controller creates the dispatcher deployment and service. 

This is a followup to the upstream changes knative-sandbox/eventing-kafka#798


mgencur:
@devguyio we vendor the forks here: https://github.com/openshift-knative/serverless-operator/blob/main/go.mod#L44 . 
Where would the patch be? You can backport the upstream fox to downstream openshift-knative/eventing-kafka (the respective branch) and then update this go.mod file to vendor the files properly.

devguyio:
@mgencur thanks! for now we don't need to update the vendored eventing in this fix. Manifests and images only.

matzew:
/retest

I can now do a `docker pull` against the 0.23.5

cardil:
I assume this should target Serverless 1.17, right?

matzew:
/test 4.8-upgrade-tests-aws-ocp-48

double checking...

matzew:
By just deplying this, I do see that we do not have `kafka-ch-dispatcher-sm` Monitor.  :+1: 

 But... might there be one hanging around, due to upgrade test, @cardil ?

matzew:
@devguyio you need to add the patch here:
https://github.com/devguyio/serverless-operator/blob/f0ca008a38030e8813c9ee409a5e827a91c6b98d/knative-operator/hack/update-manifests.sh

and once it is added, run the `make generated-files`   (that will than actually "patch" the embedded manifest)

knative-operator/hack/update-manifests.sh
```diff
@@ -37,8 +37,9 @@ function download_kafka {
...
 
-# Change the minavailable pdb for kafka-webhook to 1
-#git apply "$root/knative-operator/hack/007-eventing-kafka-pdb.patch"
+# Apply the patches
 
-# For 1.17 we still skip HPA/PDB
-git apply "$root/knative-operator/hack/001-remove_hpa_pdb.patch"
+for p in `ls $root/knative-operator/hack/*.patch|sort -n`
+do
+  git apply $p
```
matzew, line 44: let's please not do this. thet;s keep them separate, and link to issues.

This was inherited from this hackery: 
https://github.com/openshift-knative/serverless-operator/blob/main/openshift-knative-operator/hack/update-manifests.sh#L78-L86

that way, we do have context!

devguyio: I did move the comment inside the diff file itself but sure, it's a separate issue

knative-operator/hack/update-manifests.sh
```diff
@@ -37,8 +37,9 @@ function download_kafka {
...
+done
```
markusthoemmes, line 45: I believe this is equivalent to `git apply $root/knative-operator/hack/*.patch` (see https://serverfault.com/a/122743 for globs generating a sorted list)

matzew: I still like having explicit patch applies on our repo here, since that way we have a bit more context (expect for git blame)

lberk: I believe this is, yes, but applying the patches in this way can give a clearer picture of which patch specifically failed, no?

knative-operator/hack/update-manifests.sh
matzew: yeah, same w/ renaming the patch files. LEt's do that if we want to afterwards

knative-operator/hack/update-manifests.sh
devguyio: I switched back, I wanna get this PR out :D

matzew:
upgrade test now passed! Great!

/lgtm
/approve

/hold
(holding to ensure the upstream e2e is golden too)

cardil:
I'd like to run upgrade tests once more to make sure we are truely okay.

/test 4.8-upgrade-tests-aws-ocp-48

cardil:
All tests look good

/unhold

mgencur:
Suppose we need to create a follow-up JIRA for adding the ServiceMonitor back in a different place.

PR #1146, opened 2021-08-16
[SRVKS-702] Set ClusterIP service type for Kourier Gateway by default

nak3:
The service type of Kourier Gateway is currently `LoadBalancer`, even
though OpenShift Serverless uses OpenShift Route to expose the
serveerless services. Users have to take care about the extra firewall for it.

This patch changes service type of Kourier Gateway to `ClusterIP` if unset.

### DIFF
--- a/openshift-knative-operator/pkg/serving/extension.go
+++ b/openshift-knative-operator/pkg/serving/extension.go
@@ -102,6 +102,9 @@ func (e *extension) Reconcile(ctx context.Context, comp v1alpha1.KComponent) err
 	defaultToKourier(ks)
 	common.ConfigureIfUnset(&ks.Spec.CommonSpec, "network", "ingress.class", defaultIngressClass(ks))
 
+	// Apply Kourier gateway service type.
+	defaultKourierServiceType(ks)
+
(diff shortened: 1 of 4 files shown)


nak3:
/hold

I guess upstream test will fail as gRPC test uses `LoadBalancer` type as a workaround. Let me confirm the result and fix it.

nak3:
/retest

tests do not start running. (And mesh test is infra issue.)

markusthoemmes:
Quick q on this one: Can the type be changed at runtime? IIRC, some servicetypes can't be changed at runtime.

nak3:
Ah, yes. Some AFAIK, these are not allowed: 
- headless to non-headless service
- some type -> ExternalName (because ExternalName requires `spec.externalName`)

So I think we should not allow to set `ExternalName` for Kourier, which is not used in general.

markusthoemmes (approved):
/lgtm
/approve
/hold

Please unhold once the doc issue has been created to make sure users are being made aware.

markusthoemmes:
/unhold

Docs issue was opened.

PR #1147, opened 2021-08-17
Update `test/serving.bash` to use oc apply instead of ytt

nak3:
This patch updates `test/serving.bash` to use oc apply instead of ytt.

Upstream conformance allows conformance test without ytt as https://github.com/knative/serving/commit/69caf075172546ab075a3b63ab7a23aced1eb9d0.
And downstream backport it as https://github.com/openshift/knative-serving/commit/c9430691c62581d82d867be2306b90baf6d33581.

So we should not need `ytt` command anymore.


markusthoemmes (approved):
/lgtm
/approve

/hold
To confirm the upstream test.

nak3:
/hold cancel

upstream test passed.

cardil:
I think this shouldn't this be backported to release-1.17, right?

/cherrypick release-1.17

PR #1162, opened 2021-08-24
Revert [SRVKS-702] Set ClusterIP service type for Kourier Gateway by default

nak3:
This patch reverts https://github.com/openshift-knative/serverless-operator/pull/1146 because
k8s 1.20 has a bug which fails to change the service type.

/cc @markusthoemmes


markusthoemmes (approved):
/lgtm
/approve

PR #1164, opened 2021-08-25
[SRVKS-702] Set ClusterIP service type for Kourier Gateway by default

nak3:
This patch adds https://github.com/openshift-knative/serverless-operator/pull/1146 again.

Due to a bug in kubernetes, service type change from LoadBalancer to ClusterIP does not work before v1.20.0,
so this patch applies the default only when cluster is using k8s v1.20.0+.

### DIFF
--- a/openshift-knative-operator/pkg/serving/extension.go
+++ b/openshift-knative-operator/pkg/serving/extension.go
@@ -102,6 +102,9 @@ func (e *extension) Reconcile(ctx context.Context, comp v1alpha1.KComponent) err
 	defaultToKourier(ks)
 	common.ConfigureIfUnset(&ks.Spec.CommonSpec, "network", "ingress.class", defaultIngressClass(ks))
 
+	// Apply Kourier gateway service type.
+	defaultKourierServiceType(ks)
+
 	// Override the default domainTemplate to use $name-$ns rather than $name.$ns.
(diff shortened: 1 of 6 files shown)


nak3:
Umm... no trigger to test 4.6 :sweat_drops:

openshift-knative-operator/pkg/serving/extension.go
```diff
@@ -162,3 +176,41 @@ func (e *extension) fetchLoggingHost(ctx context.Context) string {
...
+
+	minimumVersion, err := semver.Make(normalizeVersion(version))
+	if err != nil {
+		return err
+	}
+
+	// If no specific pre-release requirement is set, we default to "-0" to always allow
+	// pre-release versions of the same Major.Minor.Patch version.
+	if len(minimumVersion.Pre) == 0 {
+		minimumVersion.Pre = []semver.PRVersion{{VersionNum: 0}}
```
markusthoemmes, line 200: I suppose we should do the IsNum thing we did upstream here as well?

markusthoemmes:
LGTM modulo the `IsNum` thing we did upstream.

openshift-knative-operator/pkg/serving/extension.go
nak3: Thank you! Yes it should be added. Updated.

markusthoemmes (approved):
/lgtm
/approve

PR #1178, opened 2021-08-30
[RELEASE-1.17] Revert "Keep kafkas at one replica (#1132)"

matzew:
This reverts commit e64724db951b4946911ce9a517f91f9e0e7e5074.

Original thought was, we should be going with just one instead. However, looks like the PDBs might cause some regression here.:

> having it at 1 might actually be a regression and cause a problem, because the kafka-webhook PodDisruptionBudget created in serverless 1.16 will still exist after upgrading to 1.17 , so if KnativeKafka is reinstalled after serverless 1.17 upgrade, it might hit the PDB's allowedDisruptions limit (by having a single replica, with the PDB not allowing a single disruption)

Which is sad.... Not sure if we could control that better.

For sanity, we stick w/ the same value we had on the 1.16 release....


matzew:
Or we set the PDB to `minAvailable: 0` ? 

/hold

PR #1181, opened 2021-08-30
[SRVKS-782][UI] Cleanup state before the test execution

cardil:
Following Cypress advise, moving the cleanup code to `beforeEach` block, at the same time ensuring it will skip if no cleaning is needed.


test/ui/cypress/integration/serving.spec.js
```diff
@@ -73,50 +64,66 @@ describe('OCP UI for Serverless', () => {
...
 
   const showcaseKsvc = new ShowcaseKservice()
 
-  it('can deploy kservice and scale it', () => {
+  beforeEach(() => {
     describe('with authenticated via Web Console', cy.login)
+    describe('remove kservice', showcaseKsvc.removeApp)
+  })
+
+  afterEach(() => {
```
mgencur, line 122: You're removing the app both beforeEach and afterEach. Why ?

PR #1185, opened 2021-08-31
Remove kafka dispatcher deployment image reference

matzew:
With the 1.17 branch we no longer ship the manifest for the `kafka-ch-dispatcher`, therefore we do no longer need to replace its' image.

The controller is creating/owning the deployment, and thae dispatcher image is listed there:
 https://github.com/openshift-knative/serverless-operator/blob/main/knative-operator/deploy/resources/knativekafka/1-channel-consolidated.yaml#L881-L882

This image we still replace via: https://github.com/matzew/serverless-operator/blob/ac1e9134ebd07febfd768e8037b4fe328d7eb540/hack/generate/csv.sh#L72

/kind clean-up


markusthoemmes (approved):
/lgtm
/approve

PR #1191, opened 2021-09-06
SRVCOM-1454 Ensure alerts about deprecated API usage are caught

mgencur:
* The default alerts are fired after 60 minutes and it is too long
because our test suite runs more quickly. Setting the interval to 1
minute.

https://issues.redhat.com/browse/SRVCOM-1454


markusthoemmes (approved):
/lgtm
/approve

PR #1192, opened 2021-09-07
SRVCOM-1511 Use topology list URL conditionally

cardil:
The topology URL has change and from 4.9+ old URL isn't longer working. New URL needs to be used - see [SRVCOM-1511](https://issues.redhat.com/browse/SRVCOM-1511) for details.

Also, to accomplish it, a code refactor was done. Now, it's much easier to conditionally write code without Cypress promises, which are hard to chain and not required to resolve OCP version compatibility. Also, it's easier to navigate the code in IDE, because of regular classes.


test/ui/cypress/code/environment.js
```diff
@@ -0,0 +1,28 @@
+import SemverResolver from "./semverResolver";
+
+class Environment {
+  ocpVersion() {
+    if (!Environment.__ocpVersion) {
+      Environment.__ocpVersion = new SemverResolver(Cypress.env('OCP_VERSION'))
+    }
+    return Environment.__ocpVersion
```
mgencur, line 8: It's just a single class initialization plus reading the env variable so I'm wondering if we could simplify this whole function to just `return new SemverResolver(Cypress.env('OCP_VERSION'))`. Saving the value for later might be unnecessary here. After all, in each class you initialize the Environment class once and then mostly call ocpVersion() only 1x.

mgencur (commented):
One minor nit. Otherwise looks good.

test/ui/cypress/code/environment.js
cardil: Yes. We could.

Initially, I thought of momization for this as this was a promise, but moving all that code to simple classes made the code promise'less. So, this is left over from those experiments.

PR #1195, opened 2021-09-09
Update operator CRDs to 0.24

markusthoemmes:
As per title.

@matzew we've talked about dropping the sources field in the past I think? It gets added now so a followup might be in order to drop it.


lberk:
aiui, not related to the eventing bits being upgraded
/test 4.8-upgrade-tests-aws-ocp-48

lberk:
@markusthoemmes given the addition of the source fields, do we have to taken any action elsewhere in the operator to ensure the sources we're not supporting in SO are disabled?

markusthoemmes:
@lberk nope we haven't and I think we might not even want to. The approach we've taken so far is that we've dropped the fields from the CRD "documentation" as to not advertise them. In theory, a "power user" who'd know could still use them though. The reasoning behind that is, that in some cases (like the registry field, which we're dropping) it could come in very handy to use for people, that want to try some configuration we haven't tested but we might want people to still be able to use it as part of a PoC or thelike.

Hence the ping towards Matze (who'm I discussed this with the last time) that y'all might want to file a PR after this one to drop the sources fields if you think we shouldn't support it.

lberk:
Thanks for the info Markus!
I've filed SRVKE-932 to ensure we review that in the correct timeframes
/lgtm

PR #1196, opened 2021-09-13
[RELEASE-1.17] SRVKE-934: skip HPA 

matzew:
This reverts commit 06d8c964aea1de2d5efa248484e7aa73f772b2a2.


openshift-knative-operator/hack/update-manifests.sh
```diff
@@ -107,6 +107,9 @@ git apply "$root/openshift-knative-operator/hack/001-eventing-namespace-deletion
 # TODO: propose to upstream
 git apply "$root/openshift-knative-operator/hack/002-openshift-eventing-role.patch"
 
+# For SRVKE-629 we disable HPA:
```
maschmid, line 110: we should update the comment to refer to SRVKE-914 , as that is the new reason...

matzew: ah, yeah. updated the file

PR #1197, opened 2021-09-13
[SRVCOM-1518]  Drop eventing sources

matzew:
adding patch to remove the entire `sources` field


matzew:
@markusthoemmes can you take alook?  

Since we already have https://github.com/openshift-knative/serverless-operator/blob/35c080444b0c8e0d7b0498c76fc03ae2a338478b/olm-catalog/serverless-operator/manifests/operator_v1alpha1_knativeeventing_crd.yaml#L47  I am "just" removing the `sources` object

markusthoemmes (approved):
/lgtm
/approve

PR #1199, opened 2021-09-13
SRVKE-935 Wait for KnativeKafka to be ready in upgrade tests

mgencur:
https://issues.redhat.com/browse/SRVKE-935

### DIFF
--- a/test/upgrade/installation/serverless.go
+++ b/test/upgrade/installation/serverless.go
@@ -1,7 +1,14 @@
 package installation
 
 import (
+	"context"
+	"fmt"
 	"strings"
+
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/openshift-knative/serverless-operator/test"
 	v1a1test "github.com/openshift-knative/serverless-operator/test/v1alpha1"
@@ -42,6 +49,24 @@ func UpgradeServerless(ctx *test.Context) error {
 		ctx.T.Error("Eventing upgrade failed:", err)
 	}
 
+	// KnativeKafka doesn't provide the version to wait for. Choosing a well-known image and
+	// waiting for all pods to be updated to this image before proceeding.
+	kafkaWebhookImage, err := test.ImageFromCSV(ctx,
+		test.Flags.CSV,
+		test.OperatorsNamespace,
+		"kafka-webhook")
+	if err != nil {
+		ctx.T.Error(err)
+	}
+
+	if err = WaitForPodsWithImage(ctx,
+		knativeEventing,
+		"app=kafka-webhook",
+		"kafka-webhook",
+		kafkaWebhookImage); err != nil {
+		ctx.T.Error(err)
+	}
+
 	if _, err := v1a1test.WaitForKnativeKafkaState(ctx,
 		"knative-kafka",
 		knativeEventing,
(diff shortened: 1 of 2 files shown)


test/upgrade/installation/serverless.go
```diff
@@ -42,5 +42,13 @@ func UpgradeServerless(ctx *test.Context) error {
 		ctx.T.Error("Eventing upgrade failed:", err)
 	}
 
+	if _, err := v1a1test.WaitForKnativeKafkaState(ctx,
+		"knative-kafka",
+		knativeEventing,
+		v1a1test.IsKnativeKafkaReady,
```
markusthoemmes, line 48: Do we need to validate the version here as well, like we do in Serving and Eventing?

cardil: I think we do. The SRVKE-935 is there just because it was still running on the previous version of the KnativeKafka (0.23, instead of 0.24)

mgencur: The KnativeKafka CR's status doesn't have "version" field :-/

mgencur: Trying to see if the KnativeKafka resource was not Ready in those failed runs which would indicate that waiting only for "Ready" would be enough. But I don't see KnativeKafka CR in must-gather logs. Maybe the must-gather image doesn't collect KnativeKafka CR. Will check.
If we want to wait for the version as well we need to add the version handling to the operator.

mgencur: Sent https://github.com/openshift-knative/must-gather/pull/5 to fix must-gather

skonto: There is also an alternative to use here. Just scrape the related metric [here](https://github.com/openshift-knative/serverless-operator/blob/3c0fc76d0c2e68a3b91e27075c9517256df6f22a/knative-operator/pkg/controller/knativekafka/knativekafka_controller.go#L152). We already use this method here: https://github.com/openshift-knative/serverless-operator/blob/3c0fc76d0c2e68a3b91e27075c9517256df6f22a/test/e2ekafka/knative_kafka_test.go#L97

matzew:
@mgencur is this ready? OR did you want to add check for version testing, from CSV?

mgencur:
@matzew I wanted to add the version check but I won't get to it until tomorrow.

test/upgrade/installation/serverless.go
```diff
@@ -52,3 +77,31 @@ func UpgradeServerless(ctx *test.Context) error {
...
+			}
+		}
+		return true, nil
+	}); waitErr != nil {
+		return fmt.Errorf("containers %s in pods with label selector %s do not have the expected image: %w", containerName, podSelector, waitErr)
+	}
+	return nil
+}
+
+func PodWithImage(c *v1alpha1.ClusterServiceVersion, err error) (bool, error) {
```
markusthoemmes: This seems unused.

test/upgrade/installation/serverless.go
```diff
@@ -42,5 +49,51 @@ func UpgradeServerless(ctx *test.Context) error {
...
+			return false, err
+		}
+		for _, pod := range podList.Items {
+			for _, c := range pod.Spec.Containers {
+				if c.Name == containerName && c.Image != expectedImage {
+					return false, nil
+				}
+			}
+		}
+		return true, nil
```
markusthoemmes, line 94: This will be true if the amount of pods is 0. Can that happen?

Also: Do we need to wait for the pods to get ready?

test/upgrade/installation/serverless.go
```diff
@@ -42,5 +49,51 @@ func UpgradeServerless(ctx *test.Context) error {
...
+		ctx.T.Error(err)
+	}
+
+	if err = WaitForPodsWithImage(ctx,
+		knativeEventing,
+		"app=kafka-webhook",
+		"kafka-webhook",
+		kafkaWebhookImage); err != nil {
+		ctx.T.Error(err)
+	}
```
markusthoemmes, line 68: To address the edge-cases mentioned below, should we instead check the deployment to be updated and then rolled out? Rollout can be detected via https://github.com/knative/serving/blob/1b945a0942ef9638c3df9af68a5f91b573ee3aad/test/e2e/autoscale.go#L424-L444. So it'd be

1. Wait for the expected image to appear in the deployment spec
2. Wait for the rollout to finish

test/upgrade/installation/serverless.go
mgencur: Hmm. This is during upgrades so it should never be 0. And if it's 0 then Kafka if not deployed and it can return success, IMO. But I could check if the number of pods is 0 at the beginning.
I don't think we need to wait for the pods to get ready. This should be covered by waiting for KnativeKafka CR.

test/upgrade/installation/serverless.go
mgencur: That's an option too but not sure it's necessary. See my previous comment.

markusthoemmes (approved):
/lgtm
/approve

mgencur:
Thanks! :)

PR #1201, opened 2021-09-15
Setup deprecation alerts only for OCP >= 4.8

mgencur:
Fixes issues like this one: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-knative-serverless-operator-main-4.6-e2e-aws-ocp-46-continuous/1437929380566274048

### DIFF
--- a/test/lib.bash
+++ b/test/lib.bash
@@ -302,7 +302,7 @@ function check_serverless_alerts {
 function setup_quick_api_deprecation_alerts {
   local ocp_version=$(oc get clusterversion version -o jsonpath='{.status.desired.version}')
   # Setup deprecation alerts for OCP >= 4.8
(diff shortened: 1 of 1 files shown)


nak3:
LGTM.
So, v4.7-'s prometheusrules does not have `api-usage`, correct? as per the error log:
```
Error from server (NotFound): prometheusrules.monitoring.coreos.com "api-usage" not found
```

nak3:
/foo

I guess there is no trigger to test 4.7 or before but just in case...

nak3:
No trigger to test for v4.7- but LGTM for the script change.

mgencur:
Thanks. I tried putting several different version in the condition, including 4.7.x and the result was that the alert setup is skipped for anything older than 4.8. We will see in the nightly job...

PR #1208, opened 2021-09-20
Add KnativeKafka version to it's status and assert after upgrade

markusthoemmes:
As per title, we've been having trouble figuring out when Knative Kafka would be upgraded and worked around it not surfacing a version by deep-checking the cluster's state.
This makes KnativeKafka be on-par with serving and eventing by surfacing its version itself so we can wait for that signal.

### DIFF
--- a/hack/generate/csv.sh
+++ b/hack/generate/csv.sh
@@ -95,7 +95,7 @@ function add_related_image {
 EOF
 }
 
-function add_downstream_operator_deployment_image {
+function add_downstream_operator_deployment_env {
   cat << EOF | yq write --inplace --script - "$1"
 - command: update
   path: spec.install.spec.deployments(name==knative-openshift).spec.template.spec.containers(name==knative-openshift).env[+]
@@ -109,7 +109,7 @@ EOF
 # we don't add scope prefixes to image overrides here. We don't have a clash anyway without any scope prefixes!
 # there was a naming clash between eventing and kafka, but we won't provide the Kafka overrides to the
 # midstream operator.
-function add_upstream_operator_deployment_image {
+function add_upstream_operator_deployment_env {
   cat << EOF | yq write --inplace --script - "$1"
 - command: update
   path: spec.install.spec.deployments(name==knative-operator).spec.template.spec.containers(name==knative-operator).env[+]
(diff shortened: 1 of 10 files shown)


hack/generate/csv.sh
```diff
@@ -125,17 +125,20 @@ cp "$template" "$target"
...
 # don't add Kafka image overrides to upstream operator
 for name in "${kafka_images[@]}"; do
   echo "kafka Image: ${name} -> ${kafka_images_addresses[$name]}"
   add_related_image "$target" "KAFKA_IMAGE_${name}" "${kafka_images_addresses[$name]}"
-  add_downstream_operator_deployment_image "$target" "KAFKA_IMAGE_${name}" "${kafka_images_addresses[$name]}"
+  add_downstream_operator_deployment_env "$target" "KAFKA_IMAGE_${name}" "${kafka_images_addresses[$name]}"
 done
 
+# Add Knative Kafka version to the downstream operator
+add_downstream_operator_deployment_env "$target" "KNATIVE_EVENTING_KAFKA_VERSION" "$(metadata.get dependencies.eventing_kafka)"
```
rhuss, line 140: Don't know the function for fetching the metadata, but looks ok for me. (guess it's battle-tested already elsewhere)

rhuss (approved):
Looks good to me and is a nice simple solution (with a little duplication). Ideally one could extract the version from an image's tag of the Kafka operator (so that there is not even a possibility of a conflicting configuration), but since setting the env variable is part of automation that looks robust, it's totally fine for me.

matzew (approved):
/lgtm

test/upgrade/installation/serverless.go
```diff
@@ -47,28 +47,10 @@ func UpgradeServerless(ctx *test.Context) error {
...
-		"kafka-webhook",
-		kafkaWebhookImage); err != nil {
-		return err
-	}
-
 	if _, err := v1a1test.WaitForKnativeKafkaState(ctx,
 		"knative-kafka",
 		knativeEventing,
-		v1a1test.IsKnativeKafkaReady,
+		v1a1test.IsKnativeKafkaWithVersionReady(strings.TrimPrefix(test.Flags.KafkaVersion, "v")),
```
matzew, line 53: nice to see this coming in. :tada:

PR #1217, opened 2021-09-23
Bump operator reference to 0.25.2

markusthoemmes:
Just for completeness. Seems like the CRDs didn't change at all, but the reference should still be correct.

/assign @devguyio


matzew (approved):
/lgtm
/approve

PR #1219, opened 2021-09-23
Update the linters used in our CI to their latest versions

markusthoemmes:
As per title, a bit of a preparation step to potentially get more out of operator-sdk for example.


matzew (approved):
/lgtm
/approve

thanks for doing these updates :tada:

PR #1222, opened 2021-09-27
Explicitly remove obsolete kourier resources after update is done

markusthoemmes:
This moves us onto the new renamed Kourier resources with a bit of operator-internal orchestration to make the rollout smooth and impactless.

1. Removes the "old" Kourier manifest to not auto-delete anything.
2. Explicitly removes the renamed resources **after** the rollout is complete
3. Bends the existing control service to the new pods to allow for a smooth takeover

Once 1.19 ships and we start moving onto 0.26, it should be safe to just remove all of the code marked with `TODO` in this PR and we should be back to clean upstream usage.


nak3:
/retest

`TestServerlessUpgrade/VerifyContinualTests/AutoscaleSustainingTest` failed due to missing SLO late 0.003. Let's see another test result. 

```
    autoscaler.go:63: Error:  request success rate under SLO: total = 51984, errors = 184, rate = 0.996460, SLO = 0.999000
````

markusthoemmes:
Yeah, I've seen this. I don't think the result is acceptable yet :cry:

markusthoemmes:
/unhold

Passed twice in a row. Let's get this in and watch the nighlies. If we see request dropping, we should consider to roll this back.

PR #1225, opened 2021-09-27
Fix support email address and tighten bundle checks

markusthoemmes:
As per title, the "wrong" email address was blocking the check tightening before.


lance (commented):
/lgtm

PR #1226, opened 2021-09-28
[SRVKS-550] Set termination policy to OCP Route based on HTTPOption

nak3:
This patch sets termination policy to OCP Route based on ingress's `Spec.HTTPOption`.

Also, it sets `httpProtocol` in config-network to `Redirected`. (I expect some test in
`test/servinge2e/kourier` fail but other test pass.)

### DIFF
--- a/serving/ingress/pkg/reconciler/ingress/resources/route.go
+++ b/serving/ingress/pkg/reconciler/ingress/resources/route.go
@@ -111,6 +111,11 @@ func makeRoute(ci *networkingv1alpha1.Ingress, host string, rule networkingv1alp
 		return nil, ErrNoValidLoadbalancerDomain
 	}
 
+	terminationPolicy := routev1.InsecureEdgeTerminationPolicyAllow
+	if ci.Spec.HTTPOption == networkingv1alpha1.HTTPOptionRedirected {
+		terminationPolicy = routev1.InsecureEdgeTerminationPolicyRedirect
+	}
+
 	route := &routev1.Route{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:        name,
@@ -130,7 +135,7 @@ func makeRoute(ci *networkingv1alpha1.Ingress, host string, rule networkingv1alp
 			},
 			TLS: &routev1.TLSConfig{
 				Termination:                   routev1.TLSTerminationEdge,
-				InsecureEdgeTerminationPolicy: routev1.InsecureEdgeTerminationPolicyAllow,
+				InsecureEdgeTerminationPolicy: terminationPolicy,
 			},
 			WildcardPolicy: routev1.WildcardPolicyNone,
 		},
--- a/serving/ingress/pkg/reconciler/ingress/resources/route.go
+++ b/serving/ingress/pkg/reconciler/ingress/resources/route.go
@@ -119,7 +119,7 @@ func makeRoute(ci *networkingv1alpha1.Ingress, host string, rule networkingv1alp
 	// TODO: Remove this annotation handling after serving 0.26+.
 	// Ingress configures the HTTPOption based on the annotation.
 	// https://github.com/knative/serving/commit/d9c1342b5761afdac88c563535885e37fae27c7e
-	if len(annotations) != 0 && annotations[networking.HTTPOptionAnnotationKey] != "" {
+	if annotations[networking.HTTPOptionAnnotationKey] != "" {
 		annotation := annotations[networking.HTTPOptionAnnotationKey]
 		switch strings.ToLower(annotation) {
 		case "enabled":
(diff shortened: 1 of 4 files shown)


nak3:
`TestKnativeServiceHTTPS` failed as expected because it redirects to HTTPS.

markusthoemmes:
Wasn't there an annotation introduced to control this per Service? If so, we could now change the test to make sure HTTP is served in one case and redirected in the other?

serving/ingress/pkg/reconciler/ingress/resources/route.go
```diff
@@ -111,6 +111,26 @@ func makeRoute(ci *networkingv1alpha1.Ingress, host string, rule networkingv1alp
...
+	// https://github.com/knative/serving/commit/d9c1342b5761afdac88c563535885e37fae27c7e
+	if len(annotations) != 0 && annotations[networking.HTTPOptionAnnotationKey] != "" {
+		annotation := annotations[networking.HTTPOptionAnnotationKey]
+		switch strings.ToLower(annotation) {
+		case "enabled":
+			terminationPolicy = routev1.InsecureEdgeTerminationPolicyAllow
+		case "redirected":
+			terminationPolicy = routev1.InsecureEdgeTerminationPolicyRedirect
+		default:
+			return nil, fmt.Errorf("incorrect HTTPOption annotation:" + annotation)
```
markusthoemmes, line 130:
```suggestion
			return nil, fmt.Errorf("incorrect HTTPOption annotation: " + annotation)
```

serving/ingress/pkg/reconciler/ingress/resources/route.go
```diff
@@ -111,6 +111,26 @@ func makeRoute(ci *networkingv1alpha1.Ingress, host string, rule networkingv1alp
 		return nil, ErrNoValidLoadbalancerDomain
 	}
 
+	terminationPolicy := routev1.InsecureEdgeTerminationPolicyAllow
+	if ci.Spec.HTTPOption == networkingv1alpha1.HTTPOptionRedirected {
+		terminationPolicy = routev1.InsecureEdgeTerminationPolicyRedirect
+	}
+
+	// TODO: Remove this annotation handling after serving 0.26+.
+	// Ingress configures the HTTPOption based on the annotation.
```
markusthoemmes, line 120: Nice one!

serving/ingress/pkg/reconciler/ingress/resources/route.go
```diff
@@ -111,6 +111,26 @@ func makeRoute(ci *networkingv1alpha1.Ingress, host string, rule networkingv1alp
...
+	// https://github.com/knative/serving/commit/d9c1342b5761afdac88c563535885e37fae27c7e
+	if len(annotations) != 0 && annotations[networking.HTTPOptionAnnotationKey] != "" {
```
markusthoemmes, line 122:
```suggestion
	if annotations[networking.HTTPOptionAnnotationKey] != "" {
```

accessing nil and/or empty maps works gracefully. Only adding to them needs an explicit nil-check.

markusthoemmes (commented):
A few last nits, this looks great!

nak3:
/test 4.8-operator-e2e-aws-ocp-48

setup failed.

serving/ingress/pkg/reconciler/ingress/resources/route.go
```diff
@@ -111,6 +111,26 @@ func makeRoute(ci *networkingv1alpha1.Ingress, host string, rule networkingv1alp
...
+	// Ingress configures the HTTPOption based on the annotation.
+	// https://github.com/knative/serving/commit/d9c1342b5761afdac88c563535885e37fae27c7e
+	if annotations[networking.HTTPOptionAnnotationKey] != "" {
+		annotation := annotations[networking.HTTPOptionAnnotationKey]
+		switch strings.ToLower(annotation) {
+		case "enabled":
+			terminationPolicy = routev1.InsecureEdgeTerminationPolicyAllow
+		case "redirected":
+			terminationPolicy = routev1.InsecureEdgeTerminationPolicyRedirect
+		default:
```
markusthoemmes, line 129: Is `"disabled"` no a valid value here?

nak3: Yeah, we can support `disable` on Downstream but upstream does not support it and the `disable` annotation is invalid value. So I think we should align with upstream and we can consider to support it if customers requested in the future.

markusthoemmes: ack, okay!

nak3:
Oh, OpenShift's Ingress (HAProxy) returns `302` rather than `301`.

```` 
    verify_http_and_https_test.go:75: The Route at domain http://redirect-service-serverless-tests.apps.ci-ocp-4-8-amd64-aws-us-east-1-tkzcz.hive.aws.ci.openshift.org didn't serve the expected status code got=302, want=301
````

nak3:
Thank you! I think this is ready now.

markusthoemmes (approved):
/lgtm
/approve

PR #1227, opened 2021-09-28
Use actual certs instead of skip insecure

nak3:
Current downstream kourier tests HTTPS with skip insecure.
But now the certs are prepared by `trust_router_ca()` so we can use it.

### DIFF
--- a/test/servinge2e/helpers.go
+++ b/test/servinge2e/helpers.go
@@ -17,9 +17,6 @@ const (
 	helloworldService2    = "helloworld-go2"
 	kubeHelloworldService = "kube-helloworld-go"
 	helloworldText        = "Hello World!"
-
-	caSecretNamespace = "cert-manager"
-	caSecretName      = "ca-key-pair"
 )
 
 func WaitForRouteServingText(t *testing.T, caCtx *test.Context, routeURL *url.URL, expectedText string) {
--- a/test/servinge2e/helpers.go
+++ b/test/servinge2e/helpers.go
@@ -44,13 +44,6 @@ func WaitForRouteServingText(t *testing.T, caCtx *test.Context, routeURL *url.UR
 	}
 }
 
-func insecureSkipVerify() spoof.TransportOption {
-	return func(transport *http.Transport) *http.Transport {
-		transport.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
-		return transport
-	}
-}
-
 // addRootCAtoTransport returns TransportOption when HTTPS option is true. Otherwise it returns plain spoof.TransportOption.
 func addRootCAtoTransport(ctx context.Context, logf logging.FormatLogger, clients *test.Clients) spoof.TransportOption {
 	return func(transport *http.Transport) *http.Transport {
--- a/test/servinge2e/helpers.go
+++ b/test/servinge2e/helpers.go
@@ -2,18 +2,12 @@ package servinge2e
 
 import (
 	"context"
-	"crypto/tls"
-	"crypto/x509"
-	"net/http"
 	"net/url"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


test/servinge2e/helpers.go
```diff
@@ -31,15 +38,38 @@ func WaitForRouteServingText(t *testing.T, caCtx *test.Context, routeURL *url.UR
...
-		insecureSkipVerify(),
+		addRootCAtoTransport(context.Background(), t.Logf, caCtx.Clients),
 	); err != nil {
 		t.Fatalf("The Route at domain %s didn't serve the expected text \"%s\": %v", routeURL, expectedText, err)
 	}
 }
 
-func insecureSkipVerify() spoof.TransportOption {
+// addRootCAtoTransport returns TransportOption when HTTPS option is true. Otherwise it returns plain spoof.TransportOption.
+func addRootCAtoTransport(ctx context.Context, logf logging.FormatLogger, clients *test.Clients) spoof.TransportOption {
```
markusthoemmes, line 48: Can we not use upstream's `AddRootCAToTransport` for this?

nak3: Yeah, I wanted to use upstream's `AddRootCAtoTransport` but the `Clients` which is the third argument of the func causes the type mismatch.

We defined our own `Clients` https://github.com/openshift-knative/serverless-operator/blob/4cdbaafd13b3ad84346a2b3cd8f91cc2ff40d8d4/test/clients.go#L38 and I think it is not easy to change.

markusthoemmes: I see, that makes sense. I think the following should work tho

```golang
servingTest.AddRootCAtoTransport(context.Background(), t.Logf, &servingTest.Clients{KubeClient: caCtx.Clients.Kube}, true),
```

Ultimately, the function only needs the KubeClient, so passing this specific object through should be sufficient, I think.

nak3: Oh! Thank you :bow: It works :heart_eyes:  Updated.

nak3:
(mesh test also passed https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift-knative_serverless-operator/1227/pull-ci-openshift-knative-serverless-operator-main-4.8-upstream-e2e-mesh-aws-ocp-48/1442805874048372736 )

test/servinge2e/helpers.go
```diff
@@ -19,6 +17,9 @@ const (
 	helloworldService2    = "helloworld-go2"
 	kubeHelloworldService = "kube-helloworld-go"
 	helloworldText        = "Hello World!"
+
+	caSecretNamespace = "cert-manager"
+	caSecretName      = "ca-key-pair"
```
markusthoemmes, line 22: These are now unused, I think.

nak3: Yes, thank you!

test/servinge2e/kourier/verify_http_and_https_test.go
```diff
@@ -22,46 +18,13 @@ func TestKnativeServiceHTTPS(t *testing.T) {
...
-		t.Fatalf("Request to %v failed, err: %v", httpsURL, err)
-	}
-	if resp.StatusCode != http.StatusOK {
-		body, err := ioutil.ReadAll(resp.Body)
-		if err != nil {
-			t.Error("Failed to read body", err)
-		}
-		t.Fatalf("Response failed, status %v, body %v", resp.StatusCode, string(body))
-	}
+	httpsURL.Scheme = "http"
```
nak3, line 27: Umm... I think here needs to be changed. It still tests with `https`.

nak3:
@markusthoemmes This is ready now. Thank you.

markusthoemmes (approved):
/lgtm
/approve

PR #1229, opened 2021-09-30
Download zipkin from Github's container registry

matzew:

as per title... to avoid DockerHub limit problems.

/assign @cardil


cardil (approved):
/lgtm

PR #1232, opened 2021-10-04
[main] Roll in the Eventing-Kafka patch for 857

matzew:
Applying one more patch on the `KafkaChannel` manifest that we ship via product.

Reason: Yaml change in here: https://github.com/knative-sandbox/eventing-kafka/pull/857

/assign @aliok


matzew:
@aliok upstream has it for 0.24

Sounds like we should patch the 1.18 branch of this repo as well

PR #1234, opened 2021-10-05
Remove unnecessary RBAC to delete services

markusthoemmes:
We've deleted the Knative Service a while ago and the respective RBAC should no longer be necessary. Upgrade test will tell if that's true.


dsimansk:
Can be the Serving API types dropped as well, right?

https://github.com/openshift-knative/serverless-operator/blob/main/knative-operator/pkg/apis/addtoscheme_openshift_v1.go

markusthoemmes:
Indeed, pushed!

dsimansk:
@markusthoemmes rebase due to Eventing bump merged?

markusthoemmes:
Yep, already done :)

PR #1242, opened 2021-10-08
Add new label to show grafana dashboards in ODC

jerolimov:
Fixes https://issues.redhat.com/browse/ODC-6340 similar to https://github.com/openshift/cluster-monitoring-operator/pull/1294 by adding a new `console.openshift.io/odc-dashboard` label to two grafana dashboards (ConfigMaps).

This label was already observed by the OpenShift developer console. With this change ODC shows these dashboards in the "Observe" navigation entry as dropdown like this:

image

(Tested this by adding the labels manually on a cluster to the ConfigMaps.)

### DIFF
--- a/knative-operator/deploy/resources/dashboards/serving/grafana-dash-knative-serving-queue-proxy.yaml
+++ b/knative-operator/deploy/resources/dashboards/serving/grafana-dash-knative-serving-queue-proxy.yaml
@@ -5,6 +5,7 @@ metadata:
   namespace: openshift-config-managed
   labels:
     console.openshift.io/dashboard: "true"
+    console.openshift.io/odc-dashboard: "true"
 data:
   queue-proxy-dashboard.json: |+
     {
--- a/knative-operator/deploy/resources/dashboards/serving/grafana-dash-knative-serving-resources.yaml
+++ b/knative-operator/deploy/resources/dashboards/serving/grafana-dash-knative-serving-resources.yaml
@@ -5,6 +5,7 @@ metadata:
   namespace: openshift-config-managed
   labels:
     console.openshift.io/dashboard: "true"
+    console.openshift.io/odc-dashboard: "true"
 data:
   resource-dashboard.json: |+
     {


markusthoemmes (commented):
/ok-to-test
/assign @skonto

jerolimov:
@skonto Sorry for the delay! I will take a new look this week. 

I created this [bugzilla issue #2013990 "Observe dashboard crashs on reload when perspective has changed (in another tab)"](https://bugzilla.redhat.com/show_bug.cgi?id=2013990) already, but we are not working on this yet. 

So this problem happen esp. then when you comparing both dashboards (in Admin and Dev perspective) and switching the perspective and then reload the other browser tab. Also if this doesn't affect users which stay in their perspective, a bug we should fix.

But I still need to understand you https://github.com/openshift-knative/serverless-operator/pull/1242#issuecomment-940846202, this is something we should address as well? I need to check here is this is a feature or a bug.

skonto:
@jerolimov thanks for looking into this. The problem is that the same dashboard that works under admin view as expected does not work under developer view. It feels like a bug to me.

"I observed that the PromQL queries are modifed. Compare
sum(container_memory_usage_bytes{namespace="", pod=~".*", container != "POD", container != ""}) by (container)
vs the original:
sum(rate(container_cpu_usage_seconds_total{namespace=\"$namespace\", pod=~\"$revision.*\", container != \"POD\", container != \"\"}[1m])) by (container)"
The pod label regex is different, the original depends on a template variable. Although this is true also for the admin case the dashboard there works as expected."

This means that somehow it ignores some parts of the content. As you see the dashboard under developer view ignores the template var.

skonto:
@jerolimov any news here? Wdyth?

jerolimov:
Hey @skonto, apologize again for the delay.

We had many issues with our nightly builds where I hoped daily that we get a working version. And without these fixes we have already done it was difficult to verify your reported problems.

I believe the latest issue you mentioned was this bug we have now fixed in our master/main branch. Different dashboards had the issue to not work fine within a namespace.

1. **[Bugzilla 2022707 Observe / monitoring dashboard shows forbidden errors on Dev Sandbox](https://bugzilla.redhat.com/show_bug.cgi?id=2022707)**
    * https://github.com/openshift/console/pull/10460
    * https://github.com/openshift/console/pull/10548

This is the old issue you mentioned earlier, which is now also fixed:

2. **[Bugzilla 2013990 Observe dashboard crashs on reload when perspective has changed (in another tab)](https://bugzilla.redhat.com/show_bug.cgi?id=2013990)**
    * https://github.com/openshift/console/pull/10526
    * https://github.com/openshift/console/pull/10586

jerolimov:
/cc @skonto 
:)

skonto:
Back on this one after the time off :)

skonto:
@jerolimov I tested on 4.9.13 (latest release) still there are errors when trying to refresh the page for the queue proxy metrics dashboard. Pls check [here](https://gist.github.com/skonto/f573fc9780059feaab0d7a5b41abda1c).
Also here is the screenshot for the issue:
image
Under the admin view even with a page refresh (browser refresh) things look ok:
image

My biggest concern here is what will happen if users try to install the latest S-O on some older cluster. Afaik the serverless operator supports OCP from [4.6 to 4.9](https://github.com/openshift-knative/serverless-operator/blob/2e4715a8054586aa2a36f542cb8d2927ecc7d58a/olm-catalog/serverless-operator/project.yaml#L20) at the moment. Would not that be problematic as bugs are fixed on the latest OCP versions?

skonto:
Actually the above is going to be a problem until we have the fixes backported to 4.9. AFAIK the plan is to have the following upgrade path: 1.20 (OCP 4.9) -> 1.21 (OCP 4.9) -> 1.21 (OCP 4.10)

skonto:
@jerolimov I tested with `4.10.0-0.nightly-2022-01-10-101431`. It seems that the queue proxy metrics dashboard is loaded under the developer view but I noticed that the `time range` and the `refresh interval` are not respected or they dont have the same behavior as in admin view.
Check for example for time range last 5 minutes, refresh interval 15secs:

admin:
image

developer:
image

Obviously under developer view the time range is different for some reason.

At some point I also hit a case where nothing was being loaded after a browser refresh (could be net I have no idea but never happened under admin view):

image

I overcome this by selecting the right dashboard on the empty drop down menu (it is not empty but looks so).

jerolimov:
Hey @skonto, thanks again for the detailed testing.

What you find is the fix https://github.com/openshift/console/pull/10526 for issue [Bug 2013990 Observe dashboard crash on reload when perspective has changed (in another tab)](https://bugzilla.redhat.com/show_bug.cgi?id=2013990)  only fixed the crash.

We have still the issue that the time-range changes after browser reload -- when perspective is changed in another tab -- doesn't work. I created a new issue for that: [Bug 2039770 Observe dashboard doesn't react on time-range changes after browser reload when perspective is changed in another tab](https://bugzilla.redhat.com/show_bug.cgi?id=2039770)

Here is also a small video that explains the problem so that whoever picks this up can reproduce it easily. [Uploaded to dropbox, but public visible](https://www.dropbox.com/s/6gpsk9ojel0chqy/explaining-dashboard-perspective-issue.mp4?dl=0)

It's not related to the Serverless dashboards. I tested this with one of the four default dashboards.

I think this is an edge case that someone switches the perspective in one browser tab and reloads the observe dashboard in another. I mean this can happen. We should fix this. But I believe it happens even more in the case when testing and comparing both dashboards.

jerolimov:
I'm not sure what happened exactly on your 3rd screenshot. But at least I can reproduce this when changing the URL to a non-existing dashboard. While testing, maybe you tried to remove the time range and removed too much or something?

I created also a low-severity issue for this so that we can improve this case. I think also this is an edge-case that should not block this PR, or? Wdyt? :)

[Bug 2039776 Observe dashboard shows nothing if the URL links to an non existing dashboard](https://bugzilla.redhat.com/show_bug.cgi?id=2039776)

jerolimov:
Hey, based on our latest slack conversation, here is a list of all issues we logged as part of your findings. We plan to backport all of them. But our release process requires that we release 4.10 before these get merged and released of an 4.9 z-stream release.

You can see here the prepared PRs for 4.9. We expect that they get automatically merged in the next weeks.

1. **[Bugzilla 2015806 - Metrics view in Deployment reports "Forbidden" when not cluster-admin](https://bugzilla.redhat.com/show_bug.cgi?id=2015806)**
    * https://github.com/openshift/console/pull/10344
    * https://github.com/openshift/console/pull/10396
    * https://github.com/openshift/console/pull/10496

2. **[Bugzilla 2022707 - Observe / monitoring dashboard shows forbidden errors on Dev Sandbox](https://bugzilla.redhat.com/show_bug.cgi?id=2022707)**
    * https://github.com/openshift/console/pull/10460
    * https://github.com/openshift/console/pull/10548

This is the old issue you mentioned earlier, which is now also fixed:

3. **[Bugzilla 2013990 - Observe dashboard crashs on reload when perspective has changed (in another tab)](https://bugzilla.redhat.com/show_bug.cgi?id=2013990)**
    * https://github.com/openshift/console/pull/10526
    * https://github.com/openshift/console/pull/10586

4. **[Bugzilla 2039770 - Observe dashboard doesn't react on time-range changes after browser reload when perspective is changed in another tab](https://bugzilla.redhat.com/show_bug.cgi?id=2039770)**
    * https://github.com/openshift/console/pull/10836
    * TODO, will update this comment asap.

5. **[Bugzilla 2039776 - Observe dashboard shows nothing if the URL links to an non existing dashboard](https://bugzilla.redhat.com/show_bug.cgi?id=2039776)**
    * https://github.com/openshift/console/pull/10851
    * https://github.com/openshift/console/pull/10857

skonto:
Thanks a lot @jerolimov. /lgtm

rhuss (commented):
/approve
/lgtm

PR #1245, opened 2021-10-11
Log prometheus API errors but not fail

mgencur:
Account for Prometheus startup time.
The test will fail after expiring the timeout anyway.


test/monitoringe2e/helpers.go
```diff
@@ -112,7 +112,8 @@ func VerifyHealthStatusMetric(caCtx *test.Context, label string, expectedValue s
 	if err := wait.PollImmediate(test.Interval, prometheusTargetTimeout, func() (bool, error) {
 		value, _, err := pc.Query(context.Background(), fmt.Sprintf(`knative_up{type="%s"}`, label), time.Time{})
 		if err != nil {
-			return false, err
+			caCtx.T.Log("Error:", err)
```
markusthoemmes, line 115:
```suggestion
			caCtx.T.Log("Error querying prometheus metrics:", err)
```

same below.

PR #1255, opened 2021-10-14
Move common constants for labels into the top-level pkg

markusthoemmes:
As per title, this shuffles some constants around so we guarantee consistency between them. Essentially, this is a prefactor for #1254


rhuss (commented):
/approve
/lgtm

PR #1256, opened 2021-10-14
Pull in the newest operator additions

markusthoemmes:
As per title. This picks up a cherry-pick I requested upstream and fixes the slight break in API for us. Prefactor for #1254 and useful for future releases anyway (as they'd break otherwise).

### DIFF
--- a/vendor/knative.dev/operator/pkg/reconciler/knativeeventing/controller.go
+++ b/vendor/knative.dev/operator/pkg/reconciler/knativeeventing/controller.go
@@ -59,10 +59,10 @@ func NewExtendedController(generator common.ExtensionGenerator) injection.Contro
 		c := &Reconciler{
 			kubeClientSet:     kubeClient,
 			operatorClientSet: operatorclient.Get(ctx),
-			extension:         generator(ctx),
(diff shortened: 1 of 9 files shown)


vendor/knative.dev/operator/pkg/reconciler/knativeeventing/controller.go
```diff
@@ -59,10 +59,10 @@ func NewExtendedController(generator common.ExtensionGenerator) injection.Contro
 		c := &Reconciler{
 			kubeClientSet:     kubeClient,
 			operatorClientSet: operatorclient.Get(ctx),
-			extension:         generator(ctx),
 			manifest:          manifest,
 		}
 		impl := knereconciler.NewImpl(ctx, c)
+		c.extension = generator(ctx, impl)
```
rhuss, line 65: Probably unrelated (as it's a rebase to upstream, right ?), but this kind of cyclic dependency is typically an architectural smell. Not sure if this has to be like this or whether there is a good reason for this (one drawback of course is, that later refactoring gets much more difficult if those objects are weaved together so closely.

PR #1268, opened 2021-10-29
Make all Kube API imports consistent as

markusthoemmes:
A ton of code already imports packages this way naturally. `corev1` and `metav1` for instance have been mostly consistent. This makes the imports consistent for all APIs we use to make it much easier to follow code and copy code around (been bitten by this recently quite hard).

No actual changes intended. Only import renames.


nak3:
LGTM except for the "false" warning for the misspelling of importas :sweat_smile:

PR #1278, opened 2021-11-03
[SRVKS-822] Use CheckEndpointState instead of WaitForEndpointState

markusthoemmes:
As per title, this aligns us with upstream.


matzew (approved):
/lgtm
/approve

PR #1279, opened 2021-11-04
Add k8s version patch to run OCP 4.6

nak3:
As per title, OCP 4.6 is based on k8s 1.19 so the operator fails to start without patch.


nak3:
The issue is observed in https://github.com/openshift/knative-serving/pull/966#issuecomment-960394489

markusthoemmes (approved):
/lgtm
/approve

PR #1280, opened 2021-11-04
Run clean metadata before installing nodejs

nak3:
As per title this patch fixes the following issue in current CI:

```
https://rpm.nodesource.com/pub_14.x/el/7/x86_64/nodejs-14.18.1-1nodesource.x86_64.rpm: [Errno -1] Package does not match intended download. Suggestion: run yum --enablerepo=nodesource clean metadata
Trying other mirror.


Error downloading packages:
  2:nodejs-14.18.1-1nodesource.x86_64: [Errno 256] No more mirrors to try.
```


nak3:
The `Dockerfile` change will be affected after merge, I guess :thinking:

nak3:
Closing :cry:

PR #1284, opened 2021-11-04
Bump Node.js to 16 (LTS)

markusthoemmes:
Node.js 14 installation started to fail

```
https://rpm.nodesource.com/pub_14.x/el/7/x86_64/nodejs-14.18.1-1nodesource.x86_64.rpm: [Errno -1] Package does not match intended download. Suggestion: run yum --enablerepo=nodesource clean metadata
Trying other mirror.


Error downloading packages:
  2:nodejs-14.18.1-1nodesource.x86_64: [Errno 256] No more mirrors to try.
```

While that might be temporary, it seemed reasonable to me to update onto a LTS release anyway, and that seems to continue to work.


nak3:
We also need to update `templates/build-image.Dockerfile` :cat:

PR #1288, opened 2021-11-05
Add func 0.19.0 dependency

jrangelramos:
As per title, adding `func` dependency to `project.yaml`


jrangelramos:
I found another way for QE automation without need to add this entry to this repo. This change can be postponed in future when this info is relevant on this repo.

PR #1289, opened 2021-11-05
Add func 0.19.0 dependency

jrangelramos:
As per title, adding `func` dependency to `project.yaml`

### DIFF
--- a/olm-catalog/serverless-operator/project.yaml
+++ b/olm-catalog/serverless-operator/project.yaml
@@ -30,3 +30,4 @@ dependencies:
   eventing_kafka: 0.26.0
   cli: 0.25.1
   operator: 0.26.3
+  func: 0.19.0


markusthoemmes:
Does this have any perceivable effect on anything? Seems like it ain't used anywhere in this repo :thinking:

jrangelramos:
No it will not. QE jobs make use of version from this file for internal tests. I just want to concentrate versions on the same place.

olm-catalog/serverless-operator/project.yaml
```diff
@@ -30,3 +30,4 @@ dependencies:
   eventing_kafka: 0.26.0
   cli: 0.25.1
   operator: 0.26.3
+  func: 0.19.0
```
markusthoemmes, line 33: Let's add a comment if this is unused for now, like

```
# currently only used for QE automation
```

jrangelramos:
I found another way for QE automation without need to add this entry to this repo. This change can be postponed in future when this info is relevant on this repo.

PR #1290, opened 2021-11-05
[SRVCOM-1069] Stop stamping defaults into serving and eventing

markusthoemmes:
As per title. Most of this defaulting is better of "silent" (like the webhook memory override) anyway. The only real usability concern (where I'd like to keep the default stamping) was the defaulting for Kourier which looks weird if not in place like that (see https://issues.redhat.com/browse/SRVKS-832).

All of these defaults are still going to be applied through the openshift-knative-operator at reconcile time. Dropping this stamping will allow us to open all these options up towards our users without having to try to distinguish values we stamped from values the users gave us.

To be a little more graphic, the user currently sees this after applying an empty KnativeServing:

```yaml
...
  spec:
    config:
      domain:
        apps.ci-op-f2vqwm3n-73568.origin-ci-int-aws.dev.rhcloud.com: ""
    controller-custom-certs:
      name: config-service-ca
      type: ConfigMap
    deployments:
    - name: domain-mapping
      replicas: 2
    high-availability:
      replicas: 2
    registry:
      override:
        APISERVER_RA_IMAGE: registry.ci.openshift.org/openshift/knative-v0.26.0:knative-eventing-apiserver-receive-adapter
        DISPATCHER_IMAGE: registry.ci.openshift.org/openshift/knative-v0.26.0:knative-eventing-channel-dispatcher
        KUBE_RBAC_PROXY: registry.ci.openshift.org/origin/4.7:kube-rbac-proxy
        activator: registry.ci.openshift.org/openshift/knative-v0.26.0:knative-serving-activator
        autoscaler: registry.ci.openshift.org/openshift/knative-v0.26.0:knative-serving-autoscaler
        autoscaler-hpa: registry.ci.openshift.org/openshift/knative-v0.26.0:knative-serving-autoscaler-hpa
        controller/controller: registry.ci.openshift.org/openshift/knative-v0.26.0:knative-serving-controller
        domain-mapping: registry.ci.openshift.org/openshift/knative-v0.26.0:knative-serving-domain-mapping
        domainmapping-webhook: registry.ci.openshift.org/openshift/knative-v0.26.0:knative-serving-domain-mapping-webhook
        eventing-controller/eventing-controller: registry.ci.openshift.org/openshift/knative-v0.26.0:knative-eventing-controller
        eventing-webhook/eventing-webhook: registry.ci.openshift.org/openshift/knative-v0.26.0:knative-eventing-webhook
        imc-controller/controller: registry.ci.openshift.org/openshift/knative-v0.26.0:knative-eventing-channel-controller
        imc-dispatcher/dispatcher: registry.ci.openshift.org/openshift/knative-v0.26.0:knative-eventing-channel-dispatcher
        kourier-control: registry.ci.openshift.org/openshift/knative-v0.26.0:kourier
        kourier-gateway: quay.io/maistra/proxyv2-ubi8:2.1.0
        mt-broker-controller/mt-broker-controller: registry.ci.openshift.org/openshift/knative-v0.26.0:knative-eventing-mtchannel-broker
        mt-broker-filter/filter: registry.ci.openshift.org/openshift/knative-v0.26.0:knative-eventing-mtbroker-filter
        mt-broker-ingress/ingress: registry.ci.openshift.org/openshift/knative-v0.26.0:knative-eventing-mtbroker-ingress
        net-istio-controller/controller: registry.ci.openshift.org/openshift/knative-v0.26.0:net-istio-controller
        net-istio-webhook/webhook: registry.ci.openshift.org/openshift/knative-v0.26.0:net-istio-webhook
        net-kourier-controller/controller: registry.ci.openshift.org/openshift/knative-v0.26.0:kourier
        pingsource-mt-adapter/dispatcher: registry.ci.openshift.org/openshift/knative-v0.26.0:knative-eventing-mtping
        queue-proxy: registry.ci.openshift.org/openshift/knative-v0.26.0:knative-serving-queue
        storage-version-migration-eventing-eventing-0.26.0/migrate: registry.ci.openshift.org/openshift/knative-v0.26.0:knative-eventing-storage-version-migration
        storage-version-migration-serving-serving-0.26.0/migrate: registry.ci.openshift.org/openshift/knative-v0.26.0:knative-serving-storage-version-migration
        sugar-controller/controller: registry.ci.openshift.org/openshift/knative-v0.26.0:knative-eventing-sugar-controller
        webhook/webhook: registry.ci.openshift.org/openshift/knative-v0.26.0:knative-serving-webhook
    resources:
    - container: webhook
      limits:
        memory: 1Gi
```

And after this, it'll  be... well... empty. I argue the user's don't need all of the above in there. Almost all of it is internal details and not supposed to be user-facing.

**Note:** If we're fine with this, I'm happy to clean out Knative Kafka as well.

/assign @matzew @rhuss


rhuss:
+1 in principle, but I wonder what this means for updates. As the default is now stamped in the resources, there is no way to distinguish what has been provided by the user explicitly or what was added because it was a default. This also means, we can't easily change the defaults, since an update will stick to the old default value because it's provided explicitly in the spec. 

We have this issue now anyway, so I'm +1 to revert the behavior so that at least for future installations we are able to make this distinction so that we might even be able to change defaults.

Codewise the PR looks good and straightforward.

Waiting still for @matzew to chime in :)

knative-operator/pkg/webhook/knativeeventing/webhook_mutating.go
```diff
@@ -34,7 +33,10 @@ func (v *Configurator) Handle(ctx context.Context, req admission.Request) admiss
 		return admission.Errored(http.StatusBadRequest, err)
 	}
 
-	common.MutateEventing(ke)
+	// Unset the entire registry section. We used to override it anyway, so there can't
+	// be any userdata in there.
+	// TODO: Remove in the 1.21 release to potentially make this usable for users.
+	ke.Spec.CommonSpec.Registry.Override = nil
```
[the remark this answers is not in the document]
markusthoemmes: None I can necessarily think of (we even hide it in the CRD schema), but leaving our current "defaults" in there results in a huge mess of YAML which I'd rather get rid of as part of this

markusthoemmes:
To clarify here (we discussed a bit in Slack): Upgrades behave exactly the same with or without this. We're currently always force-overriding fields we think we should set anyway and are not responsive to changes by the user. With this change in place we can potentially think about doing a better job there though.

knative-operator/pkg/common/eventing.go
```diff
@@ -1,45 +0,0 @@
...
-	eventingImagesFromEnviron(ke)
-	ensureEventingWebhookMemoryLimit(ke)
-	ensureEventingWebhookInclusionMode(ke)
-	defaultToEventingHa(ke)
-}
-
-func defaultToEventingHa(ke *operatorv1alpha1.KnativeEventing) {
-	if ke.Spec.HighAvailability == nil {
-		ke.Spec.HighAvailability = &operatorv1alpha1.HighAvailability{
-			Replicas: 2,
```
matzew, line 21: Ah, so we get rid of this duplicaton, since we really wanna do it here:
https://github.com/openshift-knative/serverless-operator/blob/5454c1798b236e613daeb3a2629038258fef9416/openshift-knative-operator/pkg/eventing/extension.go#L63-L68

Is that the motivation here?

knative-operator/pkg/webhook/knativeeventing/webhook_mutating.go
```diff
@@ -34,7 +33,10 @@ func (v *Configurator) Handle(ctx context.Context, req admission.Request) admiss
 		return admission.Errored(http.StatusBadRequest, err)
 	}
 
-	common.MutateEventing(ke)
+	// Unset the entire registry section. We used to override it anyway, so there can't
+	// be any userdata in there.
+	// TODO: Remove in the 1.21 release to potentially make this usable for users.
```
matzew, line 38: Should we track that as a JIRI to noforget it? :see_no_evil:

knative-operator/pkg/webhook/knativeeventing/webhook_mutating.go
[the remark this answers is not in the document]
matzew: So... we cant' currently basically do ths: https://knative.dev/docs/install/operator/configuring-eventing-cr/#private-repository-and-private-secrets  ?

Not sure I undrstand this here - sorry

knative-operator/pkg/webhook/knativeserving/webhook_mutating.go
```diff
@@ -37,14 +33,30 @@ func (v *Configurator) Handle(ctx context.Context, req admission.Request) admiss
...
 		return admission.Errored(http.StatusInternalServerError, err)
 	}
 	return admission.PatchResponseFromRaw(req.AdmissionRequest.Object.Raw, marshaled)
 }
+
+// Technically this does nothing in terms of behavior (the default is assumed in the
+// extension code of openshift-knative-operator already), but it fixes a UX nit where
+// Kourier would be shown as enabled: false to the user if the ingress object is
+// specified.
+func defaultToKourier(ks *operatorv1alpha1.KnativeServing) {
```
matzew, line 54: why not in the "extension" ?

knative-operator/pkg/common/eventing.go
markusthoemmes: Partially. HA setup is one spot where we might think about **actually** keep the stamping, as it feels kinda sensible. For other parts though (default domains, image overrides etc.) it doesn't make any sense at all, as those pieces are supposed to change with version and cluster environment.

knative-operator/pkg/webhook/knativeserving/webhook_mutating.go
markusthoemmes: You read the comment? This is supposed to be user-visible so it has to be in the webhook.

knative-operator/pkg/webhook/knativeeventing/webhook_mutating.go
markusthoemmes: I guess we can.

knative-operator/pkg/webhook/knativeeventing/webhook_mutating.go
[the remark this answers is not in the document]
markusthoemmes: Correct, we can't (not today and not with this PR).

knative-operator/pkg/common/eventing.go
matzew: but on the extension also set memory, b/c the openshift needs a bit more than upstream.

This is also done in the "to be removed" file: https://github.com/openshift-knative/serverless-operator/blob/94fa6bb25e9ddff08728f1b7fab20f9701407397/knative-operator/pkg/common/eventing.go#L38

and the extension: https://github.com/openshift-knative/serverless-operator/blob/5454c1798b236e613daeb3a2629038258fef9416/openshift-knative-operator/pkg/eventing/extension.go#L56


> HA setup is one spot where we might think about actually keep the stamping
do you want to remove more of the "overrides" we are currently doing (in the extension)? 

sorry, if I miss something here

knative-operator/pkg/webhook/knativeserving/webhook_mutating.go
matzew: yes, but I think your latest comment here did make clear to me, why this is (still) needed

knative-operator/pkg/common/eventing.go
markusthoemmes: We're not removing any overrides in the extension.

knative-operator/pkg/webhook/knativeeventing/webhook_mutating.go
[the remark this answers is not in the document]
markusthoemmes: Sure, but today we even can't (and we stamp a lot of noise into the YAML (see description)). After this PR, we can decide if we want to support it

matzew (approved):
/lgtm
/approve
/abfahrt

PR #1293, opened 2021-11-09
[SRVKS-825] Remove annotation handling for termination policy

markusthoemmes:
0.26 shipped **actual** handling of the fields, so we don't need explicit annotation handling anymore.

/assign @nak3


nak3:
/lgtm

Thank you!

PR #1294, opened 2021-11-09
Update eventing-kafka midstream deps 

matzew:

On one of the test runs I noticed that there was still some `docker.io` image, which I replaced by our stash already on midstream. see: https://github.com/openshift-knative/eventing-kafka/pull/426

As per title, this updates that dependency, by running
```
./hack/update-deps.sh --upgrade
```
to "consume" the midstream patch/fix.


markusthoemmes (approved):
/lgtm
/approve

PR #1296, opened 2021-11-10
Make image transformation code much simpler

markusthoemmes:
This code (copied from upstream, where I've sent a similar fix) was way too complex to read. This aims to fix that.

/assign @aliok

### DIFF
--- a/knative-operator/pkg/controller/knativekafka/images.go
+++ b/knative-operator/pkg/controller/knativekafka/images.go
@@ -1,6 +1,8 @@
 package knativekafka
 
 import (
+	"fmt"
+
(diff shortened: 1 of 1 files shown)


matzew:
OK, thanks!

I like the more compact, inlined container section:
https://github.com/openshift-knative/serverless-operator/blob/597cae1f8686e86b39c0ba8c7781a2d8761aba0c/knative-operator/pkg/controller/knativekafka/images.go#L54-L71

rather than the old `updateRegistry` and its two func, that were mostly identical

LGTM

aliok:
/lgtm
/approve

Thanks Markus

PR #1301, opened 2021-11-12
[SRVCOM-1069] Drop webhook defaulting for KnativeKafka

markusthoemmes:
As per title and as a followup to https://github.com/openshift-knative/serverless-operator/pull/1290 this stops stamping defaults into the KnativeKafka resource as well. Since the webhook is now a noop, it's dropped as well.

/assign @aliok


matzew (approved):
thanks for the clean-up work

/lgtm
/approve

PR #1303, opened 2021-11-15
[SRVKE-812] Applying new multiline cfg style for eventing-kafka

matzew:

Removing patch that did kept the old cfg style


go.mod
```diff
@@ -4,6 +4,7 @@ go 1.16
 
 require (
 	github.com/blang/semver/v4 v4.0.0
+	github.com/ghodss/yaml v1.0.0
```
markusthoemmes, line 7: Please do not use this YAML package. It's unmaintained and stuff. Use https://github.com/kubernetes-sigs/yaml instead. Same interface, and maintained.

matzew: Thanks! I will update here!

FWIW, I just sent a PR for `eventing-kafka`, which uses this package instead as well

Will do the same for the `knative/operator` as well

pierDipi (approved):
/lgtm

PR #1306, opened 2021-11-16
Read folders for kafkas

matzew:
reading from `dir`, instead of directly from filez

Background:
Soon we have more files for channel/source:
- post job
- channel/source itself

As well the Knative-Kafka-Broker has `n` files.

For this reason, we read from the folders, instead of directly from files


matzew:
/assign @markusthoemmes 

I did the prefactorings

PR #1310, opened 2021-11-17
Remove secret as we not seem to use SASL/TLS on the test...

matzew:

as per title


matzew:
All passed 

Now just rebased...

PR #1314, opened 2021-11-23
SRVCOM-1615 Check for Spec.Replicas in verify_correct_deployment_shap…

mgencur:
…e test

* the actual Status.ReadyReplicas might be affected by slow environments
where the test could fail
* also improve the error message to actually show which deployment
doesn't have the desired scale

### DIFF
--- a/test/e2e/knative_eventing_test.go
+++ b/test/e2e/knative_eventing_test.go
@@ -49,10 +49,10 @@ func TestKnativeEventing(t *testing.T) {
 	})
 
 	t.Run("verify correct deployment shape", func(t *testing.T) {
-		// Check the status of scaled deployments in the knative eventing namespace
+		// Check the desired scale of deployments in the knative eventing namespace
 		for _, deployment := range []string{"eventing-controller", "eventing-webhook", "imc-controller", "imc-dispatcher", "mt-broker-controller", "sugar-controller"} {
 			if err := test.CheckDeploymentScale(caCtx, eventingNamespace, deployment, eventingHaReplicas); err != nil {
-				t.Fatalf("Failed to verify default HA settings for `%s`: %v", deployment, err)
+				t.Fatalf("Failed to verify default HA settings for %q: %v", deployment, err)
 			}
 		}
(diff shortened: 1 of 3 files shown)


test/e2e/knative_eventing_test.go
```diff
@@ -49,10 +49,10 @@ func TestKnativeEventing(t *testing.T) {
 	})
 
 	t.Run("verify correct deployment shape", func(t *testing.T) {
-		// Check the status of scaled deployments in the knative eventing namespace
+		// Check the desired scale of deployments in the knative eventing namespace
 		for _, deployment := range []string{"eventing-controller", "eventing-webhook", "imc-controller", "imc-dispatcher", "mt-broker-controller", "sugar-controller"} {
 			if err := test.CheckDeploymentScale(caCtx, eventingNamespace, deployment, eventingHaReplicas); err != nil {
-				t.Fatalf("Failed to verify default HA settings for `%s`: %v", deployment, err)
+				t.Fatalf("Failed to verify default HA settings for %q: %v", deployment, err)
```
skonto, line 55: In `manifest_test.go` we have also `'%s'` could you fix that too?

skonto:
I guess this changes the semantics a bit since we test for the desired number of pods of the deployment spec to be equal to the pass scaled number. Do we care about readiness?

test/e2e/knative_eventing_test.go
mgencur: Yes, sure.

mgencur (commented):
Hey @skonto , my point was to test only the desired scale. The actual scaling of Deploymets is kuberetes' business and I thought we don't have to test it. If the containers didn't start at all due to failure we would hit the issue pretty soon elsewhere, IMO.
Anyway, I've pushed a change to address your concern. The code first checks the desired scale for some deployments that define HA. And later it waits for all deployments to achieve the desired scale. And did this for ingress deployments as well. This is probably more robust and safe. WDYT?

skonto:
@mgencur yes looks good. 

/lgtm 
/approve

PR #1319, opened 2021-11-30
Make sure nothing is left in the serving-tests namespace before teardown

mgencur:
* Calling teardown while there are active knative resources being
deleted might leave some inconsistent resources there and prevent
deleting CRDs in the end.

This is to address https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-knative-serverless-operator-main-4.7-upgrade-tests-aws-ocp-47-continuous/1465470749794897920
There was still an active Configuration and Revision but not Service even though blockOwnerDeletion: true is set on Configuration and the Service should only be removed once the Configuration is deleted. I suppose it's a race condition between deleting resources and uninstalling Knative Serving. There are no Knative Serving pods to search for logs. Waiting for the namespace to be deleted seems like a more robust solution - it will either wait for all the resources to be gone before proceeding or time out, and in this case we'll be able to look at logs because it will all happen before Serverless teardown.

### DIFF
--- a/test/lib.bash
+++ b/test/lib.bash
@@ -220,8 +220,9 @@ function run_rolling_upgrade_tests {
     --resolvabledomain \
     --https
 
-  # Delete the leftover namespace.
-  oc delete namespace serving-tests
+  # Delete the leftover services.
+  oc delete ksvc all -n serving-tests
+  timeout 120 "[[ \$(oc get all --no-headers -n ${SERVING_NAMESPACE} | wc -l) != 0 ]]"
 
   logger.success 'Upgrade tests passed'
 }
--- a/test/lib.bash
+++ b/test/lib.bash
@@ -221,7 +221,7 @@ function run_rolling_upgrade_tests {
     --https
 
   # Delete the leftover services.
-  oc delete ksvc --all -n ${SERVING_NAMESPACE}
+  oc delete ksvc --all -n "${SERVING_NAMESPACE}"
   timeout 120 "[[ \$(oc get all --no-headers -n ${SERVING_NAMESPACE} | wc -l) != 0 ]]"
 
   logger.success 'Upgrade tests passed'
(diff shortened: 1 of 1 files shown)


mgencur:
Hmm. The latest run of upgrade tests is about 3 minutes longer than previous runs. Maybe it's enough to wait for no resources to be in that namespace. That would be quicker

test/lib.bash
```diff
@@ -221,7 +221,8 @@ function run_rolling_upgrade_tests {
     --https
 
   # Delete the leftover services.
-  oc delete ksvc --all -n serving-tests
+  oc delete ksvc all -n serving-tests
```
markusthoemmes, line 224:
```suggestion
  oc delete ksvc --all -n serving-tests
```

mgencur: Thanks

mgencur:
I think the latest solution is better, waiting for resources took 41 seconds (vs. 3 minutes for removing the whole namespace).

nak3:
> it will either wait for all the resources to be gone before proceeding or time out, and in this case we'll be able to look at logs because it will all happen before Serverless teardown.

That makes sense. But users also have to delete all existing ksvc before uninstalling the serverless-operator?

Is it not good idea to verify that we can uninstall the operator even if some ksvc are remained? If it is not possible, we need to mention it in the [document](https://docs.openshift.com/container-platform/4.9/serverless/install/removing-openshift-serverless.html) that users have to remove all ksvc before the uninstall like our test does.

mgencur:
@nak3 the step that fails is removing the CRD at the end. The serverles-operator can be uninstalled even if Kservices are there but the final step of removing the CRD fails. We could probably mention it in docs, yeah.

nak3:
Thank you. Understand.

LGTM :+1:

markusthoemmes (approved):
/lgtm
/approve

PR #1320, opened 2021-11-30
Moving the actual broker concerns to a defaultConfig

matzew:

Better reflect the acutal config is just the global default config

### DIFF
--- a/hack/lib/serverless.bash
+++ b/hack/lib/serverless.bash
@@ -244,7 +244,8 @@ metadata:
 spec:
   broker:
     enabled: true
-    bootstrapServers: my-cluster-kafka-bootstrap.kafka:9092
+    defaultConfig:
+      bootstrapServers: my-cluster-kafka-bootstrap.kafka:9092
   source:
     enabled: true
   channel:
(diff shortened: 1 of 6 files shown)


hack/lib/serverless.bash
```diff
@@ -244,7 +244,8 @@ metadata:
 spec:
   broker:
     enabled: true
-    bootstrapServers: my-cluster-kafka-bootstrap.kafka:9092
+    defaultConfig:
+      bootstrapServers: my-cluster-kafka-bootstrap.kafka:9092
```
markusthoemmes, line 248: This seems to be a breaking change. How are customers supposed to upgrade? 🤔

matzew: Nope, b/c this was introduced a week ago or so, was not there on 1.19 (the `broker` things are new)

markusthoemmes: Ahhh, nevermind then, thanks!

matzew:
I updated the CRD... to also have the `defaultConfig` ...

matzew:
/unhold
tested it also end-to-end (via UI from Operator)

matzew:
/assign @pierDipi 
going with the nested `defaultConfig` for the "global" broker settings

pierDipi (approved):
/lgtm

PR #1324, opened 2021-12-02
Fix dependabot warnings in UI test packages

markusthoemmes:
These are non productized parts, so there's no real surface area, but keeping things clean anyway.

Ran 

```
$ npm install
$ npm audit fix
```

to generate these changes.

/assign @cardil


cardil (approved):
Looks good

PR #1326, opened 2021-12-03
WIP: Remove sugar-controller and regenerate the bits

matzew:

DO NOT MERGE - ONLY A TEST


matzew:
/hold
(only a test)

matzew:
let's see if we have much dependency on this "brain fart"

openshift-knative-operator/pkg/monitoring/monitoring_eventing_test.go
```diff
@@ -19,7 +19,7 @@ func TestLoadPlatformEventingMonitoringManifests(t *testing.T) {
 		t.Errorf("Got %d, want %d", len(manifests), 1)
 	}
 	resources := manifests[0].Resources()
-	if len(resources) != 28 {
+	if len(resources) != 25 {
```
matzew, line 22: @skonto can we document why that's actually 28? (and now, after removing 25)?
I have a bit problems with these static numbers... could be a `const` with a better name, and that could have some comments on why

rhuss: Ideally this should be calculated dynamically instead of hardcoding as a number (regardless whether its a constant or not).

matzew:
Some e2e tests (upstream) rely on this bizarre magic

openshift-knative-operator/pkg/monitoring/monitoring_eventing_test.go
skonto: I will create a PR to make it more dynamic. In general this depends on the number of Eventing deployments as we create monitoring resources per deployment.

matzew:
/close
we keep it (for now)

PR #1328, opened 2021-12-03
Delete serving-tests ns after upgrade tests

mgencur:
Compared with previous solution this also works on OCP 4.6 (Kube 1.19).

### DIFF
--- a/test/lib.bash
+++ b/test/lib.bash
@@ -221,8 +221,8 @@ function run_rolling_upgrade_tests {
     --resolvabledomain \
     --https
 
-  # Delete the leftover services.
-  oc delete ksvc --all --cascade='foreground' -n serving-tests
+  # Delete the leftover namespace with services.
+  oc delete namespace serving-tests
 
   logger.success 'Upgrade tests passed'
 }


test/lib.bash
```diff
@@ -221,8 +221,8 @@ function run_rolling_upgrade_tests {
     --resolvabledomain \
     --https
 
-  # Delete the leftover services.
-  oc delete ksvc --all --cascade='foreground' -n serving-tests
+  # Delete the leftover namespace with services.
+  oc delete namespace serving-tests
```
markusthoemmes, line 225: So we keep iterating on this: Why is there services left here in the first place? Is some test not doing what it should?

mgencur: This is how upstream Serving tests work. They start the services in PreUpgrade phase and don't delete them in PostUpgrade phase. Maybe the PostUpgrade phase should delete them in upstream...

markusthoemmes: I'd think it absolutely should, yes.

mgencur: Alright. Can we merge this and then I can work upstream on deleting the services?

markusthoemmes (approved):
/lgtm
/approve

mgencur:
@markusthoemmes Thanks, I'm already working on this upstream: https://github.com/knative/serving/pull/12395

PR #1331, opened 2021-12-07
Remove fixed namespace from kourier-bootstrap

nak3:
When controller and gateway are deployed in the same namespace,
the namespace (`knative-serving-ingress`) should not need to be specified.

serverless-operator deploys kourier gateway and controller in the same
namespace so it could be omitted.


markusthoemmes (approved):
/lgtm
/approve

PR #1332, opened 2021-12-07
SRVCOM-1620 Reusable upgrade tests

mgencur:
https://issues.redhat.com/browse/SRVCOM-1620

* Allow calling this function directly from downstream wrapper scripts
which prevents script incompatibility, missed arguments and all kinds of
problems. It also simplifies downstream scripts a lot.
* Optionally call cluster upgrade tests, this is used downstream.
* Serverless upgrade and cluster upgrade can run one after another or
independently (widely used downstream)
* Remove saving the rootdir, no longer needed when test prepare scripts
are not called

### DIFF
--- a/test/lib.bash
+++ b/test/lib.bash
@@ -201,24 +201,23 @@ function run_rolling_upgrade_tests {
   export GO_TEST_VERBOSITY=standard-verbose
   export SYSTEM_NAMESPACE="$SERVING_NAMESPACE"
 
-  common_opts="./test/upgrade -tags=upgrade \
-    --kubeconfigs=${KUBECONFIG} \
-    --channels=${channels} \
-    --imagetemplate=${image_template} \
-    --catalogsource=${OLM_SOURCE} \
-    --upgradechannel=${OLM_UPGRADE_CHANNEL} \
-    --csv=${CURRENT_CSV} \
-    --servingversion=${KNATIVE_SERVING_VERSION} \
-    --eventingversion=${KNATIVE_EVENTING_VERSION} \
-    --kafkaversion=${KNATIVE_EVENTING_KAFKA_VERSION} \
+  common_opts=(./test/upgrade -tags=upgrade \
+    --kubeconfigs="${KUBECONFIG}" \
+    --channels="${channels}" \
+    --imagetemplate="${image_template}" \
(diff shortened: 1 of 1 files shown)


test/lib.bash
```diff
@@ -185,41 +185,54 @@ function downstream_monitoring_e2e_tests {
...
+    --kafkaversion=${KNATIVE_EVENTING_KAFKA_VERSION} \
     --resolvabledomain \
-    --https
+    --https"
+
+  if [[ "${UPGRADE_SERVERLESS}" == "true" ]]; then
+    # TODO: Remove creating the NS when this commit is backported: https://github.com/knative/serving/commit/1cc3a318e185926f5a408a8ec72371ba89167ee7
+    oc create namespace serving-tests
+    # shellcheck disable=SC2086
+    go_test_e2e -run=TestServerlessUpgrade -timeout=30m $common_opts
```
maschmid, line 221: would break if KUBECONFIG has a space in path?

test/lib.bash
```diff
@@ -185,41 +185,54 @@ function downstream_monitoring_e2e_tests {
...
+  fi
+
+  # For reuse in downstream test executions. Might be run after Serverless
+  # upgrade or independently.
+  if [[ "${UPGRADE_CLUSTER}" == "true" ]]; then
+    if oc get namespace serving-tests &>/dev/null; then
+      oc delete namespace serving-tests
```
skonto, line 228: Do we have to wait for deletion or we assume its empty?

test/lib.bash
mgencur: Probably. We've been using this approach in Eventing for a few releases (in e2e-commons.sh) and nobody complained. Curious we we need to account for this special case.

test/lib.bash
mgencur: This operation waits for the namespace to be gone (and so all its resources) before proceeding further.

test/lib.bash
mgencur: I could pull the KUBECONFIG param from the common params and pass it separately via `--kubeconfig="$KUBECONFIG"` .

mgencur: Or, let me try to use arrays...

mgencur: Pushed.

PR #1333, opened 2021-12-08
SRVCOM-1623 Separate tests and install

mgencur:
https://issues.redhat.com/browse/SRVCOM-1623 (see issue description for more information)

This should allow for:
* Quicker execution of E2E tests in CI due to removing redundant installs/uninstalls. Preliminary results show the runs are 3-4 minutes faster than before, for the operator-e2e check.
* Creating an image with tests which we can use to execute only tests, not installations. And all of that within OpenShift cluster

Changes:
* Use more specific Makefile targets. The ones with `-testonly` suffix only run tests. They do not install the product.
* Make it possible to contact kube API server from inside a Pod when creating new users for tests.
* Separate "system" and "test" namespaces and their creation.
* Move "Ensure no operators present" check from Golang to teardown script. The rest of the checks in teardown already included what we had in the Golang tests (there was redundancy).
* Make sure the "install.sh" scripts is called before tests and "teardown" after tests.
* Add `kafka-ch-dispatcher` among the tested deployments here: https://github.com/openshift-knative/serverless-operator/pull/1333/files#diff-a3c3e9170b364a9794db5994c3032aa25383b6816c9d3d3be51a52e8fc397660R23 (it was removed before but since the tests still create an extra Kafka channel to trigger creation of this kafka-ch-dispatcher they should also check it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,7 @@ test-e2e-testonly:
 
 test-e2e: install
 	./test/e2e-tests.sh
-	./hack/teardown
+	./hack/teardown.sh
 
 # Run E2E tests from the current repo for serving+eventing+knativeKafka
 test-e2e-with-kafka-testonly:
@@ -84,7 +84,6 @@ test-e2e-with-mesh-testonly:
 test-e2e-with-mesh: install-full-mesh
 	FULL_MESH=true ./hack/install.sh
 	FULL_MESH=true ./test/e2e-tests.sh
-	./hack/teardown.sh
 
 # Run both unit and E2E tests from the current repo.
 test-operator: test-unit test-e2e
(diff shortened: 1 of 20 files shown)


mgencur:
Looks like the operator tests are now faster by at least 3 minutes: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift-knative_serverless-operator/1333/pull-ci-openshift-knative-serverless-operator-main-4.9-operator-e2e-aws-ocp-49/1468567812397076480 (compare this run with previous ones)

maschmid:
So, my only concern is this would prevent us from adding additional variants of the CR creation tests in test/e2e  (e.g. testing with different options)  OTOH, if we want to keep the test/e2e quick (to be usable as a smoke test), we'd have to move such install/reinstall tests into a separate testsuite anyway... so 

/lgtm

mgencur:
@maschmid Thanks for the review. And yeah, testing with different variants of our custom resources would probably belong to a separate (specialized) test suite.

PR #1334, opened 2021-12-09
Patch sequence yaml

matzew:
Follow up for https://github.com/openshift/knative-eventing/pull/1549

/hold

once the midstream fix/backport is in, we need this patch for the manifest

/assign @pierDipi


pierDipi (approved):
/lgtm

PR #1337, opened 2021-12-13
kn-event sender image

cardil:
Adding kn-event sender image

Requires merge of https://github.com/openshift/release/pull/24476

### DIFF
--- a/hack/generate/csv.sh
+++ b/hack/generate/csv.sh
@@ -12,6 +12,8 @@ registry="registry.ci.openshift.org/openshift"
 serving="${registry}/knative-v$(metadata.get dependencies.serving):knative-serving"
 eventing="${registry}/knative-v$(metadata.get dependencies.eventing):knative-eventing"
 eventing_kafka="${registry}/knative-v$(metadata.get dependencies.eventing_kafka):knative-eventing-kafka"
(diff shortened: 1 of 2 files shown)


mgencur:
Just for my understanding. Is the image used by the Serverless operator on server side? Isn't it something like e.g. our `must-gather` image which we produce but it's not mentioned in CSV because users will call it directly (on client side)?

cardil:
At @Kaustubh-pande: As you can see, the CSV generation mechanics in Serverless operator adds `IMAGE_` to the front of the name of the image. In other places we have the same name but with `_IMAGE` as suffix instead.

WDYT of changing it in other places to this naming. In places like: https://github.com/openshift/knative-client/blob/f2b87473f6c69f413a2de69efeef82d51bf3c474/hack/build.sh.d/kn-plugin-event.sh#L7

cardil:
After changes in https://github.com/openshift/release/pull/24549 images was promoted properly:

```
$ skopeo inspect docker://registry.ci.openshift.org/knative/release-0.26:client-plugin-event-sender \
  | jq -r .Digest
sha256:4da4400a0c74d6d0fdd65d527c21bf5af187aa84d30de6d2dd593042fc1b7599
```

/unhold

PR #1359, opened 2022-01-04
Avoid Makefile target dependencies

mgencur:
Avoid issues like this one: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-knative-serverless-operator-main-4.9-e2e-aws-ocp-49-continuous/1478335600389001216
CI runs two makefile targets: make test-e2e-with-kafka test-upstream-e2e-no-upgrade. But they both depend on install-all which should be executed before both of the targets but it is only executed at the beginning.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,8 @@ install:
 install-operator:
 	INSTALL_SERVING="false" INSTALL_EVENTING="false" ./hack/install.sh
 
-install-all: install-strimzi
+install-all:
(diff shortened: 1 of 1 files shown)


mgencur:
Another option would be to create a specific makefile target for CI that would combine `test-e2e-with-kafka` `test-upstream-e2e-no-upgrade` in the right way. But I think the current solution is better.

pierDipi (approved):
/lgtm

PR #1360, opened 2022-01-04
Actually install Mesh for e2e tests

mgencur:
This is a follow-up to https://github.com/openshift-knative/serverless-operator/pull/1359 which had a small bug.


pierDipi (approved):
/lgtm

PR #1361, opened 2022-01-05
Splitting the broker resources into control-plane and data-plane folders

matzew:

Prefactorings for better plug-ability of future components.

For all Kafka-related bits (like the sink, or the future channel/source) we need the CP part of the "eventing-kafka-broker" repo,
and only based on selection we will be adding the data plane bits.

This PR lays out the foundation to enable the `KafkaSink`, by splitting the control-plane manifest into a separate folder, to make the actual Sink PR simpler


pierDipi (approved):
/lgtm

/hold waiting for all jobs to pass

PR #1363, opened 2022-01-05
Add disabled controllers

matzew:
another prefactor PR for enabling the sink.

We just want to have those controllers enabled, that we actually have "enabled". Following changes:
* patch the manifest to disable all controllers for broker/sink
* if broker is enabled, update the deployment and just keep the `sink` disabled

Follow up PR: Enabling the sink

### DIFF
--- a/knative-operator/deploy/resources/knativekafka/controller/1-eventing-kafka-controller.yaml
+++ b/knative-operator/deploy/resources/knativekafka/controller/1-eventing-kafka-controller.yaml
@@ -951,7 +951,7 @@ spec:
         - name: controller
           image: TO_BE_REPLACED
           args:
-            - "--disable-controllers=broker-controller, trigger-controller, sink-controller"
+            - "--disable-controllers=broker-controller,trigger-controller,sink-controller"
           imagePullPolicy: IfNotPresent
           env:
             - name: BROKER_DATA_PLANE_CONFIG_MAP_NAMESPACE
--- a/knative-operator/deploy/resources/knativekafka/controller/1-eventing-kafka-controller.yaml
+++ b/knative-operator/deploy/resources/knativekafka/controller/1-eventing-kafka-controller.yaml
@@ -950,6 +950,8 @@ spec:
       containers:
         - name: controller
           image: TO_BE_REPLACED
+          args:
+            - "--disable-controllers=broker-controller, trigger-controller, sink-controller"
           imagePullPolicy: IfNotPresent
           env:
             - name: BROKER_DATA_PLANE_CONFIG_MAP_NAMESPACE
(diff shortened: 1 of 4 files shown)


knative-operator/deploy/resources/knativekafka/controller/1-eventing-kafka-controller.yaml
```diff
@@ -950,6 +950,8 @@ spec:
       containers:
         - name: controller
           image: TO_BE_REPLACED
+          args:
+            - "--disable-controllers=broker-controller, trigger-controller, sink-controller"
```
pierDipi, line 954: Spaces aren't allowed:
```suggestion
            - "--disable-controllers=broker-controller,trigger-controller,sink-controller"
```

knative-operator/hack/002-broker-disabled-controllers.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/knative-operator/deploy/resources/knativekafka/controller/1-eventing-kafka-controller.yaml b/knative-operator/deploy/resources/knativekafka/controller/1-eventing-kafka-controller.yaml
+index 67c586a06..93388a995 100644
+--- a/knative-operator/deploy/resources/knativekafka/controller/1-eventing-kafka-controller.yaml
++++ b/knative-operator/deploy/resources/knativekafka/controller/1-eventing-kafka-controller.yaml
+@@ -950,6 +950,8 @@ spec:
+       containers:
+         - name: controller
+           image: TO_BE_REPLACED
++          args:
++            - "--disable-controllers=broker-controller, trigger-controller, sink-controller"
```
pierDipi, line 10: Spaces aren't allowed:
```suggestion
+          args:
+            - "--disable-controllers=broker-controller,trigger-controller,sink-controller"
```

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -475,6 +475,20 @@ func updateEventingKafka(kafkachannel serverlessoperatorv1alpha1.Channel) mf.Tra
...
+		// patch the deployment and enable the relevant controllers
+		if u.GetKind() == "Deployment" && u.GetName() == "kafka-controller" {
+
+			var deployment = &appsv1.Deployment{}
+			if err := scheme.Scheme.Convert(u, deployment, nil); err != nil {
+				return err
+			}
+
+			// we just keep the sink controller disabled
+			deployment.Spec.Template.Spec.Containers[0].Args = []string{"--disable-controllers=sink-controller"}
```
pierDipi, line 487: Shouldn't this be based on the sink or broker flags?

matzew: yes, but we have no sink flag yet - that will be on the follow up PR.

Now, when once install from master: all is installed. 

this PR changes that, to only enalbe the broker/trigger (and keep the sink disabled).

The follow up PR, will tweak around this section, but I want to land a little better separation (as I now do get the sink, which I do not want)

pierDipi: OK

/lgtm

matzew:
setting `broker.enabled`, I get this on the `Deployment`:

```
...
      containers:
      - args:
        - --disable-controllers=sink-controller
...
```

So, if we decide to _not_ ship the sink in 1.21 - we are safe here now :sweat_smile:

PR #1364, opened 2022-01-05
Remove duplicated Kafka CRDs 

matzew:
/hold

needs to go in _after_ #1363

See https://github.com/openshift-knative/serverless-operator/commit/38e305b2640d1cc2a9e79a38fa8a606a41047a87 for details

We remove the duplicated CRDs for:
* KafkaBindings (deprecated)
* KafkaSource
* KafkaChannel

Since we do not have any direct usage of those _yet_, on the "broker" component 

All three CRDs (`KafkaBinding`, `KafkaSource` and `KafkaChannel`) are still shipped via the "eventing-kafka" repo bits.

**NOTE:** This is only for the broker repo, which ships them too, for some reasons


matzew:
/unhold
I've rebased the PR /cc @pierDipi

pierDipi (approved):
/lgtm

PR #1365, opened 2022-01-06
Changing name of functions

matzew:

A bit of a nit, but changing the names of different "configure kafkas" functions to reflect the new bits and the current/legacy bits different configuration approach

/assign @pierDipi


aliok (approved):
/lgtm
/approve

PR #1366, opened 2022-01-06
Adding toggle for kafka sink enablement

matzew:

Adding the raw `KafkaSink` as a new tool for the serverless operator.

Follow-up PR: adding some smoke tests for the `KafkaSink` type, and enable it on test deployment

### DIFF
--- a/knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
+++ b/knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
@@ -6,9 +6,10 @@ import (
 	"os"
 	"strconv"
 
+	mfc "github.com/manifestival/controller-runtime-client"
+
 	operatorv1alpha1 "knative.dev/operator/pkg/apis/operator/v1alpha1"
 
-	mfc "github.com/manifestival/controller-runtime-client"
 	mf "github.com/manifestival/manifestival"
 	serverlessoperatorv1alpha1 "github.com/openshift-knative/serverless-operator/knative-operator/pkg/apis/operator/v1alpha1"
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/common"
@@ -78,10 +79,15 @@ func newReconciler(mgr manager.Manager) (*ReconcileKnativeKafka, error) {
 
 	kafkaControllerManifest, err := mf.ManifestFrom(mf.Path(os.Getenv("KAFKACONTROLLER_MANIFEST_PATH")))
 	if err != nil {
-		return nil, fmt.Errorf("failed to load KafkaBroker manifest: %w", err)
+		return nil, fmt.Errorf("failed to load Kafka Control-Plane manifest: %w", err)
 	}
 
 	kafkaBrokerManifest, err := mf.ManifestFrom(mf.Path(os.Getenv("KAFKABROKER_MANIFEST_PATH")))
+	if err != nil {
+		return nil, fmt.Errorf("failed to load KafkaBroker manifest: %w", err)
+	}
+
+	kafkaSinkManifest, err := mf.ManifestFrom(mf.Path(os.Getenv("KAFKASINK_MANIFEST_PATH")))
 	if err != nil {
 		return nil, fmt.Errorf("failed to load KafkaBroker manifest: %w", err)
 	}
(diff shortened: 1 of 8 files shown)


knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -483,8 +506,17 @@ func configureEventingKafka(spec serverlessoperatorv1alpha1.KnativeKafkaSpec) mf
...
+			if spec.Broker.Enabled && spec.Sink.Enabled {
+				// all: nothing to disable
+				deployment.Spec.Template.Spec.Containers[0].Args = nil
+			} else if spec.Broker.Enabled {
+				// only Broker: we disable the sink controllers
+				deployment.Spec.Template.Spec.Containers[0].Args = []string{"--disable-controllers=sink-controller"}
+			} else if spec.Sink.Enabled {
+				// only sink: we disable the Broker controllers
+				deployment.Spec.Template.Spec.Containers[0].Args = []string{"--disable-controllers=broker-controller,trigger-controller"}
+			}
```
pierDipi, line 518: As we add more controllers (the new channel, new source, and scheduler controllers) this `if` will get complex since we need every combination.

I'm thinking to give this a structure as follows (see also inline comments):

```golang
brokerControllers = []string{"broker-controller, trigger-controller}
sinkControllers = []string{"sink-controller"}

disabledControllers := make([]string, 0, len(brokerControllers) + len(sinkControllers))

if !spec.Broker.Enabled {
  disabledControllers = append(disabledControllers, brokerControllers...)
}

if !spec.Sink.Enabled {
  disabledControllers = append(disabledControllers, sinkControllers...)
}

// ...

if len(disabledController) == 0 {
  // Note, `deployment.Spec.Template.Spec.Containers[0].Args = nil` is wrong since we might have other args defined
  // we need to find the disabled controllers flag and if defined we need to remove it from the list.
} else {
  // Note, we append arguments instead of overriding them since there might be already `Args` defined.
  deployment.Spec.Template.Spec.Containers[0].Args = append(deployment.Spec.Template.Spec.Containers[0].Args, "--disabled-controllers="+strings.Join(disabledControllers, ","))
}
```

what do you think?

pierDipi: Actually, this is wrong too:
```golang
else {
  // Note, we append arguments instead of overriding them since there might be already `Args` defined.
  deployment.Spec.Template.Spec.Containers[0].Args = append(deployment.Spec.Template.Spec.Containers[0].Args, "--disabled-controllers="+strings.Join(disabledControllers, ","))
}
```

We need to find the disabled controllers flag and if defined we need to update it accordingly in the else too.

matzew:
```
  disabledControllers = append(disabledControllers, brokerControllers...)
```

I thought about that also.  I had a slice in mind, like:

disabledControllers {all controllers in there}

and than, based on enablement remove them. 

For now the toggle is OK, but w/ the Source and channel, it becomes nasty

matzew: > deployment.Spec.Template.Spec.Containers[0].Args = nil
nothing else is currently supported - and we own that bit, so not that bad

pierDipi: > nothing else is currently supported - and we own that bit, so not that bad

It's easy to forget about this in the future.

pierDipi: Do you want to keep this logic as it is for now?

pierDipi: But I think we should do what we have discussed here, so let's make sure we capture this somewhere.

matzew: @pierDipi I recently saw a `StringMap` in the knative `kn` client. 

I think we want something like this, in a follow up PR: 
https://go.dev/play/p/QEzd5shVxr2

than we also address the "not nil" for `Args` - ok?

pierDipi: /lgtm

PR #1370, opened 2022-01-07
SRVKE-1079 Enable Broker continual and pre-upgrade tests

mgencur:
https://issues.redhat.com/browse/SRVKE-1079

### DIFF
--- a/test/upgrade/upgrade_test.go
+++ b/test/upgrade/upgrade_test.go
@@ -63,10 +63,6 @@ func TestServerlessUpgrade(t *testing.T) {
 					if err := installation.UpgradeServerless(ctx); err != nil {
 						c.T.Error("Serverless upgrade failed:", err)
 					}
-					// Remove this when upgrading from Serverless 1.20 to 1.21.
-					if err := installation.EnableKafkaBroker(ctx); err != nil {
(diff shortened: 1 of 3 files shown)


test/upgrade/upgrade_test.go
```diff
@@ -54,6 +54,7 @@ func TestServerlessUpgrade(t *testing.T) {
 				},
 				kafkaupgrade.ChannelContinualTests(continual.ChannelTestOptions{}),
 				kafkaupgrade.SourceContinualTests(continual.SourceTestOptions{}),
+				kafkabrokerupgrade.BrokerContinualTests(),
```
pierDipi, line 57: should we delete this too and maybe enable the broker from the start of the test?
```golang
					// Remove this when upgrading from Serverless 1.20 to 1.21.
					if err := installation.EnableKafkaBroker(ctx); err != nil {
						c.T.Error("Failed to enable Kafka Broker on KnativeKafka resource:", err)
					}
```

mgencur: Right, I was too quick. Thanks

pierDipi (approved):
Thanks Martin!

/lgtm

PR #1372, opened 2022-01-10
Remove the patch for the disabled controllers

matzew:

as per title: Removing unneeded patch for the kafka controller manifest 

/assign @pierDipi


pierDipi (approved):
/lgtm

Thanks!

PR #1374, opened 2022-01-12
Adding `config` field, leveraging tested upstream operator APIs

matzew:

/assign @mgencur

### DIFF
--- a/knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
+++ b/knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
@@ -6,6 +6,9 @@ import (
 	"os"
 	"strconv"
 
+	operatorcommon "knative.dev/operator/pkg/reconciler/common"
+	"knative.dev/pkg/logging"
+
 	mfc "github.com/manifestival/controller-runtime-client"
 
 	operatorv1alpha1 "knative.dev/operator/pkg/apis/operator/v1alpha1"
@@ -272,6 +275,7 @@ func (r *ReconcileKnativeKafka) transform(manifest *mf.Manifest, instance *serve
 		}),
 		setKafkaDeployments(instance.Spec.HighAvailability.Replicas),
 		configureLegacyEventingKafka(instance.Spec.Channel),
+		operatorcommon.ConfigMapTransform(instance.Spec.Config, logging.FromContext(context.TODO())),
 		configureEventingKafka(instance.Spec),
 		ImageTransform(common.BuildImageOverrideMapFromEnviron(os.Environ(), "KAFKA_IMAGE_"), log),
 		replicasTransform(manifest.Client),
--- a/knative-operator/pkg/apis/operator/v1alpha1/knativekafka_types.go
+++ b/knative-operator/pkg/apis/operator/v1alpha1/knativekafka_types.go
@@ -23,6 +23,10 @@ type KnativeKafkaSpec struct {
 	// +optional
 	Channel Channel `json:"channel,omitempty"`
 
+	// A means to override the corresponding entries in the upstream configmaps
+	// +optional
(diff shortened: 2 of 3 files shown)


mgencur:
Related to https://issues.redhat.com/browse/SRVCOM-1630

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -274,6 +277,7 @@ func (r *ReconcileKnativeKafka) transform(manifest *mf.Manifest, instance *serve
 		configureLegacyEventingKafka(instance.Spec.Channel),
 		configureEventingKafka(instance.Spec),
 		ImageTransform(common.BuildImageOverrideMapFromEnviron(os.Environ(), "KAFKA_IMAGE_"), log),
+		operatorcommon.ConfigMapTransform(instance.Spec.Config, logging.FromContext(context.TODO())),
```
mgencur, line 280: It is now possible to configure kafka-broker-config in two ways from the same KnativeKafka config and the new way you just implemented takes precedence:
```spec:
  broker:
    enabled: true
    defaultConfig:
      bootstrapServers: "my-cluster-kafka-bootstrap.kafka:9092"
      numPartitions: "10"
      replicationFactor: "3"
  config:
    kafka-broker-config:
      default.topic.partitions: "10"
      default.topic.replication.factor: "3"
      bootstrap.servers: "my-cluster-kafka-bootstrap.kafka:9092"
```
I think this is alright. Just curious if you've thought about it.

mgencur: Maybe there could be a warning if both broker.defaultConfig and config.kafka-broker-config exist. But not necessary

matzew: I could try exclude the `kafka-broker-config` CM from that `config` thingy :thinking: 
/cc @pierDipi 

OR :-) We remove the `defaultConfig` bits? But I think those are more "first class", than the CM messy config syntax?

mgencur: The CM syntax is awful, yeah. For example, the channel configuration is much better with the KnativeKafka CR bits:
```
spec:
  channel:
    authSecretName
```
vs.
```
spec:
  config:
    config-kafka:
      eventing-kafka: |
        kafka:
          authSecretName: kafka-cluster
```
I was thinking about configuring as much as possible through the ConfigMap syntax and removing the [legacy spec.channel](https://github.com/openshift-knative/serverless-operator/blob/main/knative-operator/pkg/controller/knativekafka/knativekafka_controller.go#L471) configuration as well. But I'm not so sure now.
Perhaps it makes sense to keep the current "first class" configs and also allow to configure anything else by the ConfigMap syntax. I'd probably keep the current PR as is then.

pierDipi: I think this is alright. 
What I'd expect is that `defaultConfig` has priority over the other `config.kafka-broker-config` if they are both set.

pierDipi: I'd merge the 2 configurations if they are both set.

mgencur: > I think this is alright.
> What I'd expect is that `defaultConfig` has priority over the other `config.kafka-broker-config` if they are both set.

Then I guess, the new stuff must be registered *before* the `configureEventingKafka(instance.Spec) `

matzew: @mgencur I updated the PR and registered the ConfigMapTransformer _before_

PR #1375, opened 2022-01-13
Increase FinishedSleep for eventing upgrade tests

mgencur:
* Sometimes the wathola fetcher is fetching events from wathola-receiver
even before wathola-receiver gets all the events.
Example error: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/25022/rehearse-25022-periodic-ci-openshift-knative-serverless-operator-main-4.10-upgrade-tests-aws-ocp-410-continuous/1481535296045584384 (the logs of wathola-receiver and wathola-fetcher show bad timing for fetching events)


pierDipi (commented):
Should we change this upstream as well?

mgencur:
> Should we change this upstream as well?
Probably not, it's all about timing and it seems this problem doesn't happen upstream (the product upgrade might be faster there and upstream also runs fewer tests in parallel that would be using Kafka cluster).

mgencur:
/hold cancel

CI passed

PR #1376, opened 2022-01-18
Verify net-istio deployments when FULL_MESH is true

nak3:
When enabled istio, serverless operator test has to verify `net-istio-*` deployments instead of kourier deployments.

This patch fixes it.

Note: Previously the issue did not happen because `serverless_operator_e2e_tests` was called before enabling net-istio.
https://github.com/openshift-knative/serverless-operator/blob/release-1.20/test/e2e-tests.sh#L23

### DIFF
--- a/test/e2e/knative_serving_test.go
+++ b/test/e2e/knative_serving_test.go
@@ -1,6 +1,7 @@
 package e2e
 
 import (
+	"os"
 	"testing"
 
 	"github.com/openshift-knative/serverless-operator/test"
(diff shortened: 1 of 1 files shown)


nak3:
/cc @skonto @mgencur 

I have found another nit https://github.com/openshift-knative/serverless-operator/issues/1377 but should be fixed in a different PR.

nak3:
> Note: Previously the issue did not happen because serverless_operator_e2e_tests was called before enabling net-istio.
https://github.com/openshift-knative/serverless-operator/blob/release-1.20/test/e2e-tests.sh#L23

Hmm... I assumed that but https://github.com/openshift-knative/serverless-operator/pull/1360 ran `/test 4.9-upstream-e2e-mesh-aws-ocp-49` :thinking:

test/e2e/knative_serving_test.go
```diff
@@ -36,17 +37,35 @@ func TestKnativeServing(t *testing.T) {
...
+					t.Fatalf("Failed to verify default HA settings for %q: %v", deployment, err)
+				}
 			}
-		}
-		// Check the status of deployments in the ingress namespace.
-		for _, deployment := range ingressDeployments {
-			if _, err := test.WithDeploymentReady(caCtx, deployment, servingNamespace+"-ingress"); err != nil {
-				t.Fatalf("Deployment %s is not ready: %v", deployment, err)
+			// Check the status of ingress deployments in the knative serving namespace
+			for _, deployment := range []string{"net-istio-controller", "net-istio-webhook"} {
```
rhuss, line 51: would it make sense to combine both loops to a single loop as you always have to iterate over all deployments, so if the number of deployments changes you are not in danger to forgot to update one loop ?

rhuss: or put the deployments in a slice and reuse it for both loops, like it is done below for the kourier case

nak3: net-kourier deployments are deployed in `knative-serving-ingress` ns while `net-istio` is deployed in `knative-serving` ns.
So we need to different `test.CheckDeploymentScale(caCtx, servingNamespace)` vs `test.CheckDeploymentScale(caCtx, servingNamespace+"-ingress")`, deployment, haReplicas)` and it cannot make the combined loop simple, I think.

nak3: But give me a sec. I will try...

nak3: Thank you! Updated.

nak3:
/retest

The unit test error is not related to this PR. I pinged on Slack https://coreos.slack.com/archives/CD87JDUB0/p1642552817074500

test/e2e/knative_serving_test.go
```diff
@@ -36,16 +37,25 @@ func TestKnativeServing(t *testing.T) {
...
 		for _, deployment := range ingressDeployments {
-			if err := test.CheckDeploymentScale(caCtx, servingNamespace+"-ingress", deployment, haReplicas); err != nil {
+			if err := test.CheckDeploymentScale(caCtx, ingressNamespace, deployment, haReplicas); err != nil {
 				t.Fatalf("Failed to verify default HA settings: %v", err)
 			}
 		}
 		// Check the status of deployments in the ingress namespace.
 		for _, deployment := range ingressDeployments {
-			if _, err := test.WithDeploymentReady(caCtx, deployment, servingNamespace+"-ingress"); err != nil {
+			if _, err := test.WithDeploymentReady(caCtx, deployment, ingressNamespace); err != nil {
```
skonto, line 58: Can't we have one loop here?

nak3: Yes, we can. But it should be aligned with other codes like above L27-39  and the same test in `knative_eventing_test.go`. So if we want to change it, I think another refactoring PR is better.

skonto: ok no problem let's get this in.

mgencur: The above lines L27-39 have two loops because the sets of pods are different in each loop. I believe for this ingress stuff we can have just one loop because the sets are same. Btw, the sets are also different in Eventing tests. So, this is probably the only place where it can be merged into one loop.

mgencur: But it can be done in a different PR, of course :)

skonto: Indeed @mgencur, @nak3 if you can fix it now would be nice.

nak3: > the sets of pods are different in each loop

I think we can use the same set of pods(deployments) now. Upstream operator supports all deployments by HA config. I have opened for it in https://github.com/openshift-knative/serverless-operator/issues/1377

skonto: ok let's deal with this in another pr.

PR #1379, opened 2022-01-19
Ensure ordering in test

matzew:

Tmp quick fix for CI nightly/periodic:
since on the runtime the order is pretty much  irrelevant, I change the test to ensure the order of "all" disabled controllers.

/assign @pierDipi

### DIFF
--- a/knative-operator/pkg/controller/knativekafka/knativekafka_controller_test.go
+++ b/knative-operator/pkg/controller/knativekafka/knativekafka_controller_test.go
@@ -584,7 +584,8 @@ func TestDisabledControllers(t *testing.T) {
 	}}
 
 	for _, test := range tests {
-		defaultDeployment := makeEventingKafkaDeployment(t, "")
+		// by default we assume all disabled:
+		defaultDeployment := makeEventingKafkaDeployment(t, "broker-controller,trigger-controller,sink-controller")
 		t.Run(test.name, func(t *testing.T) {
 			err := configureEventingKafka(test.knativeKafka)(defaultDeployment)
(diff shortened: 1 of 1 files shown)


knative-operator/pkg/controller/knativekafka/knativekafka_controller_test.go
```diff
@@ -584,7 +584,8 @@ func TestDisabledControllers(t *testing.T) {
 	}}
 
 	for _, test := range tests {
-		defaultDeployment := makeEventingKafkaDeployment(t, "")
+		// by default we assume all disabled:
+		defaultDeployment := makeEventingKafkaDeployment(t, "broker-controller,trigger-controller,sink-controller")
```
pierDipi, line 588: Do you think this fixes the problem?

Isn't the issue about [`StringValues`](https://github.com/openshift-knative/serverless-operator/blob/a482801e95f7b24a70f3e67ad92fc882fd1ca59e/knative-operator/pkg/controller/knativekafka/knativekafka_controller.go#L531) that might return results in a different order since [iterating a map](https://github.com/openshift-knative/serverless-operator/blob/a482801e95f7b24a70f3e67ad92fc882fd1ca59e/knative-operator/pkg/common/util.go#L30-L38) isn't guaranteed to be done in order?

matzew: I know, but the order is irrelvant, on run time. This is really to fix the flake. as a work-around

pierDipi (approved):
/lgtm
/approve

PR #1385, opened 2022-01-27
[SRVKS-854] Remove domain-mapping replicas=2 from test yaml

nak3:
Since domain-mapping has been included in `spec.replicas` by https://github.com/knative/operator/commit/ba3e058a20893580a98a731080bd7b20433eaa6b, we do not need to set it in yaml.


skonto:
Let's merge so we can also test with the other PR.

PR #1390, opened 2022-01-27
[prefactor] Separate Eventing-Kafka and Eventing-Kafka-Broker versions

matzew:

Since `eventing-kafka` and `eventing-kafka-broker` are different repos, they should be separated by their versions.

Since for 1.20 and current `main`, we had a match between, this issue was overseen. Preparing work for 1.22 branch, this issue did show up :sweat_smile:


pierDipi (approved):
/lgtm

PR #1399, opened 2022-02-15
kntest moved to tools

maschmid:
Looks like kntest has been moved, https://github.com/knative/test-infra/commit/1715ad474811ff608d1581adba5044959ab60592

causing 

```
go get: module knative.dev/test-infra@upgrade found (v0.0.0-20220215044403-4226a7ba1173), but does not contain package knative.dev/test-infra/kntest/cmd/kntest
```

Attepting to fix by updating the path.


mgencur:
Merged manually as the changes will be visible/effective only after the build-image is pushed after running post-submit job.

PR #1402, opened 2022-02-15
Patch kafka broker manifests

pierDipi:
In midstream, we're using v1.2.2 code, so we need
additional env variables for the controller and 
read-only access to brokers for the webhook

### DIFF
--- a/knative-operator/hack/update-manifests.sh
+++ b/knative-operator/hack/update-manifests.sh
@@ -68,3 +68,6 @@ git apply "$root/knative-operator/hack/001-broker-config-tracing.patch"
 
 # For now we remove the CRDs, since the "broker" does not yet do anything with them
 git apply "$root/knative-operator/hack/003-broker-remove-duplicated-crds.patch"
+
+# For now we need to add prober env variables.
+git apply "$root/knative-operator/hack/004-kafka-broker-prober-env.patch"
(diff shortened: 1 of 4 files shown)


matzew:
We are currently on the broker 1.0.0 version.
usually we align all `knative` versions, per release 
Not sure what exactly is needed here

matzew:
Let's just patch for the ports

knative-operator/hack/update-manifests.sh
```diff
@@ -68,3 +68,9 @@ git apply "$root/knative-operator/hack/001-broker-config-tracing.patch"
 
 # For now we remove the CRDs, since the "broker" does not yet do anything with them
 git apply "$root/knative-operator/hack/003-broker-remove-duplicated-crds.patch"
+
+# For now we need to add prober env variables.
+git apply "$root/knative-operator/hack/004-kafka-broker-prober-env.patch"
+
+# For now we need to add broker read access to the webhook.
+git apply "$root/knative-operator/hack/005-kafka-broker-webhook-role.patch"
```
matzew, line 76: doing the patches, for "matching" upstream manifest seems to exactly what we want. 

Thanks for going that route after we discussed

matzew (approved):
/lgtm
/approve

PR #1410, opened 2022-02-21
SRVCOM-1666 update rest of the quickstart links to 4.9

maschmid:


knative-operator/deploy/resources/quickstart/serverless-application-quickstart.yaml
```diff
@@ -213,4 +213,4 @@ spec:
       summary:
         failed: Try the steps again.
         success: You just deleted your Serverless app!
-      title: Deleting your application
\ No newline at end of file
+      title: Deleting your application
```
skonto, line 216: If you want to keep the last part as is one option is:
```
 perl -pi -e 'chomp if eof' <filename>
```

PR #1418, opened 2022-02-23
Disable Channel controller

matzew:

add the `channel-controller` to the list of _disabled_ controllers - since we do not do any kafkachannel yet...

/assign @pierDipi


matzew:
/close ...

that one comes with 1.2 ... I was on the wrong branch :sweat_smile:

PR #1420, opened 2022-02-24
Go style fixes

skonto:
As per title. It covers:
- unused params. Methods that implement some interface will keep for now the unused param. We could however replace with `_` or consider simplifying them in another round.
- unhandled errors
- deprecations


matzew (approved):
/lgtm
/approve

failure is not related ...

PR #1414, opened 2022-02-22
[SRVKE-1152] Land new java kafka source

matzew:
https://issues.redhat.com/browse/SRVKE-1152

PR introduces Java-based kafka source from `broker` repo, and a migration job.

Based on:
* https://github.com/openshift-knative/serverless-operator/pull/1393
* https://github.com/openshift-knative/serverless-operator/pull/1413
* https://github.com/openshift-knative/serverless-operator/pull/1417

Summary, with a working 1.1.x of Knative Eventing, we move on to _replace_ the source implementation :tada:

PR #1423, opened 2022-03-01
Migrate KafkaSource dashboards to Broker components

pierDipi:
- Enable RBAC proxy injection for KafkaSource (temporarly
  disabled in #1414
- Use the new metric names for KafkaSource

Images:
kafka-source-dashboards-namespace-selection

kafka-source-metrics-dashboards

### DIFF
--- a/knative-operator/deploy/resources/dashboards/eventing/grafana-dash-knative-eventing-source.yaml
+++ b/knative-operator/deploy/resources/dashboards/eventing/grafana-dash-knative-eventing-source.yaml
@@ -866,7 +866,7 @@ data:
              "tableColumn": "",
              "targets": [
                {
-                 "expr": "sum(rate(kafkasource_event_count{namespace_name=\"$eventNamespace\"}[1m]))",
+                 "expr": "sum(rate(event_count_1_total{job=\"kafka-source-dispatcher-sm-service\", namespace=\"knative-eventing\", namespace_name=\"$eventNamespace\"}[1m]))",
                  "format": "time_series",
                  "instant": false,
                  "refId": "A"
@@ -875,7 +875,7 @@ data:
              "thresholds": "",
              "timeFrom": null,
              "timeShift": null,
-             "title": "Kafka: Event Count (avg/sec, over 1m window)",
+             "title": "KafkaSource: Event Count (avg/sec, over 1m window)",
              "type": "singlestat",
              "valueFontSize": "80%",
              "valueMaps": [
@@ -954,7 +954,7 @@ data:
              "tableColumn": "",
              "targets": [
                {
-                 "expr": "sum(rate(kafkasource_event_count{namespace_name=\"$eventNamespace\", response_code_class=\"2xx\"}[1m])) / sum(rate(kafkasource_event_count{namespace_name=\"$eventNamespace\"}[1m]))",
+                 "expr": "sum(rate(event_count_1_total{job=\"kafka-source-dispatcher-sm-service\", namespace=\"knative-eventing\", namespace_name=\"$eventNamespace\", response_code_class=\"2xx\"}[1m])) / sum(rate(event_count_1_total{job=\"kafka-source-dispatcher-sm-service\", namespace=\"knative-eventing\", namespace_name=\"$eventNamespace\"}[1m]))",
                  "format": "time_series",
                  "instant": false,
                  "refId": "A"
@@ -1020,7 +1020,7 @@ data:
              "steppedLine": false,
              "targets": [
                {
-                 "expr": "sum(rate(kafkasource_event_count{namespace_name=\"$eventNamespace\"}[1m])) by (response_code_class)",
+                 "expr": "sum(rate(event_count_1_total{job=\"kafka-source-dispatcher-sm-service\", namespace=\"knative-eventing\", namespace_name=\"$eventNamespace\"}[1m])) by (response_code_class)",
                  "format": "time_series",
                  "hide": false,
                  "instant": false,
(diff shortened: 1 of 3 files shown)


knative-operator/deploy/resources/dashboards/eventing/grafana-dash-knative-eventing-source.yaml
```diff
@@ -875,7 +875,7 @@ data:
              "thresholds": "",
              "timeFrom": null,
              "timeShift": null,
-             "title": "Kafka: Event Count (avg/sec, over 1m window)",
+             "title": "KafkaSource: Event Count (avg/sec, over 1m window)",
```
pierDipi, line 878: I changed this for consistency with other titles.

knative-operator/deploy/resources/dashboards/eventing/grafana-dash-knative-eventing-source.yaml
```diff
@@ -866,7 +866,7 @@ data:
              "tableColumn": "",
              "targets": [
                {
-                 "expr": "sum(rate(kafkasource_event_count{namespace_name=\"$eventNamespace\"}[1m]))",
+                 "expr": "sum(rate(event_count_1_total{job=\"kafka-source-dispatcher-sm-service\", namespace=\"knative-eventing\", namespace_name=\"$eventNamespace\"}[1m]))",
```
skonto, line 869: Upstream there was a PR/discussion for making the component optional as a prefix of the metric name. The long term replacement would be to pass it as a label. Is `event_count_1_total` the official name of the metric? Is the metrics API captured somewhere to be able to review it too eg. knative docs?

pierDipi: > Upstream there was a PR/discussion for making the component optional as a prefix of the metric name. The long term replacement would be to pass it as a label.

For these components, we're using a different monitoring stack/library (micrometer), and as Knative changes the monitoring API, we will change them separately for these components.

In general, I'm +1 on this since the `kafkasource` prefix (or any other component specific prefix) doesn't make a lot of sense

> Is event_count_1_total the official name of the metric?

On this particular metric, I tried really hard in many ways to make it like other Knative components but I haven't found a way to make Micrometer behaves the way I want on naming.

The tldr is:
In these components, we have the problem that we're translating Kafka clients metrics (JMX-based naming) to the Prometheus naming convention and, at the same time, `event_count_1_total` is (custom) defined as `event_count` but the transformer adds the `_total` suffix and the unit (dimensionless) `_1`.
So, if I disable the Prometheus naming translation, I get errors due to JMX metrics not being scrapeables since they contain characters that Prometheus doesn't allow, so we need to keep this `_1_total` suffix.

pierDipi: > Is the metrics API captured somewhere to be able to review it too eg. knative docs?

I still need to document this upstream.

pierDipi: > Is event_count_1_total the official name of the metric?

The only docs that I found is this https://knative.dev/docs/eventing/observability/metrics/eventing-metrics/#broker-filter
The official name is `event_count` and it is used for sources, brokers, channels, etc.

skonto:
@pierDipi Verified that data do show up in the dashboard as expected:
image
My concern is with the metric name that is a bit "cryptic" imho.

matzew:
> My concern is with the metric name that is a bit "cryptic" imho.

cryptic, as in?

skonto:
@matzew it has a counter that does not seem consistent with current naming (I suspect it comes from an auto-increment?).
Also the prefix of the component is missing (not consistent with the current metrics api). Metrics API is a contract with customer tooling as well. I would like to review it before exposing it.

pierDipi:
> @matzew it has a counter (substring) that is not consistent with current metrics naming (I suspect it comes from an auto-increment?). Also the prefix of the component is missing (not consistent with the current metrics api). Metrics API is a contract with customer tooling as well. Let's document the document/review the metrics before exposing it via dashboards. I was asking about it in a previous PR as well.

See my comment https://github.com/openshift-knative/serverless-operator/pull/1423#discussion_r816679957

pierDipi:
@skonto all Kafka components will use this new naming as we migrate all components to use this new data plane.

I'm not following what the real problem is.

Prometheus naming best practises is to use (see https://prometheus.io/docs/practices/naming/), like for counters:
```
process_cpu_seconds_total
```
scheme is:
```
<metric_name>_<unit>_total
```
so the metric
```
event_count_1_total
```
is following it.

I'd argue that the existing metrics don't follow best practises for Prometheus metrics naming

pierDipi:
> If you feel strongly about this, I can remove the `1` in `event_count_1_total`.
> 
> Let me add more details on why is `1`.
> 
> The [unified code for units of measure](https://ucum.org/ucum.html) defines units and in one section "PREFIXES AND UNITS USED IN INFORMATION TECHNOLOGY" they write:
> 
> > ... Therefore we define a second symbol for bit, without the suffix, to be the dimensionless unit 1.

Of course, this is only for the `<unit>` part in the scheme `<metric_name>_<unit>_total`

knative-operator/deploy/resources/dashboards/eventing/grafana-dash-knative-eventing-source.yaml
skonto: Ok I see there is an open issue for this https://github.com/micrometer-metrics/micrometer/issues/2881. Looking at the [micrometer code](https://github.com/micrometer-metrics/micrometer/blob/598a6b4efe3dcf32c5b5ae03bdc5e73a45a1b697/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusNamingConvention.java#L58) they implement the [OpenMetrics standard](https://github.com/micrometer-metrics/micrometer/issues/2482#issue-820964481) directly.
At the Knative side (pkg) we dont impose this since we use opencensus client lib where we map metrics to different exporters. The prometheus exporter there does not enforce this as it seems you need to [explicit enable the standard](https://github.com/prometheus/client_golang/issues/942) at the golang client for [some scenarios] (https://github.com/prometheus/client_golang/blob/main/prometheus/promhttp/http.go#L355-L364).
Opencensus does not add any options [there](https://github.com/knative/pkg/blob/fe264173447bda1c4215a5ef7a010958ec5d7596/vendor/contrib.go.opencensus.io/exporter/prometheus/prometheus.go#L70).

Omitting the [suffixes](https://github.com/prometheus/client_golang/issues/725#issuecomment-598069346) above is pretty valid although not encouraged. 
The problem is that imho as a Knative ecosystem we need to have a consistent API and follow one or more standards. It is a bit weird to have different conventions and standards per component. So we either enforce Openmetrics at the golang side for Knative pkg too or change micrometer to make the suffixes optional. I guess the former is more realistic given the Openmetrics Standard evolution.

I strongly recommend we discuss this upstream and one opportunity is the documentation PR. 

 > The official name is event_count and it is used for sources, brokers, channels, etc.

My point is that this needs to be documented upstream (as you mentioned). I am sure users will ask the same because they might expect the current format `{component}_event_count{}`.

PR #1425, opened 2022-03-02
[SRVKS-877] Add common label `knative.openshift.io/part-of: openshift-serverless` for namespaces deployed by operator

nak3:
This patch adds common label `knative.openshift.io/part-of:
openshift-serverless` 
for resources deployed by operator.

Currently users needs to add `knative.openshift.io/system-namespace`
label manually but they can use `knative.openshift.io/part-of` label
instead after this PR.

### DIFF
--- a/openshift-knative-operator/pkg/common/label.go
+++ b/openshift-knative-operator/pkg/common/label.go
@@ -5,12 +5,15 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
-var commonLabel = map[string]string{"app.openshift.io/part-of": "openshift-serverless"}
-
 // InjectEnvironmentIntoDeployment injects the common label into the resources.
 func InjectCommonLabel() mf.Transformer {
 	return func(u *unstructured.Unstructured) error {
-		u.SetLabels(commonLabel)
+		labels := u.GetLabels()
+		if labels == nil {
+			labels = make(map[string]string, 1)
+		}
+		labels["app.openshift.io/part-of"] = "openshift-serverless"
+		u.SetLabels(labels)
 		return nil
 	}
 }
--- a/openshift-knative-operator/pkg/common/label.go
+++ b/openshift-knative-operator/pkg/common/label.go
@@ -1,0 +1,16 @@
+package common
+
+import (
+	mf "github.com/manifestival/manifestival"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+var commonLabel = map[string]string{"app.openshift.io/part-of": "openshift-serverless"}
+
+// InjectEnvironmentIntoDeployment injects the common label into the resources.
+func InjectCommonLabel() mf.Transformer {
+	return func(u *unstructured.Unstructured) error {
+		u.SetLabels(commonLabel)
+		return nil
+	}
+}
--- a/openshift-knative-operator/pkg/common/label.go
+++ b/openshift-knative-operator/pkg/common/label.go
@@ -2,6 +2,7 @@ package common
 
 import (
 	mf "github.com/manifestival/manifestival"
+	socommon "github.com/openshift-knative/serverless-operator/pkg/common"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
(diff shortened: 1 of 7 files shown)


openshift-knative-operator/pkg/common/label.go
```diff
@@ -0,0 +1,20 @@
+package common
+
+import (
+	mf "github.com/manifestival/manifestival"
+	socommon "github.com/openshift-knative/serverless-operator/pkg/common"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// InjectEnvironmentIntoDeployment injects the common label into the resources.
```
skonto, line 9: should be `InjectCommonLabel`

nak3: Thank you!

test/servicemesh.go
```diff
@@ -139,7 +135,7 @@ func AllowFromServingSystemNamespaceNetworkPolicy(namespace string) *networkingv
 						{
 							NamespaceSelector: &metav1.LabelSelector{
 								MatchLabels: map[string]string{
-									KnativeSystemNamespaceKey: "true",
+									"knative.openshift.io/part-of": "openshift-serverless",
```
skonto, line 138: maybe reuse the constants? eg. `common.ServerlessCommonLabelKey: common.ServerlessCommonLabelValue`

skonto:
@nak3 seen before?

```
    source_native_kafka_broker_ksvc_test.go:125: ConfigMap still contains the finalizer kafka.brokers.eventing.knative.dev/serverless-tests-smoke-test-native-kafka-broker
    
```
/retest

skonto:
All tests pass now @nak3 should we unhold this?

openshift-knative-operator/pkg/common/label.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	mf "github.com/manifestival/manifestival"
+	socommon "github.com/openshift-knative/serverless-operator/pkg/common"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// InjectCommonLabel injects the common label into the resources.
+func InjectCommonLabel() mf.Transformer {
+	return func(u *unstructured.Unstructured) error {
```
skonto, line 11: Should not the transformer be applied on namespaces only?

nak3:
Let's wait for this until early next week. Roland and some folks are on PTO.

>  Is the new label used mainly for setting up the net policies on Serverless namespaces only?

At this moment, the label is used for the networkpolicies, yes. 

> How about the user namespaces should still the label be applied manually?

I'm sure we should not touch user's namespaces.

openshift-knative-operator/pkg/common/label.go
nak3: Do you have any reason why you think so? I don't have any strong opinion actually so it is alright to apply on namespace, though.

skonto: My thinking was that might be too much to add a label everywhere a bit verbose that is why I asked the rationale behind it.
But also it can be useful in certain cases eg. query by label etc.

nak3: Okay, I think it is alright to add the labels into namespaces only so updated the PR.

test/servinge2e/kourier/servicemesh_test.go
```diff
@@ -72,8 +72,6 @@ func setupNamespaceForServiceMesh(ctx *test.Context, serviceMeshNamespace, testN
 	test.CreateServiceMeshMemberRollV1(ctx, test.ServiceMeshMemberRollV1("default", serviceMeshNamespace, testNamespace))
 
 	test.CreateNetworkPolicy(ctx, test.AllowFromServingSystemNamespaceNetworkPolicy(testNamespace))
-	test.LabelNamespace(ctx, "knative-serving", test.KnativeSystemNamespaceKey, "true")
```
skonto, line 75: We dont want to verify `socommon.ServerlessCommonLabelKey`?

nak3: No, `test.LabelNamespace` does not verify the label but just added the manual label for the NetworkPolicy. 
We don't need to add the label anymore. I removed `LabelNamespace` helper function as well.

skonto: But we do add a label in the namespace resources we dont want an e2e test for that?

PR #1428, opened 2022-03-07
SRVCOM-1733 Fix formatting of Knative Eventing section in CSV

mgencur:
Remove the unordered list which is unnecessary in this case and align
the style with other bullets in the list.

https://issues.redhat.com/browse/SRVCOM-1733

PR #1430, opened 2022-03-07
[release-1.21] Fix formatting of Knative Eventing section in CSV

mgencur:
Same as https://github.com/openshift-knative/serverless-operator/pull/1428, only for 1.21 branch

Remove the unordered list which is unnecessary in this case and align
the style with other bullets in the list.


matzew (approved):
/lgtm
/approve

PR #1432, opened 2022-03-07
[KafkaSource] Update post-install / storage migrator jobs

matzew:

Adding new image to the generated CSV for completeness.

* Removing the old `storage-version-migration-kafka-source` post job from the `csv` generator;
* Do not use midstream images in manifest, but `TO_BE_REPLACED` so that our script does replace them (helps also witth productization)
* adding new post-job and a storage version migrator for the new source implementation to the CSV generator;
 
**NOTE:** For the storage version migrator we current reuse the image that we do already build for knative eventing-core. This might change in the future.


devguyio (approved):
/lgtm
/hold

pierDipi (approved):
/lgtm
/unhold

PR #1436, opened 2022-03-16
Fix invalid makefile target install-with-mesh-enabled

nak3:
As per title, current `install-with-mesh-enabled` target has a broken indent and it does not work as below:

```
$ make install-with-mesh-enabled 
make: Nothing to be done for 'install-with-mesh-enabled'.
```

This patch fixes it.

/cc @mgencur


nak3:
/retest

infra.

```
The connection to the server api.ci-ocp-4-10-amd64-aws-us-east-1-vgjfx.hive.aws.ci.openshift.org:6443 was refused - did you specify the right host or port?
error when retrieving current configuration of:
```

PR #1438, opened 2022-03-17
Going with Apache Kafka 3.1

matzew:

### DIFF
--- a/hack/lib/strimzi.bash
+++ b/hack/lib/strimzi.bash
@@ -48,6 +48,7 @@ function install_strimzi_cluster {
           offsets.topic.replication.factor: 3
           transaction.state.log.replication.factor: 3
           transaction.state.log.min.isr: 2
+          auto.create.topics.enable: false
           inter.broker.protocol.version: "3.1"
         storage:
           type: jbod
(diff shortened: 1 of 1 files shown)


matzew:
```
    knative_eventing_test.go:34: Failed to verify that health metrics work correctly for Eventing failed to access the Prometheus API endpoint and get the metric value expected: timed out waiting for the condition
```

/retest

hack/lib/strimzi.bash
```diff
@@ -48,9 +48,7 @@ function install_strimzi_cluster {
           offsets.topic.replication.factor: 3
           transaction.state.log.replication.factor: 3
           transaction.state.log.min.isr: 2
-          inter.broker.protocol.version: "3.0"
-          log.message.format.version: "3.0"
-          auto.create.topics.enable: "false"
```
pierDipi, line 53: is this setting gone? it's nice to have that one to make sure we're actually creating topics.

matzew: >           auto.create.topics.enable: "false"

jikes! I removed that by accident.

Added it back

pierDipi (approved):
/lgtm
/approve

PR #1449, opened 2022-03-22
[SRVCOM-1745] Add valid subscription annotation

skonto:
This is mandatory to avoid users deploying products without subscription.


skonto:
@nak3 @mgencur @rhuss gentle ping.

PR #1450, opened 2022-03-22
[SRVKS-890] OpenShift Route support for encrypted backends

nak3:


serving/ingress/pkg/reconciler/ingress/resources/route.go
```diff
@@ -142,6 +142,19 @@ func makeRoute(ci *networkingv1alpha1.Ingress, host string, rule networkingv1alp
 		},
 	}
 
+	// Use passthrough type for OpenShift Ingress -> Kourier Gateawy to encrypt the traffic
```
skonto, line 145: s/Gateawy/Gateway

serving/ingress/pkg/reconciler/ingress/resources/route.go
```diff
@@ -142,6 +142,19 @@ func makeRoute(ci *networkingv1alpha1.Ingress, host string, rule networkingv1alp
...
+	// when Internal TLS is enabled.
+	// In other words, do not use edge termination which makes plain traffic.
+	if isInternalEncryptionEnabled(rule) {
+		route.Spec.Port = &routev1.RoutePort{
+			TargetPort: intstr.FromString(HTTPSPort),
+		}
+		route.Spec.TLS = &routev1.TLSConfig{
+			Termination:                   routev1.TLSTerminationPassthrough,
```
skonto, line 153: Is this ok if `EnablePassthroughRouteAnnotation` is not there?

serving/ingress/pkg/reconciler/ingress/resources/route.go
```diff
@@ -154,6 +167,23 @@ func makeRoute(ci *networkingv1alpha1.Ingress, host string, rule networkingv1alp
 	return route, nil
 }
 
+// isInternalEncryptionEnabled determines whether internal-encryption is enabled or not.
+// In general, we can determine it by the value internal-encryption in config-network, however the serverless ingress does not
+// watch the ConfigMap. Therefore we determine it by ServiceHTTPSPort(443) port for the backend in Kingress.
+func isInternalEncryptionEnabled(rule networkingv1alpha1.IngressRule) bool {
```
skonto, line 173: Can we add a watch for the configmap or is this robust enough?
A more general question. In the future what stops us from moving the ingress stuff to the knative operator besides the overhead of the resources? 
As a side note information such as if the internal encryption is on are provided for free at the Serving CR level.
I am asking because I am working on the vision of having one operator for all stuff. Wdyth?

skonto:
Validation will pass when net-kourier branch for 1.4 has https://github.com/openshift-knative/net-kourier/pull/23 in. I am wondering though if we should first test here and then merge the other PRs to make sure all tests here pass.

serving/ingress/pkg/reconciler/ingress/resources/route.go
nak3: I think it is robust enough. Actually adding a watch for the configmap has two problems:
1. We need to pull informer/listner code.
2. It would cause a conflictions between Kingress and OpenShift Ingress in just a few (milli)seconds. (Because they are watching the configmap via different controller.)
So, watching the ingress would be better.

nak3: I think it is better to answer in another Slack or somewhere? 
If we started the thread here, we cannot merge this :sweat_smile:

serving/ingress/pkg/reconciler/ingress/resources/route.go
nak3: It creates the same route eventually. So I merged it into the same `if` condition.

skonto:
@nak3 I guess we will need another PR to fetch the manifest at the S-O side under kodata.

nak3:
Yes, after we merged https://github.com/openshift-knative/net-kourier/pull/23 we need it.

nak3:
/retest

I think current error is infra.

```
12:31:12.856 INFO:    Build the image in the cluster-internal registry.
Unable to connect to the server: dial tcp: i/o timeout
```

skonto:
```
The connection to the server api.ci-ocp-4-10-amd64-aws-us-east-1-v8c2c.hive.aws.ci.openshift.org:6443 was refused - did you specify the right host or port?
```

nak3:
/hold cancel

All eventing failure. I don't want to merge this with this status but I confirmed that eventing team is working on the CI issue on Slack.

pierDipi:
> /hold cancel
> 
> All eventing failure. I don't want to merge this with this status but I confirmed that eventing team is working on the CI issue on Slack.

Yes, I'm working on it.

PR #1454, opened 2022-03-25
[SRVKE-1161] Use `generate-name` for eventing jobs to avoid upgrade errors

pierDipi:
As per title.


test/upgrade/upgrade_test.go
```diff
@@ -136,6 +137,10 @@ func postUpgradeTests(ctx *test.Context) []pkgupgrade.Operation {
 		kafkabrokerupgrade.BrokerPostUpgradeTest(),
 		kafkabrokerupgrade.SinkPostUpgradeTest(),
 	)
+	tests = append(tests, upgrade.VerifyPostInstallJobs(upgrade.VerifyPostJobsConfig{
+		Namespace:    "knative-eventing",
+		FailOnNoJobs: true,
+	}))
```
pierDipi, line 143: I think this test might be useful for serving/networking too.

cc @nak3  @skonto

mgencur:
This looks good to me but someone else should review it as well and approve.

knative-operator/deploy/resources/knativekafka/channel/2-channel-post-install.yaml
```diff
@@ -109,7 +109,7 @@ roleRef:
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: storage-version-migration-kafka-channel-1.1.0
+  generateName: storage-version-migration-kafka-channel-
```
aliok, line 112: Not sure if we should remove the version here. What was the reason we had the version in the name originally?

knative-operator/pkg/controller/knativekafka/images.go
```diff
@@ -57,6 +57,8 @@ func ImageTransform(overrideMap map[string]string) mf.Transformer {
 			// Replace direct image YAML references.
 			if image, ok := overrideMap[obj.GetName()+delimiter+container.Name]; ok {
 				container.Image = image
+			} else if image, ok := overrideMap[obj.GetGenerateName()+delimiter+container.Name]; ok {
+				container.Image = image
```
aliok, line 61: 👍

aliok:
/lgtm
/approve
/hold

Added a minor comment. Feel free to ignore that and unhold

knative-operator/deploy/resources/knativekafka/channel/2-channel-post-install.yaml
pierDipi: I believe it is to have a different image for each version of this job. 
The problem is that having the version suffix hardcoded forces us to change this at every release (even patch releases).

PR #1458, opened 2022-03-30
Enable tracing for eventing by patching knativeeventing CR

mgencur:
Configuring tracing via the ConfigMap in knative-eventing namespace doesn't work as it is overwritten later by reconciling KnativeEventing CR. It worked so far only because there were tests running from eventing-kafka which were patching the knativeeventing CR. However, there's another [PR](https://github.com/openshift-knative/eventing-kafka/pull/625) for eventing-kafka that separates E2E tests and installing tracing. And the patching would not be done anymore. So, I'm fixing this here.

### DIFF
--- a/hack/lib/tracing.bash
+++ b/hack/lib/tracing.bash
@@ -63,19 +63,8 @@ EOF
 
 function enable_eventing_tracing {
   logger.info "Configuring tracing for Eventing"
-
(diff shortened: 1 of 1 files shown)


pierDipi (approved):
/lgtm
/approve

PR #1459, opened 2022-03-30
Replace generateName with version-based name for Kafka-related jobs

pierDipi:
Marek saw that we're running the post-install jobs too many times,
I believe because we're using generate name while upstream is
replacing the `generateName` field with a `name` based on the
`KnativeServing` or `KnativeEventing` version [1].

The fix is to do a similar transformation for Kafka-related jobs.

[1] https://github.com/knative/operator/blob/0f37c1e64154f4a78675feffe81c7e0c6fe3b39c/pkg/reconciler/common/job.go#L32-L57

### DIFF
--- a/hack/generate/csv.sh
+++ b/hack/generate/csv.sh
@@ -153,6 +153,7 @@ for name in "${kafka_images[@]}"; do
 done
 
 # Add Knative Kafka version to the downstream operator
+add_downstream_operator_deployment_env "$target" "CURRENT_VERSION" "$(metadata.get project.version)"
 add_downstream_operator_deployment_env "$target" "KNATIVE_EVENTING_KAFKA_VERSION" "$(metadata.get dependencies.eventing_kafka)"
 
 # Override the image for the CLI artifact deployment
--- a/knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
+++ b/knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"strconv"
 
+	batchv1 "k8s.io/api/batch/v1"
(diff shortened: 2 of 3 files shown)


knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -291,6 +293,26 @@ func (r *ReconcileKnativeKafka) transform(manifest *mf.Manifest, instance *serve
...
+	version := os.Getenv("CURRENT_VERSION")
+	return func(u *unstructured.Unstructured) error {
+		if u.GetKind() == "Job" {
+			job := &batchv1.Job{}
+			if err := scheme.Scheme.Convert(u, job, nil); err != nil {
+				return err
+			}
+			if job.GetName() == "" && job.GetGenerateName() != "" {
+				job.SetName(fmt.Sprintf("%s%s", job.GetGenerateName(), version))
+				job.SetGenerateName("")
```
matzew, line 306: this is a bit different than upstreams. `jobs.go` transformer

should we also align that upstream?

hack/generate/csv.sh
```diff
@@ -153,6 +153,7 @@ for name in "${kafka_images[@]}"; do
 done
 
 # Add Knative Kafka version to the downstream operator
+add_downstream_operator_deployment_env "$target" "CURRENT_VERSION" "$(metadata.get project.version)"
```
matzew, line 156: is `CURRENT_VERSION` a good name?

/cc @mgencur

matzew (approved):
/lgtm
/approve

/hold
if others want to also comment on this

hack/generate/csv.sh
pierDipi: We can iterate on the name if we have a better name, any suggestions?

PR #1465, opened 2022-04-05
Bump golangci-lint to latest

skonto:
Hitting this [multiple times](https://github.com/openshift-knative/serverless-operator/runs/5829404731?check_suite_focus=true). See discussion [here](https://coreos.slack.com/archives/CD87JDUB0/p1649082942219109).

PR #1466, opened 2022-04-05
Update setup-go and golangci-lint actions

dsimansk:
Per title.

/cc @matzew

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -19,7 +19,7 @@ jobs:
       GOPATH: ${{ github.workspace }}
     steps:
       - name: Setup Golang
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@v3
         with:
           go-version: 1.17.x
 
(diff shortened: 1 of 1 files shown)


.github/workflows/validate.yaml
```diff
@@ -112,7 +112,7 @@ jobs:
                 -level="error"
 
       - name: Go Lint - knative-operator
-        uses: golangci/golangci-lint-action@v2
+        uses: golangci/golangci-lint-action@v3
```
dsimansk, line 115: >v3.0.0+ requires explicit setup-go installation step prior to using this action: uses: actions/setup-go@v2.
The skip-go-installation option has been removed.

Fixes the issue with too fresh go version being installed for golangci-lint runs.

.github/workflows/validate.yaml
```diff
@@ -112,7 +112,7 @@ jobs:
...
         with:
           version: v1.43
```
dsimansk, line 117: @matzew @skonto @pierDipi should we stick with whatever upstream actions use or just go latest by default?

skonto: I updated to latest #1465 ;)

pierDipi: I have no real opinion as long as it works and is useful.

dsimansk:
nevermind :)
/close

PR #1479, opened 2022-04-07
Bump to 1.23

matzew:
let's have these as single, atomic PRs, and not as a big all update shot PR


skonto:
Usually we do this after https://github.com/openshift/release/pull/27522 is merged. I guess it will not create issues since we dont touch the knative operator version.

matzew:
@skonto usually this was the very first Markus and did in the past 

the version of Serverless Operator is independent of any operands - hence, IMO this is the canonical first shot

pierDipi (approved):
/lgtm
/approve

PR #1482, opened 2022-04-08
Get eventing core 1.2.1

matzew:
as per title, updating eventing core

PR #1487, opened 2022-04-08
Bump serving core to 1.2.0

skonto:
As per title. Mostly path renames and some patch updates.


skonto:
@nak3 pls take a look.

olm-catalog/serverless-operator/project.yaml
```diff
@@ -22,13 +22,14 @@ requirements:
...
+  net_istio: 1.2.0
   maistra: 2.1.0
 
   eventing: 1.2.1
   eventing_kafka: 1.1.0
   eventing_kafka_broker: 1.1.3
+
   cli: 1.1.0
-  operator: 1.0.1
+  operator: 1.2.0
```
skonto, line 35: We need to update the operator to address the issue with the limits applying to all containers across containers.

skonto: This slightly touches the eventing crd too @matzew could you check pls?

matzew:
/hold
https://github.com/openshift-knative/serverless-operator/pull/1482
was merged too early - (since the not required jobs did fail).

I am reverting that PR now ...

matzew:
Let's please do this _AFTER_ this revert is in: https://github.com/openshift-knative/serverless-operator/pull/1488

olm-catalog/serverless-operator/project.yaml
matzew: Would it be possible to do that bump on a separate PR ? 
Or is there a hard dependency that it must be done in this "serving and friends" bump?

skonto: Ok I can try that.

skonto:
```
    revision_timeout_test.go:170: Failed to create Service: admission webhook "webhook.metadata-webhook.example.com" denied the request: mutation failed: cannot decode incoming new object: json: unknown field "maxDurationSeconds"
```

The error with the maxDuration field exists because in 1.2.0 the field was removed: https://github.com/knative/serving/pull/12640.

Also the S-O CI job on main still uses 1.1.2.

```
[36mINFO[0m[2022-04-08T09:41:50Z] Running root, [input:openshift_release_golang-1.17], [input:base], [input:cli], [input:openshift_knative-v1.1.0_knative-eventing-kafka-src], [input:openshift_knative-v1.1.2_knative-serving-src], [input:openshift_knative-v1.1.3_knative-eventing-kafka-broker-src], [input:openshift_knative-v1.2.1_knative-eventing-src], src, knative-openshift-ingress, knative-operator, openshift-knative-operator, serverless-operator-src, upstream-e2e-mesh-aws-ocp-410 
[36mINFO[0m[2022-04-08T09:41:50Z] Tagging openshift/knative-v1.2.1:knative-eventing-src into pipeline:openshift_knative-v1.2.1_knative-eventing-src. 
```
When all repos are aligned it should go away.

olm-catalog/serverless-operator/project.yaml
matzew: No, but not sure why the operator is also bumped, when we bump serving (PR says: serving bump).

I personally like these to be all small, and less or more atomic, instead of a huge PR that bumps `n` components, and errors are harder to isolate

matzew:
> The error with the maxDuration field exists because it is new in [1.2.0](https://github.com/openshift/knative-serving/commits/release-v1.2.0?after=c6952920993cea26498db4a32f5c817c62bf59fe+69&branch=release-v1.2.0) but was not there in 1.1.2. Btw this revision field was removed in 1.2.2!

sounds like this breaks  - should something be done for it?

matzew:
> ```
> �[36mINFO�[0m[2022-04-08T09:41:50Z] Running root, [input:openshift_release_golang-1.17], [input:base], [input:cli], 
> [input:openshift_knative-v1.1.0_knative-eventing-kafka-src], [input:openshift_knative-v1.1.2_knative-serving-src], 
> [input:openshift_knative-v1.1.3_knative-eventing-kafka-broker-src], [input:openshift_knative-v1.2.1_knative-eventing-src], src, 
> knative-openshift-ingress, knative-operator, openshift-knative-operator, serverless-operator-src, upstream-e2e-mesh-aws-
> ocp-410 
```

I think that line is misleading. I think I had some struggle with old versions logged for SRC image before. But I forgot.

I think @mgencur could help in the past

**Comment by skonto** [comment]
> sounds like this breaks - should something be done for it?

The migration path after 1.22 is here: https://github.com/knative/serving/issues/12634. but not activity.

**Comment by skonto** [comment]
@mgencur is on pto afaik.

**Comment by matzew** [comment]
> but no activity.

quesstion is: do we need to do something here? I guess accepting the failure is not OK?

**Comment by skonto** [comment]
> I guess accepting the failure is not OK?
The failure exists because we are not using the 1.2,0 image in the ci job.

**Comment by matzew** [comment]
> The failure exists because we are not using the 1.2.0 source image in the ci job. The tests in 1.1.2 use that field.

ah, ok. I think Martin may know more on this subject :smile:

**Comment by maschmid** [comment]
So, yes, in general, image sources are taken from the branch, not from the PR , so the changes in `openshift/ci-operator/source-image/Dockerfile` will only by applied in next PRs after this is merged.

**Comment by skonto** [comment]
@maschmid the error is:
```
    revision_timeout_test.go:170: Failed to create Service: admission webhook "webhook.metadata-webhook.example.com" denied the request: mutation failed: cannot decode incoming new object: json: unknown field "maxDurationSeconds"
```
So the tests actually try to create a service with the new field added. Webhook denies it. The new field is in 1.2.0 not the previous 1.1.2. I am wondering what is happening.

skonto:
Ok I figured it out. This is not the Servign webhook that fails.
Actually its the metadata webhook that denies it :D We need to update the dependencies as it gets the version from ehre: https://github.com/openshift-knative/serverless-operator/blob/eb2ff38bec2c48f28b0a6752e95b2c931430c014/serving/metadata-webhook/cmd/webhook/main.go#L21.

maschmid:
"webhook.metadata-webhook.example.com" is probably the https://github.com/nak3/metadata-webhook/

skonto:
Yeah we need to update the dependencies. I will close this PR and start a new one with limited changes and then will create a bigger with all the deps updates in for Serving.

skonto:
@matzew ^^^

skonto:
@nak3 could the webhook share the same manifests as the knative operator to avoid such issues in the future?

nak3:
The metadata-webhook code was moved to https://github.com/openshift-knative/serverless-operator/tree/main/serving/metadata-webhook 

I will archive https://github.com/nak3/metadata-webhook/ :sweat_smile:

skonto:
@nak3 @maschmid I just found out that running `make generated-files` (required by our validation action) will cause the source image to change to 1.2.0. That means we still need to update the deps. So the smallest PR I can do would require vendor dir changes.

PR #1500, opened 2022-04-12
[SRVKE-1202] Remove KafkaBinding's podspecable ClusterRoleBinding

pierDipi:
This is basically the companion PR for
https://github.com/openshift-knative/eventing-kafka-broker/pull/201

Signed-off-by: Pierangelo Di Pilato <pierdipi@redhat.com>

PR #1501, opened 2022-04-12
Bump knative operator crds to 1.2.2

skonto:
As per title. The major addition that is needed downstream is this [commit](https://github.com/knative/operator/commit/8528bd581eab6d9f855290a1e98f70b6ab4cfa8d). We need to align versions anyway.
Deps are already updated.

### DIFF
--- a/olm-catalog/serverless-operator/project.yaml
+++ b/olm-catalog/serverless-operator/project.yaml
@@ -31,4 +31,4 @@ dependencies:
   eventing_kafka: 1.1.0
   eventing_kafka_broker: 1.1.3
   cli: 1.1.0
-  operator: 1.2.0
+  operator: 1.2.2
--- a/go.mod
+++ b/go.mod
@@ -170,7 +170,7 @@ replace (
 	// Knative forks.
 	knative.dev/eventing => github.com/openshift/knative-eventing v0.99.1-0.20220331162435-b93f6ffdf2e5
 	knative.dev/eventing-kafka => github.com/openshift-knative/eventing-kafka v0.19.1-0.20220331132250-5e770359a891
-	knative.dev/eventing-kafka-broker => github.com/openshift-knative/eventing-kafka-broker v0.25.1-0.20220407092930-534b61b8c609
+	knative.dev/eventing-kafka-broker => github.com/openshift-knative/eventing-kafka-broker v0.25.1-0.20220412082434-af580b16c8a1
 	knative.dev/serving => github.com/openshift/knative-serving v0.10.1-0.20220407204305-c6952920993c
 )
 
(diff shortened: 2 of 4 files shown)


olm-catalog/serverless-operator/hack/002-serving-drop-unsupported-fields.patch
```diff
@@ -31,8 +31,8 @@ index aa8f3e459..4b4f9b580 100644
                config:
                  additionalProperties:
                    additionalProperties:
-@@ -127,13 +108,8 @@ spec:
-                       type: object
+@@ -770,13 +751,8 @@ spec:
+                       type: array
                ingress:
                  description: The ingress configuration for Knative Serving
 +                x-kubernetes-preserve-unknown-fields: true # To allow for some fields we've deleted.
```
pierDipi, line 38: perhaps adding which fields might help the reader?

pierDipi: or a link for additional context.

olm-catalog/serverless-operator/project.yaml
```diff
@@ -31,4 +31,4 @@ dependencies:
   eventing_kafka: 1.1.0
   eventing_kafka_broker: 1.1.3
   cli: 1.1.0
-  operator: 1.0.1
+  operator: 1.2.0
```
pierDipi, line 34: Since there is 1.2.2, should we use that one?
https://github.com/knative/operator/releases/tag/knative-v1.2.2

skonto: Serving is on 1.2.0 but I guess should not matter?

olm-catalog/serverless-operator/hack/002-serving-drop-unsupported-fields.patch
skonto: This pre-existed, not part of this PR, let's open another issue for this.

pierDipi: Ok, reviewing a patch of patches isn't easy and I thought it was introduced here.

olm-catalog/serverless-operator/project.yaml
pierDipi: Yes, they can be 2 separate versions (in theory)

skonto: Ok no problem let me do it.

skonto: @pierDipi done

matzew: +1 on using the latest version. we should never rely on all versions in the same `x.y.z` range, since they are independent (in theory)

skonto: The patch version should not matter but it is a good thing in general to have versions close to each other. We dont want all versions to be on latest but have a very old component obviously.

matzew:
`kafka-controller-post-install-1.22.0` that does not make sense ..

skonto:
@matzew should not this be fixed with the Eventing 1.2.x updates?

pierDipi:
The issue is that we didn't merge https://github.com/openshift-knative/serverless-operator/pull/1500 before merging midstream https://github.com/openshift-knative/eventing-kafka-broker/pull/201, indeed the error is:
```
{"severity":"EMERGENCY","timestamp":"2022-04-12T12:52:34.132730986Z","logger":"kafka-broker-post-install","caller":"post-install/main.go:47","message":"source migration failed: failed to delete ClusterRoleBinding eventing-sources-kafka-controller-podspecable-binding: clusterrolebindings.rbac.authorization.k8s.io \"eventing-sources-kafka-controller-podspecable-binding\" is forbidden: User \"system:serviceaccount:knative-eventing:knative-kafka-controller-post-install\" cannot delete resource \"clusterrolebindings\" in API group \"rbac.authorization.k8s.io\" at the cluster scope","stacktrace":"main.main\n\t/go/src/knative.dev/eventing-kafka-broker/control-plane/cmd/post-install/main.go:47\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:255"}
```

pierDipi:
We should wait for https://github.com/openshift-knative/serverless-operator/pull/1500 to go in first.

skonto:
Rebased let's see.

matzew (approved):
/lgtm
/approve

PR #1453, opened 2022-03-25
Tests for Distributed Tracing

mgencur:
https://issues.redhat.com/browse/SRVCOM-1732

* Separate Tracing installation from running tests. Tracing is installed first and tests use it (previously the Eventing test functions were also installing tracing by themselves
* OpenTelemetry tracing is installed for E2E tests from serverless-operator repo
* Zipkin tracing is still used for upstream tests and upgrade tests because those tests depend on the existence of Zipkin pod
* Migrate tests from [ocf-qe-testsuite](https://gitlab.cee.redhat.com/ocf-qe/ocf-qe-testsuite/-/blob/master/test/operator/tracing_test.go) except for tracePropagatedByUserContainer which requires a new image and does not test anything related to Knative (IMO). Run the tests as part of E2E test suite (test/servinge2e/
* Minor refactoring - move the definition of testNamespace to a common place (this is done via a separate commit for easier review.
* Update dependencies for the new Jaeger Query API (done via a separate commit)

PR #1510, opened 2022-04-14
Prevent deleting knative-serving NS when uninstalling Service Mesh

mgencur:
Pulling this commit out of https://github.com/openshift-knative/serverless-operator/pull/1453 and sending it separately.


nak3:
This change is good to me.
But I'm prefer to separating the namespace into a differet yaml like `hack/lib/mesh_resources/namespace.yaml` and update `hack/lib/mesh.bash` as applying `namespace.yaml` but not deleting it.

mgencur:
@nak3 Something like the last commit above?

nak3:
/lgtm
/approve
/hold

Yes, exactly! Thank you so much. (`/hold` for CI.)

PR #1515, opened 2022-04-15
Bump dependencies to release-1.2

pierDipi:
Bump only vendored dependencies to 1.2

See individual commits for details.


pierDipi:
```

error: unable to recognize "STDIN": Get "https://api.ci-ocp-4-10-amd64-aws-us-west-1-mlpbc.hive.aws.ci.openshift.org:6443/api?timeout=32s": dial tcp 54.215.62.53:6443: connect: connection refused
```

/test 4.10-operator-e2e-aws-ocp-410

go.mod
```diff
@@ -180,8 +170,8 @@ replace (
 	github.com/go-logr/zapr => github.com/go-logr/zapr v0.4.0
 	github.com/manifestival/manifestival => github.com/manifestival/manifestival v0.7.0
 	k8s.io/klog/v2 => k8s.io/klog/v2 v2.8.0
-
-	// Needed by metadata webhook and other components
-	knative.dev/networking => knative.dev/networking v0.0.0-20211101215640-8c71a2708e7d
-	knative.dev/pkg => knative.dev/pkg v0.0.0-20211210132429-e86584fd3c69
 )
+
+replace knative.dev/hack => knative.dev/hack v0.0.0-20220118141833-9b2ed8471e30
```
skonto, line 175: That might help let's see!

pierDipi: Yes, this is done automatically and is based on the actual version on the branch
https://github.com/openshift-knative/serverless-operator/pull/1515/files#diff-ff3402c713d8a504795b7115caa1af088b1bfc583d43adb5d0bc5a83aaa03bc1R34-R35

hack/update-deps.sh
```diff
@@ -22,17 +22,17 @@ SERVING_VERSION="release-v1.2.0"
...
   "knative.dev/operator@${KN_VERSION}"
-  "knative.dev/pkg@v0.0.0-20220223180940-4fcbc1bc12e8"
 )
 
 FLOATING_FORK_DEPS=(
   "knative.dev/eventing-kafka-broker=github.com/openshift-knative/eventing-kafka-broker@${EVENTING_KAFKA_BROKER_VERSION}"
   "knative.dev/eventing=github.com/openshift/knative-eventing@${EVENTING_VERSION}"
   "knative.dev/eventing-kafka=github.com/openshift-knative/eventing-kafka@${EVENTING_KAFKA_VERSION}"
   "knative.dev/serving=github.com/openshift/knative-serving@${SERVING_VERSION}"
+  "knative.dev/pkg=knative.dev/pkg@${KN_VERSION}"
```
skonto, line 34: This is supposed to be the openshift ones. What does it mean to move this here?

skonto: Also the name now becomes a bit irrelevant.
Could be `FLOATING_DEPS_FORCED_VERSIONS`?

pierDipi: This means force using whatever is in the `KN_VERSION` branch, so it fixes the fact that serving and eventing are at different hashes of these 2 packages.

I can introduce a different variable for these but the result is the same.

skonto: Ok I guess the end result is the same hash though without requiring it upfront.

pierDipi: or just `REPLACE_DEPS`

pierDipi: but I think it's a very minor comment, I can do it next, it's not worth waiting another 4 hours for running e2e tests :)

PR #1522, opened 2022-04-20
Remove redundant ContinualTest from knative/eventing

mgencur:
The test is identical to BrokerBackedByChannelTest because in
serverless-operator repo we set up KafkaChannel as the default channel
before the test suite. Both tests deploy Broker and Trigger and use
KafkaChannel in the background. The tests are vendored from different
repositories and due to the settings we have they test the same
configuration.


matzew (approved):
/lgtm
/approve

matzew:
> upgrade-tests-aws-ocp-410

passed :heavy_check_mark:

PR #1523, opened 2022-04-21
Install distributed tracing

mgencur:
Pulling two commits from https://github.com/openshift-knative/serverless-operator/pull/1453 and sending them separately:
*Install Distributed Tracing
*Wait for OpenTelemetryCollector CR to be successfully applied

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,12 +115,9 @@ test-upstream-e2e-mesh-testonly:
 
 test-upstream-e2e-mesh:
 	FULL_MESH="true" UNINSTALL_MESH="false" ./hack/mesh.sh
-	./hack/tracing.sh
-	FULL_MESH=true ./hack/install.sh
+	TRACING_BACKEND=zipkin TRACING_NAMESPACE=knative-eventing ./hack/tracing.sh
+	FULL_MESH=true TRACING_BACKEND=zipkin TRACING_NAMESPACE=knative-eventing ./hack/install.sh
 	FULL_MESH=true ./test/e2e-tests.sh
-	./hack/teardown.sh
-	TRACING_BACKEND=zipkin ./hack/tracing.sh
-	FULL_MESH=true TRACING_BACKEND=zipkin ./hack/install.sh
 	FULL_MESH=true TEST_KNATIVE_KAFKA=false TEST_KNATIVE_UPGRADE=false ./test/upstream-e2e-tests.sh
 
 # Run upstream E2E tests without upgrades.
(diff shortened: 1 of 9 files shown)


skonto:
This is awesome thanks @mgencur!

hack/lib/serverless.bash
```diff
@@ -155,6 +155,10 @@ function deploy_knativeserving_cr {
     deploy_with_kourier
   fi
 
+  if [[ $ENABLE_TRACING == "true" ]]; then
+    enable_serving_tracing
```
skonto, line 159: @pierDipi do you think Eventing should also integrate here? Afaik Eventing has its own setup for tracing tests?

Makefile
```diff
@@ -77,6 +90,7 @@ test-e2e-with-kafka-testonly:
 
 test-e2e-with-kafka:
 	UNINSTALL_STRIMZI="false" ./hack/strimzi.sh
+	./hack/tracing.sh
```
skonto, line 93: Is this equivalent to `install-tracing-opentelemetry` here? Should we replace with that to make it more explicit?

Makefile
```diff
@@ -100,8 +115,12 @@ test-upstream-e2e-mesh-testonly:
 
 test-upstream-e2e-mesh:
 	FULL_MESH="true" UNINSTALL_MESH="false" ./hack/mesh.sh
+	./hack/tracing.sh
 	FULL_MESH=true ./hack/install.sh
 	FULL_MESH=true ./test/e2e-tests.sh
+	./hack/teardown.sh
+	TRACING_BACKEND=zipkin ./hack/tracing.sh
```
skonto, line 122: Should we just call `install_zipkin_tracing`?

Makefile
skonto, line 93: I think a lot of the rules in this file could be optimized:
test-e2e-with-kafke: uninstall-strimzi uninstall-tracing-opentelemetry 
There is a catch with dependencies ordering and the fact that they are only built once but are not we use the targets once per time?

hack/lib/serverless.bash
```diff
@@ -155,6 +155,10 @@ function deploy_knativeserving_cr {
     deploy_with_kourier
   fi
 
+  if [[ $ENABLE_TRACING == "true" ]]; then
```
skonto, line 158: Do we actually need to disable this since by default we enable it? What cases you had in mind?

Makefile
skonto: cool thanks for the history lesson :)

hack/lib/serverless.bash
skonto: ok

skonto:
let's see if tests pass this time

PR #1530, opened 2022-04-26
Replace community image with OpenShift's

nak3:
This patch replaces community image with OpenShift's.

CI sometimes fails due to timeout with `ContainerCreating` as ([log](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift-knative_serverless-operator/1530/pull-ci-openshift-knative-serverless-operator-main-4.10-operator-e2e-aws-ocp-410/1518842222026428416)):

```
serverless-tests2                                  kube-helloworld-go-5d86cf49cd-fh77q                               0/1     ContainerCreating   0             8m7s    <none>         ip-10-0-136-64.ec2.internal    <none>           <none>
```

The CI log did not say the exact reason why it took so long to start the container but the failed pod always uses community image which is larger.

(community image) https://console.cloud.google.com/gcr/images/knative-samples/global/helloworld-go
(openshift image) https://quay.io/repository/openshift-knative/hello-openshift?tab=tags

So it is worth to use OpenShift image.

Also, I think we should use OpenShift image rather than community image.

PR #1531, opened 2022-04-26
use quay mirror of helloworld-go in tests

maschmid:
gcr.io seems to be terribly slow today, causing timeouts. We should use quay by default anyway.

### DIFF
--- a/test/eventinge2e/source_to_ksvc_test.go
+++ b/test/eventinge2e/source_to_ksvc_test.go
@@ -14,7 +14,7 @@ import (
 
 const (
 	pingSourceName    = "smoke-test-ping"
-	image             = "gcr.io/knative-samples/helloworld-go"
+	image             = "quay.io/openshift-knative/helloworld-go"
 	helloWorldService = "helloworld-go"
 	helloWorldText    = "Hello World!"
 	ksvcAPIVersion    = "serving.knative.dev/v1"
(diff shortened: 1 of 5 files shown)


nak3:
Oh, thank you! I also saw an issue in our CI ans sent https://github.com/openshift-knative/serverless-operator/pull/1530.
But your PR is better :smile: Let's merge yours.

maschmid:
> Oh, thank you! I also saw an issue in our CI and sent #1530. But your PR is better smile Let's merge yours.

Ah! Sorry, I didn't notice that one. I agree, let's merge this one if it works first, and we can perhaps switch to "hello-openshift" later to consolidate the test images.

nak3:
`openshift-knative/helloworld-go` does not have `latest` tag so it seems CI fails to pull the image :thinking: 
https://quay.io/repository/openshift-knative/helloworld-go?tab=tags

I also cannot pull it in my local :sweat_drops: 
```
$ docker pull quay.io/openshift-knative/helloworld-go
Using default tag: latest
Error response from daemon: manifest for quay.io/openshift-knative/helloworld-go:latest not found: manifest unknown: manifest unknown
```

maschmid:
OK, I have pushed :latest (retagged the current gcr.io/knative-samples/helloworld-go:latest )

PR #1537, opened 2022-04-29
Disable tracing by default

mgencur:
* This will only install Tracing and enable it in KnativeServing and KnativeEventing CR when running tests in CI
* The make targets that are usually used locally (make install etc.) will not enable tracing
* The `install-all` target has been modified to install everything including Tracing


matzew (approved):
/lgtm
/approve

mgencur:
/hold
Let's run all tests (including upstream).

matzew (approved):
/lgtm
/approve

PR #1541, opened 2022-05-02
Extend user permission tests to cover all types

mgencur:
https://issues.redhat.com/browse/SRVCOM-854

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -159,7 +159,7 @@ require (
 	k8s.io/gengo v0.0.0-20211129171323-c02415ce4185 // indirect
 	k8s.io/klog/v2 v2.40.1 // indirect
 	k8s.io/kube-openapi v0.0.0-20211115234752-e816edb12b65 // indirect
-	k8s.io/utils v0.0.0-20211208161948-7d6a63dca704
+	k8s.io/utils v0.0.0-20211208161948-7d6a63dca704 // indirect
 	knative.dev/caching v0.0.0-20220223181140-b3aed5c4e283 // indirect
 	knative.dev/reconciler-test v0.0.0-20220118183433-c8bfbe66bada
 	sigs.k8s.io/structured-merge-diff/v4 v4.1.2 // indirect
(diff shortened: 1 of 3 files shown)


mgencur:
Rebased. Let's see what happens. The failures looked unrelated.

mgencur:
The error in the upgrade tests is: `prober.go:103: actual unavailable period 1m1.032662026s is over down time limit of 1m0s` 
We've seen it a few times and we'll need to decide if the 1minute limit is OK or should be increased with the latest migration to the new Kafka Broker components.
Anyway, this failure has nothing to do with this PR. I think it can be merged.

test/extensione2e/kafka/user_permissions_test.go
```diff
@@ -49,31 +52,62 @@ func TestKafkaUserPermissions(t *testing.T) {
...
+		Spec: kafkachannelv1beta1.KafkaChannelSpec{
+			NumPartitions:     1,
+			ReplicationFactor: 1,
+		},
+	}
+
+	kafkaSink := &kafkasinksv1alpha1.KafkaSink{
+		Spec: kafkasinksv1alpha1.KafkaSinkSpec{
+			Topic:             "my-topic",
+			NumPartitions:     pointer.Int32Ptr(10),
```
skonto, line 65: How about using `ptr.Int32(10)` and import "knative.dev/pkg/ptr" instead?
Same for ` ptr.String(kafkasinksv1alpha1.ModeStructured)`.

mgencur: I could do that. The two functions have absolutely same result. Just for my understanding - is there any benefit?

go.mod
```diff
@@ -159,7 +159,7 @@ require (
 	k8s.io/gengo v0.0.0-20211129171323-c02415ce4185 // indirect
 	k8s.io/klog/v2 v2.40.1 // indirect
 	k8s.io/kube-openapi v0.0.0-20211115234752-e816edb12b65 // indirect
-	k8s.io/utils v0.0.0-20211208161948-7d6a63dca704 // indirect
+	k8s.io/utils v0.0.0-20211208161948-7d6a63dca704
```
skonto, line 162: Update this too?

test/extensione2e/kafka/user_permissions_test.go
skonto: Avoid introducing new deps.

go.mod
mgencur: Yeah. Now I understand. Thanks. Pushed.

mgencur:
/retest
Infra issue.

PR #1542, opened 2022-05-03
Make target for installing ClusterLogging

mgencur:
Might be handy e.g. in debugging upgrade tests where the wathola-sender is shutdown during the test and the logs are lost. By installing ClusterLogging the logs are made available via Kibana.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,6 +68,9 @@ install-tracing-opentelemetry:
 uninstall-tracing-opentelemetry:
 	UNINSTALL_TRACING=true ./hack/tracing.sh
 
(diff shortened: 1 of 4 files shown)


skonto:
The logs exist only during the lifetime of the cluster or we could retrieve them on demand?
Btw as a side note it would be nice to inject the logs of CI jobs into elastic search and failures with context (maybe via predefined queries etc to avoid digging into logs manually).

mgencur:
AFAIK, the logs are queryable/available only when the cluster is up.

pierDipi:
@skonto can we merge this PR?

PR #1547, opened 2022-05-04
Add timeout option and verbose flag to golangci-lint action

skonto:
For more check [here](https://github.com/openshift-knative/serverless-operator/pull/1546#issuecomment-1117174138).
Default time out is 1m. Setting it to 10m, a bit generous to cover slow runs (but should take less).


skonto:
```
  level=info msg="Execution took 5m30.29484176s"
  
  golangci-lint found no issues
  Ran golangci-lint in 335111ms
```

PR #1552, opened 2022-05-06
Increase Zipkin -Xmx from 5G to 7G

mgencur:
This is for upgrade tests to properly gather tracing info


pierDipi (approved):
/lgtm
/approve

PR #1555, opened 2022-05-12
Separate building and deploying bundle and index images for CatalogSo…

mgencur:
…urce

https://issues.redhat.com/browse/SRVCOM-1861

Changes:
* build bundle and index images when running in CI or when using DOCKER_REPO_OVERRIDE (then the command would be `DOCKER_REPO_OVERRIDE=docker.io/mgencur make images dev` for installing the CatalogSource
* when not running in CI or with DOCKER_REPO_OVERRIDE use the latest nightly build of the index and bundle images (this also allows for using the `make dev` target from other repositories when the goal is simply to install the latest nightly build of serverless)
* use the new file-based catalogs as described in [OpenShift docs](https://docs.openshift.com/container-platform/4.10/operators/admin/olm-managing-custom-catalogs.html#olm-managing-custom-catalogs-fb) - this format should be fully functional since >=OCP 4.6

For deploying CatalogSource with the latest nightly build we still need to add some configuration to openshift/release to let CI build and promote the index image. The PR is available at https://github.com/openshift/release/pull/28552


mgencur:
/hold
I will prepare PR for openshift/release to have everything ready when we merge this.

matzew:
/hold

Let's please wait for the `KafkaChannel` migration PR

matzew:
/unhold

(I just see that has been merged)

olm-catalog/serverless-operator/index/Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM openshift/origin-base as builder
```
skonto, line 1: Why not just start with `FROM quay.io/operator-framework/opm:latest`? 
This is what we end-up anyway.

hack/lib/catalogsource.bash
```diff
@@ -12,133 +12,59 @@ function ensure_catalogsource_installed {
...
-          exec:
-            command:
-            - grpc_health_probe
-            - -addr=localhost:50051
-        command:
-        - /bin/sh
-        - -c
-        - |-
-          podman login -u $pull_user -p $token image-registry.openshift-image-registry.svc:5000 && \
-          /bin/opm registry add -d index.db --container-tool=podman --mode=replaces -b registry.ci.openshift.org/openshift/openshift-serverless-v$PREVIOUS_VERSION:serverless-stop-bundle,image-registry.openshift-image-registry.svc:5000/$OLM_NAMESPACE/serverless-bundle && \
```
skonto, line 122: I guess removing commands such as `opm registry` also makes us [compatible](https://github.com/redhat-openshift-ecosystem/community-operators-prod/discussions/512) with FBC index.

hack/lib/catalogsource.bash
```diff
@@ -156,6 +82,43 @@ EOF
...
+    find "${build_dir}" -type f -exec sha1sum {} + \
+      > "${rootdir}/_output/${name}.sha1sum"
+    oc -n "${OLM_NAMESPACE}" delete configmap "${name}-sha1sums" --ignore-not-found=true
+    oc -n "${OLM_NAMESPACE}" create configmap "${name}-sha1sums" \
+      --from-file="${rootdir}/_output/${name}.sha1sum"
+    rm -f "${rootdir}/_output/${name}.sha1sum"
+  else
+    logger.info "${name} build is up-to-date."
+  fi
+}
```
pierDipi, line 119: we are using this approach because it's not possible to inject the previously built images using CI operator in a different image build context, right?

pierDipi (commented):
Thanks!

hack/lib/catalogsource.bash
mgencur: Exactly. The `registry` command is for the old format. We're now using `init` and `render` to generate the new format.

hack/lib/catalogsource.bash
mgencur: Correct, I did not find a way to get the exact pullspec for the image so that it could be used in the following image build. The pull spec is only available in the "test" containers when running tests, through `dependencies` section of the ci-operator config. And did not get any answer for this on #forum-testplatform: https://coreos.slack.com/archives/CBN38N3MW/p1652275831765409

pierDipi (approved):
/lgtm

/hold (unhold once you're ready to merge)

PR #1561, opened 2022-05-16
[WIP] Manual tests

aliok:


matzew:
Not sure what this PR is really about.

The `Validate / Generated files are commited` of course fails, b/c we touch `vendor` files

If we just wanna trigger a build, we should do something like:

```shell
date > ci && git add ci && git commit -m ":robot: CI was here...:fire:" ci
```

PR #1569, opened 2022-05-17
Pull the previous bundle image from a new namespace "knative"

mgencur:
This is the namespace where images are now promoted since release-1.22

This PR depends on another one: The namespace was changed in https://github.com/openshift/release/pull/28552 because the old namespace openshift is no longer allowed.

When this PR is merged the serverless-index image will be built in CI and published. Then `make install` without DOCKER_REPO_OVERRIDE should work seemlessly.

### DIFF
--- a/hack/lib/catalogsource.bash
+++ b/hack/lib/catalogsource.bash
@@ -58,7 +58,7 @@ function install_catalogsource {
     cp "${index_build_dir}/Dockerfile" "${rootdir}/_output/bkp.Dockerfile"
 
     # Replace the nightly bundle reference with the previously built bundle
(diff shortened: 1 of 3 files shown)


mgencur:
There's one upgrade test failing but it's related to duplicate events which are delivered, not to installing catalogsource, it can be seen that the previous version was installed correctly

skonto:
Does that mean we need to change the path for other images too like https://github.com/openshift-knative/serverless-operator/blob/bd939a49e0feaf508e496f85f11e7eac9cc8a0c0/openshift/ci-operator/source-image/Dockerfile#L4?

mgencur:
@skonto We don't need the path in the other images. We only changed the namespace for `serverless-operator` itself (main, and release-1.22) branch. They're independent

matzew (approved):
/lgtm
/approve

I could pull these images now...

matzew:
/hold

feel free to cancle the hold.......

PR #1574, opened 2022-05-18
Proper cleanup for CatalogSource and Tracing

mgencur:
* Clean also the index image build and configmap
* Do not fail if jaeger.jaegertracing.io CRD doesn't exist

### DIFF
--- a/hack/lib/tracing.bash
+++ b/hack/lib/tracing.bash
@@ -232,10 +232,12 @@ function teardown_opentelemetry_tracing {
     oc delete -n openshift-operators "${csv}" --ignore-not-found
   fi
 
+  # Do not remove Jaeger if it's part of Service Mesh
   if [[ $(oc get crd servicemeshcontrolplanes.maistra.io --no-headers | wc -l) != 1 ]]; then
-    # Do not remove Jaeger if it's part of Service Mesh
-    oc delete -n "${TRACING_NAMESPACE}" jaeger.jaegertracing.io jaeger
-    timeout 600 "[[ \$(oc get -n ${TRACING_NAMESPACE} deployment jaeger --no-headers | wc -l) != 0 ]]"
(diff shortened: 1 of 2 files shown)


matzew (approved):
/lgtm
/approve

/hold
just for the two more test runs

mgencur:
@matzew Sorry, there was still a problem. It should be fine now. Pushed changes.

hack/lib/tracing.bash
```diff
@@ -232,16 +218,20 @@ function teardown_opentelemetry_tracing {
     oc delete -n openshift-operators "${csv}" --ignore-not-found
   fi
 
+  # Do not remove Jaeger if it's part of Service Mesh
   if [[ $(oc get crd servicemeshcontrolplanes.maistra.io --no-headers | wc -l) != 1 ]]; then
-    # Do not remove Jaeger if it's part of Service Mesh
-    oc delete -n "${TRACING_NAMESPACE}" jaeger.jaegertracing.io jaeger
-    timeout 600 "[[ \$(oc get -n ${TRACING_NAMESPACE} deployment jaeger --no-headers | wc -l) != 0 ]]"
+    if [[ $(oc get -n "${TRACING_NAMESPACE}" jaeger.jaegertracing.io jaeger --no-headers | wc -l) != 0 ]]; then
+      oc delete -n "${TRACING_NAMESPACE}" jaeger.jaegertracing.io jaeger --ignore-not-found
```
skonto, line 224: the `ignore-not-found` is added because we previously deleted stuff even if they were not previously installed?

mgencur: I guess on this line we could remove the `--ignore-not-found` because we check one line above that it really exists. But it shouldn't harm. WDYT?

skonto: Ok does not harm.

skonto:
Still:
```
{  failed to wait for the created cluster claim to become ready: timed out waiting for the condition}
```

PR #1576, opened 2022-05-18
use new "/knative/" instead of "/openshift/" index image by default

maschmid:
The default nightly index image has been moved to "knative" namespace.

## Proposed Changes
- :bug: Use proper index image since it was renamed to "knative" namespace


matzew (approved):
was able to pull the image from the reg.

Thanks @maschmid

mgencur:
Thanks!

PR #1577, opened 2022-05-19
Increase Zipkin to -Xmx9G

mgencur:
Fixes JIRA #

## Proposed Changes

- :broom: Increase the Zipkin Xmx limit to 9G as we're still hitting OOM errors when running upgrade tests and collecting traces as can be seen [here](https://github.com/openshift-knative/serverless-operator/pull/1541). Another option would be to send fewer events. We're now sending events every 10 ms. We could probably set this to 20 ms but first try increasing the memory limits.

### DIFF
--- a/hack/lib/tracing.bash
+++ b/hack/lib/tracing.bash
@@ -53,12 +53,12 @@ spec:
               apiVersion: v1
               fieldPath: metadata.namespace
         - name: JAVA_OPTS
-          value: '-Xms128m -Xmx7G -XX:+ExitOnOutOfMemoryError'
+          value: '-Xms128m -Xmx9G -XX:+ExitOnOutOfMemoryError'
         - name: MEM_MAX_SPANS
           value: '10000000'
         resources:
           limits:
-            memory: 8Gi
+            memory: 10Gi
           requests:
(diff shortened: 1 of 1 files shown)


matzew (approved):
/lgtm
/approve

hack/lib/tracing.bash
```diff
@@ -53,12 +53,12 @@ spec:
               apiVersion: v1
               fieldPath: metadata.namespace
         - name: JAVA_OPTS
-          value: '-Xms128m -Xmx7G -XX:+ExitOnOutOfMemoryError'
```
skonto, line 56: I guess we could avoid the mem if we change the sample rate in Serving and Eventing configs to be `0.1`.
The latter is the value used upstream by default.

mgencur: We need "1.0" and store every single trace. In upgrades it is quite common that only one or two events are not delivered. So we need to trace every event. If we set this to "0.1" we'd not be able to get the required traces most of the time.

skonto: Ok although we could make this probabilistic in the future.

PR #1583, opened 2022-05-24
Make sure OPENSHIFT_BUILD_NAME is not unbound in local runs

mgencur:
## Proposed Changes

:bug:
- We run with some bash options which do not allow undefined variables. Make sure we have a default value in each case
-


matzew (approved):
/lgtm
/approve

I could now move forward and the unbound failure is no longer there

PR #1589, opened 2022-05-30
[release-1.22] SRVCOM-1873 change storage-version-migration names for eventing/serving

matzew:
Fixes SRVCOM-1873 

## Proposed Changes

- :bug: fix the name of the job temporarily for the upstream `storage-version-migration` components


openshift-knative-operator/hack/010-hack-storage-version-migrator-name.patch
```diff
@@ -0,0 +1,66 @@
...
++# /tmp/tmp.Edjrsh4tnc/serving-strg-version-migration.yaml
+ # Copyright 2020 The Knative Authors
+ #
+ # Licensed under the Apache License, Version 2.0 (the "License");
+@@ -18,12 +18,12 @@
+ apiVersion: batch/v1
+ kind: Job
+ metadata:
+-  generateName: storage-version-migration-serving-
++  generateName: strg-version-migration-serving-
```
skonto, line 44: Could we simply change only the job names usign some suffix specific to the release here?

mgencur: I suppose we'd need a patch "template" like we do with CSV. But it should be possible.
We'd first create the patch from the template (by adding new stuff in the `release-files` make target), then we'd have the correct patch to be applied in `generated-files` make target.

mgencur:
/lgtm
Let's merge this for now because we're in rush. @matzew is working on a proper fix within Golang which will be a little later.

PR #1593, opened 2022-05-31
[release-1.22] Use knative namespace for operator images

mgencur:
That's the namespace where operator images are promoted now.

## Proposed Changes

:bug:
- This is same as https://github.com/openshift-knative/serverless-operator/commit/84b6d1a66846d854fe2fbd70fe035cb5313db472 , just for release-1.22 branch.
- This change should not have any effect on the product. This should only help fix CI issues, especially in upgrade tests where we upgrade from 1.22 to a newer version. The serverless-bundle and serverless-stop-bundle must point to correct up-to-date images. So far they were pointing to a little outdated operator images from "openshift" namespace.


pierDipi (approved):
/lgtm
/approve

PR #1599, opened 2022-06-02
upgrade tests image_template

maschmid:
Currently upgrade tests were broken on `main`, as they tried to use a :v1.3 kafka broker image, even though we're not yet updated eventing-kafka-broker dep to 1.3

## Proposed Changes
- :bug: Sets proper image_template for upgrade tests in cases serving/eventing/kafka versions differ.


mgencur:
LGTM
Let's just get the upgrade tests passing. Maybe it's again caused by Nodes not available?

maschmid:
we may need to get https://github.com/openshift-knative/serverless-operator/pull/1600 into `main` first

PR #1600, opened 2022-06-02
Scale up workers in upgrade test

dsimansk:
## Proposed Changes

- Bring back scale up workers script

/cc @maschmid @mgencur

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -138,7 +138,7 @@ test-upstream-upgrade-testonly:
 test-upstream-upgrade:
 	UNINSTALL_STRIMZI="false" ./hack/strimzi.sh
 	TRACING_BACKEND=zipkin ./hack/tracing.sh
-	INSTALL_PREVIOUS_VERSION="true" INSTALL_KAFKA="true" TRACING_BACKEND=zipkin ENABLE_TRACING=true ./hack/install.sh
+	INSTALL_PREVIOUS_VERSION="true" INSTALL_KAFKA="true" TRACING_BACKEND=zipkin ENABLE_TRACING=true SCALE_UP=6 ./hack/install.sh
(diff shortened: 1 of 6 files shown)


mgencur:
/hold
For upgrade tests to pass.

mgencur:
hi @dsimansk , I think this PR is still missing some checks. The history of this PR shows that at least 2 from the last 6 runs failed with Nodes not ready:
{code}
{
  "lastHeartbeatTime": "2022-06-02T20:33:27Z",
   "lastTransitionTime": "2022-06-02T20:34:55Z",
   "message": "Kubelet stopped posting node status.",
   "reason": "NodeStatusUnknown",
   "status": "Unknown",
   "type": "Ready"
}
{code}
Also, the "ready machine count" is not the desired one as can be seen [here](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-knative_serverless-operator/1600/pull-ci-openshift-knative-serverless-operator-main-4.10-upgrade-tests-aws-ocp-410/1532439941689118720/artifacts/upgrade-tests-aws-ocp-410/gather-extra/artifacts/oc_cmds/machineconfigpools)  (ready 5 vs. desired 6)

maschmid:
In some of our downstream tests we use

```
function machineconfig_updates {
  # There should be only "True" and there should be at least one
  [ True = "$(oc get machineconfigpool --no-headers=true  '-o=custom-columns=UPDATED:.status.conditions[?(@.type=="Updated")].status' | uniq)" ]
}
```

maybe we should add a similar wait for the Updated condition = True here as well.

dsimansk:
@mgencur @maschmid 
I'd wonder if ran into something like AWS resource limit during worker node scaling. 

Should we maybe also try installed cluster with bigger worker nodes?

PR #1602, opened 2022-06-06
Bump serving deps to 1.3.0

skonto:
## Proposed Changes

- Bumps Serving to 1.3. 
- Deps like Knative Operator can be updated in a subsequent PR as there are some breaking changes in that package.
For example several types moved to a base pkg etc.
- Split into two PRs for easier review.


go.mod
```diff
@@ -158,29 +156,35 @@ require (
...
 	k8s.io/klog/v2 => k8s.io/klog/v2 v2.8.0
 )
 
-replace knative.dev/hack => knative.dev/hack v0.0.0-20220118141833-9b2ed8471e30
+replace knative.dev/hack => knative.dev/hack v0.0.0-20220224013837-e1785985d364
+
+replace knative.dev/pkg => knative.dev/pkg v0.0.0-20220314165618-a637a96a1bd9
 
-replace knative.dev/pkg => knative.dev/pkg v0.0.0-20220222214439-083dd97300e1
+replace knative.dev/operator v0.30.2 => knative.dev/operator v0.29.2
```
skonto, line 190: This can be removed if we update the operator in another PR.

go.mod
```diff
@@ -158,29 +156,35 @@ require (
...
 
 replace (
 	// Adjustments to align transitive deps
 	github.com/go-logr/logr => github.com/go-logr/logr v0.4.0
 	github.com/go-logr/zapr => github.com/go-logr/zapr v0.4.0
 	github.com/manifestival/manifestival => github.com/manifestival/manifestival v0.7.0
+
+	k8s.io/api v0.23.4 => k8s.io/api v0.22.5
+	k8s.io/apimachinery v0.23.4 => k8s.io/apimachinery v0.22.5
+	k8s.io/client-go v0.23.4 => k8s.io/client-go v0.22.5
```
skonto, line 182: The `k8s.io` replacements probably can be eliminated if we bump Eventing too.

nak3:
LGTM if CI passed.

But... Actually I still don't know how this PR was created :sweat_smile: The [README.md](https://github.com/openshift-knative/serverless-operator#3-update-the-golang-dependencies
) says:

> To update the dependencies, update the KN_VERSION variable in [update-deps.sh](https://github.com/openshift-knative/serverless-operator/blob/main/hack/update-deps.sh). Then run ./hack/update-deps.sh --upgrade (like we do upstream) to pull in all the correct versions. It's encouraged to commit the generated changes separately, to ease review.

But I cannot create the same change with this PR by `./hack/update-deps.sh --upgrade` and I cannot find the procedure in https://github.com/openshift-knative/serverless-release-manual as well.

Maybe SIG Release team knows how to create this PR? :sweat_smile: It would be great if we can find the docs for "How to Bump deps in S-O".

hack/update-deps.sh
```diff
@@ -13,11 +13,11 @@ set -o pipefail
 cd "${ROOT_DIR}"
 
 # This controls the knative release version we track.
-KN_VERSION="release-1.2"
+KN_VERSION="release-1.3"
 EVENTING_VERSION="release-v1.2"
 EVENTING_KAFKA_VERSION="release-v1.2"
 EVENTING_KAFKA_BROKER_VERSION="release-v1.2"
-SERVING_VERSION="release-v1.2.0"
+SERVING_VERSION="release-v1.3"
```
pierDipi, line 20: IMHO, dependencies should be bumped for every component together and we can remove replace directives.

skonto: I am in favor of this but I think the last time we didnt want to do a big PR for all deps.

go.mod
pierDipi: I'd rather bump every component dependencies in one go.

hack/update-deps.sh
skonto: Yes as I mentioned I prefer that, I will bump Eventing too. I added the replacements so Eventing can do the rest of the bumping anyway.

skonto:
When updating Eventing and removing replacements I hit multiple times:
```
# knative.dev/pkg/client/injection/kube/client
../../../vendor/knative.dev/pkg/client/injection/kube/client/client.go:205:5: cannot use (*wrapClient)(nil) (type *wrapClient) as type kubernetes.Interface in assignment:
	*wrapClient does not implement kubernetes.Interface (missing AutoscalingV2 method)
``` 
Reason is: this Serving PR https://github.com/knative/serving/commit/4126928029b952206f592f3670ccbe4a6089d9c8
updated K8s version on release-1.3 branch to v0.23.4 while on release-1.3 on knative.dev/pkg we are still on v0.22.5 https://github.com/knative/pkg/blob/a637a96a1bd980f099d76d07e7a965847114700e/go.mod#L46. This causes this failure. 
I guess we need to keep replacements fix Serving and then remove replacements?

pierDipi:
> @nak3 @pierDipi When updating Eventing and removing replacements I hit multiple times:
> 
> ```
> # knative.dev/pkg/client/injection/kube/client
> ../../../vendor/knative.dev/pkg/client/injection/kube/client/client.go:205:5: cannot use (*wrapClient)(nil) (type *wrapClient) as type kubernetes.Interface in assignment:
> 	*wrapClient does not implement kubernetes.Interface (missing AutoscalingV2 method)
> ```
> 
> Reason is: this Serving PR [knative/serving@4126928](https://github.com/knative/serving/commit/4126928029b952206f592f3670ccbe4a6089d9c8) updated K8s version on release-1.3 branch to v0.23.4 while on release-1.3 on knative.dev/pkg we are still on v0.22.5 https://github.com/knative/pkg/blob/a637a96a1bd980f099d76d07e7a965847114700e/go.mod#L46. This causes this failure. I guess we need to keep replacements in this PR, fix Serving and then remove replacements. WDYTH?

I'd fix/bump serving first

skonto:
> I'd fix/bump serving first

If it is fast enough sure. I will do the PRs upstream/downstream.

pierDipi:
what's the reason for using a previous k8s version?
if we cannot bump k8s versions but then the eventing bump won't work ...

skonto:
> what's the reason for using a previous k8s version?

I use whatever exists per brunch, each release branch is cut against a specific K8s version afaik. Eventing on release-1.3 should use the same no? It seems it does: https://github.com/knative/eventing/blob/release-1.3/go.mod#L42.

pierDipi:
I see, then with that commit in serving we won't get https://github.com/openshift-knative/serverless-operator/pull/1602#issuecomment-1147266863 anymore, right?

skonto:
> I see, then with that commit in serving we won't get https://github.com/openshift-knative/serverless-operator/pull/1602#issuecomment-1147266863 anymore, right?

The replacements exist both upstream and downstream on release-1.3 branch for Serving eg. https://github.com/openshift/knative-serving/blob/release-v1.3/go.mod. But in S-O we have to use them again.

pierDipi:
So, in my opinion, this repository (ideally) shouldn't have any replace directive for transitive dependencies and just inherit whatever serving/eventing use that would simplify the deps bump on every release because we don't need to update the replace directive each time (and we did miss this step in a past release, so to avoid it at all we should keep this ).

pierDipi:
It seems wrong to me that to fix serving (or whatever other component) we're hacking this repository, that's why I'm saying let's align serving/eventing to whatever works here, that's cleaner because the fix is localized to a specific component.

skonto:
> It seems wrong to me that to fix serving (or whatever other component) we're hacking this repository, that's why I'm saying let's align serving/eventing to whatever works here, that's cleaner because the fix is localized to a specific component.

Serving upstream had to update stuff as a backport and then added these replacements directives due to that google auth package for good or bad. I dont expect this to happen for 1.4 check here: https://github.com/knative/serving/blob/release-1.4/go.mod, Serving on release-1.4 is back to normal again. So this is an exception imho.

pierDipi:
> > It seems wrong to me that to fix serving (or whatever other component) we're hacking this repository, that's why I'm saying let's align serving/eventing to whatever works here, that's cleaner because the fix is localized to a specific component.
> 
> Serving upstream had to update stuff as a backport and then added these replacements directives due to that google auth package for good or bad. This makes S-O to pick up the wrong version sicne it has no knowledge of the replacement directives at the Serving side. I dont expect this to happen for 1.4 check here: https://github.com/knative/serving/blob/release-1.4/go.mod, Serving on release-1.4 is back to normal again, uses `v0.23.5`. So this is an exception imho. What alternative do you see (we don't want to revert the upstream backport because it fixes the tag resolution which is pretty important)? From my pov there is nothing to fix unless we find a different version for `github.com/google/go-containerregistry` that uses 0.22.5 and we backport upstream/downstream. I doubt it because Dave would have already done so but anyway and without the right version probably the upstream registry fix wont work.
> 
> However, I dont see why Serving/Eventing are not aligned here? Both use v0.22.5 on release-1.3. These are temporary replacements as I explained that will go away with release-1.4.

Is serverless operator involved in that feature? afaik, that fix is only needed in the serving controller which isn't built from this repository, so I don't get why replace directives here should fix that issue.

pierDipi (approved):
Ok, thanks for sharing the details and for the bump!

/lgtm
/hold

`/unhold` if it's ready to go in.

skonto:
@nak3 @mgencur any comment for getting this in?

skonto:
@matzew any comment wrt Eventing? Ok to unhold?

matzew:
/unhold

looks like for next release things are more smooth :pray: 

thanks for the questions here, @pierDipi  and the details, @skonto

mgencur:
It would be great to setup image mirroring in https://github.com/openshift/release/blob/master/core-services/image-mirroring/knative/mapping_knative_v1_3_quay

PR #1608, opened 2022-06-07
[wip] Bump upstream operator to 1.3

skonto:
## Proposed Changes

- Bumps knative operator to 1.3
- Removes the related replacement in go.mod
- Updates several files due to changes in operator's pkg structure. Some stuff like `Resources` are being deprecated and being replaced by `DeploymentsOverride` which means in another PR we would need to use the new field if not already.
- Split into several commits for smoother review.


knative-operator/pkg/apis/operator/v1alpha1/knativekafka_types.go
```diff
@@ -2,7 +2,7 @@ package v1alpha1
 
 import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	operatorv1alpha1 "knative.dev/operator/pkg/apis/operator/v1alpha1"
```
matzew, line 5: so nasty :-)

matzew:
Should we update our own CRDs, from upstream due to these changes:
https://github.com/knative/operator/pull/969/files

BUT: not sure they actually run the storage version migrator over there...

skonto:
> Should we update our own CRDs, from upstream due to these changes:
https://github.com/knative/operator/pull/969/files

My first question is whether we need a conversion webhook to migrate to beta.  If so it can be either at the knative-oeprator side or the openshift-knative-operator side (upstream operator already implements) one. I will open a jira to track the migration because it involves multiple steps it seems. Also the tests above fail due to the use of beta in controller's logic.

skonto:
Opened SRVCOM-1911 to track crd migration etc. Closing this.

PR #1611, opened 2022-06-07
Change sample-rate for tracing in upgrade tests to 0.3

mgencur:
This is to prevent OOM issues in Zipkin.

Having sample-rate 0.1 produces about 2,106.8 MiB of data in Zipkin. We
currently have -Xmx9G for Zipkin so setting this parameter to 0.3 in
order to have some buffer for cases when the tests take longer and
produce more data.

Fixes https://issues.redhat.com/browse/SRVCOM-1907

:bug:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -141,7 +141,7 @@ test-upstream-upgrade-testonly:
 test-upstream-upgrade:
 	UNINSTALL_STRIMZI="false" ./hack/strimzi.sh
 	TRACING_BACKEND=zipkin ./hack/tracing.sh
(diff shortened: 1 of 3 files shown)


pierDipi (approved):
/lgtm
/approve

pierDipi:
```
tracing_test.go:72: Error verifying traces: timed out waiting for the condition: expected at least 100 traces, got 34
```
This seems related and it seems the expected number

matzew (approved):
/lgtm
/approve

PR #1613, opened 2022-06-08
Place Zipkin pod on a dedicated node

mgencur:
This is a follow up on https://github.com/openshift-knative/serverless-operator/pull/1611
Fixes https://issues.redhat.com/browse/SRVCOM-1908

If the Zipkin is placed on the same node as Kafka and collect a lot of traces, there might be all kinds of intermittent errors in tests. The nodes can stop reacting if the memory reaches limits.

## Proposed Changes

- Label one node for Zipkin and also place a taint on it
- Force placing the Zipkin pod on the node via nodeAffinity
- Prevent other pods from running on the same node
- Do the setup before installing other components so that they're placed on different nodes
- Place Zipkin pod on a dedicated node only for upgrade tests where we gather a lot of traces, for other test suites it's not a problem

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,8 +139,8 @@ test-upstream-upgrade-testonly:
 	TEST_KNATIVE_KAFKA=true TEST_KNATIVE_E2E=false TEST_KNATIVE_UPGRADE=true ./test/upstream-e2e-tests.sh
 
 test-upstream-upgrade:
+	TRACING_BACKEND=zipkin ZIPKIN_DEDICATED_NODE=true ./hack/tracing.sh
 	UNINSTALL_STRIMZI="false" ./hack/strimzi.sh
-	TRACING_BACKEND=zipkin ./hack/tracing.sh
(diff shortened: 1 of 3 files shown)


dsimansk:
@mgencur would this change help and even work scaling workers? Wdyt?

mgencur:
@dsimansk Can you rephrase it? I don't understand.

pierDipi (approved):
/lgtm
/approve

pierDipi:
> @mgencur would this change help and even work scaling workers? Wdyt?

as I understand your question, I don't think it helps to scale workers (it might only make things worst), we're grabbing a worker node (whatever the worker poll / machine set is) from the cluster to place only Zipkin, so if we're scaling to 6 nodes, 1 node is only for Zipkin, so we're kinda back at using 5 nodes.

mgencur:
The Zipkin Pod could have been causing a lot of issues. Scaling to 6 might not be necessary. I want to test going back to 4 workers (1 for zipkin and 3 for other pods). But let me do this via a separate PR.

PR #1614, opened 2022-06-08
[wip] [SRVKS-900] Backport env override in upstream operator

skonto:
## Proposed Changes

- Enables overriding of env vars in Knative operator.
- Backports https://github.com/knative/operator/pull/1085. The approach here is preferred compared to having an annotation on CR https://github.com/knative/operator/pull/1083, (note that the env setting cannot be replaced with a setting in config-net-istio as we need to apply informer filtering when net-istio controller starts at the sharedmain call).
- Patching the crds shouldnt create compatibility issues as the fields should be preserved if operator is downgraded. 
Also in the past the crds have been patched multiple times eg. Resource overrides.
- Discussion for backporting [here](https://coreos.slack.com/archives/CF5ANN61F/p1654114246116849) as it is hard to backport upstream anyway.
- This PR can be reverted once operator moves to 1.5 (we are a few versions behind).
- To test:
``` 
export KO_DOCKER_REPO=...
make images
make install-full-mesh
FULL_MESH=true INSTALL_SERVING="false" INSTALL_EVENTING="false" ./hack/install.sh
```
```
Apply:

kind: KnativeServing
metadata:
  name: knative-serving
  namespace: knative-serving
spec:
  ingress:
    istio:
      enabled: true
  deployments:
    - annotations:
        sidecar.istio.io/inject: "true"
        sidecar.istio.io/rewriteAppHTTPProbers: "true"
      name: activator
    - annotations:
        sidecar.istio.io/inject: "true"
        sidecar.istio.io/rewriteAppHTTPProbers: "true"
      name: autoscaler
    - annotations:
        test: "true"
      name: net-istio-controller
    - env:
        - container: controller
          envVars:
            - name: ENABLE_SECRET_INFORMER_FILTERING_BY_CERT_UID
              value: "true"
      name: net-i
```
Observe deployment env:

```
$ oc get pods  net-istio-controller-57dff5786d-sq79s     -n knative-serving -o jsonpath='{.spec.containers[].env}' | jq .
[
  {
    "name": "SYSTEM_NAMESPACE",
    "valueFrom": {
      "fieldRef": {
        "apiVersion": "v1",
        "fieldPath": "metadata.namespace"
      }
    }
  },
  {
    "name": "CONFIG_LOGGING_NAME",
    "value": "config-logging"
  },
  {
    "name": "CONFIG_OBSERVABILITY_NAME",
    "value": "config-observability"
  },
  {
    "name": "METRICS_DOMAIN",
    "value": "knative.dev/net-istio"
  },
  {
    "name": "ENABLE_SECRET_INFORMER_FILTERING_BY_CERT_UID",
    "value": "true"
  }
]
```
Full test using the net-istio issue [here](https://gist.github.com/skonto/6010c0da420b044dd768927408f9472b).


nak3:
Not sure about upstream, but our downstream will default `ENABLE_SECRET_INFORMER_FILTERING_BY_CERT_UID` to `true` very soon, right?
(Note, we have the breaking change process as https://spaces.redhat.com/pages/viewpage.action?spaceKey=SRVLS&title=Breaking+change+and+deprecation+feature+process so we can change it 1.25 at the earliesst.)

If so I wonder that we can use your [annotation idea](https://github.com/knative/operator/pull/1083 ) in the meantime instead of this huge backport.

The implementation also could be simple on downstream :partying_face:  as we just support it around `openshift-knative-operator/pkg/serving/extension.go` by `InjectEnvironmentIntoDeployment`.

skonto:
@nak3 
> Not sure about upstream, but our downstream will default ENABLE_SECRET_INFORMER_FILTERING_BY_CERT_UID to true very soon, right?

Depends on the feedback but this should be the case for both upstream and downstream. Downstream sooner.

> (Note, we have the breaking change process as https://spaces.redhat.com/pages/viewpage.action?spaceKey=SRVLS&title=Breaking+change+and+deprecation+feature+process so we can change it 1.25 at the earliesst.)

If this is a breaking change then https://github.com/openshift-knative/serverless-operator/commit/1ca4d072bec7aa40b7b5b4942d26a478344b1fb1 was not? I don't think it is otherwise all changes in crds we have done should have been so. This field is optional. /cc @matzew wdyth?

> If so I wonder that we can use your https://github.com/knative/operator/pull/1083 in the meantime instead of this huge backport. The implementation also could be simple on downstream partying_face as we just support it around openshift-knative-operator/pkg/serving/extension.go by InjectEnvironmentIntoDeployment.

Yes we can do that too. Yeah I am aware of that injection func.

nak3:
> If this is a breaking change then https://github.com/openshift-knative/serverless-operator/commit/1ca4d072bec7aa40b7b5b4942d26a478344b1fb1 was not? I don't think it is otherwise all changes in crds we have done should have been so. This field is optional. /cc @matzew wdyth?

No, the point is not CRD change. Changing `ENABLE_SECRET_INFORMER_FILTERING_BY_CERT_UID` to `true` by default is the breaking change.

skonto:
> No, the point is not CRD change. Changing ENABLE_SECRET_INFORMER_FILTERING_BY_CERT_UID to true by default is the breaking change.

Ok I misread then. Yeah that is one. So for 1.24 my thinking is to have it off, let customer try it and then at 1.25 change to true and announce the change. Thanks for the pointer.

skonto:
Closing this in favor of #1616

PR #1619, opened 2022-06-13
[main] Test downgrades

mgencur:
Fixes https://issues.redhat.com/browse/SRVCOM-1863

## Proposed Changes

- perform downgrade via OLM
- include post-downgrade tests from upstream
- include previous versions of components in project.yaml

### DIFF
--- a/olm-catalog/serverless-operator/project.yaml
+++ b/olm-catalog/serverless-operator/project.yaml
@@ -40,7 +40,7 @@ dependencies:
   operator: 1.2.2
   # Previous versions required for downgrade testing
   previous:
-    serving: 1.1.2
-    eventing: 1.1.0
+    serving: 1.2.0
+    eventing: 1.2.1
     eventing_kafka: 1.1.0
-    eventing_kafka_broker: 1.1.3
+    eventing_kafka_broker: 1.2.3
--- a/olm-catalog/serverless-operator/project.yaml
+++ b/olm-catalog/serverless-operator/project.yaml
@@ -38,3 +38,9 @@ dependencies:
   eventing_kafka_broker_artifacts_branch: release-v1.2
   cli: 1.2.0
   operator: 1.2.2
+  # Previous versions required for downgrade testing
+  previous:
+    serving: 1.1.2
+    eventing: 1.1.0
+    eventing_kafka: 1.1.0
+    eventing_kafka_broker: 1.1.3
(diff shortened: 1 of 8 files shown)


olm-catalog/serverless-operator/project.yaml
```diff
@@ -38,3 +38,9 @@ dependencies:
   eventing_kafka_broker_artifacts_branch: release-v1.2
   cli: 1.2.0
   operator: 1.2.2
+  # Previous versions required for downgrade testing
+  previous:
```
skonto, line 42: This is fine, just curious is there a way to get previous in some other way?

mgencur: I'm not aware of a good way to get that info. It might also be beneficial to have this info explicit here. Whenever someone updates the "current" version they should also update the "previous" version in the same pull request.
But, if you're aware of a better way let me know.

test/upgrade/installation/serverless.go
```diff
@@ -53,3 +53,64 @@ func UpgradeServerless(ctx *test.Context) error {
...
 }
+
+func DowngradeServerless(ctx *test.Context) error {
+	const subscription = "serverless-operator"
+
+	if err := test.DeleteSubscription(ctx, subscription, test.OperatorsNamespace); err != nil {
+		return err
+	}
+
+	if err := test.DeleteClusterServiceVersion(ctx, test.Flags.CSV, test.OperatorsNamespace); err != nil {
```
skonto, line 64: So the idea is to keep the CRDs, install previous and test again. In the future we plan to use new crd versions and even if we have a conversion webhook when we move to a previous version I suspect the CR will not get downgraded.

olm-catalog/serverless-operator/project.yaml
skonto: I was thinking of the olm bundle image or something similar but dont think it is better, even if it was possible. Better be explicit.

test/lib.bash
```diff
@@ -256,7 +260,7 @@ EOF
       oc delete namespace serving-tests
     fi
     oc create namespace serving-tests
-    go_test_e2e -run=TestClusterUpgrade -timeout=190m "${common_opts[@]}" \
+    go_test_e2e -run=TestClusterUpgrade -timeout=220m "${common_opts[@]}" \
```
matzew, line 263: wise to increase :sweat_smile:

matzew (approved):
/lgtm
/approve

/unhold

mgencur:
@skonto @matzew Thanks for reviews.

PR #1634, opened 2022-06-29
Enable tracing when installing custom resources instead of patching t…

mgencur:
…hem later

Fixes [SRVCOM-1947](https://issues.redhat.com/browse/SRVCOM-1947)

## Proposed Changes

- Configure Tracing from the beginning rather than patching the custom resource later when it already exists


pierDipi (approved):
/lgtm
/approve

mgencur:
@pierDipi Thanks!

PR #1642, opened 2022-07-11
Remove istio, kourier patches

skonto:
## Proposed Changes

- Part of the work for the SIG release
- Removes the remaining bits
- Moves manifests to the midstream repos

Pending:
1) PR at serving/net-istio,kourier side
2) ingress image injection

PR #1644, opened 2022-07-13
Use an amd64 tag for the opm image

skonto:
## Proposed Changes

- Facing the issue [here](https://github.com/openshift/knative-serving/pull/1169#issuecomment-1182599769). Our jobs use `amd64` afaik. The image changed a few hours ago and things broke.
- `opm:latest` is derived from an arm64 arch check [here](https://quay.io/repository/operator-framework/opm?tab=tags).
- Discussion on slack [here](https://coreos.slack.com/archives/C3VS0LV41/p1657671513007229).

### DIFF
--- a/olm-catalog/serverless-operator/index/Dockerfile
+++ b/olm-catalog/serverless-operator/index/Dockerfile
@@ -1,6 +1,6 @@
 FROM openshift/origin-base as builder
 
-COPY --from=quay.io/operator-framework/opm:latest /bin/opm /bin/opm
+COPY --from=quay.io/operator-framework/opm:v1.23.2 /bin/opm /bin/opm
 
 # Copy declarative config root into image at /configs
 COPY configs /configs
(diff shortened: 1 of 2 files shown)


skonto:
Hitting it here too: #1642  in another form:
```
[1/2] STEP 4/5: RUN /bin/opm init serverless-operator --default-channel=stable --output yaml >> /configs/index.yaml
/bin/sh: /bin/opm: cannot execute binary file: Exec format error
```

olm-catalog/serverless-operator/index/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM openshift/origin-base as builder
 
-COPY --from=quay.io/operator-framework/opm:latest /bin/opm /bin/opm
+COPY --from=quay.io/operator-framework/opm:v1.23.2-amd64 /bin/opm /bin/opm
```
pierDipi, line 3: `:master` is multi arch, can we use that one?

pierDipi: or `v1.23`

skonto: Is it safe to use master? that tag is multi arch too.

pierDipi: `v1.23.2-amd64` isn't `v1.23` is multi arch

pierDipi: > Is it safe to use master?

No, it's not but so far we have used "latest" which isn't safe as well

pierDipi: I'd go with `v1.23`

skonto:
flacky test
/retest

pierDipi (approved):
/lgtm
/approve

PR #1649, opened 2022-07-18
SRVCOM-1981 Use origin-operator-registry that includes opm binary

mgencur:
https://issues.redhat.com/browse/SRVCOM-1981

* Use specific version of the downstream origin-operator-registry image that includes OPM binary instead of "upstream latest" version which might change.

Related Slack discussion: https://coreos.slack.com/archives/C3VS0LV41/p1657876220715119

### DIFF
--- a/olm-catalog/serverless-operator/index/Dockerfile
+++ b/olm-catalog/serverless-operator/index/Dockerfile
@@ -1,17 +1,17 @@
 FROM openshift/origin-base as builder
 
-COPY --from=quay.io/operator-framework/opm:v1.23.2 /bin/opm /bin/opm
+COPY --from=quay.io/openshift/origin-operator-registry:4.11 /bin/opm /bin/opm
 
 # Copy declarative config root into image at /configs
 COPY configs /configs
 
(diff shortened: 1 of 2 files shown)


mgencur:
We're operating with internal registry and we need the `--skip-tls-verify` option for `opm render`. This option is available since `origin-operator-registry:4.11` but our latest ocp version is 4.10. When I bump the max ocp version to 4.11 some files are re-generated and e.g. documentation links pointing to 4.11 don't work yet (4.11 is supposed to be released on Aug 3).
One option would be to wait with this PR until 4.11 is live. Then it would work as it is now in this PR.
As an intermediate solution, we could use a specific version of the "upstream" opm (e.g. v1.13) or we can leave the code base as is because the "latest" currently works well.

mgencur:
/hold
(should be merged when OCP 4.11 goes live)

mgencur:
We've decided to hardcode the 4.11 version in the dockerfile as it is the only version that currently works without issues.

skonto:
Let's also add SRVCOM-1981 at the title of the PR.

PR #1658, opened 2022-07-26
[SRVKE-1192] Remove sugar-controller

pierDipi:
sugar-controller is no longer a separate deployment in 1.24.


aliok (approved):
/lgtm
/approve

PR #1660, opened 2022-07-27
Replicate legacy sugar controller config during tests

pierDipi:
Copied from https://github.com/knative/eventing/blob/main/test/config/configmaps/config-sugar.yaml


aliok (approved):
/lgtm
/approve

PR #1679, opened 2022-08-24
Use backwards compatible way to get prometheus token

mgencur:
This is deprecated on OCP 4.11:
Command "get-token" is deprecated, and will be removed in the
future version. Use oc create token instead


mgencur:
This will be required in the near future. When I use client and serverless of version 4.11 and call the original command I get this:
```
Command "get-token" is deprecated, and will be removed in the future version. Use oc create token instead.
error: could not find a service account token for service account "prometheus-k8s"
```
So, on 4.11 the token doesn't seem to be included in the service account.

mgencur:
There seems to be a cluster pool for 4.11 now: https://docs.ci.openshift.org/docs/how-tos/cluster-claim/#existing-cluster-pools . We should switch CI to use it soon.

matzew (approved):
/lgtm
/approve

PR #1682, opened 2022-08-26
[SRVKS-955] Allow to override CERTS_SECRET_NAME value via config-network

nak3:
The default ingress certificate name is configurable via IngressController CR from `router-certs-default`.

When users changed the name, Kourier also needs to point to the customized name. So this patch allows users to
point to the name via `"openshift-ingress-default-certificate"` in `config-network`.

### DIFF
--- a/openshift-knative-operator/pkg/serving/kourier.go
+++ b/openshift-knative-operator/pkg/serving/kourier.go
@@ -1,6 +1,8 @@
 package serving
 
 import (
+	"strings"
+
 	mf "github.com/manifestival/manifestival"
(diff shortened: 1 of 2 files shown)


skonto:
Probe test is failing (we already know that) also lint seems timing out.

nak3:
lint was fixed by another PR.

skonto:
Saw the other PR so why lint fails? Needs rebasing?

skonto:
@nak3 So this PR is for setting some cert other than the OCP ingress one to be used by internal encryption. Correct?

mgencur:
I tested this PR locally with my cluster with those specific certificates and it worked!

PR #1695, opened 2022-09-05
[SRVCOM-1911] Upgrade Openshift Knative Operator CRDs to v1beta1

skonto:
## Proposed Changes

- Upgrades Knative operator to 1.4.2
- Migrates CRDs and code base to v1beta1
- Adds conversion webhook logic as part of the knative operator deployment
- Updates csv content for conversion webhook definition
- Handles upgrade/downgrade paths
- Ideally it should go in after 1.25 is cut and before the usual bumps.
- Most of the files updates are vendor file updates and other updates due to the cr version change. It is easier to review the PR by checking the files instead of the commits.
- Upstream webhook config is [here](https://github.com/knative/operator/blob/main/config/webhook/webhook.yaml). We dont set `terminationGracePeriodSeconds: 300` for now, since it affects knative-operator pod too.
Note: pls consult the related internal ft doc for more info and details.


skonto:
```
Unable to connect to the server: dial tcp: lookup api.ci-ocp-4-10-amd64-aws-us-west-1-qvkfz.hive.aws.ci.openshift.org on 172.30.0.10:53: read udp 10.128.44.2:52235->172.30.0.10:53: i/o timeout
```

nak3:
`346 file` changes :sweat_smile: But maybe a lot of changes are just `s/v1alpha1/v1beta1/g`?

We probably should review carefully in `openshift-knative-operator/cmd/operator/main.go`? If there is any code which you changed the logic, please let us know. I may overlook them.

skonto:
@nak3 

> 346 file changes sweat_smile But maybe a lot of changes are just s/v1alpha1/v1beta1/g?

As I mentioned there are ~280 files changed only because of the go.mod changes. Knative operator version change brings vendor changes. Also it is not just version changes. The codebase uses v1alpha1 everywhere, there is webhook addition and other changes pls refer to the ft track doc. 

> We probably should review carefully in openshift-knative-operator/cmd/operator/main.go? If there is any code which you changed the logic, please let us know. I may overlook them.

Check the file changes and navigate to the right [file](https://github.com/openshift-knative/serverless-operator/pull/1695/files#diff-bb8a07b2f24ca76693efa95a4d2a63685cc9e9af7f24f462a19b8c6e89e89a6e) pls. Not sure why its not visible :D 
image

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
```diff
@@ -709,14 +759,15 @@ spec:
       type: ValidatingAdmissionWebhook
       deploymentName: knative-openshift
       admissionReviewVersions:
+        - v1alpha1
         - v1beta1
```
pierDipi, line 763: `admissionReviewVersions` have nothing to do with CRD version, so we shouldn't add v1alpha1 here.

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
```diff
@@ -728,14 +779,15 @@ spec:
...
```
pierDipi, line 783: Same here.

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
```diff
@@ -766,14 +818,15 @@ spec:
       type: MutatingAdmissionWebhook
       deploymentName: knative-openshift
       admissionReviewVersions:
+        - v1alpha1
         - v1beta1
```
pierDipi, line 822: Same here.

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
```diff
@@ -785,21 +838,46 @@ spec:
...
```
pierDipi, line 842: Same here.

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
```diff
@@ -785,21 +838,46 @@ spec:
...
+      deploymentName: knative-operator-webhook
+      sideEffects: None
+      webhookPath: /resource-conversion
+      conversionCRDs:
+        - knativeservings.operator.knative.dev
+    - generateName: conversion.knativeeventings.operator.knative.dev
+      type: ConversionWebhook
+      admissionReviewVersions:
+        - v1alpha1
+        - v1beta1
```
pierDipi, line 873: Same here.

hack/patches/008-migration.patch
```diff
@@ -0,0 +1,88 @@
...
++			})
++			if err != nil {
++				logger.Fatalw("Unable to migrate existing custom resources", zap.Error(err))
++			}
++		}()
+
+-		err = common.MigrateCustomResource(ctx, dynamicclient.Get(ctx), apixclient.NewForConfigOrDie(injection.GetConfig(ctx)))
+-		if err != nil {
+-			logger.Fatalw("Unable to migrate existing custom resources", zap.Error(err))
+-		}
```
pierDipi, line 85: why moving the migration to another go routine?

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
pierDipi: These versions tell the API server which request version the webhook supports, Knative webhooks support v1beta1 and v1 admission review versions

matzew:
@skonto is this for 1.25?

docs/full-mesh.md
```diff
@@ -66,7 +66,7 @@ Then, create or update KnativeServing with the following options.
 
 ```shell
 cat <<EOF | oc apply -f -
-apiVersion: operator.knative.dev/v1alpha1
```
matzew, line 69: Do we have a 1.26 ticket for updating docs?

go.mod
```diff
@@ -27,17 +27,19 @@ require (
...
 	knative.dev/eventing-kafka v0.32.0
 	knative.dev/eventing-kafka-broker v0.0.0-00010101000000-000000000000
-	knative.dev/hack v0.0.0-20220524153203-12d3e2a7addc
+	knative.dev/hack v0.0.0-20220601014117-bafe4b15ca50
 	knative.dev/networking v0.0.0-20220412163509-1145ec58c8be
 	knative.dev/operator v0.31.2
-	knative.dev/pkg v0.0.0-20220524202603-19adf798efb8
-	knative.dev/serving v0.30.0
+	knative.dev/pkg v0.0.0-20220601013938-bac16f2394ce
+	knative.dev/serving v0.32.0
```
matzew, line 34: The bump of serving, does it need to be in here? Or can it be separated? E.g. eventing seems to be already on 0.32 :man_shrugging:

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
skonto: Ok will remove that part.

hack/patches/008-migration.patch
skonto: Because it blocks the controller creation and since the migration requires this webhook itself in order to work this never moves on. It is a deadlock.

go.mod
skonto: Yes I tried to align those. v0.32.0 is Serving 1.5 afaik but for operator is not. I could replace it to keep it as is.

skonto:
> @skonto is this for 1.25?
This goes to 1.26 as discussed.

go.mod
skonto: Actually no, we already replace the versions with the midstream ones so we are ok.

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
skonto: @pierDipi Actually reading [here](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/):

```
API servers send the first AdmissionReview version in the admissionReviewVersions list they support. If none of the versions in the list are supported by the API server, the configuration will not be allowed to be created. If an API server encounters a webhook configuration that was previously created and does not support any of the AdmissionReview versions the API server knows how to send, attempts to call to the webhook will fail and be subject to the [failure policy](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy).
```
The failure policy we have is ignore so it makes sense to have that list of versions.

pierDipi: > The failure policy we have is ignore so it makes sense why we v1beta1 is not failing so far. According to the above quote it does not make sense what we have there so far. 

v1beta1 is not supposed to fail (regardless of the policy), why do you think it should fail on it or why doesn't it make sense?

> Now that we have both v1alpha1 and v1beta1 this should be ok.

Nope, the subject of the paragraph is `admissionReviewVersions` and not CRD version, v1alpha1 is not supported by Knative webhooks so it should not be there.

docs/full-mesh.md
skonto: Created SRVCOM-2105 for this.

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
skonto: These are not knative webhooks it is Knative operator's webhook. The version requested at the admissionReview is via an object version that the webhook should support.

skonto: Ok I guess you are refering to the k8s requirements to set [v1beta1 and v1 only](https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/request/admissionreview.go#L150). Ok will change.

matzew (approved):
/lgtm
/approve

PR #1698, opened 2022-09-07
Enable internal encryption by default in tests

mgencur:
## Proposed Changes

- Enable internal encryption by default
- Disable encryption for Service Mesh
- Disable tracing test which doesn't allow Activator in path
- Remove TestKnativeServiceHTTPS: The Route is passthrough with insecureEdgeTerminationPolicy:Redirect. The kourier gateway serves only HTTPS and it requires certificates.


nak3:
/lgtm
/hold

Thank you! Feel free to unhold.

mgencur:
@nak3 Thanks for review. Just curious. Why the hold?

PR #1709, opened 2022-09-15
[SRVKE-1303] Document how to install and upgrade midstream

pierDipi:
Documentation for:
- Installation
- Upgrade
- Downgrade

Fixes JIRA #SRVKE-1303

### DIFF
--- a/docs/install-midstream.md
+++ b/docs/install-midstream.md
@@ -113,7 +113,7 @@ spec:
 
 ### Downgrade
 
-OLM doesn't support downgrades, so referencing in a Subscription a previous version isn't supported, so we need a few
+OLM doesn't support downgrades, so referencing a previous version in a Subscription isn't supported, so we need a few
 steps to work around the limitation.
 
 1. Delete the `Subscription`
--- a/docs/install-midstream.md
+++ b/docs/install-midstream.md
@@ -87,7 +87,6 @@ spec:
   # Change the CatalogSource name
   source: serverless-operator-v1-24-0
   sourceNamespace: openshift-marketplace
-  # Automatic is fine and safe since there is only one version in the referenced catalog
   installPlanApproval: Automatic
 ```
 
@@ -107,7 +106,6 @@ spec:
   # Change the CatalogSource name
   source: serverless-operator-v1-25-0
   sourceNamespace: openshift-marketplace
-  # Automatic is fine and safe since there is only one version in the referenced catalog
   installPlanApproval: Automatic
 ```
 
@@ -146,6 +144,5 @@ KnativeEventing, and KnativeKafka resources won't work but Knative components wi
     # Change the CatalogSource name
     source: serverless-operator-v1-24-0
     sourceNamespace: openshift-marketplace
-    # Automatic is fine and safe since there is only one version in the referenced catalog
     installPlanApproval: Automatic
   ```
--- a/docs/install-midstream.md
+++ b/docs/install-midstream.md
@@ -111,7 +111,7 @@ spec:
 
 ### Downgrade
 
-OLM doesn't support downgrades, so referencing a previous version in a Subscription isn't supported, so we need a few
+OLM **does not** support downgrades, so referencing a previous version in a Subscription isn't supported, so we need a few
 steps to work around the limitation.
 
 1. Delete the `Subscription`
--- a/docs/install-midstream.md
+++ b/docs/install-midstream.md
@@ -1,0 +1,151 @@
+# Installing and upgrading OpenShift Serverless midstream
+
+Navigate to resources directory:
+
+```shell
+cd resources-install-midstream
+```
+
+## Installation
+
+Create the `openshift-serverless` namespace:
+
+```shell
+oc apply -f 100-oss-namespace.yaml
+```
+
+Create an `OperatorGroup`:
+
+```shell
+oc apply -f 200-oss-operator-group.yaml
+```
+
+### Create a CatalogSource for each version
+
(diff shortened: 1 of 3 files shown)


pierDipi:
@r00ta @danielezonca can you please take a look at this PR?

pierDipi:
Waiting for feedback

/hold

docs/install-midstream.md
```diff
@@ -0,0 +1,151 @@
...
+  name: serverless-operator-nightly
+  namespace: openshift-marketplace
+spec:
+  displayName: Serverless Operator
+  # Replace the version with whatever version you want to install
+  image: registry.ci.openshift.org/knative/openshift-serverless-nightly:serverless-index
+  publisher: Red Hat
+  sourceType: grpc
+```
+
```
matzew, line 70: Let's add a note here, that `main` is not recommended, since it likely more unstable than actual branches which usually back a released/supported version

docs/install-midstream.md
```diff
@@ -0,0 +1,151 @@
...
+  # Change the CatalogSource name
+  source: serverless-operator-v1-25-0
+  sourceNamespace: openshift-marketplace
+  # Automatic is fine and safe since there is only one version in the referenced catalog
+  installPlanApproval: Automatic
+```
+
+### Downgrade
+
+OLM doesn't support downgrades, so referencing in a Subscription a previous version isn't supported, so we need a few
```
matzew, line 116: I'd write does not in bold :smile: (**does not**)

matzew: I am also not sure if we should really document this, if the underlying framework has it.

Perhaps we can move this to a different markdown-file, and decorate it with warnings.

I am a bit afraid of folks reading and trying this, while we (as a product) are not really supporting that.


It is for sure valuable information, but perhaps make some more noise on what we want and what not

matzew:
Very nice work on getting this so properly documented! I added a few suggestions comments, but this is very useful information

matzew:
/retest
FWIW... :smile:

docs/install-midstream.md
mgencur, line 116:
```suggestion
OLM doesn't support downgrades, so referencing a previous version in a Subscription isn't supported, so we need a few
```

mgencur: Nit ^^

mgencur (commented):
Hey, some minor stuff. Otherwise looks ok.

docs/install-midstream.md
```diff
@@ -0,0 +1,151 @@
...
+metadata:
+  name: serverless-operator
+  namespace: openshift-serverless
+spec:
+  channel: stable
+  name: serverless-operator
+  # Change the CatalogSource name
+  source: serverless-operator-v1-25-0
+  sourceNamespace: openshift-marketplace
+  # Automatic is fine and safe since there is only one version in the referenced catalog
```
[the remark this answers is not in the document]
pierDipi: Ok, I'll remove it.

docs/install-midstream.md
```diff
@@ -0,0 +1,151 @@
...
+metadata:
+  name: serverless-operator
+  namespace: openshift-serverless
+spec:
+  channel: stable
+  name: serverless-operator
+  # Change the CatalogSource name
+  source: serverless-operator-v1-24-0
+  sourceNamespace: openshift-marketplace
+  # Automatic is fine and safe since there is only one version in the referenced catalog
```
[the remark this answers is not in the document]
pierDipi: Done, thanks!

docs/install-midstream.md
pierDipi: Done

docs/install-midstream.md
pierDipi: We're testing this downgrade path and it is the only viable option we're recommending

pierDipi: See SRVCOM-2037 and our downgrade tests

matzew: I am ok adding this here. Specificially this is our own midstream repo. No product doc.

I am just concerned of a boomerang. Generally good to add some extra "warning" to path like this.

docs/install-midstream.md
```diff
@@ -0,0 +1,148 @@
...
+
+2. Delete the `ClusterServiceVersion` associated with the specific release you have installed:
+
+  ```shell
+  oc delete csv -n openshift-serverless serverless-operator.v1.25.0
+  ```
+
+This step deletes the pods in `openshift-serverless`, these aren't critical since they provide a way to install Knative
+components, that means that until step 3 is complete, installation and configuration functionality on KnativeServing,
+KnativeEventing, and KnativeKafka resources won't work but Knative components will be up and running as previously.
```
danielezonca, line 131: This looks a bit hack-ish but it should be fine for now.
Are we testing this scenario in some way? 
Just to be sure that this doesn't break in the future

danielezonca (approved):
LGTM 👍

docs/install-midstream.md
pierDipi: That's what the OLM team suggested as an escape rule and we're testing it here [1] [while "Continual" tests run](https://github.com/openshift-knative/serverless-operator/blob/f2bd3b46b9774cb0b51afd64f1b57cb3f5b935fd/test/upgrade/upgrade_test.go#L51-L60).

[1] https://github.com/openshift-knative/serverless-operator/blob/f2bd3b46b9774cb0b51afd64f1b57cb3f5b935fd/test/upgrade/installation/serverless.go#L57-L116

danielezonca: Ok thanks for the clarification :)

PR #1714, opened 2022-09-19
Delete eventing ns from manifests

pierDipi:
Same as https://github.com/openshift/knative-eventing/pull/1877 but for SO due to https://github.com/openshift/knative-eventing/pull/1877#discussion_r968329524

### DIFF
--- a/vendor/knative.dev/operator/pkg/reconciler/knativeeventing/knativeeventing.go
+++ b/vendor/knative.dev/operator/pkg/reconciler/knativeeventing/knativeeventing.go
@@ -104,6 +104,7 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ke *v1alpha1.KnativeEven
 		return err
 	}
 	stages := common.Stages{
+		common.FilterNamespace("knative-eventing"),
 		common.AppendTarget,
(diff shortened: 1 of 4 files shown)


vendor/knative.dev/operator/pkg/reconciler/knativeeventing/knativeeventing.go
```diff
@@ -104,6 +104,7 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ke *v1alpha1.KnativeEven
 		return err
 	}
 	stages := common.Stages{
+		common.FilterNamespace("knative-eventing"),
```
matzew, line 107: So we exclude the namespace kind with the name.

Is patching vendored code really better, than manfiests?

pierDipi: Yes, because manifests are in a versioned path directory

pierDipi: which forces us to update patches every bump.

matzew (approved):
/lgtm
/approve

PR #1724, opened 2022-09-22
[main] Set memory requests to 10Gi when on dedicated node

mgencur:
The Zipkin pod was recently evicted and traces for events lost:
```
istio-system                   18m         Warning   Evicted                                      pod/zipkin-58fcd685d-shtj4                                         The node was low on resource: memory. Container zipkin was using 8725020Ki, which exceeds its request of 256Mi.
```
This PR should minimizes the chance of evicting the Pod.
Related discussion: https://coreos.slack.com/archives/CLUCMK7R6/p1663832224355229

## Proposed Changes

-
-
-

### DIFF
--- a/hack/lib/tracing.bash
+++ b/hack/lib/tracing.bash
@@ -2,7 +2,6 @@
 
 function install_tracing {
   if [[ "${TRACING_BACKEND}" == "zipkin" ]]; then
-    ZIPKIN_MEMORY_REQUESTS="256Mi"
     if [[ "$ZIPKIN_DEDICATED_NODE" == "true" ]]; then
       dedicate_node_to_zipkin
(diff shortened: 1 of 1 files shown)


hack/lib/tracing.bash
```diff
@@ -90,7 +92,7 @@ spec:
           limits:
             memory: 10Gi
           requests:
-            memory: 256Mi
+            memory: ${ZIPKIN_MEMORY_REQUESTS}
```
maschmid, line 95: should we define a default for ZIPKIN_MEMORY_REQUESTS in install_zipkin_tracing , in case it is invoked on its own, and not inside `install_tracing`

mgencur: OK. Pushed.

PR #1726, opened 2022-09-22
Make WaitForKnativeKafkaState wait longer

mgencur:
The original 5 minutes might not be enough in some cases, e.g. during downgrades in downstream.

The problem with short timeout can be seen e.g. here: https://master-jenkins-csb-serverless-qe.apps.ocp-c1.prod.psi.redhat.com/job/functional_tests/job/stream1_25/job/rolling-upgrade-1.25/1, the KnativeKafka pods are started about 5 minutes after other Eventing pods during downgrade. In the end the KnativeKafka CR status gets the right version but it's after the current timeout.

Fixes JIRA #

## Proposed Changes

-
-
-


pierDipi (approved):
/lgtm
/approve

PR #1727, opened 2022-09-23
Restricted pod admission policy for Zipkin pod

mgencur:
To prevent warnings on OCP 4.11 and make it work on 4.12.

Fixes JIRA #

## Proposed Changes

-
-
-

### DIFF
--- a/hack/lib/tracing.bash
+++ b/hack/lib/tracing.bash
@@ -73,6 +73,10 @@ spec:
       annotations:
         sidecar.istio.io/inject: "false"
     spec:
+      securityContext:
+        runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
       containers:
       - name: zipkin
         image: ghcr.io/openzipkin/zipkin:2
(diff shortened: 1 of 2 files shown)


hack/lib/tracing.bash
```diff
@@ -73,6 +73,10 @@ spec:
       annotations:
         sidecar.istio.io/inject: "false"
     spec:
+      securityContext:
```
skonto, line 76: Not sure if things will fail on 4.12 by default (need to test), It would fail if we set `restricted` on the tracing namespace for sure though. However, it does not do any harm I suppose.

mgencur:
@skonto Do you have an idea if this change will work on OCP 4.6 ? Otherwise I need to try before merging this.

mgencur:
This PR if for main (Serverless 1.26) which is supposed to work on OCP 4.8 and later. But I want to backport it to release-1.25 as well to get rid of the warnings. Will try locally on 4.6

mgencur:
@skonto Indeed, it was failing on 4.6. Pushed a fix and another fix for kafka-ui as well.

skonto:
@mgencur should we rebase the eror is related to `KafkaSinkSourceContinualTest`.

mgencur:
@skonto Why rebase? We know this test has been failing.

skonto:
/retest
ok thought it was fixed in the mean time.

skonto:
@mgencur stamping it feel free to unhold if everything is ok in previous releases.
/lgtm

PR #1730, opened 2022-09-23
allow strimzi users to delete topics

maschmid:
## Proposed Changes
- :bug: Allow the strimzi kafkausers to delete topics

### DIFF
--- a/hack/lib/strimzi.bash
+++ b/hack/lib/strimzi.bash
@@ -123,6 +123,12 @@ spec:
           name: "*"
         operation: Describe
         host: "*"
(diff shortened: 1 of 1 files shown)


maschmid:
Required for a kafka broker created with a my-tls|sasl-user to be deleted. Currently the SO tests don't need it, but our downstream soak tests do. Note that upstream eventing-kafka-broker also has it in their test kafka config https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/test/kafka/user-tls.yaml#L65-L72

pierDipi (approved):
/lgtm
/approve

PR #1736, opened 2022-09-27
Configure min.insync.replicas for Kafka cluster

pierDipi:
Increase data availability.


hack/lib/strimzi.bash
```diff
@@ -53,6 +53,7 @@ function install_strimzi_cluster {
           transaction.state.log.min.isr: 2
           inter.broker.protocol.version: "3.2"
           auto.create.topics.enable: "false"
+          min.insync.replicas: 2
```
matzew, line 56: I think I tried this before, and there were failures...

pierDipi: Ok, I'll check the errors

pierDipi:
Closing for now since failures aren't strictly related to this problem

PR #1739, opened 2022-09-27
[release-1.25] Update eventing-kafka-broker dependency

mgencur:
The goal is to get the latest test/pkg/addressable/send.go with this fix:
openshift-knative/eventing-kafka-broker@a034b9f

Fixes JIRA #

## Proposed Changes

-
-
-


pierDipi (commented):
To update deps we need to change this file to point to the correct midstream branch:
https://github.com/openshift-knative/serverless-operator/blob/398f94893fd951140ccd89ce10d8164001d49ca5/hack/update-deps.sh#L19

mgencur:
Isn't it the right branch? release-v1.4 
I just need the latest from that branch so I ran `go mod edit -replace knative.dev/eventing-kafka-broker=github.com/openshift-knative/eventing-kafka-broker@release-v1.4` and then `make generated-files` . Is there a better way? Thanks

pierDipi (approved):
/lgtm
/approve

PR #1742, opened 2022-09-29
[release-1.25] Fix massive kafka-controller restarts after toggling KnativeKafka components

matzew:
Fixes JIRA #1320

## Proposed Changes

- adding a sort by values, so we have an expected ordering of the strings that call out the controllers that are disabled
- only render the `args` if we have controllers that are disabled, otherwise do not render the `args`


pierDipi (approved):
/lgtm

PR #1749, opened 2022-10-03
Fix sample rate and Zipkin memory request

mgencur:
## Proposed Changes

- use `debug: false` for tracing which allows sample-rate to be used
- get rid of high memory request for Zipkin as it often fails to schedule


mgencur:
/hold
I want to see if traces are properly gathered in upgrade tests.

mgencur:
OK. Zipkin didn't run out of memory and there's a good number of collected traces. I think it works.

PR #1752, opened 2022-10-03
Bump deps to 1.5

skonto:
## Proposed Changes

- As per title. This completes the setup for 1.26.
- Split in two commits for easier review.


hack/update-deps.sh
```diff
@@ -33,6 +32,7 @@ REPLACE_DEPS=(
   "knative.dev/serving=github.com/openshift/knative-serving@${SERVING_VERSION}"
   "knative.dev/pkg=knative.dev/pkg@${KN_VERSION}"
   "knative.dev/hack=knative.dev/hack@${KN_VERSION}"
+  "knative.dev/networking=knative.dev/networking@${KN_VERSION}"
```
skonto, line 35: If we don't move there is a conflict with networking versions.

hack/update-deps.sh
```diff
@@ -13,16 +13,15 @@ set -o pipefail
 cd "${ROOT_DIR}"
 
 # This controls the knative release version we track.
-KN_VERSION="release-1.4"
-EVENTING_VERSION="release-v1.4"
+KN_VERSION="release-1.5"
+EVENTING_VERSION="release-v1.5"
 EVENTING_KAFKA_VERSION="release-v1.4"
```
skonto, line 18: @pierDipi do we have a plan to remove the deprecated eventing_kafka bits. Should I go on and do it in another PR?

pierDipi: I think we can remove the code dependency now, thanks for your help on this!

pierDipi (approved):
Thanks! 

If CI is happy, I'm happy with this PR, so
/lgtm

skonto:
As a reminder for future lint updates TODO:

> $ golangci-lint run
knative-operator/cmd/manager/main.go:132:13: G114: Use of net/http serve function that has no support for setting timeouts (gosec)
		if err := http.ListenAndServe(":8080", nil); err != nil && !errors.Is(err, http.ErrServerClosed) {
		          ^
$ /golangci-lint version
golangci-lint has version 1.49.0 built from cc2d97f3 on 2022-08-24T10:24:37Z

We are on 1.45 so it is ok for now.

hack/update-deps.sh
pierDipi: We should still keep the test-source-image reference, only the code/vendor dependency can be removed for now.

skonto:
@pierDipi ready for another stamp.

hack/update-deps.sh
skonto: We use the dep in tests too: https://github.com/openshift-knative/serverless-operator/blob/main/test/upgrade/upgrade_test.go#L27, so let's remove that in another PR.

pierDipi (approved):
/lgtm

hack/update-deps.sh
pierDipi: that code dependency is transitively imported by eventing-kafka-broker, so we will keep it but it's not gonna be enforced anymore

skonto: I can remove the one line if that is the goal here. It has no effect, it does not touch go.mod. Is that what you are describing?

PR #1755, opened 2022-10-05
[SRVKE-1324] Avoid reconciliation loops

pierDipi:
This is a  fix for SRVKE-1324

## Proposed Changes

- set creation timestamp to 0 on every resource
- global resync only when some config maps change (config-tracing and kafka-config-logging)
- remove match expression from webhook definition since it is changed by Knative webhook logic

### DIFF
--- a/knative-operator/deploy/resources/knativekafka/controller/eventing-kafka-controller.yaml
+++ b/knative-operator/deploy/resources/knativekafka/controller/eventing-kafka-controller.yaml
@@ -1648,7 +1648,6 @@ webhooks:
     reinvocationPolicy: IfNeeded
     matchPolicy: Equivalent
     namespaceSelector:
-      matchExpressions: [ ]
       matchLabels:
         app.kubernetes.io/name: knative-eventing
     objectSelector:
--- a/knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
+++ b/knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"strconv"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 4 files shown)


knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -681,3 +683,14 @@ func executeStages(instance *serverlessoperatorv1alpha1.KnativeKafka, manifest *
 	}
 	return nil
 }
+
+func removeCreationTimestamp(manifest *mf.Manifest, _ *serverlessoperatorv1alpha1.KnativeKafka) error {
+	// Avoid multiple unnecessary resource updates and reconciler looping
+	// due to creationTimestamp set to `null` on some resources.
```
skonto, line 689: When `null` is set? Could you use a predicate to filter these changes on meta instead?

pierDipi: I'm using the same approach the other functions are using, see 
https://github.com/openshift-knative/serverless-operator/blob/26a310ee58a38e70d3f7bd1d1d057891d06e287f/knative-operator/pkg/controller/knativekafka/images.go#L86-L88
(there are many other instances of the same logic)

matzew (approved):
/lgtm
/approve

knative-operator/deploy/resources/knativekafka/controller/eventing-kafka-controller.yaml
```diff
@@ -1648,7 +1648,6 @@ webhooks:
     reinvocationPolicy: IfNeeded
     matchPolicy: Equivalent
     namespaceSelector:
-      matchExpressions: [ ]
```
matzew, line 1651: upstream removes this here: https://github.com/knative-sandbox/eventing-kafka-broker/pull/2715

PR #1757, opened 2022-10-05
[SRVCOM-2118] Install console resources only if related crds are available

skonto:
## Proposed Changes

- This is meant for the case where we don't have console available.
- Tested as described [here](https://gist.github.com/skonto/1af0ad93d8a3b7bb1d6f5169f0ab7a72) with 4.12 nightly. We had to patch a few manifests due to SRVKS-934.
[knative-openshift.log](https://github.com/openshift-knative/serverless-operator/files/9727810/os.log), 
[knative-operator-webhook.log](https://github.com/openshift-knative/serverless-operator/files/9727813/opw.log)

Paths tested:

a) no console -> install S-O -> Install Console -> Install Serving/Eventing
b) no console -> install S-O -> install Serving/Eventing ->  Install Console
c)  Install Console -> install S-O -> Install Serving/Eventing
To install console just use:

```
  oc patch clusterversion/version --type merge -p '{"spec":{"capabilities":{"additionalEnabledCapabilities":["Console"]}}}'

```
-  If users have no console installed and they decide to install it later we require the knative-openshift pod to be restarted.
- Fixes the issue with cache sync as described in the ticket: 
```
failed to wait for knativeserving-controller caches to sync: no matches for kind "ConsoleCLIDownload" in version "console.openshift.io/v1"
```

### DIFF
--- a/knative-operator/cmd/manager/main.go
+++ b/knative-operator/cmd/manager/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"errors"
 	"flag"
 	"fmt"
@@ -10,6 +11,7 @@ import (
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/apis"
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/common"
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/controller"
+	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/controller/knativeserving/consoleclidownload"
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/monitoring"
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/monitoring/dashboards/health"
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/webhook/knativeeventing"
@@ -18,7 +20,11 @@ import (
 	"github.com/spf13/pflag"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
+	apiextension "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/rest"
+
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
@@ -166,8 +172,21 @@ func setupServerlesOperatorMonitoring(cfg *rest.Config) error {
 		return fmt.Errorf("failed to setup monitoring resources: %w", err)
 	}
 
-	if err := health.InstallHealthDashboard(cl); err != nil {
-		return fmt.Errorf("failed to setup the Knative Health Status Dashboard: %w", err)
+	apiExtensionClient, err := apiextension.NewForConfig(cfg)
+	if err != nil {
+		return err
 	}
+
+	if _, err = apiExtensionClient.ApiextensionsV1().CustomResourceDefinitions().Get(context.Background(), consoleclidownload.CLIDownloadCRDName, metav1.GetOptions{}); err == nil {
+		common.ConsoleInstalled.Store(true)
+		if err := health.InstallHealthDashboard(cl); err != nil {
+			return fmt.Errorf("failed to setup the Knative Health Status Dashboard: %w", err)
+		}
+	} else {
+		if !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to fetch ConsoleCLIDownload CRDs: %w", err)
+		}
+	}
+
 	return nil
 }
--- a/knative-operator/cmd/manager/main.go
+++ b/knative-operator/cmd/manager/main.go
@@ -12,6 +12,7 @@ import (
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/common"
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/controller"
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/controller/knativeserving/consoleclidownload"
+	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/controller/knativeserving/quickstart"
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/monitoring"
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/monitoring/dashboards/health"
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/webhook/knativeeventing"
@@ -177,10 +178,25 @@ func setupServerlesOperatorMonitoring(cfg *rest.Config) error {
 		return err
 	}
 
+	var installedDashboard bool
 	if _, err = apiExtensionClient.ApiextensionsV1().CustomResourceDefinitions().Get(context.Background(), consoleclidownload.CLIDownloadCRDName, metav1.GetOptions{}); err == nil {
-		common.ConsoleInstalled.Store(true)
+		consoleclidownload.SetConsoleCRDInstalled(consoleclidownload.CLIDownloadCRDName)
 		if err := health.InstallHealthDashboard(cl); err != nil {
 			return fmt.Errorf("failed to setup the Knative Health Status Dashboard: %w", err)
+		}
+		installedDashboard = true
+	} else {
+		if !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to fetch ConsoleCLIDownload CRDs: %w", err)
+		}
+	}
+
+	if _, err = apiExtensionClient.ApiextensionsV1().CustomResourceDefinitions().Get(context.Background(), quickstart.QuickStartsCRDName, metav1.GetOptions{}); err == nil {
+		consoleclidownload.SetConsoleCRDInstalled(quickstart.QuickStartsCRDName)
+		if !installedDashboard {
+			if err := health.InstallHealthDashboard(cl); err != nil {
+				return fmt.Errorf("failed to setup the Knative Health Status Dashboard: %w", err)
+			}
 		}
 	} else {
 		if !apierrors.IsNotFound(err) {
--- a/knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
+++ b/knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
@@ -10,6 +10,7 @@ import (
 	socommon "github.com/openshift-knative/serverless-operator/pkg/common"
 	consolev1 "github.com/openshift/api/console/v1"
 	routev1 "github.com/openshift/api/route/v1"
+	apiextensionv1 "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -26,13 +27,20 @@ var (
 	log               = common.Log.WithName("consoleclidownload")
 )
 
-// Apply installs kn ConsoleCLIDownload and its required resources
-func Apply(instance *operatorv1beta1.KnativeServing, apiclient client.Client, scheme *runtime.Scheme) error {
+// Apply installs kn ConsoleCLIDownload and its required resources when applicable
+func Apply(instance *operatorv1beta1.KnativeServing, apiclient client.Client, apiExtensionV1Client apiextensionv1.ApiextensionsV1Interface) error {
+	// Install the cli download route even if there is no console available to allow fetching the related binaries
 	route, err := reconcileKnConsoleCLIDownloadRoute(apiclient, instance)
 	if err != nil {
 		return err
 	}
-
+	// Skip installing console cli download resources if there are no related CRDs available eg. cluster is installed without console
+	if _, err = apiExtensionV1Client.CustomResourceDefinitions().Get(context.Background(), "consoleclidownloads.console.openshift.io", metav1.GetOptions{}); err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return fmt.Errorf("failed to fetch ConsoleCLIDownload CRDs: %w", err)
+	}
 	return reconcileKnConsoleCLIDownload(apiclient, instance, route)
 }
 
--- a/knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
+++ b/knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
@@ -10,7 +10,7 @@ import (
 	socommon "github.com/openshift-knative/serverless-operator/pkg/common"
 	consolev1 "github.com/openshift/api/console/v1"
 	routev1 "github.com/openshift/api/route/v1"
-	apiextensionv1 "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
+	"go.uber.org/atomic"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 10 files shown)


skonto:
infra
/retest

skonto:
```

File: knative-operator/pkg/controller/knativeserving/knativeserving_controller.go [hunk 1]
```diff
@@ -323,6 +331,13 @@ func (r *ReconcileKnativeServing) installQuickstarts(instance *operatorv1beta1.K
 
 // installKnConsoleCLIDownload creates CR for kn CLI download link
 func (r *ReconcileKnativeServing) installKnConsoleCLIDownload(instance *operatorv1beta1.KnativeServing) error {
+	// Skip installing console cli download if there are no related CRDs available eg. cluster is installed without console
+	if _, err := r.apiExtensionV1Client.CustomResourceDefinitions().Get(context.Background(), "consoleclidownloads.console.openshift.io", metav1.GetOptions{}); err != nil {
+		if errors.IsNotFound(err) {
+			return nil
+		}
+		return fmt.Errorf("failed to fetch ConsoleCLIDownload CRDs: %w", err)
+	}
```
dsimansk, line 340: We talked with @rhuss briefly when the original issue was raised. That maybe we should only skip creation of `ConsoleCliDownloads` resource, but keep the download route itself. Nested in the `.Apply()` function. Therefore download URL could be easily inferred with enougt insight about it. 

https://github.com/openshift-knative/serverless-operator/blob/main/knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go#L29-L36

cardil: @dsimansk Without the Console, the download link will not work.

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
```diff
@@ -26,13 +27,20 @@ var (
...
+// Apply installs kn ConsoleCLIDownload and its required resources when applicable
+func Apply(instance *operatorv1beta1.KnativeServing, apiclient client.Client, apiExtensionV1Client apiextensionv1.ApiextensionsV1Interface) error {
+	// Install the cli download route even if there is no console available to allow fetching the related binaries
 	route, err := reconcileKnConsoleCLIDownloadRoute(apiclient, instance)
 	if err != nil {
 		return err
 	}
-
+	// Skip installing console cli download resources if there are no related CRDs available eg. cluster is installed without console
+	if _, err = apiExtensionV1Client.CustomResourceDefinitions().Get(context.Background(), "consoleclidownloads.console.openshift.io", metav1.GetOptions{}); err != nil {
```
cardil, line 38: It would be a good idea to check this, and if CRD exists, don't check anymore.

The console could be not enabled during the installation. Then on running cluster, the user could enable the console. Afterwards, the console could not be removed anymore.

That's why it's pointless to check for every reconciliation.

cardil, line 38: Nit: Maybe a variable for `"consoleclidownloads.console.openshift.io"` to use here and in tests?

cardil (changes requested):

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
```diff
@@ -112,9 +118,21 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	}
 
 	gvkToResource := map[schema.GroupVersionKind]client.Object{
-		consolev1.GroupVersion.WithKind("ConsoleCLIDownload"): &consolev1.ConsoleCLIDownload{},
-		routev1.GroupVersion.WithKind("Route"):                &routev1.Route{},
+		routev1.GroupVersion.WithKind("Route"): &routev1.Route{},
+	}
+
+	// If console is installed add the ccd resource watcher, otherwise remove it avoid manager exiting due to kind not found.
+	// If we install the console later, this pod needs to be restarted as dynamically adding a watcher won't help since Serving reconciliation may not happen.
```
skonto, line 125: This needs doc. We can't add a watcher for the console crds as we don't own them. Also web console cannot be uninstalled once isntalled so this should be enough. Also we use pod restarting in other occasions like tls secrets for internal encryption so not unexpected.

knative-operator/pkg/controller/knativeserving/consoleclidownload/consoleclidownload.go
```diff
@@ -24,15 +25,20 @@ const knCLIDownload = "kn"
 var (
 	operatorNamespace = os.Getenv(common.NamespaceEnvKey)
 	log               = common.Log.WithName("consoleclidownload")
+	ConsoleInstalled  = atomic.NewBool(false)
```
skonto, line 28: not ideal to keep state but this is special case.

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
cardil: This is unfortunate, really. I think, we can't add watcher to Console CRDs because they don't exist.

I got other idea how this could be done. It should be possible to watch on [ClusterVersion](https://docs.openshift.com/container-platform/4.11/rest_api/config_apis/clusterversion-config-openshift-io-v1.html) resource (`name == 'version'`). We could wait until its `.status.capabilities.enabledCapabilities` contains the `Console` capability. If such capability just became enabled, we probably still should wait a little (simple `.Get()` and `time.Sleep(250)` loop) until `ConsoleCLIDownload` is defined by Console Operator to avoid race with it. It should take at most a couple of seconds. After that from our point of view the Console is installed, and we could go ahead to create `ConsoleCLIDownload` resources.

knative-operator/pkg/controller/knativeserving/knativeserving_controller_test.go
```diff
@@ -124,6 +125,7 @@ func TestExtraResourcesReconcile(t *testing.T) {
 			ns := &dashboardNamespace
 
 			cl := fake.NewClientBuilder().WithObjects(ks, ingress, ns, &servingNamespace).Build()
+			consoleclidownload.ConsoleInstalled.Store(true)
```
cardil, line 128: We should adjust the test to simulate no console. Then console should be enabled, after a while (by go routine). In that way, we could test proper waiting for the console to became available.

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
```diff
@@ -323,7 +343,7 @@ func (r *ReconcileKnativeServing) installQuickstarts(instance *operatorv1beta1.K
 
 // installKnConsoleCLIDownload creates CR for kn CLI download link
 func (r *ReconcileKnativeServing) installKnConsoleCLIDownload(instance *operatorv1beta1.KnativeServing) error {
-	return consoleclidownload.Apply(instance, r.client, r.scheme)
+	return consoleclidownload.Apply(instance, r.client)
```
cardil, line 346: This file does more than just installing `ConsoleCLIDownload` resources. It also installs Quickstarts, and Dashboard: https://github.com/openshift-knative/serverless-operator/blob/26a310ee58a38e70d3f7bd1d1d057891d06e287f/knative-operator/pkg/controller/knativeserving/knativeserving_controller.go#L320-L333

All of them have no sense without Console, and probably will fail as well.

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
skonto: I dont feel this is the way to go (I thought about it regarding crds as well) as the reconciliation code requires a namespaced object related to Serving not clusterversion.
We depend on something we dont own and we need to get that information at installation time or after installation from an external resource. I am wondering how the OCP operators which provide oc deal with that?

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
skonto: I didnt notice a failure anywhere but can remove them as they depend on the console.

knative-operator/pkg/controller/knativeserving/knativeserving_controller_test.go
skonto: Ok we can try that but I was expecting a e2e test for this anyway.

cardil: Yeah. An e2e test would be better, I think.

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
skonto: Ok console operator deals with the oc binaries: https://github.com/openshift/console-operator/blob/master/pkg/console/controllers/clidownloads/controller.go

skonto: > This is unfortunate, really. I think, we can't add watcher to Console CRDs because they don't exist.
We still can do `Watch(&source.Kind{Type: &apisextensionv1.CustomResourceDefinition{}}` I will take a look to see if we can implement it as a watch or an additional watcher.

knative-operator/pkg/controller/knativeserving/knativeserving_controller_test.go
skonto: I am thinking about it and it is not that easy either. Have to run without console and then apply it. We need a separate ci flow for this since we can just uninstall console.

skonto:
@cardil 
a) I changed the logic to make it automatic the whole process for all resources.
b) Updated description with paths tested on 4.12 nightly.
c) I updated the unit test. 

In general dashboards are just cms in openshift-config-managed ns so nothing is broken but for the sake of completeness I did that extra work to fix that too.

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
```diff
@@ -112,16 +123,45 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+
+	// If console is installed add the ccd resource watcher, otherwise remove it avoid manager exiting due to kind not found.
+	if common.ConsoleInstalled.Load() {
+		gvkToResource[consolev1.GroupVersion.WithKind("ConsoleCLIDownload")] = &consolev1.ConsoleCLIDownload{}
+	} else {
+		enqueueRequests := handler.MapFunc(func(obj client.Object) []reconcile.Request {
+			if obj.GetName() == consoleclidownload.CLIDownloadCRDName {
+				log.Info("Serving, processing crd request", "name", obj.GetName())
+				common.ConsoleInstalled.Store(true)
+				_ = health.InstallHealthDashboard(r.(*ReconcileKnativeServing).client)
```
skonto, line 137: We don't need this if we dont check for dashboards.

skonto:
`{Failed  === RUN   TestServerlessDowngrade/VerifyContinualTests/KafkaSinkSourceContinualTest` fails often.
/restart

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
cardil: How about Quickstarts? I think they are a CRD https://docs.openshift.com/container-platform/4.11/rest_api/console_apis/consolequickstart-console-openshift-io-v1.html

skonto: Yes there is one I saw that when reconciled the rest but they are both installed with the console so no need to discover them explicitly.

cardil: The CRDs are deployed at random by Console operator. Even if they are deployed sequentially, we shouldn't rely on that behavior. Now you are waiting for `ConsoleCLIDownload` to be available. There is a chance the `ConsoleQuickstart` would not be deployed yet, although a small one.

Instead of waiting for `ConsoleCLIDownload` to be available, we could wait for [`ClusterOperator`](https://docs.openshift.com/container-platform/4.11/rest_api/config_apis/clusteroperator-config-openshift-io-v1.html) named `console` to be available and reported `Progressing==False && Degraded==False`. That would make sure the console is ready.

skonto: > Instead of waiting for ConsoleCLIDownload to be available, we could wait for [ClusterOperator](https://docs.openshift.com/container-platform/4.11/rest_api/config_apis/clusteroperator-config-openshift-io-v1.html) named console to be available and reported Progressing==False && Degraded==False.

I am not even sure that this makes sure that crds are there. Does it explicitly check again for crds status after they are created? Anyway I prefer the crd check instead.

skonto: I checked clusteroperator too so it does get to `Progressing==False && Degraded==False` after crds are installed but will try the crds approach first as closer to what is being installed.

skonto:
@cardil Updated the PR to cover all crds, tested with `launch 4.12.0-0.nightly-2022-10-05-053337 aws,no-capabilities` for scenarios a), b) of the PR description and `launch 4.12.0-0.nightly-2022-10-05-053337 gcp` for the third scenario.
Unlike the outcome in https://issues.redhat.com/browse/SRVKS-963 I saw no failures given that changes in the patch https://gist.github.com/skonto/6a3c9a7db386a9f3b2ca8a0ab115d58b are applied.

skonto:
@dsimansk gentle ping, pls review.

knative-operator/cmd/manager/main.go
```diff
@@ -166,8 +173,36 @@ func setupServerlesOperatorMonitoring(cfg *rest.Config) error {
...
+	var installedDashboard bool
+	if _, err = apiExtensionClient.ApiextensionsV1().CustomResourceDefinitions().Get(context.Background(), consoleclidownload.CLIDownloadCRDName, metav1.GetOptions{}); err == nil {
+		consoleclidownload.SetConsoleCRDInstalled(consoleclidownload.CLIDownloadCRDName)
+		if err := health.InstallHealthDashboard(cl); err != nil {
+			return fmt.Errorf("failed to setup the Knative Health Status Dashboard: %w", err)
+		}
+		installedDashboard = true
+	} else {
+		if !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to fetch ConsoleCLIDownload CRDs: %w", err)
```
cardil, line 190: We shouldn't return an error, if CRD does not exist. That is valid, and expected case. Instead, we should log it.

Otherwise, we will see false errors in logs.

knative-operator/cmd/manager/main.go
```diff
@@ -166,8 +173,36 @@ func setupServerlesOperatorMonitoring(cfg *rest.Config) error {
...
+	if _, err = apiExtensionClient.ApiextensionsV1().CustomResourceDefinitions().Get(context.Background(), quickstart.QuickStartsCRDName, metav1.GetOptions{}); err == nil {
+		consoleclidownload.SetConsoleCRDInstalled(quickstart.QuickStartsCRDName)
+		if !installedDashboard {
+			if err := health.InstallHealthDashboard(cl); err != nil {
+				return fmt.Errorf("failed to setup the Knative Health Status Dashboard: %w", err)
+			}
+		}
+	} else {
+		if !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to fetch ConsoleCLIDownload CRDs: %w", err)
```
cardil, line 203: Same as with CLI download CRD. We shouldn't return an error, if CRD does not exist. That is valid, and expected case. Instead, we should log it.

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
cardil: Okay. Let's go with CRDs way. However, I think it may simplify the code if we instead rely on `ClusterOperator` status.

cardil (changes requested):
Looks good. Only some small nits.

skonto:
> Okay. Let's go with CRDs way. However, I think it may simplify the code if we instead rely on ClusterOperator status.

Yeah I thought about this too. We can merge this and iterate if I dont find the cycles to do it until the cut next week.

knative-operator/cmd/manager/main.go
skonto: right ok will fix thanks.

skonto:
@cardil I will investigate clusterversion too in another PR because I see that other stuff are going to be optional like monitoring and I am not sure if we can have crds to detect everything or if the code will become too complex.

I will /hold for now give me a day or too and let's see.

cardil:
BTW. I'm having this error when trying to install this on 4.12 cluster with no capabilities:

```
Error from server (NotFound): error when creating "STDIN": namespaces "openshift-marketplace" not found
ERROR   14:25:19.497 🚨 Error (code: 1) occurred at hack/lib/catalogsource.bash:75, with command: oc apply -n "$OLM_NAMESPACE" -f -
ERROR   14:25:19.500 Stack trace:
ERROR   14:25:19.501  1: hack/lib/catalogsource.bash:9 install_catalogsource(...)
ERROR   14:25:19.503  2: hack/install.sh:22 ensure_catalogsource_installed(...)
```

Maybe you could add appropriate bash code to deploy `openshift-marketplace` namespace, if it's missing?

PR #1758, opened 2022-10-06
Use queue-sidecar-image instead of queueSidecarImage.

nak3:
As per title, this patch replaces queueSidecarImage with queue-sidecar-image.

It causes an error in midstream because upstream changed the default as https://github.com/knative/serving/commit/79a45887a7d56b305a714f21a36e736f1b9f5ebe


nak3:
/cc @skonto 

https://github.com/openshift/knative-serving/pull/1269 cannot verify this patch without merging.

PR #1765, opened 2022-10-11
Adding most simple logging enhancement

matzew:

Fixes JIRA SRVKE-1287

## Proposed Changes

- adding `spec.logging.level` for `KnativeKafka` and ensuring only reasonable values via the schema validation

### DIFF
--- a/knative-operator/pkg/apis/operator/v1alpha1/knativekafka_types.go
+++ b/knative-operator/pkg/apis/operator/v1alpha1/knativekafka_types.go
@@ -8,6 +8,9 @@ import (
 
 // KnativeKafkaSpec defines the desired state of KnativeKafka
 // +k8s:openapi-gen=true
+
+const DEFAULT_LOG_LEVEL = "INFO"
+
 type KnativeKafkaSpec struct {
 	// Allows configuration for KafkaBroker installation
 	// +optional
@@ -33,7 +36,7 @@ type KnativeKafkaSpec struct {
 
 	// Set logging configuration of the data plane.
 	// +optional
-	Logging Logging `json:"logging,omitempty"`
+	Logging *Logging `json:"logging,omitempty"`
 }
 
 // KnativeKafkaStatus defines the observed state of KnativeKafka
--- a/knative-operator/pkg/apis/operator/v1alpha1/knativekafka_types.go
+++ b/knative-operator/pkg/apis/operator/v1alpha1/knativekafka_types.go
@@ -9,7 +9,7 @@ import (
 // KnativeKafkaSpec defines the desired state of KnativeKafka
 // +k8s:openapi-gen=true
 
-const DEFAULT_LOG_LEVEL = "INFO"
+const DefaultLogLevel = "INFO"
 
 type KnativeKafkaSpec struct {
 	// Allows configuration for KafkaBroker installation
--- a/knative-operator/pkg/apis/operator/v1alpha1/knativekafka_types.go
+++ b/knative-operator/pkg/apis/operator/v1alpha1/knativekafka_types.go
@@ -30,6 +30,10 @@ type KnativeKafkaSpec struct {
 	// HighAvailability allows specification of HA control plane.
 	// +optional
 	HighAvailability *base.HighAvailability `json:"high-availability,omitempty"`
+
+	// Set logging configuration of the data plane.
+	// +optional
+	Logging Logging `json:"logging,omitempty"`
 }
 
 // KnativeKafkaStatus defines the observed state of KnativeKafka
@@ -122,6 +126,13 @@ type Channel struct {
 	// auth configuration.
 	// +optional
 	AuthSecretName string `json:"authSecretName"`
+}
+
+type Logging struct {
+	// Defines the log level. Allowed values are 'TRACE', 'DEBUG', 'INFO', 'WARN' and 'ERROR'.
+	// The default value is 'INFO'.
+	// +optional
+	Level string `json:"level,omitempty"`
 }
 
 func init() {
--- a/knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
+++ b/knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
@@ -613,6 +613,14 @@ func configureEventingKafka(spec serverlessoperatorv1alpha1.KnativeKafkaSpec) mf
 			return scheme.Scheme.Convert(deployment, u, nil)
 		}
 
+		if u.GetKind() == "ConfigMap" && u.GetName() == "kafka-config-logging" {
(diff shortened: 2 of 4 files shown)


knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -602,6 +602,14 @@ func configureEventingKafka(spec serverlessoperatorv1alpha1.KnativeKafkaSpec) mf
 			return scheme.Scheme.Convert(deployment, u, nil)
 		}
 
+		if u.GetKind() == "ConfigMap" && u.GetName() == "kafka-config-logging" {
+			log.Info("Found ConfigMap kafka-config-logging, updating it with values from spec")
+			xmlTemplate := "    <configuration>\n      <appender name=\"jsonConsoleAppender\" class=\"ch.qos.logback.core.ConsoleAppender\">\n        <encoder class=\"net.logstash.logback.encoder.LogstashEncoder\"/>\n      </appender>\n      <root level=\"" + spec.Logging.Level + "\">\n        <appender-ref ref=\"jsonConsoleAppender\"/>\n      </root>\n    </configuration>"
```
matzew, line 607: On purpose I am avoiding proper xml usage, to keep this _simple_ instead of using `encoding/xml` and friends...

olm-catalog/serverless-operator/manifests/operator_v1alpha1_knativekafka_crd.yaml
```diff
@@ -124,6 +124,20 @@ spec:
                     minimum: 1
                     type: integer
                 type: object
+              logging:
+                description: some loggings
```
matzew, line 128: that needs to change :smile:

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
aliok: Nit picking:

We can actually have a template file and use https://pkg.go.dev/embed.

I am ok with this too though.

But at least use raw string literals?

aliok:
It is a very simple change, but still, any tests would be appreciated

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
matzew: I changed to raw strings. Let's not use the template engine.

I also extracted the "xml rendering" fu to some (testable) function, to make the actual code of the transformers less verbose.

thanks @aliok

knative-operator/pkg/apis/operator/v1alpha1/knativekafka_types.go
```diff
@@ -30,6 +30,10 @@ type KnativeKafkaSpec struct {
 	// HighAvailability allows specification of HA control plane.
 	// +optional
 	HighAvailability *base.HighAvailability `json:"high-availability,omitempty"`
+
+	// Set logging configuration of the data plane.
+	// +optional
+	Logging Logging `json:"logging,omitempty"`
```
pierDipi, line 36: should this be a pointer since it's optional?
```suggestion
	Logging *Logging `json:"logging,omitempty"`
```

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -613,6 +613,13 @@ func configureEventingKafka(spec serverlessoperatorv1alpha1.KnativeKafkaSpec) mf
 			return scheme.Scheme.Convert(deployment, u, nil)
 		}
 
+		if u.GetKind() == "ConfigMap" && u.GetName() == "kafka-config-logging" {
+			log.Info("Found ConfigMap kafka-config-logging, updating it with values from spec")
+			if err := unstructured.SetNestedField(u.Object, renderLoggingConfigXML(spec.Logging.Level), "data", "config.xml"); err != nil {
```
pierDipi, line 618: This is potentially rendering an empty string

matzew: Doing some default, like for ha, in here: https://github.com/openshift-knative/serverless-operator/pull/1765/files#diff-3b74c6630d5f7a1fc62c6f999d5c1bc3572e5b8603ad46edc46c51dce4eab0adR296

aliok: @matzew Link you posted is recursive :D

I guess you mention https://github.com/openshift-knative/serverless-operator/pull/1765/files#diff-8a3b1481924d4139a463a9c68ccdaa1056686b4209ab0be3ef82b488127b8f97R127-R140

I assumed the specification in the CRD yaml that does the validation and defaulting.

Are there any cases where the created custom object escapes Kube api server validation @pierDipi ? Or, don't those validations and defaulting specified in OpenAPI definitions are not working the way we think they would?

olm-catalog/serverless-operator/manifests/operator_v1alpha1_knativekafka_crd.yaml
```diff
@@ -124,6 +124,20 @@ spec:
                     minimum: 1
                     type: integer
                 type: object
+              logging:
+                description: Set logging configuration of the data plane
```
pierDipi, line 128:
```suggestion
                description: Set logging configuration of the data plane (receivers and dispatchers). It is not recommended to use DEBUG or TRACE in production.
```

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
pierDipi: > Are there any cases where the created custom object escapes Kube api server validation @pierDipi ?

That's exactly my original question, so we should try and see what happens, for reference:
- https://github.com/kubernetes/apiextensions-apiserver/issues/28
- https://github.com/kubernetes/kubernetes/issues/62325#issuecomment-444767470

olm-catalog/serverless-operator/manifests/operator_v1alpha1_knativekafka_crd.yaml
pierDipi: Another option:
```suggestion
                description: Set logging configuration of the data plane (receivers and dispatchers). It is not recommended to use DEBUG or TRACE in production since it's too verbose and degrades performance.
```

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
matzew: We default it to `INFO` on the serverside,see the code. And if I change the value to some bogus value, the openapi validates does reject the change.

IMO this is all fine, as is now

matzew:
```
KafkaSinkSourceContinualTest
```

/test 4.11-upgrade-tests-aws-ocp-411

/unhold

PR #1787, opened 2022-11-04
[SRVCOM-2118] Install console resources only if the related clusteroperator is available

skonto:
## Proposed Changes

- Aims to replace #1757
- Uses the clusteroperator resource instead of crds so we can extend easily in the future for other capabilities like: monitoring and ingress operators which will become optional. As a side note clusterversion is also not that useful since it gets no update [about when the capability is ready](https://gist.github.com/skonto/3ebf91dc178db4b2aa0c856f978f4e64).  
The clusteroperator object has the following form: 
```
        {
            "apiVersion": "config.openshift.io/v1",
            "kind": "ClusterOperator",
            "metadata": {
                "annotations": {
                    "capability.openshift.io/name": "Console",
                    "include.release.openshift.io/ibm-cloud-managed": "true",
                    "include.release.openshift.io/self-managed-high-availability": "true",
                    "include.release.openshift.io/single-node-developer": "true"
                },
                "creationTimestamp": "2022-11-03T21:02:23Z",
                "generation": 1,
                "name": "console",
                "ownerReferences": [
                    {
                        "apiVersion": "config.openshift.io/v1",
                        "kind": "ClusterVersion",
                        "name": "version",
                        "uid": "dd22e3e6-e668-4608-8b67-7dcf9590a649"
                    }
                ],
                "resourceVersion": "43097",
                "uid": "8a989891-890e-42bc-a247-e15fc7c1966e"
            },
            "spec": {},
            "status": {
                "conditions": [
                    {
                        "lastTransitionTime": "2022-11-03T21:02:46Z",
                        "message": "All is well",
                        "reason": "AsExpected",
                        "status": "False",
                        "type": "Degraded"
                    },
                    {
                        "lastTransitionTime": "2022-11-03T21:03:08Z",
                        "message": "All is well",
                        "reason": "AsExpected",
                        "status": "False",
                        "type": "Progressing"
                    },
                    {
                        "lastTransitionTime": "2022-11-03T21:03:07Z",
                        "message": "All is well",
                        "reason": "AsExpected",
                        "status": "True",
                        "type": "Available"
                    },
                    {
                        "lastTransitionTime": "2022-11-03T21:02:46Z",
                        "message": "All is well",
                        "reason": "AsExpected",
                        "status": "True",
                        "type": "Upgradeable"
                    }
                ],
...
```
- Tested in all scenarios as described in #1757 with [4.12.0-ec.4](https://amd64.ocp.releases.ci.openshift.org/releasestream/4-dev-preview/release/4.12.0-ec.4)
Paths tested:

a) no console -> install S-O -> Install Console -> Install Serving/Eventing
b) no console -> install S-O -> install Serving/Eventing -> Install Console
c) Install Console -> install S-O -> Install Serving/Eventing
- More about post-install configuration [here](https://docs.openshift.com/container-platform/4.11/post_installation_configuration/cluster-capabilities.html#setting_additional_enabled_capabilities_cluster-capabilities). Capabilities cannot be uninstalled and this PR is based on this assumption.

### DIFF
--- a/knative-operator/pkg/controller/knativeserving/consoleutil/util.go
+++ b/knative-operator/pkg/controller/knativeserving/consoleutil/util.go
@@ -1,0 +1,32 @@
+package consoleutil
+
+import (
+	configv1 "github.com/openshift/api/config/v1"
+	"go.uber.org/atomic"
+)
+
+const ConsoleClusterOperatorName = "console"
+
+var ConsoleInstalled = atomic.NewBool(false)
+
+// SetConsoleToInstalledStatus updates to true the detected status of the console capability.
+// Once a capability is installed it cannot be uninstalled.
+func SetConsoleToInstalledStatus() {
+	ConsoleInstalled.Store(true)
+}
+
+// IsConsoleInstalled checks the detected status of the console capability.
+func IsConsoleInstalled() bool {
+	return ConsoleInstalled.Load()
+}
+
+// IsClusterOperatorAvailable iterates over conditions of the related resource
+// and checks if it is available.
+func IsClusterOperatorAvailable(status configv1.ClusterOperatorStatus) bool {
+	for _, cond := range status.Conditions {
+		if cond.Type == configv1.OperatorAvailable && cond.Status == configv1.ConditionTrue {
+			return true
+		}
+	}
+	return false
+}
--- a/olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
+++ b/olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
@@ -343,7 +343,9 @@ spec:
                 - clusteroperators
                 - clusteroperators/status
               verbs:
-                - "*"
+                - "get"
+                - "watch"
+                - "list"
             - apiGroups:
                 - route.openshift.io
               resources:
--- a/olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
+++ b/olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
@@ -338,6 +338,13 @@ spec:
               verbs:
                 - "*"
             - apiGroups:
+                - config.openshift.io
+              resources:
+                - clusteroperators
+                - clusteroperators/status
+              verbs:
+                - "*"
+            - apiGroups:
                 - route.openshift.io
               resources:
                 - routes
--- a/go.mod
+++ b/go.mod
@@ -133,7 +133,7 @@ require (
 	github.com/xdg-go/scram v1.1.1 // indirect
 	github.com/xdg-go/stringprep v1.0.3 // indirect
 	go.opencensus.io v0.23.0 // indirect
-	go.uber.org/atomic v1.9.0 // indirect
+	go.uber.org/atomic v1.9.0
 	go.uber.org/automaxprocs v1.4.0 // indirect
 	go.uber.org/multierr v1.7.0 // indirect
 	golang.org/x/crypto v0.0.0-20220214200702-86341886e292 // indirect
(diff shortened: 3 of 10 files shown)


knative-operator/pkg/controller/knativeserving/consoleutil/util.go
```diff
@@ -0,0 +1,32 @@
+package consoleutil
```
skonto, line 1: When we add new optional components we can rename this to `capabilities` and move it to some other structure.

skonto:
Operator tests timed-out.
```
13:53:20.861 DEBUG:   [[ $(oc get pods -n knative-eventing --field-selector=status.phase!=Succeeded -o jsonpath='{.items}') != '[]' ]] : Waiting until non-zero (max 700 sec.)
......................................
```

skonto:
```

=== Failed
=== FAIL: test/upgrade TestServerlessUpgrade/VerifyContinualTests/ProbeTest (0.06s)
revision:  
Warning  InternalError  28m   revision-controller  public cert in the secret is not ready yet

net-kourier:
{"severity":"WARNING","timestamp":"2022-11-05T18:07:48.175616363Z","logger":"net-kourier-controller","caller":"ingress/controller.go:123","message":"Error pushing snapshot to gateway: code: 13 message route: unknown weighted cluster 'serving-tests/pizzaplanet-upgrade-service-00002'"}

```

knative-operator/pkg/controller/knativeserving/consoleutil/util.go
cardil: +1 for capabilities package (might be a follow up)

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
```diff
@@ -337,6 +337,13 @@ spec:
                 - consoleclidownloads
               verbs:
                 - "*"
+            - apiGroups:
+                - config.openshift.io
+              resources:
+                - clusteroperators
+                - clusteroperators/status
+              verbs:
+                - "*"
```
cardil, line 346: We need only GET/WATCH, right? We shouldn't declare we change those.

cardil (changes requested):
I have given only a glance look. It looks good. Let someone else give a more thorough review... 

Some minor issue about permissions.

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
skonto: We can make it more stricter I guess the only reason I didnt do it earlier is because we need to review all those permissions  in the csv (we have many others with `*`) but let me check if get, watch is enough.

skonto: Let's do the capabilities in a follow up no rush for 1.26.

cardil: We can use `*` for our types like `KnativeServing`, or other things that we actually create like `ConsoleCLIDownload`. Here, we just read, so we should limit the permissions.

But, this isn't a pressing issue. We can fix those permissions in a follow-up PR.

matzew: > review all those permissions

Fine w/ doing this at a later stage...

skonto: You need also "list' let me check first and btw unfortunately we have other permissions open for non Knative stuff like service monitors etc;)

skonto:
@matzew @cardil ready.

skonto:
Let's merge this is ready.

cardil (approved):
/lgtm

cardil:
Closes #1757

skonto:
Teardown is getting stuck. :thinking: 

/test 4.11-operator-e2e-aws-ocp-411

skonto:
Merging manually to move with the cut.

PR #1789, opened 2022-11-04
Rebase target branch before running validate

pierDipi:
This removes the need to rebase PRs that aren't rebased 
on top of the latest target branch commit.

Copied from https://github.com/knative/actions/blob/2da2b8c17f8f1d4317e1d215e0e80bedf44274e1/.github/workflows/go-test.yaml#L40-L52

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -39,6 +39,21 @@ jobs:
         with:
           path: ./src/github.com/${{ github.repository }}
 
+      - name: Merge upstream
+        if: github.event_name == 'pull_request'
+        working-directory: ./src/github.com/${{ github.repository }}
+        run: |
+          if ! git config user.name > /dev/null; then
+            git config user.name "John Doe"
+          fi
+          if ! git config user.email > /dev/null; then
+            git config user.email "johndoe@localhost"
+          fi
+          git remote add upstream https://github.com/${{ github.repository }}.git
+          git fetch upstream ${{ github.base_ref }}
+          git pull --no-rebase --no-commit upstream ${{ github.base_ref }}
+        shell: bash
+
       - name: Regenerate all generated files
         working-directory: ./src/github.com/${{ github.repository }}
         run: make generated-files
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -51,7 +51,7 @@ jobs:
             git config user.email "johndoe@localhost"
           fi
           git remote add upstream https://github.com/${{ github.repository }}.git
-          git pull upstream ${{ github.base_ref }}
+          git pull --no-rebase upstream ${{ github.base_ref }}
         shell: bash
 
       - name: Regenerate all generated files
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -51,7 +51,6 @@ jobs:
             git config user.email "johndoe@localhost"
           fi
           git remote add upstream https://github.com/${{ github.repository }}.git
-          git fetch upstream ${{ github.base_ref }}
           git pull upstream ${{ github.base_ref }}
         shell: bash
 
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -52,7 +52,7 @@ jobs:
           fi
           git remote add upstream https://github.com/${{ github.repository }}.git
           git fetch upstream ${{ github.base_ref }}
-          git pull --no-rebase --no-commit upstream ${{ github.base_ref }}
+          git pull upstream ${{ github.base_ref }}
         shell: bash
 
       - name: Regenerate all generated files
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -35,9 +35,10 @@ jobs:
           sudo mv /usr/bin/yq /usr/local/bin/yq
 
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           path: ./src/github.com/${{ github.repository }}
+          fetch-depth: 0
 
       - name: Merge upstream
         if: github.event_name == 'pull_request'


pierDipi:
/cc @skonto @matzew

As per our discussion on Slack.

.github/workflows/validate.yaml
```diff
@@ -39,6 +39,22 @@ jobs:
...
+        run: |
+          if ! git config user.name > /dev/null; then
+            git config user.name "John Doe"
+          fi
+          if ! git config user.email > /dev/null; then
+            git config user.email "johndoe@localhost"
+          fi
+          git remote add upstream https://github.com/${{ github.repository }}.git
+          git fetch upstream ${{ github.base_ref }}
+          git pull --no-rebase --no-commit upstream ${{ github.base_ref }}
```
skonto, line 55: This is equivalent to git merge right?

pierDipi: Yes

skonto: Where do we do the push?

.github/workflows/validate.yaml
```diff
@@ -39,6 +39,22 @@ jobs:
         with:
           path: ./src/github.com/${{ github.repository }}
 
+      - name: Merge upstream
+        if: github.event_name == 'pull_request'
+        with:
+          path: ./src/github.com/${{ github.repository }}
+        run: |
+          if ! git config user.name > /dev/null; then
+            git config user.name "John Doe"
```
skonto, line 48: Could use something like: "GitHub Actions"

.github/workflows/validate.yaml
```diff
@@ -39,6 +39,22 @@ jobs:
...
+          fi
+          if ! git config user.email > /dev/null; then
+            git config user.email "johndoe@localhost"
```
skonto, line 51: Could have something like: `xxx-github-actions[bot]@users.noreply.github.com`

.github/workflows/validate.yaml
pierDipi: We don't, this is basically faking the fact that validates run on target branch + PR changes because without it a GH action will run with the content of your branch which might not be based on the latest commit.

.github/workflows/validate.yaml
pierDipi: This is not a commit we will see in the history, so it doesn't really matter

.github/workflows/validate.yaml
skonto: Correct ok it is just the validation ;)

.github/workflows/validate.yaml
skonto: Yeah just a suggestion to make it look more related anyway.

skonto:
Seems validation fails.

pierDipi:
I will check it next week.

/hold

pierDipi:
it works locally :tm:

pierDipi:
Ok, this works now, failing is related to https://github.com/openshift-knative/serverless-operator/pull/1817

.github/workflows/validate.yaml
```diff
@@ -35,9 +35,25 @@ jobs:
           sudo mv /usr/bin/yq /usr/local/bin/yq
 
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           path: ./src/github.com/${{ github.repository }}
+          fetch-depth: 0
```
skonto, line 41: This is supposed to fetch history for all branches and tags (btw we exclude tags from fetching), is this required here?

pierDipi: Yes, it was failing with `fatal: refusing to merge unrelated histories` without it, see https://github.com/openshift-knative/serverless-operator/actions/runs/3496276603/jobs/5853997426

.github/workflows/validate.yaml
```diff
@@ -35,9 +35,25 @@ jobs:
...
+        working-directory: ./src/github.com/${{ github.repository }}
+        run: |
+          if ! git config user.name > /dev/null; then
+            git config user.name "John Doe"
+          fi
+          if ! git config user.email > /dev/null; then
+            git config user.email "johndoe@localhost"
+          fi
+          git remote add upstream https://github.com/${{ github.repository }}.git
+          git fetch upstream ${{ github.base_ref }}
```
mgencur, line 54: I'm curious if "git fetch" is required here? From `git help pull` command:
```
More precisely, git pull runs git fetch with the given parameters and then depending on configuration options
       or command line flags, will call either git rebase or git merge to reconcile diverging branches.
```

mgencur: The original code from which you copied had `--no-rebase --no-commit`. I suppose especially the first option should be specified (or `--rebase`) when the branches diverged. Again, from git help:
```
If the current branch and
       the remote have diverged, the user needs to specify how to reconcile the divergent branches with --rebase or
       --no-rebase
```

pierDipi: No, it shouldn't be needed

pierDipi: Removed https://github.com/openshift-knative/serverless-operator/pull/1789/commits/66654b8a1afff56405f7b2d70f2af28546e9742e

.github/workflows/validate.yaml
```diff
@@ -35,9 +35,24 @@ jobs:
...
+        working-directory: ./src/github.com/${{ github.repository }}
+        run: |
+          if ! git config user.name > /dev/null; then
+            git config user.name "John Doe"
+          fi
+          if ! git config user.email > /dev/null; then
+            git config user.email "johndoe@localhost"
+          fi
+          git remote add upstream https://github.com/${{ github.repository }}.git
+          git pull upstream ${{ github.base_ref }}
```
pierDipi, line 54:
```suggestion
          git pull upstream --no-rebase ${{ github.base_ref }}
```

.github/workflows/validate.yaml
pierDipi: Thanks! Added --no-rebase https://github.com/openshift-knative/serverless-operator/pull/1789/commits/83966da1c26466c499ebf427e43597c7ed3023a2

pierDipi: The --no-commit is not possible because we use git to detect validate codegen

pierDipi:
/unhold

Let me know if it works as intended

PR #1802, opened 2022-11-09
Bump deps to 1.6

skonto:
## Proposed Changes

- bumps deps for all components to 1.6
- keeps compat for 1.17


skonto:
will open another

PR #1813, opened 2022-11-16
[SRVCOM-2170] Handle nil logging level on deletion

pierDipi:
As per title, spec.logging can be nil if wasn't reconciled yet by the "new version" of the controller


knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -688,7 +688,11 @@ func contains(array []string, name string) bool {
 	return false
 }
 
-func renderLoggingConfigXML(loglevel string) string {
+func renderLoggingConfigXML(logging *serverlessoperatorv1alpha1.Logging) string {
+	loglevel := "INFO"
+	if logging != nil {
```
matzew, line 693: :see_no_evil:

matzew (approved):
/lgtm
/approve

PR #1816, opened 2022-11-17
Disable KUBERNETES_MIN_VERSION by setting lower version

nak3:
Please refer to the e-mail sub: `Stop using upstream's KUBERNETES_MIN_VERSION validation`.

TODO:
- [x] unit test.


nak3:
/cc @skonto @pierDipi @matzew 

Could you please take a look?

openshift-knative-operator/pkg/common/min_version.go
```diff
@@ -0,0 +1,64 @@
...
+			}
+			podSpec = &dep.Spec.Template.Spec
+			convert = func(spec *corev1.PodSpec) error {
+				dep.Spec.Template.Spec = *podSpec
+				return scheme.Scheme.Convert(dep, u, nil)
+			}
+		}
+
+		if podSpec == nil {
+			// Do not need to inject the env valiable.
```
pierDipi, line 52:
```suggestion
			// Do not need to inject the env variable.
```

pierDipi (commented):
Can we include Eventing Kafka too?

here's the place to add it https://github.com/openshift-knative/serverless-operator/blob/eae7141412b14d3064d5e073a349080eac124370/knative-operator/pkg/controller/knativekafka/knativekafka_controller.go#L338

vendor/knative.dev/pkg/version/version.go
```diff
@@ -33,7 +33,7 @@ const (
 	// NOTE: If you are changing this line, please also update the minimum kubernetes
 	// version listed here:
 	// https://github.com/knative/docs/blob/mkdocs/docs/snippets/prerequisites.md
-	defaultMinimumVersion = "v1.19.0"
+	defaultMinimumVersion = "v1.22.0"
```
skonto, line 36: Eventing is setting this to 1.21.0?

matzew: 4.8 is 1.21, that's why

skonto: Ok then counted it wrongly so should this be 1.21.0 as well?

pierDipi: No, this is now whatever is set in code at the current version, no more patching since `getMinimumVersion` will use the injected env variable

pierDipi:
Thanks for doing this @nak3 !

vendor/knative.dev/pkg/version/version.go
skonto: Yes but for consistency reasons it would be nice to be the aligned. Even if it is not used it is an indeication of the version supported.

skonto: Yeah maybe a note in project.yaml that this file is ginored.

pierDipi: This will align itself once we remove patching from other midstream repositories, I don't think we should keep having a patch for alignment in this repo because that's defeat the purpose of the PR itself

pierDipi: +1 on adding a comment if we find that useful

nak3: When someone finds the `version.go`, I'm sure I will check if what `KUBERNETES_MIN_VERSION` is set. 

https://github.com/openshift-knative/serverless-operator/blob/5cb524816e9084acc36457df5e09c7f3c84bb5a2/vendor/knative.dev/pkg/version/version.go#L29-L36

nak3: Anyway, I added the comment as I don't want to keep this PR open...

pierDipi (approved):
/lgtm

PR #1823, opened 2022-11-21
[release-1.26] Do not use wathola-forwarder in Eventing upgrade tests

mgencur:
See [Slack discussion](https://coreos.slack.com/archives/CEXRYS5QC/p1668777165819389?thread_ts=1668499127.372969&cid=CEXRYS5QC).

## Proposed Changes

- Do not use wathola-forwarder in Eventing upgrade tests and use only Eventing components. The forwarder is a Knative Service and Knative Serving shows flakiness in its own tests during upgrades so it probably affects Eventing tests as well if the forwarder component is included.
- Exclude us-east-1e from scale-up, see the comment in code for more details.

### DIFF
--- a/test/lib.bash
+++ b/test/lib.bash
@@ -232,7 +232,7 @@ EOF
   # Test configuration. See https://github.com/knative/eventing/tree/main/test/upgrade#probe-test-configuration
   # TODO(ksuszyns): remove EVENTING_UPGRADE_TESTS_SERVING_SCALETOZERO when knative/operator#297 is fixed.
   export EVENTING_UPGRADE_TESTS_SERVING_SCALETOZERO=false
-  export EVENTING_UPGRADE_TESTS_SERVING_USE=true
+  export EVENTING_UPGRADE_TESTS_SERVING_USE=false
   export EVENTING_UPGRADE_TESTS_CONFIGMOUNTPOINT=/.config/wathola
   export GATEWAY_OVERRIDE="kourier"
   export GATEWAY_NAMESPACE_OVERRIDE="${INGRESS_NAMESPACE}"
--- a/hack/lib/scaleup.bash
+++ b/hack/lib/scaleup.bash
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 function scale_up_workers {
-  local current_total az_total replicas idx
+  local current_total az_total replicas idx ignored_zone
   if [[ "${SCALE_UP}" -lt "0" ]]; then
     logger.info 'Skipping scaling up, because SCALE_UP is negative.'
     return 0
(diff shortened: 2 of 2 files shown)


test/lib.bash
```diff
@@ -232,7 +232,7 @@ EOF
   # Test configuration. See https://github.com/knative/eventing/tree/main/test/upgrade#probe-test-configuration
   # TODO(ksuszyns): remove EVENTING_UPGRADE_TESTS_SERVING_SCALETOZERO when knative/operator#297 is fixed.
   export EVENTING_UPGRADE_TESTS_SERVING_SCALETOZERO=false
-  export EVENTING_UPGRADE_TESTS_SERVING_USE=true
+  export EVENTING_UPGRADE_TESTS_SERVING_USE=false
```
cardil, line 235: Add a comment, explaining the change. With a `TODO` with an issue link to be fixed.

mgencur: Is it necessary? Even if I mention some Serving issue we're not going to re-enable Serving because in the future there might be other issues like this. Having Serving in the path makes the test more complex and hard to debug. Issues in Serving should be spotted by different tests. Our goal is to test Eventing here.
So, mentioning some issue here doesn't make sense here. And the change on this line can be found in git history and related github PR.

cardil: I see it differently. We test Serving in serving, together with upgrades. Same for Eventing. This cross-cutting integration was written to cover that exact case - issues that may arise when using both. That's why we should have a serving case. The same is true for above case `SERVING_SCALETOZERO`.

mgencur: I think this combination of components was created with good intent. However, the recent and past problems
showed that this scenario only obfuscates the actual errors. The tests appear like they're loosing events,
which is true but the problem is not in Eventing/Kafka even though the failed tests show the problem there.
We have separate tests in Serving which we run and which showed the same problem. So, having Serving
in this Eventing upgrade scenario only made me spend several days debug Eventing/Kafka and try to reproduce
the problem, only to show it's not there. In the future, whenever we see lost events, we will need to disable
Serving again, just to be sure it's not there.

From the point of Eventing it's nothing else than sending an Event to duckv1.KReference whether it's Knative Service or regular Kube Service. There's no difference from the Sender side.
As a result, we don't test anything specific for Eventing+Serving integration and I think this way we can't
even spot any issues between these two sub system, simply because we use a generalized approach. We can only spot issues either in Eventing or - in Serving itself but for that purpose we already have other tests (more simple and more focused). 

Same applies for SERVING_SCALETOZERO. The upgrade framework first deploys the KService, then waits for it
to scale to zero and then starts the continual tests - i.e. starts sending the requests. The KService
is scaled up way before the actual upgrade and it is never scaled to zero during the upgrade because we're
continuously sending requests/events. So, effectively this is the same as a regular E2E that scales to zero and
then back to one (or more). Upgrades don't play any role here because we don't test scale-from-zero during the upgrade. There's a different test in Serving for scaling to zero, upgrading, and then scaling to one: ServicePreUpgradeAndScaleToZeroTest+ServicePostUpgradeFromScaleToZeroTest

The general integration between Eventing and Serving (events sent to Knative Services) is well tested in other E2E tests,
for example: https://github.com/openshift-knative/serverless-operator/tree/main/test/extensione2e/kafka

Last note, debugging Eventing upgrade tests is hard and bringing Serving to the scenario makes it even harder.

cardil: Lets think about it in https://issues.redhat.com/browse/SRVCOM-2176

cardil (approved):
/lgtm

PR #1826, opened 2022-11-23
:bomb: Reproduction of go run failure on OCP CI

cardil:
/hold

This is just a reproduction of `go run package@version` failure on OpenShift CI


skonto:
@cardil I suspect we should close this one right?

PR #1831, opened 2022-11-23
[WIP] [1.25] Scripting changes to be able to use EKB with a branch

aliok:
Anotner test similar to https://github.com/openshift-knative/serverless-operator/pull/1827

PR #1792, opened 2022-11-08
Bump EKB to 1.6

pierDipi:
As per title

For after the cut.

PR #1832, opened 2022-11-24
[WIP] [1.25] Scripting changes to be able to use EKB with a commit

aliok:
Same as https://github.com/openshift-knative/serverless-operator/pull/1831 except using a commit/tag


aliok:
/retest

I think upgrade tests are failing for real, but I cannot really understand

aliok:
/retest
Some infra issues

olm-catalog/serverless-operator/project.yaml
```diff
@@ -34,15 +34,15 @@ dependencies:
...
   # eventing core midstream branch name
   eventing_artifacts_branch: release-v1.4
 
   eventing_kafka: 1.1.0
-  # eventing-kafka-broker promotion version without the `knative-v` prefix
-  eventing_kafka_broker: 1.4
-  # eventing-kafka-broker midstream branch name
-  eventing_kafka_broker_artifacts_branch: release-v1.4
+  # eventing-kafka-broker promotion tag
+  eventing_kafka_broker: f24550ef10c9deea57c2306db05d813406ee73b8 # tag corresponds to a build of release v1.4
```
matzew, line 43: does _this_ require a manual "bump", if we want to ship something different? Yes, right?

aliok: Yes. We want to make a conscious decision

matzew: Doesn't this really slow us down?

Today, using `latest` we do get the updated images implicitly, now, when we want to address this - or update shipped software, we do have to constantly commit to this repo. 

Is this repo really the right repo to address this concern? 

I fear that this adds more unneeded load, or one use-cases: faster updates to "managed services"

pierDipi: I think, when we discussed this, we agreed that we keep using "latest for the branch" (aka floating) but when downstream users need a specific fix/commit form ekb, we will update it, open a PR, run the CI tests, merge (so that we can rebuild index images, etc), and then revert to floating

matzew: floating is a huge productivity factor. 

Which makes me wonder if we try to "fix/solve" the issue of "fast delivery", on a wrong level, or layer

aliok:
/close
Gonna work on this after https://github.com/openshift-knative/serverless-operator/pull/1792

PR #1841, opened 2022-11-28
Add annotation to toggle the gRPC/HTTP2 port name

nak3:
This patch introduces `serverless.openshift.io/default-enable-http2` for KnativeServing.

As current h2c protocol name breaks websocket on OCP Route, 
change the port name only when the annotation is added.
Note, this annotation should be able to drop around OCP 4.13.

Reference - https://docs.openshift.com/container-platform/4.11/networking/ingress-operator.html#nw-http2-haproxy_configuring-ingress
> Consequently, if you have an application that is intended to accept WebSocket connections,
>  it must not allow negotiating the HTTP/2 protocol or else clients will fail to upgrade to the WebSocket protocol.

### DIFF
--- a/openshift-knative-operator/pkg/serving/kourier.go
+++ b/openshift-knative-operator/pkg/serving/kourier.go
@@ -86,6 +86,15 @@ func addKourierEnvValues(ks base.KComponent) mf.Transformer {
 // addKourierAppProtocol adds appProtocol name to the Kourier service.
 // OpenShift Ingress needs to have it to handle gRPC/H2C.
 func addKourierAppProtocol(ks base.KComponent) mf.Transformer {
+	// TODO: revisit after OCP 4.13 (HAProxy 2.4) is available.
+	// As current h2c protocol name breaks websocket on OCP Route, change the port name only when the annotation is added.
+	//
+	// see - https://docs.openshift.com/container-platform/4.11/networking/ingress-operator.html#nw-http2-haproxy_configuring-ingress
+	// > Consequently, if you have an application that is intended to accept WebSocket connections,
+	// > it must not allow negotiating the HTTP/2 protocol or else clients will fail to upgrade to the WebSocket protocol.
+	if _, ok := ks.GetAnnotations()["serverless.openshift.io/default-enable-http2"]; !ok {
+		return nil
+	}
 	return func(u *unstructured.Unstructured) error {
 		if u.GetKind() != "Service" && u.GetName() != "kourier" {
 			return nil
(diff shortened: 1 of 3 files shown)


nak3:
/cc @skonto @mgencur 

I included https://github.com/openshift-knative/serverless-operator/pull/1841 to verify the websoket passed. If the commits should be separated I will update this PR.

openshift-knative-operator/pkg/serving/kourier.go
```diff
@@ -86,6 +86,15 @@ func addKourierEnvValues(ks base.KComponent) mf.Transformer {
 // addKourierAppProtocol adds appProtocol name to the Kourier service.
 // OpenShift Ingress needs to have it to handle gRPC/H2C.
 func addKourierAppProtocol(ks base.KComponent) mf.Transformer {
+	// TODO: revisit after OCP 4.13 (HAProxy 2.4) is available.
+	// As current h2c protocol name breaks websocket on OCP Route, change the port name only when the annotation is added.
+	//
+	// see - https://docs.openshift.com/container-platform/4.11/networking/ingress-operator.html#nw-http2-haproxy_configuring-ingress
+	// > Consequently, if you have an application that is intended to accept WebSocket connections,
+	// > it must not allow negotiating the HTTP/2 protocol or else clients will fail to upgrade to the WebSocket protocol.
+	if _, ok := ks.GetAnnotations()["serverless.openshift.io/default-enable-http2"]; !ok {
```
skonto, line 95: So this avoids setting the protocol for all scenarios unless the annotation is set.
What about grpc is this not supported if there is no annotation?

skonto:
> /cc @skonto @mgencur
> 
> I included #1841 to verify the websoket passed. If the commits should be separated I will update this PR.

Do you mean some other PR? Did not parse this one.

skonto:
```
05:08:23.657 DEBUG:   [[ $(oc get pods -n knative-eventing --field-selector=status.phase!=Succeeded -o jsonpath='{.items}') != '[]' ]] : Waiting until non-zero (max 700 sec.)
```

nak3:
> Do you mean some other PR? Did not parse this one.

yes, you can see https://github.com/openshift-knative/serverless-operator/pull/1841/commits/6a30b7b2e6d098ac4a94d97d3e73e1c6a350acbb which is another PR.

openshift-knative-operator/pkg/serving/kourier.go
nak3: Right, gRPC is not supported if there is no annotation.

nak3: > What about grpc is this not supported if there is no annotation?

It will break current websocket behavior as no current users have the annotation.

nak3: Sorry, I didn't get your suggestion.
You mean that we should support gRPC by default? If so, websocket will be broken and make an impact for the current users.

skonto: Ok yes that mans we dont have support for bothh by default.

PR #1845, opened 2022-11-28
[WIP] Use go install instead of go get

pierDipi:
In Go 1.18 go get produces errors


pierDipi:
Ah, the build doesn't pick up my changes, so we need to merge first to see the results

PR #1849, opened 2022-11-29
Remove the old hooks for the long ago removed/deprecated 'knative-sources' namespace

matzew:

## Proposed Changes

- as per title, we remove old hooks that tested on the deprecated `knative-sources` namespace


pierDipi (approved):
/lgtm
/approve

PR #1852, opened 2022-11-30
Set Serving previous version to 1.5.0

skonto:
- Accidentally set it to 1.6.0 now reverting back to the correct version.

### DIFF
--- a/olm-catalog/serverless-operator/project.yaml
+++ b/olm-catalog/serverless-operator/project.yaml
@@ -50,7 +50,7 @@ dependencies:
   operator: 1.6.1
   # Previous versions required for downgrade testing
   previous:
-    serving: 1.6.0
+    serving: 1.5.0
     eventing: 1.5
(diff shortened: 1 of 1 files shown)


pierDipi (approved):
/lgtm
/approve

for tests
/hold 

I fear this might fail https://github.com/openshift-knative/serverless-operator/blob/ff8ea160b20e856bb464643d7c6b0dc91a4da84b/test/upgrade/installation/serverless.go#L90-L97 because otherwise I would be surprised it is passing currently

skonto:
I think the crs once converted can be read with the new version.

pierDipi:
/unhold

I'm still confused on why it doesn't fail https://github.com/openshift-knative/serverless-operator/pull/1852#pullrequestreview-1199030577

skonto:
As discussed on slack it exits early if the previous version equals to the one before the downgrade because CR are not reconciled yet.

PR #1863, opened 2022-12-06
[WIP] Try running PrePost tests and Continual tests in separate cycles

mgencur:
This should have a number of benefits:
* The phase with continual tests run much shorter (only Serverless upgrade and downgrade is included, not the other Pre and Post upgrade/downgrade tests included in that timeframe) so it should be possible to gather more (all?) Traces during the run.
* Failed continual tests which time out won't cause all tests to have "Unknown" status
* Possibly less resources required for the cluster as fewer tests run in parallel. The Pre and Post upgrade/downgrade tests should not affect the continual tests now.

Drawbacks:
* Longer execution - in theory, longer by 3 minutes because what runs twice is upgrade (1 minute) and downgrade (2 minutes).

## Proposed Changes

-
-
-


mgencur:
Filed https://issues.redhat.com/browse/SRVCOM-2203 for the "upgrade after downgrade" scenario.

mgencur:
TODO: We need to be able to skip downgrades in the scenario where we run cluster upgrade after serverless upgrade.

mgencur:
Additional speed up is done in https://github.com/openshift-knative/serverless-operator/pull/1866

test/lib.bash
```diff
@@ -261,15 +261,10 @@ EOF
...
-    go_test_e2e -run=TestServerlessUpgrade -timeout=100m "${common_opts[@]}"
-
-    # Restart Zipkin to prevent OutOfMemory errors.
-    oc -n "${TRACING_NAMESPACE}" delete pod -l="app=zipkin"
-    oc -n "${TRACING_NAMESPACE}" wait --for=condition=ready --timeout=3m pod -l="app=zipkin"
-
-    if [[ "${SKIP_DOWNGRADE}" == "false" ]]; then
-      go_test_e2e -run=TestServerlessDowngrade -timeout=60m "${common_opts[@]}"
-    fi
+    # Run the two test suites one by one to prevent the situation when nested
```
matzew, line 264: :100:

matzew (approved):
/lgtm
/approve

matzew:
https://github.com/openshift-knative/serverless-operator/blob/970247d07461389b32b64ac13ce32177e88cce90/test/upgrade/installation/serverless.go#L103-L127

I guess some of the duplicated code in updrage/downgrade could be also extracted?

PR #1866, opened 2022-12-08
Reduce trace export limit from default 100 to 30 for each test

mgencur:
The goal is to speed up test execution as getting traces for many events
takes a lot of time.
There are two types of failures: missed events and duplicate events.
Usually, when it happens the number of missed events is very low (<10)
so this limit is for duplicate events. We allow duplicate events in the
test and traces are only informational.

## Proposed Changes

-
-
-


test/lib.bash
```diff
@@ -235,6 +235,7 @@ EOF
   # Review this line as part of SRVCOM-2176
   export EVENTING_UPGRADE_TESTS_SERVING_USE=false
   export EVENTING_UPGRADE_TESTS_CONFIGMOUNTPOINT=/.config/wathola
+  export EVENTING_UPGRADE_TESTS_TRACEEXPORTLIMIT=30
```
matzew, line 238: I reviewed the upstream already

matzew (approved):
/lgtm
/approve

PR #1872, opened 2022-12-15
[WIP] Restrict wildcard rbac verbs

ReToCode:
Fixes https://issues.redhat.com/browse/SRVKS-896

## Proposed Changes
- broom: Use specific rbac verbs instead of wildcards. According to http://static.open-scap.org/ssg-guides/ssg-ocp4-guide-cis.html#xccdf_org.ssgproject.content_rule_rbac_wildcard_use
- broom: split service account for `knative-openshift` and `knative-operator` to respect the principle of least privilege.

/hold


skonto:
cc @matzew or @pierDipi for any objection

nak3:
This PR is LGTM.

> Fixes https://issues.redhat.com/browse/SRVKS-896

Please allow me to confirm. As far as reading SRVKS-896, the customer asked us to fix the wildcard in https://github.com/knative/serving/tree/main/config/core/200-roles actually. So, this PR is a part of SRVKS-896, right?

ReToCode:
> 

For now, we just improved the situation for the roles SO. I'll need to discuss this result with Simon. The upstream roles have wildcards for `apiGroups` that are owned by Knative, which is viable to me and I think also to @skonto (the alternative would impose a long list of resources that we would need to maintain on various places). Also it's not quite clear how the customer actually tested/audited the roles. Seems at that point in time we used different roles claiming `cluster-admin`.

skonto:
Thanks for doing this @ReToCode, two questions:
a) Is it possible to define some [alias/reference for some common verbs,](https://github.com/openshift-knative/eventing/blob/release-v1.7/config/core/roles/controller-clusterroles.yaml#L145) does OLM support that in csv?
b) Is there a way to test if this is the minimum set of access rights required?

ReToCode:
a) Not that I'm aware of, also nothing in the documentation about it.
b) Also nothing that I'm aware of. The current list is a convergence approach. With `bind` and `escalate` the operator is still basically running as cluster admin, as he can create any clusterRole with any verbs.

skonto:
@ReToCode let's coordinate with Simon before moving on so we know we are solving this right or we need to do more for this part.

ReToCode:
@skonto we got the ok to go ahead with this, if there is no objection from @matzew or @pierDipi. Will also add a section  about the permission in docs.

skonto:
/lgtm
/approve

hold for @matzew if he has some last minute comment.

PR #1873, opened 2022-12-16
[WIP] Add KNATIVE_SERVING_TEST_MANIFESTS_DIR env variable

nak3:
This patch adds `KNATIVE_SERVING_TEST_MANIFESTS_DIR`

As per https://github.com/openshift-knative/serving/pull/81/files#r1049354250, we should
remove configmap approach in the midstream and we need to use env variable `KNATIVE_SERVING_TEST_MANIFESTS_DIR`
for the ingress manifests.


nak3:
/hold
Cutting 1.27 branch today and we should merge this PR once it was done.

nak3:
Thank you for the review. But I realized that this is not necessary for now. So I close here for now.

PR #1874, opened 2022-12-16
Fix set-env GitHub action

ReToCode:
## Proposed Changes
- Fix build issues with new syntax for GitHub actions ::set-env (https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files)


ReToCode:
The template in `csv.yaml` now contains a variable instead of `nightly`. Should we hard-code the version there or do we expect everyone to set `CURRENT_VERSION_IMAGES` when running `make release-files`?

pierDipi (approved):
Thanks!

/lgtm
/approve

PR #1883, opened 2022-12-19
Create eventing ns during strimzi install, already...

matzew:

Create eventing ns during strimzi install, already...


creydr (approved):
/lgtm

PR #1885, opened 2022-12-19
Create eventing namespace if it doesn't exist in install strimzi

pierDipi:
With the recent changes in [1], we're installing components in parallel which means that by the time we're trying to create Strimzi secrets in the eventing namespace, eventing is not installed, therefore there is no eventing namespace present.

[1] https://github.com/openshift-knative/serverless-operator/commit/db6b80685e4236e6de5dded74730a5106cceb781

### DIFF
--- a/hack/lib/strimzi.bash
+++ b/hack/lib/strimzi.bash
@@ -232,7 +232,7 @@ EOF
 
   logger.info "Deleting existing Kafka user secrets"
 
-  oc create namespace "${EVENTING_NAMESPACE}" --dry-run=client -o yaml | oc apply -f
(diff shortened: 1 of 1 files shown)


hack/lib/strimzi.bash
```diff
@@ -232,6 +232,8 @@ EOF
 
   logger.info "Deleting existing Kafka user secrets"
 
+  oc create namespace "${EVENTING_NAMESPACE}" --dry-run=client -o yaml | oc apply -f
```
matzew, line 235: why here? and not else, e.g. were we create other resources in this script/file ?

creydr:
Probably same as #1885

matzew:
/close

1883 got merged

PR #1890, opened 2022-12-21
[SRVCOM-2209] Support kube rbac proxy overrides

skonto:
- Allows to override kube-rbac-proxy container attributes eg. resources 
- Can't be done via a normal deployment override without this patch because knative operator applies deployment [overrides before we inject kube-rbac-proxy](https://github.com/knative/operator/blob/6f94c4f029e102e753ea4786a4780acc0ceff42e/pkg/reconciler/common/transformers.go#L49) container at the extension side. Followed the same approach for Kafka for consistency.

### DIFF
--- a/openshift-knative-operator/pkg/monitoring/helpers.go
+++ b/openshift-knative-operator/pkg/monitoring/helpers.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	mf "github.com/manifestival/manifestival"
+	"github.com/openshift-knative/serverless-operator/openshift-knative-operator/pkg/common"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
@@ -18,8 +19,6 @@ import (
 	"k8s.io/client-go/kubernetes/scheme"
 	"knative.dev/operator/pkg/apis/operator/base"
 	"knative.dev/pkg/logging"
-
-	"github.com/openshift-knative/serverless-operator/openshift-knative-operator/pkg/common"
 )
 
 const (
--- a/openshift-knative-operator/pkg/monitoring/rbac_proxy_injection_test.go
+++ b/openshift-knative-operator/pkg/monitoring/rbac_proxy_injection_test.go
@@ -13,6 +13,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/kubernetes/scheme"
+	"knative.dev/operator/pkg/apis/operator/base"
 )
 
 func TestInjectRbacProxyContainerToDeployments(t *testing.T) {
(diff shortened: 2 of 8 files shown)


openshift-knative-operator/pkg/monitoring/rbac_proxy_injection_test.go
```diff
@@ -64,6 +66,24 @@ func TestInjectRbacProxyContainerToDeployments(t *testing.T) {
...
+		Name: "activator",
+		Resources: []base.ResourceRequirementsOverride{{
+			Container: RBACContainerName,
+			ResourceRequirements: corev1.ResourceRequirements{
+				Limits: limits,
+			},
+		}},
+	}}
+
+	if manifest, err = manifest.Transform(operator.OverridesTransform(overrides, nil)); err != nil {
```
nak3, line 83:
```suggestion
	if manifest, err = manifest.Transform(operator.OverridesTransform(KubeRbacProxyOverridesOnly(overrides, sets.NewString("activator")), nil)); err != nil {
```

I think `KubeRbacProxyOverridesOnly` should be covered, thought current test (upstream's `OverridesTransform()` only) ends up passing.

openshift-knative-operator/pkg/monitoring/helpers.go
```diff
@@ -270,3 +270,13 @@ func getRBACManifest() (mf.Manifest, error) {
 	rbacManifest, err := mf.NewManifest(rbacPath)
 	return rbacManifest, err
 }
+
```
nak3, line 273: I think it is better to have the function comment which explains why this function is necessary as you explained in the PR message below.

> Can't be done via a normal deployment override without this patch because knative operator applies deployment [overrides before we inject kube-rbac-proxy](https://github.com/knative/operator/blob/6f94c4f029e102e753ea4786a4780acc0ceff42e/pkg/reconciler/common/transformers.go#L49) container at the extension side. Followed the same approach for Kafka for consistency.

openshift-knative-operator/pkg/monitoring/helpers.go
```diff
@@ -270,3 +270,13 @@ func getRBACManifest() (mf.Manifest, error) {
...
+func KubeRbacProxyOverridesOnly(overrides []base.WorkloadOverride, deployments sets.String) []base.WorkloadOverride {
```
nak3, line 274: I wondered what is the meaning of `Only` in `KubeRbacProxyOverridesOnly`, And I understood as "override the resource in the deployment "only" which has kube-rbac-proxy". Maybe it is better to add function comment.

nak3:
As per customer's RFE in JIRA ticket:

__Customer need limits because in their environment they have a limitrange object which requires set limits for every container.__

I feel that they might want us to support one common resource setting for `kube-rbac-proxy` like [queue-proxy's](https://github.com/knative/serving/blob/3b8f17cd6f97f419dc9bf8796a0b8a31335e75e5/config/core/configmaps/deployment.yaml#L55-L69
). Otherwise, the customer must configure the resource limit on kube-rbac-proxy for every single deployments :cold_sweat:

Of course, your approach allows them to configure the flexible configuration, though...

skonto:
> I feel that they might want us to support one common resource setting for kube-rbac-proxy like [queue-proxy's](https://github.com/knative/serving/blob/3b8f17cd6f97f419dc9bf8796a0b8a31335e75e5/config/core/configmaps/deployment.yaml#L55-L69). Otherwise, the customer must configure the resource limit on kube-rbac-proxy for every single deployments 

We could just set a default value as Knative does for queue proxy and have the ability to configure it via a dummy config option in Serving CR:
```
spec:
  config:
    deployment:
       kube-rbac-proxy-cpu-request: ...
       kube-rbac-proxy-cpu-limit:
       ...
```
Then we could just set the right value at the extension side during transformation phase.

openshift-knative-operator/pkg/monitoring/helpers.go
skonto: It means only consider the deployment override that has a container name `kube-rbac-proxy`. I will add a  comment.

openshift-knative-operator/pkg/monitoring/helpers.go
skonto: Sure.

openshift-knative-operator/pkg/monitoring/rbac_proxy_injection_test.go
skonto:
```	overrides := []base.WorkloadOverride{{
		Name: "activator",
		Resources: []base.ResourceRequirementsOverride{{
			Container: RBACContainerName,
			ResourceRequirements: corev1.ResourceRequirements{
				Limits: limits,
			},
		}},
	}}
```
`KubeRbacProxyOverridesOnly` does just filtering based on deployment name which is supposed to have a kube-rbac-proxy. In this cases it makes no difference if we apply the filtering or not as the returned array is the same.

skonto:
@nak3 could you review/stamp? gentle ping :)

pierDipi (approved):
/lgtm

skonto:
> And this is a OpenShift specific feature, so I wonder that we should verify this feature on e2e test. (We don't need a performance test but just checking the modified value was applied or not via API.)

I will take a look may add one.

PR #1888, opened 2022-12-20
Bump eventing and knative-operator to v1.7

ReToCode:
## Proposed Changes
- Bump all dependencies to 1.7 in `olm-catalog/serverless-operator/project.yaml` and `hack/update-deps.sh`
- Update dependencies with `make generated-files` and `./hack/update-deps.sh --upgrade`
- Updated the patch `hack/patches/004-disable-manifests-version-match.patch`
- Added another `go mod tidy -compat=1.18`, otherwise go `go mod vendor` did not run
- Updated `go mod tidy` format to `1.18`
- Introduce a new directory and env `GORUN_PATH` variable because of changes in `go_run()` function here: https://github.com/knative/hack/commit/123a2783c4d8e4091e1625e0410d9f60c9a18e1a

As discussed in Slack, I'm unable to update the go dependencies without affecting serving as well. We'll see how this merges together with serving and operator bumps.

PR #1896, opened 2022-12-22
[obsolete] Namespaced broker metrics

aliok:
Fixes JIRA #

## Proposed Changes

- Introduce webhook for KnativeKafka
- Set `namespaced-broker-resources` in KnativeKafka `spec.config`, so that we fill the `config-namespaced-broker-resources` configmap with the stuff we need for each namespaced broker dataplane in broker namespace
- For each namespaced broker dataplane, we do these:
  - Create a Kubernetes `Service` that makes the dataplane pods accessible by Prometheus. That service basically integrates with the RBAC proxy sidecar in the dataplane pods
  - Create a `ServiceMonitor` that makes Prometheus scrape our dataplane pods. It also sets up certificates for the RBAC proxy mentioned above.
  - Create a `ClusterRoleBinding` of the `rbac-proxy-reviews-prom` `ClusterRole` for the `knative-kafka-broker-data-plane` `ServiceAccount` in the broker namespace. Otherwise, RBAC proxy cannot authorize itself.
  - Set `"openshift.io/cluster-monitoring": "true",` label on the broker namespace, so that Prometheus monitors the namespace and creates certs/secrets and also scrapes the pods
- Also, now that we tell the Knative Kafka controller to do all of these stuff above, it needs RBAC permissions. I created `eventing-kafka-broker-extra.yaml` file in the KnativeKafka manifest that contains the required cluster role and its binding. Since the file is in the correct directory, Serverless Operator simply applies the content of the file.

### NOTE: Please review commit by commit

Questions:
- Is it ok to leave the `"openshift.io/cluster-monitoring": "true"` label on the namespace when all the brokers in the namespace are deleted? This is a hard problem since we won't know who actually set the label.


knative-operator/deploy/resources/knativekafka/broker/eventing-kafka-broker-extra.yaml
```diff
@@ -0,0 +1,55 @@
+# ClusterRole to be able to create resources for namespaced broker
+# Kafka controller creates some additional resources, such as ServiceMonitors, in the broker namespace
+# when a namespaced broker is created.
+# The controller will need the permissions listed in this role.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: eventing-kafka-namespaced-broker-additional-resources
```
matzew, line 8: aren't these two files the same?
should we add them on midstream, and continue to just populate this folder with manifests from midstream?

knative-operator/hack/update-manifests.sh
```diff
@@ -59,3 +59,7 @@ download_kafka source "${kafka_source_files[@]}"
 # artifacts are downloaded from midstream openshift/release/artifacts directory.
 # Before adding patches to this file consider sending a patch to midstream and then by just running
 # `make generated-files` the patch will appear in the final bundled artifacts.
+
+# Copy some additional manifest files that are only relevant for Serverless Operator.
+extra_dir="$root/knative-operator/deploy/resources/knativekafka_extra"
```
matzew, line 64: why is extra a special thing?

knative-operator/deploy/resources/knativekafka/broker/eventing-kafka-broker-extra.yaml
aliok: I thought about that, but that's will be patch then. ...which is not ideal :(

knative-operator/hack/update-manifests.sh
aliok: Because the files in `knative-operator/deploy/resources/` dir are actually downloaded from midstream within this script above. Before downloading, the script deletes the contents of the directory.

aliok: If there's already something in midstream, I would be happy to leverage that.

Or, if you have any better idea

aliok:
/test all

https://github.com/openshift-knative/eventing-kafka-broker/pull/522 is merged now. Tests here should be all green.

I will then add the e2e tests for the namespaced broker metrics.

aliok:
```
error: failed to create secret secrets "strimzi-tls-secret" already exists
```

To fix the issue above, we need to get the Eventing-Kafka PR here merged: https://github.com/knative-sandbox/eventing-kafka/pull/1277

Then, update deps in midstream Eventing-Kafka-Broker.
Then, update deps in ServerlessOperator to use newest EKB.

😢 

SO uses a script from EKB, that comes from EK

aliok:
All tests pass locally. Removed WIP status

aliok:
/hold

Gonna rebase on top of https://github.com/openshift-knative/serverless-operator/pull/1888

aliok:
/test 4.11-operator-e2e-aws-ocp-411

Irrelevant
```
kafka_source_to_ksvc_test.go:304: failed while waiting for KafkaSource to become ready: timed out waiting for the condition
``

**Comment by aliok** [comment]
/test 4.11-operator-e2e-aws-ocp-411

```
kafka_source_to_ksvc_test.go:304: failed while waiting for KafkaSource to become ready: timed out waiting for the condition
```

File: openshift-knative-operator/pkg/monitoring/monitoring_kafka.go [hunk 3]
```diff
@@ -0,0 +1,194 @@
...
+			Kind:       "Namespace",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "{{.Namespace}}",
+			Labels: map[string]string{
+				"openshift.io/cluster-monitoring": "true",
+			},
+		},
+	}
+}
```
pierDipi, line 101: The namespace is controlled by the user, are we sure we're not conflicting with whatever their config (labels/annotations, etc).

I'd honestly let them provide the `openshift.io/cluster-monitoring: true` label if they want without us assuming cluster monitoring is present (note: recently OpenShift monitoring was made an optional installation)

knative-operator/pkg/webhook/knativekafka/webhook_mutating.go
```diff
@@ -0,0 +1,47 @@
...
+	if err != nil {
+		return admission.Errored(http.StatusInternalServerError, err)
+	}
+
+	marshaled, err := json.Marshal(ke)
+	if err != nil {
+		return admission.Errored(http.StatusInternalServerError, err)
+	}
+	return admission.PatchResponseFromRaw(req.AdmissionRequest.Object.Raw, marshaled)
+}
```
pierDipi, line 47: Should we handle the case where cluster monitoring is not installed?

openshift-knative-operator/pkg/monitoring/monitoring_kafka.go
matzew: good point, especially w/ the optional factor of monitoring (did not know)

knative-operator/pkg/webhook/knativekafka/webhook_mutating.go
matzew: Good point. Question: Is this handle else already? We could separate this extra handling in a follow up PR ? :shrug:

pierDipi: There is a function `ShouldEnableMonitoring` we could leverage

pierDipi: Yes, follow up is fine too

pierDipi:
Another question that I have that is not clear from the PR description is:

why a webhook? why not adding the logic to the knativekafka reconciler?

openshift-knative-operator/pkg/monitoring/monitoring_kafka.go
aliok: If we don't provide `openshift.io/cluster-monitoring: true`, then the monitoring operator won't create the secret, which will make the dataplane pods fail I think. It is because the RBAC proxy side car for monitoring wants that secret mounted.

knative-operator/pkg/webhook/knativekafka/webhook_mutating.go
aliok: Let me check. I was thinking I already leveraged that but let me check again.

openshift-knative-operator/pkg/monitoring/monitoring_kafka.go
pierDipi: I see, I don't see logic in the existing reconciler to avoid injecting RBAC proxy when cluster monitoring is disabled

pierDipi: cc @skonto

aliok:
> Another question that I have that is not clear from the PR description is:
> why a webhook? why not adding the logic to the knativekafka reconciler?

IMO it is not a good idea to mutate the `KnativeKafka` CR in the KnativeKafka reconciler.

Why mutation? I used the existing `spec.config` feature that creates/updates configmaps.

pierDipi:
My concern is that existing installations or newer ones will apply a small KnativeKafka and they will get back a huge KnativeKafka with a field that is not supposed to be end user used

pierDipi:
So, I think hiding the implementation details from the end user is better, the implementation is similar we just need the json string `string_from_extra_file` which you already generate in `ReconcileMonitoringForNamespacedBroker`

openshift-knative-operator/pkg/monitoring/monitoring_kafka.go
aliok: > The namespace is controlled by the user, are we sure we're not conflicting with whatever their config (labels/annotations, etc).

BTW, this label is "applied". Other labels/annotations won't change

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -748,3 +751,23 @@ func removeCreationTimestamp(manifest *mf.Manifest, _ *serverlessoperatorv1alpha
...
 }
+
+func injectNamespacedBrokerMonitoring(kk *serverlessoperatorv1alpha1.KnativeKafka) mf.Transformer {
+	return func(u *unstructured.Unstructured) error {
+		if u.GetKind() != "ConfigMap" || u.GetName() != "config-namespaced-broker-resources" {
+			return nil
+		}
+		if !openshiftmonitoring.ShouldEnableMonitoring(kk.Spec.Config) {
+			return nil
+		}
```
pierDipi, line 762: KnativeKafka is not used to configure monitoring for Eventing components, but only KnativeEventing, so we should check on KnativeEventing.

However, I'm fine with a TODO / Jira and address it separately but just to clarify that this won't work as intended.

aliok: Addressed in [a4225f2](https://github.com/openshift-knative/serverless-operator/pull/1896/commits/a4225f24eee6f7b4545f18b5822046fe47982cdd)

aliok:
/test 4.11-operator-e2e-aws-ocp-411

```
kafka_source_to_ksvc_test
```

knative-operator/pkg/webhook/knativekafka/webhook_mutating.go
aliok: This is done in this PR

openshift-knative-operator/pkg/monitoring/monitoring_kafka.go
pierDipi: Ok, then we're good

pierDipi: One problem with this approach is that I cannot use user-defined project monitoring instead of cluster monitoring because the label is hardcoded, can we at least try to use whatever is set on the knative-eventing namespace instead of hardcoding it here?

https://docs.openshift.com/container-platform/4.10/monitoring/enabling-monitoring-for-user-defined-projects.html

aliok: Alternatives:

1. We don't care about user-defined project monitoring and keep things as is
2. We provide a flag in the KnativeKafka CRD to specify the label. That would apply to all namespaces though.
3. We don't set the label on the namespace. Then we need to make sure if the label is not there, we also don't inject the RBAC proxy, create the ServiceMonitor, etc. So, user needs to label the namespace, then create the namespaced broker instance. 

Option 3) sounds reasonable in the summary above, but there are some issues:
a. This would require some work on eventing-kafka-broker, as KafkaController will need to remove the RBAC proxy sidecar when propagating the dataplane deployments if there's no monitoring label set.
b. Similarly, no monitoring label is set, then the additional resources (ServiceMonitor, ClusterRoleBinding, etc.) shouldn't be created. This is again to be checked by the KafkaController.

Any other ideas @pierDipi 

Also, I am just thinking if we can merge this PR as is and implement support for user-defined project monitoring as a follow up, although it will probably need migration code as the release cut is pretty soon.

pierDipi: I'm ok with keeping as is for now, and then iterate towards 2 or 3

aliok: OK, let's merge as is.

aliok:
/test 4.11-operator-e2e-aws-ocp-411

I am starting to think there's an actual failure here though: ` kafka_source_to_ksvc_test.go:304: failed while waiting for KafkaSource to become ready: timed out waiting for the condition`

aliok:
Cherrypicked https://github.com/openshift-knative/serverless-operator/pull/1905/commits/de1a2417f3dd83020edee199e3193a211c212f4c to see if that fixes the issue

aliok:
/test 4.11-operator-e2e-aws-ocp-411

## Message 1
```
 kafka_source_to_ksvc_test.go:346: Failed to delete KafkaSource serverless-tests/smoke-ks-plain: timed out waiting for the condition
```

## Message 2
```
failed to reconcile contract: failed to reconcile egress: failed to resolve subscriber: failed to get object serverless-tests/helloworld-go-sasl: services.serving.knative.dev "helloworld-go-sasl" not found
```

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-knative_serverless-operator/1896/pull-ci-openshift-knative-serverless-operator-main-4.11-operator-e2e-aws-ocp-411/1611042684074463232/artifacts/operator-e2e-aws-ocp-411/operator-e2e/artifacts/gather-knative/quay-io-openshift-knative-must-gather-sha256-f1cd829ea23838ac3e9b7e3a42269ea34a32e7af6a5465f9d5d21a3caf3ce446/serverless-tests/kafkasources.sources.knative.dev/describe.log


## Message 3

```
{"level":"error","ts":"2023-01-05T18:16:17.644Z","logger":"kafka-broker-controller","caller":"consumergroup/reconciler.go:302","msg":"Returned an error","commit":"630a33d-dirty","knative.dev/pod":"kafka-controller-8446c8567b-grgwd","knative.dev/controller":"knative.dev.eventing-kafka-broker.control-plane.pkg.reconciler.consumergroup.Reconciler","knative.dev/kind":"internal.kafka.eventing.knative.dev.ConsumerGroup","knative.dev/traceid":"ec8f78b3-85b6-4928-b509-aad551778875","knative.dev/key":"serverless-tests/9f43dfbc-6551-4e62-bf59-a147d7068dda","targetMethod":"FinalizeKind","error":"unable to delete the consumer group smoke-cg-smoke-ks-sasl: kafka server: The group is not empty","stacktrace":"knative.dev/eventing-kafka-broker/control-plane/pkg/client/internals/kafka/injection/reconciler/eventing/v1alpha1/consumergroup.(*reconcilerImpl).Reconcile\n\t/go/src/github.com/openshift/origin/control-plane/pkg/client/internals/kafka/injection/reconciler/eventing/v1alpha1/consumergroup/reconciler.go:302\nknative.dev/pkg/controller.(*Impl).processNextWorkItem\n\t/go/src/github.com/openshift/origin/vendor/knative.dev/pkg/controller/controller.go:542\nknative.dev/pkg/controller.(*Impl).RunContext.func3\n\t/go/src/github.com/openshift/origin/vendor/knative.dev/pkg/controller/controller.go:491"}
{"level":"error","ts":"2023-01-05T18:16:17.644Z","logger":"kafka-broker-controller","caller":"controller/controller.go:566","msg":"Reconcile error","commit":"630a33d-dirty","knative.dev/pod":"kafka-controller-8446c8567b-grgwd","knative.dev/controller":"knative.dev.eventing-kafka-broker.control-plane.pkg.reconciler.consumergroup.Reconciler","knative.dev/kind":"internal.kafka.eventing.knative.dev.ConsumerGroup","knative.dev/traceid":"ec8f78b3-85b6-4928-b509-aad551778875","knative.dev/key":"serverless-tests/9f43dfbc-6551-4e62-bf59-a147d7068dda","duration":0.116967206,"error":"unable to delete the consumer group smoke-cg-smoke-ks-sasl: kafka server: The group is not empty","stacktrace":"knative.dev/pkg/controller.(*Impl).handleErr\n\t/go/src/github.com/openshift/origin/vendor/knative.dev/pkg/controller/controller.go:566\nknative.dev/pkg/controller.(*Impl).processNextWorkItem\n\t/go/src/github.com/openshift/origin/vendor/knative.dev/pkg/controller/controller.go:543\nknative.dev/pkg/controller.(*Impl).RunContext.func3\n\t/go/src/github.com/openshift/origin/vendor/knative.dev/pkg/controller/controller.go:491"}
{"level":"info","ts":"2023-01-05T18:16:17.645Z","logger":"kafka-broker-controller.event-broadcaster","caller":"record/event.go:285","msg":"Event(v1.ObjectReference{Kind:\"ConsumerGroup\", Namespace:\"serverless-tests\", Name:\"9f43dfbc-6551-4e62-bf59-a147d7068dda\", UID:\"1453dd71-7e82-4a50-b46a-dd19d52e2a2e\", APIVersion:\"internal.kafka.eventing.knative.dev/v1alpha1\", ResourceVersion:\"73205\", FieldPath:\"\"}): type: 'Warning' reason: 'InternalError' unable to delete the consumer group smoke-cg-smoke-ks-sasl: kafka server: The group is not empty","commit":"630a33d-dirty","knative.dev/pod":"kafka-controller-8446c8567b-grgwd"}
```

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-knative_serverless-operator/1896/pull-ci-openshift-knative-serverless-operator-main-4.11-operator-e2e-aws-ocp-411/1611042684074463232/artifacts/operator-e2e-aws-ocp-411/gather-extra/artifacts/pods/knative-eventing_kafka-controller-8446c8567b-grgwd_controller.log

knative-operator/pkg/controller/knativekafka/knativekafka_controller.go
```diff
@@ -748,3 +753,33 @@ func removeCreationTimestamp(manifest *mf.Manifest, _ *serverlessoperatorv1alpha
...
 }
+
+func injectNamespacedBrokerMonitoring(apiClient client.Client) mf.Transformer {
+	return func(u *unstructured.Unstructured) error {
+		if u.GetKind() != "ConfigMap" || u.GetName() != "config-namespaced-broker-resources" {
+			return nil
+		}
+
+		eventingList := &operatorv1beta1.KnativeEventingList{}
+		err := apiClient.List(context.Background(), eventingList)
```
pierDipi, line 764: With this additional logic, I think we need to watch KnativeEventing too here: https://github.com/openshift-knative/serverless-operator/blob/f1c6988f062e599cf16c210269e8d87c1479b542/knative-operator/pkg/controller/knativekafka/knativekafka_controller.go#L144-L146

aliok:
/close

Closing this as I worked on 1.27 here: https://github.com/openshift-knative/serverless-operator/pull/1914
I am currently cherrypicking that PR into main: https://github.com/openshift-knative/serverless-operator/pull/1923

I will address the comments here in this PR separately.

aliok:
Another note here:

```
    - lastTransitionTime: "2023-01-10T10:19:37Z"
      message: 'unable to transform dataplane manifest: failed to get ConfigMap knative-eventing/config-tracing:
        configmap "config-tracing" not found'
```

Looks like there's an issue when we delete KnativeEventing before KnativeKafka.

aliok:
```
    - lastTransitionTime: "2023-01-10T10:19:37Z"
      message: 'unable to transform dataplane manifest: failed to get ConfigMap knative-eventing/config-tracing:
        configmap "config-tracing" not found'
```

When I delete KnativeEventing and keep KnativeKafka, Broker deletion is not blocked. 

However, Broker reconciliation is blocked when KnativeEventing is deleted. ...which is totally expected and I don't think we should try to provide a fix here.

cc @pierDipi

PR #1859, opened 2022-12-05
:bug: [QE] Serverless is default for 4.12+ so UI tests no longer need to choose Serverless type

cardil:
Part of SRVCOM-2182

## Proposed Changes

- :bug: [QE] Serverless is default for 4.12+ so UI tests no longer need to choose Serverless type

PR #1897, opened 2022-12-22
:gift: Await at fail if debug label is present

cardil:
## Proposed Changes

- :gift: Await at fail if debug label is present

### DIFF
--- a/hack/lib/ui.bash
+++ b/hack/lib/ui.bash
@@ -65,6 +65,7 @@ function debugging.setup {
   export BASH_XTRACEFD=19
 
   error_handlers.register stacktrace
+  await-at-fail.setup
 
   # Register finish of debugging at exit
   trap debugging.finish EXIT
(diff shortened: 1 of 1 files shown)


cardil:
You could see this in action on https://github.com/openshift-knative/serverless-operator/pull/1859 tests

hack/lib/ui.bash
```diff
@@ -77,6 +78,29 @@ function debugging.finish {
...
+    error_handlers.register await-at-fail
+  else
+    logger.info 'Add "ci/await-at-fail" label to PR, if you like to await after failure to allow debugging'
+  fi
+}
+
+function await-at-fail {
+  logger.info 'Awaiting at max 1h for user to debug'
+  timeout 3600 "[[ \$(oc get configmap done -n debug -o name | wc -l) -eq 0 ]]"
+}
```
pierDipi, line 102: This seems generically useful, it doesn't seem specific to ui tests, can we put these functions in common lib?

I'd call the function and the label `await_on_error` to be consistent with `error_handlers` terminology

pierDipi:
@cardil @mgencur @skonto I'm closing this PR as there is no reason to spend CI runs on this

PR #1908, opened 2023-01-06
[SRVKS-876] Use http-protocol instead of httpOption

nak3:
This is a tiny change which uses an updated annotation key.
The key is available since upstream v1.1.


nak3:
Docs has a mistake https://issues.redhat.com/browse/SRVKS-1004

ReToCode (approved):

PR #1933, opened 2023-01-17
Bump Maistra 2.3 image for test

nak3:
As per title, this patch bumps Maistra 2.3.


olm-catalog/serverless-operator/project.yaml
```diff
@@ -35,7 +35,7 @@ dependencies:
   net_kourier_artifacts_branch: release-1.7
   net_istio: 1.7.0
   net_istio_artifacts_branch: release-1.7
-  maistra: 2.1.0
+  maistra: 2.3-daily-2023-01-16
```
nak3, line 38: On upstream, it was possible to set the prefix by `date` command https://github.com/knative-sandbox/net-kourier/blob/main/.github/workflows/kind-e2e.yaml#L127-L128 but it is not possible here as it will change the csv file and `make generated-files` produces diff every day.

Asking SM team if they can add `2.3` or `2.3-latest` tags.

/hold

skonto: Or we could replace it at the extension side no?

nak3:
/hold cancel

ServiceMesh team added the `2.3-latest` tag :tada:

PR #1934, opened 2023-01-17
Increase receiver teardown to 10 seconds

mgencur:
It's the time between receiving "Finished" event (which holds the number of sent events) and the moment when the number of actually received events is compared with the number from Finished event. The receiver needs to wait longer to receive all the events from different Kafka partisions.

This is what we already have in [eventing-kafka-broker config](https://github.com/openshift-knative/eventing-kafka-broker/blob/release-v1.6/test/upgrade/continual/channel-config.toml#L10).

I see in downstream testing that events from the test end are often marked as not delivered but the Trace for the events is complete. This setting should fix that.

## Proposed Changes

-
-
-

### DIFF
--- a/test/upgrade/continual/channel-config.toml
+++ b/test/upgrade/continual/channel-config.toml
@@ -7,3 +7,7 @@ interval = {{ .Config.Interval.Nanoseconds }}
 
 [forwarder]
 target = '{{- .ForwarderTarget -}}'
+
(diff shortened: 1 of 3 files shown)


mgencur:
/hold
For upgrade tests to pass.

cardil (approved):
/lgtm

PR #1950, opened 2023-01-30
Test EUS OpenShift upgrades

mgencur:
Fixes [SRVCOM-1737](https://issues.redhat.com/browse/SRVCOM-1737)

Tests upgrades from one EUS version to another (e.g. 4.8 -> 4.9 -> 4.10). Follows guidelines from https://docs.openshift.com/container-platform/4.8/updating/preparing-eus-eus-upgrade.html

## Proposed Changes

- If channel is EUS do two upgrades in a row (skipping a version is not possible)
- Use semantic version comparison for OpenShift versions instead of string comparison ("4.10" must be recognized as higher than "4.9")


test/upgrade/installation/openshift.go
```diff
@@ -88,3 +94,78 @@ func IsClusterVersionWithImageReady(image string) ClusterVersionInStateFunc {
...
+
+	pauseMachineConfigPool(ctx, true)
+
+	if err := UpgradeOpenShift(ctx); err != nil {
+		return fmt.Errorf("failed to upgrade to odd OpenShift release: %w", err)
+	}
+
+	if err := UpgradeOpenShift(ctx); err != nil {
+		return fmt.Errorf("failed to upgrade to even OpenShift release: %w", err)
+	}
```
cardil, line 121: Why is this done twice?

cardil (commented):
Looks nice. Just a nit, and a question...

test/upgrade/installation/openshift.go
```diff
@@ -31,7 +37,7 @@ func UpgradeOpenShift(ctx *test.Context) error {
 		desiredRelease := clusterVersion.Status.Desired
 		// Choose the highest version as the list can be unordered.
 		for _, update := range clusterVersion.Status.AvailableUpdates {
-			if update.Version > desiredRelease.Version {
+			if semver.Compare("v"+update.Version, "v"+desiredRelease.Version) == 1 {
```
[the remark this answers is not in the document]
mgencur: Thanks, I was looking at this alternative. For example, [this comparison](https://pkg.go.dev/search?q=semver&m=)shows how many projects import both packages. The number if higher for x/mod.
I think blang/semver/v4 might be better for some very general usage where you don't exactly know what the version looks like so it catches errors and does some other modifications. But in this particular case, the version format is stable. And it's a one liner - simple and clear. I don't know if using the longer approach brings any benefit here.

test/upgrade/installation/openshift.go
mgencur: OpenShift cannot skip a version during upgrades. It needs to go through each upgrade one-by-one. It's also described in the docs I linked.

test/upgrade/installation/openshift.go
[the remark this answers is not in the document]
cardil: True, as long your statement about the stability of the version format is true. Those values came from outside: either from our scripts, or from `ClusterVersion` CR.

I can easily image, we or OCP team, adding a `v` to the number. Then, the code you proposed will fail to parse it properly. And, that would be a silent fail: https://go.dev/play/p/VvjyepnG5rj

test/upgrade/installation/openshift.go
cardil: So, the `UpgradeOpenShift` func upgrade OCP to next major version. I see.

How about we make it clear in the func name: `UpgradeOpenShiftToNextMajor`?

mgencur: It's not exactly "next major". It upgrades to the last version (e.g. 4.9.54) from the next stream 4.9 and then to the last version (4.10.45) of the next stream 4.10. Not sure how to better name it than UpgradeEUS. I also included this comment:
```
// UpgradeEUS follows guidelines from https://docs.openshift.com/container-platform/4.8/updating/preparing-eus-eus-upgrade.html
// to upgrade OpenShift between two EUS versions.
```
maybe "UpgradeOpenShiftEUSToEUS` ?

cardil: I was thinking about renaming the already present `UpgradeOpenShift` func. but, maybe that would be obvious if the code that uses it be a bit clearer about the intent.

How about we make it clear and at the same time dynamic. Now, you are assuming the EUS to EUS is 2 releases apart. OCP could switch to have longer window in the future.

How about:

```golang
for upgraded := false; !upgraded; upgraded = isEUSRelease(ctx) {
  if err := UpgradeOpenShift(ctx); err != nil {
    return fmt.Errorf("failed to upgrade to OpenShift release: %w", err)
  }
}
```

mgencur: Hmm. There's no easy/reliable way to figure out if any release is EUS. The only way is to compare the MAJOR.MINOR version of CSV with the name of the channel (e.g. eus-4.10). And that sounds easier to change than changing the whole strategy by OpenShift team of having EUS for every second minor version of OCP.

FWIW, I'm not assuming anything. This is the approach that is described in [docs](https://docs.openshift.com/container-platform/4.10/updating/preparing-eus-eus-upgrade.html#updating-eus-to-eus-upgrade_eus-to-eus-upgrade) for EUS-to-EUS upgrade and it is clearly stated that every "all even-numbered minor versions of OpenShift Container Platform offer Extended Update Support (EUS)"

I know you're seeing a little duplication in the code but it is just one line (even the error message is different in this case). Using a loop only makes it more confusing by suggesting that there can be any number of openshift upgrades.

test/upgrade/installation/openshift.go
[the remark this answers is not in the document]
mgencur: Alright, pushed the change.

test/upgrade/installation/openshift.go
cardil: Okay, I get you.

Could you please just add a comment like:

```golang
// Calling UpgradeOpenShift twice to upgrade from one EUS release to another EUS release
```

mgencur: Done.

cardil: BTW. I had proposed that loop because I saw there was this function already: `installation.IsChannelEUS(ctx)`.

cardil (approved):
/lgtm

mgencur:
@cardil Thanks.

PR #1954, opened 2023-02-03
[SRVKS-944] improve secret informer filtering

skonto:
- Brings back the deprecated Istio annotation that we preserved in [1.27](https://github.com/openshift/openshift-docs/pull/54913#discussion_r1090375002). When the annotation is applied it has the same behavior as in 1.27. It will stay around for a release at least so users can be prepared.
- By default we set the filtering off. It will be enabled by default in 1.29.
- If user uses the new way of setting the env vars the annotation is ignored for Istio.
- The deprecated annotation is not introduced for Kourier since it is going to be removed.
- Added a number of test cases covering all required scenarios.
- Any configuration applied is based on the ingress enabled.

### DIFF
--- a/openshift-knative-operator/pkg/serving/extension.go
+++ b/openshift-knative-operator/pkg/serving/extension.go
@@ -71,10 +71,10 @@ func (e *extension) Transformers(ks base.KComponent) []mf.Transformer {
 		overrideKourierNamespace(ks),
 		addKourierEnvValues(ks),
 		addKourierAppProtocol(ks),
-		enableSecretInformerFiltering(ks),
 		common.VersionedJobNameTransform(),
 		common.InjectCommonEnvironment(),
 	}
+	tf = append(tf, enableSecretInformerFilteringTransformers(ks)...)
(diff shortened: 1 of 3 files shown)


nak3:
> By default we set the filtering on.

Please allow me to confirm. The 1.27 release note mentions that the default will be changed on 1.28? If not, I think we should not change it yet for the users preparation. FYI @maschmid

nak3:
LGTM if the 1.27 release note mentions that the default will be changed.

skonto:
@nak3 the release notes will mention nothing for the feature we will start by enabling this by default and deprecating the annotation: https://github.com/openshift/openshift-docs/pull/54913#discussion_r1090375002. Will that create issues?

nak3:
The 1.28 release note will be published at the same time with the 1.28 product release so users cannot prepare for the default change if they use automatic upgrade. So, I think we should announce it at least one previous version about the change. 
( @maschmid You also concerned abut it, right?)

skonto:
Ok let's enable in 1.29.

nak3:
/lgtm
Thank you so much @skonto !

PR #1956, opened 2023-02-08
Increase parallel to 16 and run autoscale tests separately

mgencur:
Related to https://issues.redhat.com/browse/SRVKS-1021

## Proposed Changes

- Increase parallel settings for E2E tests to 16
- Run autoscale tests separately with parallel==3

The autoscale test is affected by CPU resources of the client node. Increasing `parallel` to a high value might affect the test (cos other parallel tests utilize the client CPU as well) and it won't be able to generate enough load. Thus we run it separately with lower parallelism.

### DIFF
--- a/test/serving.bash
+++ b/test/serving.bash
@@ -96,7 +96,7 @@ function upstream_knative_serving_e2e_and_conformance_tests {
     parallel=2
   fi
 
-  mv ./e2e/autoscale_test.go ./e2e/autoscale_test.backup
+  mv ./test/e2e/autoscale_test.go ./test/e2e/autoscale_test.backup
 
   SYSTEM_NAMESPACE="$SERVING_NAMESPACE" go_test_e2e -tags="e2e" -timeout=30m -parallel=$parallel \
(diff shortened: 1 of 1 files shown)


mgencur:
The test container in 4.11-upstream-e2e-aws-ocp-411 from the latest run:
```
Step upstream-e2e-aws-ocp-411-serverless-e2e succeeded after 1h2m30s
```
Some older runs with the previous parallel settings (==3):
```	
Step upstream-e2e-aws-ocp-411-serverless-e2e succeeded after 1h5m0s
Step upstream-e2e-aws-ocp-411-serverless-e2e succeeded after 1h8m0s
```

mgencur:
@skonto This one is ready for merging, IMO.

mgencur:
> I guess we need this independently of the upstream fix?

Yes. This PR is more like an improvement to speed up CI. It's not really a fix for the issue. The fix will come from upstream. We will see in the future if we will need to increase "cpu request" for the CI pod that runs the test. As you said we saw the failure in CI as well even though it was rare. The upstream fix should improve the situation and I'd only increase cpu request if we still see the problem.

PR #1957, opened 2023-02-08
Issues/srvke 1290

matzew:
Fixes [SRVKE-1290](https://issues.redhat.com/browse/SRVKE-1290)

## Proposed Changes

- exclude those pods that are not considered HA/production, since we now ship Kafka for actual prod.


matzew:
/hold
we do this _after_ we have cut for 1.28, so that this will land in 1.29

/cc @maschmid

ReToCode (approved):
/lgtm
/approve

PR #1962, opened 2023-02-16
Deep Kitchensink upgrade tests

mgencur:
Fixes https://issues.redhat.com/browse/SRVCOM-1742

The test output looks [like this](https://gist.github.com/mgencur/193f9d0a80d22102c0725b65ea4b2650) . The test suite runs for ~25 mins.

## Proposed Changes

- Refactor kitchensink to split definition and test execution so that we can re-use them in upgrade tests
- Introduce kitchensink upgrade tests which does the following:
-- Create a group of features
-- Shuffle the features
-- Split the shuffled group into sub-groups
-- Perform a series of upgrades and before each upgrade install one of the sub-groups
-- After last upgrade, assert all features through their "Assert" functions, modify all resources and check the modification was successful, delete the feature namespace and check it was deleted and not hanging there

Additional changes:
- Define a series of upgrades in project.yaml, together with their CSV, source and Serving/Eventing custom resource to be used in each step
- Add Make targets to run the kitchensink upgrade tests


maschmid:
Could we perhaps do one more featureset that would contain at least one of each CR and always create this one with the initial serverless version?  (so that we can be sure we cover each kind of resource from the initial version to current?)

mgencur:
@maschmid Maybe it's not necessary. It depend how often we'll run the tests. If we run them periodically every night then it would take up to 3 days to discover issues in the worst case.

mgencur:
The tests are made to be run periodically. If we want to run them with the product during release testing only once then having additional featureset with all CRs and only installing them with the initial version is not enough anyway. We'd need to run the test suite multiple times (with current release minus 3, minus 2, minus 1). Just to make sure each version and each resource in that version is covered.
Maybe I could add a switch for one-off testing - it would make the test install all kinds of resources in each stage (version). I'm just not sure if the cluster will handle that and how long the test will run. WDYT?

maschmid:
OK, let's merge this now then, and we can revisit it later.

/lgtm

PR #1963, opened 2023-02-17
Release 1.27

MayukhEshan:
Fixes JIRA #
Contribute p/z specific test code to midstream repos
Changes regarding removing patches for operator test and installation of serverless operator from the midstream repos


hack/lib/catalogsource.bash
```diff
@@ -71,6 +71,12 @@ function install_catalogsource {
     mv "${rootdir}/_output/bkp.Dockerfile" "${index_build_dir}/Dockerfile"
   fi
 
+# IMAGE_INDEX_BUNDLE decleare in var.bash as default null then work like existing if not null it will take the value from script iib images
+if [ -n "${IMAGE_INDEX_BUNDLE:-}" ]; then
+    echo "Image index is : $IMAGE_INDEX_BUNDLE"
+    index_image="$IMAGE_INDEX_BUNDLE"
+fi
```
mgencur, line 78: Can you replaces the above with this?
```
if [ -n "${INDEX_IMAGE:-}" ]; then
    echo "Index image : $INDEX_IMAGE"
    index_image="$INDEX_IMAGE"
fi
```
and also change the name in vars.bash
Notes: We already use "serverless-index" and "serverless-bundle", these are separate images. The name you have (INDEX_IMAGE_BUNDLE) looks like it combines them together. It's a bit confusing.
The comment starting with "IMAGE_INDEX_BUNDLE decleare" is probably unnecessary, it's really clear from the code what it does.
Nit: The indentation was wrong, it needs to be shifted right.

hack/lib/vars.bash
```diff
@@ -98,3 +98,5 @@ quay.io/openshift-knative/{{- with .Name }}
 EOF
 )
 export IMAGE_TEMPLATE="${IMAGE_TEMPLATE:-"$DEFAULT_IMAGE_TEMPLATE"}"
+export IMAGE_INDEX_BUNDLE="${IMAGE_INDEX_BUNDLE:-}"
+export IMAGE_REGISTRY_NAME="${IMAGE_REGISTRY_NAME:-quay.io}"
```
mgencur, line 102: The IMAGE_REGISTRY_NAME variable is unused. Please remove it from this file.

test/lib.bash
```diff
@@ -113,14 +123,20 @@ function downstream_serving_e2e_tests {
   done
   kubeconfigs_str="$(array.join , "${kubeconfigs[@]}")"
 
+  # check for test flags
+  RUN_FLAGS=("-failfast -timeout=30m -parallel=1")
```
mgencur, line 127: You're changing the default from 60m to 30m here. Please put back 60m

test/lib.bash
```diff
@@ -142,7 +158,13 @@ function downstream_eventing_e2e_tests {
   SYSTEM_NAMESPACE="${SYSTEM_NAMESPACE:-"knative-eventing"}"
   export SYSTEM_NAMESPACE
 
-  go_test_e2e -failfast -timeout=60m -parallel=1 ./test/eventinge2e \
+  # check for test flags
+  RUN_FLAGS=("-failfast -timeout=30m -parallel=1")
```
mgencur, line 162: again: You're changing the default from 60m to 30m here. Please put back 60m

test/lib.bash
```diff
@@ -163,7 +185,13 @@ function downstream_knative_kafka_e2e_tests {
   SYSTEM_NAMESPACE="${SYSTEM_NAMESPACE:-"knative-eventing"}"
   export SYSTEM_NAMESPACE
 
-  go_test_e2e -failfast -timeout=60m -parallel=1 ./test/extensione2e/kafka \
+  # check for test flags
+  RUN_FLAGS=("-failfast -timeout=30m -parallel=1")
```
mgencur, line 189: again: You're changing the default from 60m to 30m here. Please put back 60m

test/lib.bash
```diff
@@ -95,7 +100,12 @@ function serverless_operator_kafka_e2e_tests {
   done
   kubeconfigs_str="$(array.join , "${kubeconfigs[@]}")"
 
-  go_test_e2e -failfast -tags=e2e -timeout=30m -parallel=1 ./test/e2ekafka \
+  RUN_FLAGS=("-failfast -timeout=30m -parallel=1")
```
mgencur, line 103: With this approach, all the test suites now share the same RUN_FLAGS. So, you can't set it differently for individual test suites. I guess this is alright if it works on your side. We don't need to change the flags, we're fine with the defaults.

mgencur: Can you please use just strings instead of arrays in all places? And then use it as "${RUN_FLAGS}" I've put the same comments on the other PR from you team: https://github.com/openshift-knative/eventing/pull/121#discussion_r1107205782

PR #1964, opened 2023-02-20
p/z specific test code for operator  midstream repos

MayukhEshan:
Contribute p/z specific test code to midstream repos
Changes regarding removing patches for operator test and installation of serverless operator from the midstream repos

### DIFF
--- a/test/lib.bash
+++ b/test/lib.bash
@@ -77,7 +77,12 @@ function serverless_operator_e2e_tests {
   done
   kubeconfigs_str="$(array.join , "${kubeconfigs[@]}")"
 
-  go_test_e2e -failfast -tags=e2e -timeout=30m -parallel=1 ./test/e2e \
+  RUN_FLAGS="-failfast -timeout=30m -parallel=1"
+  if [ -n "${OPERATOR_TEST_FLAGS:-}" ]; then
+    RUN_FLAGS="${OPERATOR_TEST_FLAGS}"
+  fi
+
+  go_test_e2e -tags=e2e ${RUN_FLAGS} ./test/e2e \
     --channel "$OLM_CHANNEL" \
     --kubeconfigs "${kubeconfigs_str}" \
     --imagetemplate "${IMAGE_TEMPLATE}" \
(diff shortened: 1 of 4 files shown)


mgencur:
Hmm, as I expected. The shellcheck won't allow unquoted `$RUN_ARGS`. You might really need to use arrays, please check how we do it e.g. [here](https://github.com/openshift-knative/serverless-operator/blob/main/test/lib.bash#L258)

test/lib.bash
```diff
@@ -77,7 +77,12 @@ function serverless_operator_e2e_tests {
   done
   kubeconfigs_str="$(array.join , "${kubeconfigs[@]}")"
 
-  go_test_e2e -failfast -tags=e2e -timeout=30m -parallel=1 ./test/e2e \
+  RUN_FLAGS="-failfast -timeout=30m -parallel=1"
+  if [ -n "${OPERATOR_TEST_FLAGS:-}" ]; then
+    RUN_FLAGS="${OPERATOR_TEST_FLAGS}"
+  fi
```
mgencur, line 83: To pass shellcheck it could be something like this:
```
RUN_FLAGS=(-failfast -timeout=30m -parallel=1)
if [ -n "${OPERATOR_TEST_FLAGS:-}" ]; then
  RUN_FLAGS=("${OPERATOR_TEST_FLAGS}")
fi
...
go_test_e2e -tags=e2e "${RUN_FLAGS[@]}" ...
```

MayukhEshan: yes in previous pr i did something similar. https://github.com/openshift-knative/serverless-operator/pull/1963
but in the eventing repo i saw this changes and then modified the existing ones
https://github.com/openshift-knative/serverless-operator/pull/1963
ok i am changing

mgencur: Please check my updated example above. The shellcheck is very strict about how the params can be passed.

mgencur: AFAIK, in the eventing repo there's no shell check. So, it will allow for doing things that are not allowed by shellcheck.

test/lib.bash
```diff
@@ -100,12 +100,12 @@ function serverless_operator_kafka_e2e_tests {
...
 
-  RUN_FLAGS="-failfast -timeout=30m -parallel=1"
+  RUN_FLAGS=(-failfast -timeout=30m -parallel=1)
   if [ -n "${OPERATOR_TEST_FLAGS:-}" ]; then
-    RUN_FLAGS="${OPERATOR_TEST_FLAGS}"
+    IFS=" " read -r -a RUN_FLAGS <<< "$OPERATOR_TEST_FLAGS"
   fi
 
-  go_test_e2e -tags=e2e ${RUN_FLAGS} ./test/e2ekafka \
+  go_test_e2e -tags=e2e "${RUN_FLAGS}" ./test/e2ekafka \
```
mgencur, line 108: This needs to be `"${RUN_FLAGS[@]}"` . And there are more occurrences in this file

PR #1966, opened 2023-02-21
workaround for SRVKE-1415, move KnativeKafka install after... 

maschmid:
...KnativeEventing is ready

workaround for SRVKE-1415

- :broom: Move KnativeKafka install after KnativeEventing is ready, to make sure it has a complete config-observability


pierDipi (approved):
/lgtm
/approve

PR #1968, opened 2023-02-22
Prometheus to have permissions for endpoinds in user namespace

aliok:
Fixes JIRA https://issues.redhat.com/browse/SRVKE-1408

- Works with https://github.com/knative-sandbox/eventing-kafka-broker/pull/2979

Basically:
- We create a new ClusterRole that gives `get`, `list` and `watch` permissions on some resources that Prometheus needs
- We create a new RoleBinding to that ClusterRole for the user namespace but for the `prometheus-k8s` SA in `openshift-monitoring` namespace
- The ClusterRole is created when KnativeKafka is installed. It is in the manifest.
- The RoleBinding is created using the "additional resources" mechanism in EKB namespaced broker.

These are all according to the fix described in JIRA. However, instead of a Role, I use a ClusterRole for reuse purposes. Then we create RoleBinding in the user namespace for that ClusterRole.

There's one caveat...
In EKB controller, we use Manifestival to create resources that are in "additional resources". When we have the RoleBinding like this in the Manifest:

```
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: knative-prometheus-k8s
roleRef:
  kind: Role
  name: knative-prometheus-k8s
subjects:
- kind: ServiceAccount
  name: prometheus-k8s
  namespace: openshift-monitoring           # --> changed
```

will become:

```
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: knative-prometheus-k8s
roleRef:
  kind: Role
  name: knative-prometheus-k8s
subjects:
- kind: ServiceAccount
  name: prometheus-k8s
  namespace: <user-namespace>        # --> changed
```

This is done here https://github.com/manifestival/manifestival/blob/82196bd303e23c641fda3d002c0663b892092e9c/transform.go#L66-L73 

So, that's why we need https://github.com/knative-sandbox/eventing-kafka-broker/pull/2979

### DIFF
--- a/knative-operator/deploy/resources/knativekafka_extra/broker/eventing-kafka-broker-extra.yaml
+++ b/knative-operator/deploy/resources/knativekafka_extra/broker/eventing-kafka-broker-extra.yaml
@@ -22,6 +22,7 @@ rules:
       - "rbac.authorization.k8s.io"
     resources:
       - clusterrolebindings
+      - rolebindings
     verbs:
       - get
       - list
@@ -38,6 +39,14 @@ rules:
       - list
       - watch
       - update
+  - apiGroups:
+      - ""
+    resources:
+      - endpoints
+    verbs:
+      - get
+      - list
+      - watch
 
 ---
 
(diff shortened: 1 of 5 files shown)


mgencur:
@aliok I guess you want to re-enable the metrics test in this PR? https://github.com/openshift-knative/serverless-operator/blob/main/test/extensione2e/kafka/source_native_kafka_broker_ksvc_test.go#L45

knative-operator/deploy/resources/knativekafka_extra/broker/eventing-kafka-broker-extra.yaml
```diff
@@ -22,6 +22,7 @@ rules:
       - "rbac.authorization.k8s.io"
```
matzew, line 22: I am still confused with `extra` folder vs `extra` manifest, but OK

matzew (approved):
/lgtm
/approve

pierDipi (commented):
/lgtm

PR #1973, opened 2023-02-28
Update metadata for 1.27.1 release

Kaustubh-pande:
Signed-off-by: Kautsubh pande [kpande@redhat.com](mailto:kpande@redhat.com)

cc @matzew @dsimansk @mgencur

/hold
Not before https://github.com/openshift/release/pull/36799

### DIFF
--- a/olm-catalog/serverless-operator/index/Dockerfile
+++ b/olm-catalog/serverless-operator/index/Dockerfile
@@ -8,9 +8,9 @@ COPY --from=opm /bin/opm /bin/opm
 COPY configs /configs
 
 RUN /bin/opm init serverless-operator --default-channel=stable --output yaml >> /configs/index.yaml
-RUN /bin/opm render --skip-tls-verify -o yaml registry.ci.openshift.org/knative/openshift-serverless-v1.26.0:serverless-stop-bundle \
-      registry.ci.openshift.org/knative/openshift-serverless-v1.27.0:serverless-bundle >> /configs/index.yaml || \
-    /bin/opm render --skip-tls-verify -o yaml registry.ci.openshift.org/knative/openshift-serverless-v1.26.0:serverless-stop-bundle \
+RUN /bin/opm render --skip-tls-verify -o yaml registry.ci.openshift.org/knative/openshift-serverless-v1.27.0:serverless-stop-bundle \
+      registry.ci.openshift.org/knative/openshift-serverless-v1.27.1:serverless-bundle >> /configs/index.yaml || \
+    /bin/opm render --skip-tls-verify -o yaml registry.ci.openshift.org/knative/openshift-serverless-v1.27.0:serverless-stop-bundle \
       registry.ci.openshift.org/knative/openshift-serverless-nightly:serverless-bundle >> /configs/index.yaml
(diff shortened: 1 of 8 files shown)


mgencur:
@Kaustubh-pande You need to update files in https://github.com/openshift-knative/serverless-operator/tree/release-1.27/templates and then run `make release-files` to generate the CSV etc. Right now you're directly modifying the generated files and that's why the check fails.

olm-catalog/serverless-operator/index/Dockerfile
```diff
@@ -8,9 +8,9 @@ COPY --from=opm /bin/opm /bin/opm
 COPY configs /configs
 
 RUN /bin/opm init serverless-operator --default-channel=stable --output yaml >> /configs/index.yaml
-RUN /bin/opm render --skip-tls-verify -o yaml registry.ci.openshift.org/knative/openshift-serverless-v1.26.0:serverless-stop-bundle \
+RUN /bin/opm render --skip-tls-verify -o yaml registry.ci.openshift.org/knative/openshift-serverless-v1.27.0:serverless-stop-bundle \
```
mgencur, line 11: Perhaps, when we're changing 1.26.0 to 1.27.0 in this file we should also change 1.27.0 to 1.27.1 on this line: https://github.com/openshift-knative/serverless-operator/blob/2566f7641e913c8a917cc0fb3ed35d9598479a0b/olm-catalog/serverless-operator/index/Dockerfile#L12

mgencur:
In order to fix the Github action failure, I think the github action should be changed in this way:
```
diff --git a/.github/workflows/validate.yaml b/.github/workflows/validate.yaml
index 4485fb458..8c4884b62 100644
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -55,11 +55,12 @@ jobs:
         shell: bash
 
       - name: Set CURRENT_VERSION_IMAGES on push
         if: (github.event_name == 'push' || github.event_name == 'workflow_dispatch') && github.ref_name != 'main'
         run: |
-          branch="${{ github.ref_name }}.0" # example output: "release-1.26.0"
-          # Replace "release-" with "v", so that the final variable is like "v1.27.0"
-          echo "CURRENT_VERSION_IMAGES=${branch/release-/v}" >> $GITHUB_ENV
+          branch="$(yq read olm-catalog/serverless-operator/project.yaml project.version)" # example output: "1.26.0"
+          # Ad "v", so that the final variable is like "v1.27.0"
+          echo "CURRENT_VERSION_IMAGES=v${branch}" >> $GITHUB_ENV
         shell: bash
 
       - name: Set CURRENT_VERSION_IMAGES on PRs
```
Then we'll need this on main branch as well for the future cuts.
/cc @pierDipi 
The current approach doesn't work with patch releases like 1.27.1 because it always uses zero at this line: https://github.com/openshift-knative/serverless-operator/commit/aff24df21d6d38b957f79814b961ae7d76f46ba9#diff-9237aa56078dfbd8fcac3198563f306ada599aabaa43aca884905129779c5846R60

pierDipi:
+1

PR #1986, opened 2023-03-07
Allow increasing Z-stream versions in generated Index Dockerfile

mgencur:
Porting changes from https://github.com/openshift-knative/serverless-operator/pull/1973 to the main branch for the future patch releases. These changes had to be done to allow for producing 1.27.1.

The Index image Dockerfile only allowed for `.0`  as the Z-stream. The CatalogSource  for 1.27.1 was not installing correctly.

## Proposed Changes

-
-
-


pierDipi (approved):
/lgtm

PR #1985, opened 2023-03-06
Select testsuites to be run

mgencur:
Fixes https://issues.redhat.com/browse/SRVCOM-2347

The idea is to select test suites based on our knowledge of this repository and paths within the repo. It is similar to [Prow's triggering jobs based on changes ](https://docs.prow.k8s.io/docs/jobs/#triggering-jobs-based-on-changes). But it is more fine-grained. It doesn't require individual test suites to run in separate Prow checks which would bloat the ci-operator config.

This PR depends on change in openshift-knative/hack: [PR](https://github.com/openshift-knative/hack/pull/27)

## Proposed Changes

- Introduce testsuites.yaml which defines relation between path expressions and test suites to run (these are simple strings)
- Leverage logic from openshift-knative/hack to generate a list of tests which should be run for the given changes
- Create `should_run` function which can check if the given bash function should run, otherwise it returns quickly

Usage:
* Define a test suite like this:
```
- name: "Eventing Kafka"
  run_if_changed:
    - "^knative-operator/pkg/controller/knativekafka/"
  tests:
    - serverless_operator_kafka_e2e_tests
```
* Add `should_run` function inside the bash function which runs the tests:
```
function serverless_operator_kafka_e2e_tests {
  #Note: ${FUNCNAME[0]} resolves to "serverless_operator_kafka_e2e_tests"
  should_run "${FUNCNAME[0]}" || return  
  ...
}
```
* For the example above, if the changes in the given pull request match `^knative-operator/pkg/controller/knativekafka/` then the function `serverless_operator_kafka_e2e_tests will run.`
* More details in the provided testsuite.yaml in this PR

PR #2003, opened 2023-03-16
Use GOPATH for running testselect

mgencur:
Do not rely on PATH having the binary because other repositories might set GOPATH to a different directory and then "go get" will install it t here. The directory might not be on path.

Fixes the problem from https://github.com/openshift-knative/serverless-operator/pull/1985/files#r1137359087

Fixes JIRA #

## Proposed Changes

-
-
-


pierDipi (approved):
/lgtm
/approve

PR #2014, opened 2023-03-22
Drop Kourier image for `IMAGE_kourier-control`

nak3:
This patch drops `IMAGE_kourier-control` which should be covered by `IMAGE_net-kourier-controller__controller`.


ReToCode (approved):
/lgtm

PR #2020, opened 2023-03-27
Revert "Issues/srvke 1290 (#1957)"

matzew:
This reverts commit 3fe2e7c8496d57b20d868bf9d70f13fe87087bd5.

## Proposed Changes

- this revert prevents that IMC are not scalable anymore. The original change for https://github.com/openshift-knative/serverless-operator/pull/1957 did cause this problem: only controller/webhook are scaled, which is not desired.

Proper solution: define `WorkloadOverrides` instead of blindly setting the HA replica


pierDipi (commented):
@matzew there are conflicts.

you don't want to fix the original issue anymore? why?

pierDipi (approved):
/lgtm
/approve

PR #2023, opened 2023-03-29
SRVCOM-2371 TestKsvcWithServiceMeshSidecar fails with subtest may hav…

mgencur:
…e called FailNow on a parent test

Fixes https://issues.redhat.com/browse/SRVCOM-2371

## Proposed Changes

-
-
-


mgencur:
/cherry-pick release-1.28
The bug was hit in INTEROP testing and we will be using 1.28 there for some time so we need the backport to older branch.

PR #2024, opened 2023-03-29
[SRVCOM-2370] Update build Go version to 1.19

dsimansk:
Per SIG release call discussion, bumping Go to 1.19.

@ReToCode @matzew @pierDipi @skonto

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -24,7 +24,7 @@ jobs:
       - name: Setup Golang
         uses: actions/setup-go@v2
         with:
-          go-version: 1.18.x
+          go-version: 1.19.x
 
       - name: Install prerequisites
         env:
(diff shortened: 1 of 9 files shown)


dsimansk:
> Looks good so far, but we also need bumps here:
> 
>     * https://github.com/openshift-knative/serverless-operator/blob/main/go.mod#L3?
> 
>     * https://github.com/openshift-knative/serverless-operator/blob/main/hack/update-deps.sh#L57
> 
>     * https://github.com/openshift-knative/serverless-operator/blob/main/hack/update-deps.sh#L64
> 
> 
> and run `./hack/update-deps.sh --upgrade`

The `go.mod` version should be orthogonal to the CI infra version, e.g. for compatibility reason etc. Therefore we can keep the separate if needed. But probably as a rule of thumb, bump every time it's doable?

Strangely the generate check is failing upon .go code formatting being different. However, I can't see any changes locally on 1.19 or 1.20.

ReToCode:
I'd still align those. That does not hurt and does definitely not cause fancy issues with formatting stuff differently on various go versions. This might even be this? Do you run `make generated-files` on the same go version as the GH-action? Looks like the indentation is different.

dsimansk:
@ReToCode figured out the formatting, for some I needed to rerun `gofmt` locally to get the fixed output. Otherwise it's ready.

skonto:
I thoutht we were building with 1.19.x not bumping, this creates issues on midstream repos that are not on 1.19 eg. 1.9.

PR #2029, opened 2023-03-31
Use golang 1.19 just in docker builds

ReToCode:
As per slack discussion: https://redhat-internal.slack.com/archives/CF5ANN61F/p1680251118091129?thread_ts=1680244739.280449&cid=CF5ANN61F

## Changes
- Partially reverts https://github.com/openshift-knative/serverless-operator/pull/2024
- Sets golang version 1.19 in all places used to build docker images

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -22,9 +22,9 @@ jobs:
       GOPATH: ${{ github.workspace }}
     steps:
       - name: Setup Golang
-        uses: actions/setup-go@v4
+        uses: actions/setup-go@v2
         with:
-          go-version: 1.19.x
+          go-version: 1.18.x
 
       - name: Install prerequisites
         env:
@@ -113,10 +113,10 @@ jobs:
     env:
       GOPATH: ${{ github.workspace }}
     steps:
-      - name: Setup Golang
-        uses: actions/setup-go@v4
+      - name: Set up Go 1.18.x
+        uses: actions/setup-go@v2
         with:
-          go-version: 1.19.x
+          go-version: 1.18.x
 
       - name: Checkout
         uses: actions/checkout@v2
--- a/olm-catalog/serverless-operator/project.yaml
+++ b/olm-catalog/serverless-operator/project.yaml
@@ -21,7 +21,7 @@ requirements:
     # is ignored as it is overridden by fake version via KUBERNETES_MIN_VERSION.
     # This value is used for CSV's min version validation.
     minVersion: 1.21.0
-  golang: '1.18'
+  golang: '1.19'
   nodejs: 16.x
   ocpVersion:
     min: '4.8'
--- a/openshift/ci-operator/build-image/Dockerfile
+++ b/openshift/ci-operator/build-image/Dockerfile
@@ -1,6 +1,6 @@
 # Dockerfile to bootstrap build and test in openshift-ci
 
-FROM registry.ci.openshift.org/openshift/release:golang-1.18
+FROM registry.ci.openshift.org/openshift/release:golang-1.19
 
 # Add kubernetes repository
 ADD openshift/ci-operator/build-image/kubernetes.repo /etc/yum.repos.d/
--- a/pkg/client/clientset/versioned/fake/register.go
+++ b/pkg/client/clientset/versioned/fake/register.go
@@ -23,14 +23,14 @@ var localSchemeBuilder = runtime.SchemeBuilder{
 // AddToScheme adds all types of this clientset into the given scheme. This allows composition
 // of clientsets, like in:
 //
-//	import (
-//	  "k8s.io/client-go/kubernetes"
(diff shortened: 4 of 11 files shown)


dsimansk:
/lgtm
For tests and other reviews
/hold

.github/workflows/validate.yaml
```diff
@@ -22,9 +22,9 @@ jobs:
       GOPATH: ${{ github.workspace }}
     steps:
       - name: Setup Golang
-        uses: actions/setup-go@v4
+        uses: actions/setup-go@v2
         with:
-          go-version: 1.19.x
+          go-version: 1.18.x
```
dsimansk, line 27: Per Kenjiros' comment about Serving repo. It's probably OK to have this actions run on 1.19.

olm-catalog/serverless-operator/project.yaml
```diff
@@ -21,7 +21,7 @@ requirements:
     # is ignored as it is overridden by fake version via KUBERNETES_MIN_VERSION.
     # This value is used for CSV's min version validation.
     minVersion: 1.21.0
-  golang: '1.18'
+  golang: '1.19'
```
dsimansk, line 24: I wonder if the property is useful anywhere though.

openshift/ci-operator/build-image/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # Dockerfile to bootstrap build and test in openshift-ci
 
-FROM registry.ci.openshift.org/openshift/release:golang-1.18
+FROM registry.ci.openshift.org/openshift/release:golang-1.19
```
dsimansk, line 3: I think the build image aka base should be 1.19 as well. Otherwise we'll create a version hell for use slowly. :)

.github/workflows/validate.yaml
ReToCode: I imagine that could increase the confusion a bit more :) Now we use the new version for builds images, everything else stays at 1.18.

olm-catalog/serverless-operator/project.yaml
ReToCode: Yes it is used to specify the verison in `openshift/ci-operator/build-image/Dockerfile`. See https://github.com/openshift-knative/serverless-operator/blob/2157cc9bed970860264b22649a78be4416024ebd/hack/generate/dockerfile.sh#L20 and https://github.com/openshift-knative/serverless-operator/blob/506f33152372b36703304f103894e99231fbfd63/templates/build-image.Dockerfile#L3

pkg/client/clientset/versioned/fake/register.go
```diff
@@ -23,14 +23,14 @@ var localSchemeBuilder = runtime.SchemeBuilder{
 // AddToScheme adds all types of this clientset into the given scheme. This allows composition
 // of clientsets, like in:
 //
-//	import (
-//	  "k8s.io/client-go/kubernetes"
-//	  clientsetscheme "k8s.io/client-go/kubernetes/scheme"
-//	  aggregatorclientsetscheme "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/scheme"
-//	)
+//   import (
```
skonto, line 26: Alignment diffs?

ReToCode: That is the revert of https://github.com/openshift-knative/serverless-operator/pull/2024. And what I meant with, golang versions might have different formatting. https://github.com/openshift-knative/serverless-operator/pull/2024 did format them with 1.19, this reverts it back to the style of 1.18.

olm-catalog/serverless-operator/project.yaml
dsimansk: Thanks, for the pointer that's a interesting coupling. I don't see any other use of the variable. Because we probably want to control all of the Go build versions at once.

dsimansk:
From my comment on the Slack thread:
>My understanding is that go.mod go 1.18 is a min supported version for the module to be buildable on. E.g. declaring this module is using APIs from 1.18, not present on previous Go version. Therefore it needs to be increased only if module needs/wants to consumer newer APIs 1.19+. However, that should not prevent local or CI builds from using newer Go version to have CVEs fixes and improvements, that doesn't break API contract with our module and 1.18. Hence what we currently have in the upstream. All the infra Prow or GH actions are on Go 1.20.

I think only hack scripts and `go.mod` should be reverted to the previous state. Otherwise everything else is infra and can be 1.19.

olm-catalog/serverless-operator/project.yaml
ReToCode: Yeah not ideal. It's quite confusing at how may places we have to update the version :)

ReToCode:
> I think only hack scripts and go.mod should be reverted to the previous state. Otherwise everything else is infra and can be 1.19.

I am going to merge this to get CI back working again. Lets discuss in the next SIG Release meeting then decide and update the repos accordingly.

/unhold

PR #2034, opened 2023-04-04
Index image includes current bundle and two versions back

mgencur:
Fixes https://issues.redhat.com/browse/SRVCOM-2378

## Proposed Changes

- index.yaml now includes the latest version and two versions back (e.g. 1.29, 1.28, 1.27) so that we can cover the time period when the latest live release is 1.27 and on main branch we already have 1.29. The CatalogSource we're creating will have this index.yaml and will be able to upgrade from 1.27 (redhat-operators catalogsource) to 1.28 (serverless-operator catalogsource)
- Do not use stop-bundle at all. The metadata with information about bundles and their replacements is in index.yaml. We can use a regular bundle in place of the stop-bundle.
 
I will follow up with a PR to openshift/release to remove [building the stop-bundle](https://github.com/openshift/release/blob/master/ci-operator/config/openshift-knative/serverless-operator/openshift-knative-serverless-operator-main__4.12.yaml#L93-L96) and then in serverless-operator to remove [Dockerfile for stop-bundle](https://github.com/openshift-knative/serverless-operator/blob/main/olm-catalog/serverless-operator/stopbundle.Dockerfile), it won't be required anymore.


olm-catalog/serverless-operator/project.yaml
```diff
@@ -14,6 +14,9 @@ olm:
     default: 'stable'
     list:
       - 'stable'
+  previous:
+    replaces: 1.27.0
+    skipRange: '>=1.27.0 <1.28.0'
```
pierDipi, line 19: I'm assuming that this needs to be bumped whenever project.version is bumped? Should we increase the comment on project.version?

We can follow up on this, no reason to block this PR

mgencur: Yes. It needs to be bumped as you said. Will add the comment in a follow-up.

PR #2036, opened 2023-04-05
Add eventing-istio to test image

pierDipi:
As per title.

This is the first step to run the eventing-istio tests in the mesh test job


pierDipi:
Fixed a typo `isto`

PR #2042, opened 2023-04-11
Minimal ServiceEntry for Kafka cluster

pierDipi:
Reduce the ServiceEntry hosts configuration to the minimum.


hack/lib/mesh_resources/kafka-service-entry.yaml
```diff
@@ -5,9 +5,6 @@ metadata:
 spec:
   hosts:
     - my-cluster-kafka-bootstrap.kafka
```
matzew, line 7: so we end up w/ less resources?  
IMO generally a good pattern to use only one format of URLs for the same host

matzew (approved):
/lgtm
/approve

PR #2051, opened 2023-04-20
feat: adds tekton resources for building functions

lance:
These resources should be installed in the cluster when Serverless is installed.


dsimansk:
Does it require OpenShift Pipelines/Tekton to be installed on the cluster already? Ohterwise adding unrecognized custom resource.

Moreover we should probably create a templated files in `templates` dir and at least make image params replaced by the values from `project.yaml`.

lance:
> Does it require OpenShift Pipelines/Tekton to be installed on the cluster already? Ohterwise adding unrecognized custom resource.

I was under the impression that pipelines are always installed on an OpenShift cluster, but I think that is wrong. I need to check to see if it is installed.

openshift-knative-operator/cmd/operator/kodata/knative-functions/1-tekton-task.yaml
cardil, line 1: Those tasks should be deployed dynamically, same as we deploy `ConsoleCliDownload` CR only if console is present. See: https://github.com/openshift-knative/serverless-operator/pull/1787

cardil (changes requested):

lance:
@dsimansk @cardil thanks for your feedback. I've made some changes based on your suggestions and would appreciate any additional feedback. This is my first time working with this repository, and it is a bit of black magic to me. Please point me in the right direction if I've gone astray here.

knative-operator/cmd/manager/main.go
```diff
@@ -187,3 +195,49 @@ func setupServerlesOperatorMonitoring(cfg *rest.Config) error {
...
+
+// SetupFunctionTasksAndPipelines checks to see if Tekton is installed
+// and if so creates the necessary resources for on-cluster function builds
+func SetupFunctionTasksAndPipelines(cfg *rest.Config) error {
+	cl, err := client.New(cfg, client.Options{})
+	if err != nil {
+		return fmt.Errorf("failed to create a client: %w", err)
+	}
+
+	// Check to see if Tekton is installed by looking for the pipelines CRD
```
skonto, line 207: I suggest we have a common logic for OCP feature discovery.
We check the cluster operator status [here](https://github.com/openshift-knative/serverless-operator/blob/main/knative-operator/cmd/manager/main.go#L176-L187) for console for example.

lance:
Closing this. After spending some time chatting with @cardil I think it may not make sense to install these resources via the operator, but instead have them be user-installed via artifacthub.io.

PR #2052, opened 2023-04-24
Wait for Serving config propagation

skonto:
## Proposed Changes

- Similar to https://github.com/openshift-knative/serving/pull/273. See discussion [here](https://redhat-internal.slack.com/archives/CF5ANN61F/p1682322827453259).
- Go 1.19 seems to allow for faster [test execution](https://redhat-internal.slack.com/archives/CF5ANN61F/p1680757892190399?thread_ts=1680614138.940309&cid=CF5ANN61F) and so config is not ready when tests are run.
- We should have this anyway for consistency reasons.

### DIFF
--- a/test/serving.bash
+++ b/test/serving.bash
@@ -163,19 +163,15 @@ function upstream_knative_serving_e2e_and_conformance_tests {
 }
 
 function enable_feature_flags {
-  local failed=0
-
   for feature in "$@"; do
     echo "Enabling feature: $feature"
-    configure_cm features "$feature":enabled || failed=1
+    configure_cm features "$feature":enabled
   done
   # Allow settings to be picked up
   sleep 30
-  return $failed
 }
 
 function configure_cm {
-  local failed=0
   local cm="$1"
   local patch=""
   declare -A json_properties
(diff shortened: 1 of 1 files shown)


test/serving.bash
```diff
@@ -184,3 +161,38 @@ function upstream_knative_serving_e2e_and_conformance_tests {
...
+function enable_feature_flags {
+  local failed=0
+
+  for feature in "$@"; do
+    echo "Enabling feature: $feature"
+    configure_cm features "$feature":enabled || failed=1
+  done
+  # Allow settings to be picked up
+  sleep 30
+  return $failed
```
skonto, line 174: Midstream we fail tests if anything returns !=0, do we need this here too?

mgencur: No because we already have it. We define flags such as `set -e` which make the tests fail fast. And we don't want to gather the additional state which is done in "fail_test" as we run serverless must gather ourselves.

And for the same reason, we should probably remove all occurrences of `|| failed=1` from this PR. The `failed` flag is redundant here.

mgencur (commented):
Hey @skonto , except for using the `failed` flag it's looking ok. Please see my commen in the code.

test/serving.bash
```diff
@@ -184,3 +161,38 @@ function upstream_knative_serving_e2e_and_conformance_tests {
...
+  for property in "${@:2}"; do
+    KEY="${property%%:*}"
+    VALUE="${property##*:}"
+    patch=${patch:+$patch,}"\"$KEY\": \"$VALUE\""
+    # escape in case property contains dots eg. kubernetes.pod-spec
+    j_property="$(echo "'$KEY'" | sed "s/\./\\\./g")"
+    json_properties["$j_property"]="$VALUE"
+  done
+
+  oc -n ${SYSTEM_NAMESPACE} patch knativeserving/knative-serving --type=merge --patch="{\"spec\": {\"config\": { \"$cm\": {$patch} }}}" || failed=1
```
mgencur: We use "${SERVING_NAMESPACE}" in this repo. That's also the reason why the tests failed.

mgencur:
/lgtm
/hold
Hold for tests.

PR #2064, opened 2023-05-09
Rename channel name as per recommendation

Kaustubh-pande:
As per the Slack discussion, we agree to use Channel naming as stable, stable-1.29


dsimansk:
/approve
/lgtm

For ACK from all stakeholders.
/hold

dsimansk:
/approve
/lgtm

For ACK from all stakeholders.
/hold

PR #2073, opened 2023-05-15
Update S-O version to next release cycle

dsimansk:
Per title. Sorry for the delay. 

/cc @pierDipi @matzew @ReToCode @skonto

PR #2076, opened 2023-05-15
Introduce bump SO automation

### LINKED ISSUE #2074: Update Serverless Operator version with a tool
Looking at PRs, like:
* https://github.com/openshift-knative/serverless-operator/pull/2073
* https://github.com/openshift-knative/serverless-operator/pull/1997

We should have some proper automation for the version bump.

The actual flow is to adjust the version on the `project,yaml` file (ideally do a commit), and run `make generated-files` (and do a separate commit here too).

However, some changes on the `project.yaml` bump can be derived from the actual bump. Ideally we should just to:

```
  version: 1.CURRENT+1.0
```

other updates like `replaces`, `skipRange`, or `csv / source`, are following a predictable schema required by the actual version bump.

pierDipi:
Fixes #2074

This PR is introducing an automation that allows
bumping SO automatically. 

Such bump is often error prone and time consuming,
in addition to being something to learn as more people
join us in the release coordination.

The automation is automatically triggered when a new
branch is created (similar to https://github.com/openshift-knative/serverless-operator/pull/2061) with an option of 
manually triggering the workflow in case we ever need to
re-trigger it for  whatever reason.

This is the diff I'm getting by running this tool on top of the
current PR:

```diff
diff --git a/olm-catalog/serverless-operator/project.yaml b/olm-catalog/serverless-operator/project.yaml
index dd67fddcc..4ce5427d1 100644
--- a/olm-catalog/serverless-operator/project.yaml
+++ b/olm-catalog/serverless-operator/project.yaml
@@ -4,18 +4,19 @@ project:
     # all components in `dependencies.previous` to the same versions as `dependencies` in the same PR.
     # Otherwise, the upgrade tests will not pass, as we have a different SO version with the same bundle contents.
     # Also make sure to update values under `olm.previous` by copying from `olm.replaces` and `olm.skipRange`.
-    version: 1.29.0
+    version: 1.30.0
 olm:
-    replaces: 1.28.0
-    skipRange: '>=1.28.0 <1.29.0'
+    replaces: 1.29.0
+    skipRange: '>=1.29.0 <1.30.0'
     channels:
         default: 'stable'
         list:
-            - 'stable'
-            - 'stable-1.29'
+            - stable
+            - stable-1.29
+            - stable-1.30
     previous:
-        replaces: 1.27.0
-        skipRange: '>=1.27.0 <1.28.0'
+        replaces: 1.28.0
+        skipRange: '>=1.28.0 <1.29.0'
 requirements:
     kube:
         # The min version validation in `vendor/knative.dev/pkg/version/version.go`
@@ -55,9 +56,9 @@ dependencies:
     operator: 1.6.0
     # Previous versions required for downgrade testing
     previous:
-        serving: 1.7.0
-        eventing: 1.7
-        eventing_kafka_broker: 1.7
+        serving: knative-v1.8
+        eventing: knative-v1.8
+        eventing_kafka_broker: knative-v1.8
 upgrade_sequence:
     - csv: serverless-operator.v1.26.0
       source: redhat-operators
@@ -67,3 +68,5 @@ upgrade_sequence:
       source: redhat-operators
     - csv: serverless-operator.v1.29.0
       source: serverless-operator
+    - csv: serverless-operator.v1.30.0
+      source: serverless-operator
```

Which is similar to: https://github.com/openshift-knative/serverless-operator/pull/2073

### DIFF
--- a/.github/workflows/bump.yaml
+++ b/.github/workflows/bump.yaml
@@ -39,9 +39,9 @@ jobs:
         uses: peter-evans/create-pull-request@v4
         with:
           path: ./src/github.com/${{ github.repository }}
-          branch: auto/bump-so-version-${{ env.SO_VERSION }}
-          title: "[${{ github.ref_name }}] Bump SO version to version ${{ env.SO_VERSION }}"
-          commit-message: "Bump SO version to version ${{ env.SO_VERSION }}"
+          branch: auto/bump-so-version-${{ github.ref_name }}
+          title: "[${{ github.ref_name }}] Bump Serverless Operator version"
+          commit-message: "Bump SO version"
           delete-branch: true
           body: |
             Bump SO version + run `make generated-files`
--- a/.github/workflows/bump.yaml
+++ b/.github/workflows/bump.yaml
@@ -1,0 +1,47 @@
+---
+name: Bump
+
+on:
+  push:
+    branches:
+      - '**'
+  workflow_dispatch: # Manual workflow trigger
+
+jobs:
+  bump-so-version:
+    name:
+    runs-on: ubuntu-latest
+    env:
+      GOPATH: ${{ github.workspace }}
+    steps:
+      - name: Setup Golang
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.20.x
+
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          # Always checkout main, we don't support bumping for patch releases (for now)
+          ref: main
+          path: ./src/github.com/${{ github.repository }}
+          fetch-depth: 0
+
+      - name: Bump SO
+        working-directory: ./src/github.com/${{ github.repository }}
+        run: go run hack/cmd/bumpso/bumpso.go
+
+      - name: Regenerate all generated files
+        working-directory: ./src/github.com/${{ github.repository }}
+        run: make generated-files
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
+        with:
+          path: ./src/github.com/${{ github.repository }}
+          branch: auto/bump-so-version-${{ env.SO_VERSION }}
+          title: "[${{ github.ref_name }}] Bump SO version to version ${{ env.SO_VERSION }}"
+          commit-message: "Bump SO version to version ${{ env.SO_VERSION }}"
+          delete-branch: true
+          body: |
+            Bump SO version + run `make generated-files`
--- a/hack/cmd/bumpso/bumpso.go
+++ b/hack/cmd/bumpso/bumpso.go
@@ -1,0 +1,173 @@
+package main
+
+import (
+	"bytes"
+	"flag"
+	"fmt"
+	"log"
+	"os"
+	"path/filepath"
+
+	"github.com/coreos/go-semver/semver"
+	"gopkg.in/yaml.v3"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func main() {
+	if err := run(); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func run() error {
+	project := make(map[string]interface{}, 8)
+	wd, err := os.Getwd()
+	if err != nil {
+		log.Fatal(err)
+	}
+	projectPath := filepath.Join(wd, "olm-catalog/serverless-operator/project.yaml")
+
+	flag.StringVar(&projectPath, "project-path", projectPath, "")
+	flag.Parse()
+
+	var node yaml.Node
+
+	file, err := os.ReadFile(projectPath)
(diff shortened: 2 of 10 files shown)


olm-catalog/serverless-operator/project.yaml
```diff
@@ -1,78 +1,69 @@
----
 project:
-  name: serverless-operator
-
-  # When bumping the Operator to a new version (major and minor), make sure to also update
-  # all components in `dependencies.previous` to the same versions as `dependencies` in the same PR.
-  # Otherwise, the upgrade tests will not pass, as we have a different SO version with the same bundle contents.
-  # Also make sure to update values under `olm.previous` by copying from `olm.replaces` and `olm.skipRange`.
-  version: 1.29.0
-
```
matzew, line 10: should we add a linter that nobody adds empty lines etc?

matzew:
```
-        serving: 1.7.0
-        eventing: 1.7
-        eventing_kafka_broker: 1.7
+        serving: knative-v1.8
+        eventing: knative-v1.8
+        eventing_kafka_broker: knative-v1.8
```
is this really the expected format for the `previous`?

Looking at older PRs, it seems that the `knative-v` is not used here.

pierDipi:
> ```
> -        serving: 1.7.0
> -        eventing: 1.7
> -        eventing_kafka_broker: 1.7
> +        serving: knative-v1.8
> +        eventing: knative-v1.8
> +        eventing_kafka_broker: knative-v1.8
> ```
> 
> is this really the expected format for the `previous`?
> 
> Looking at older PRs, it seems that the `knative-v` is not used here.

Yes, it should be fine

olm-catalog/serverless-operator/project.yaml
pierDipi: Seems like a high-effort / low impact improvement, so maybe later ?

.github/workflows/bump.yaml
```diff
@@ -0,0 +1,47 @@
...
+  bump-so-version:
+    name:
+    runs-on: ubuntu-latest
+    env:
+      GOPATH: ${{ github.workspace }}
+    steps:
+      - name: Setup Golang
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.20.x
```
dsimansk, line 20: I'd go with `stable` to have it always on a floating version to avoid bumps in the future.  
```suggestion
          go-version: stable
```

.github/workflows/bump.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+      - name: Regenerate all generated files
+        working-directory: ./src/github.com/${{ github.repository }}
+        run: make generated-files
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
+        with:
+          path: ./src/github.com/${{ github.repository }}
+          branch: auto/bump-so-version-${{ env.SO_VERSION }}
```
dsimansk, line 42: Is the `$SO_VERSION` exported somewhere?

hack/cmd/bumpso/bumpso.go
```diff
@@ -0,0 +1,174 @@
...
+	currentVersion.Patch = 0
+
+	newVersion := &semver.Version{
+		Major:      currentVersion.Major,
+		Minor:      currentVersion.Minor,
+		Patch:      currentVersion.Patch,
+		PreRelease: currentVersion.PreRelease,
+		Metadata:   currentVersion.Metadata,
+	}
+	newVersion.BumpMinor()
```
dsimansk, line 66: I wonder if should detect something like `CURRENT_BRANCH==release-*` and bump patch version. But it's kind of niche use case with usual cadence. :)

.github/workflows/bump.yaml
matzew: stable seems a bit vague? I do like explicit better

ReToCode (commented):
LGTM. Thank you for this!

.github/workflows/bump.yaml
pierDipi: I removed this usage, I believe there is no easy way to get the branch that triggered the pull request / to which version we're bumping at this stage of the workflow, so we will use the target branch

matzew (approved):
/lgtm
/approve

/hold
in case there is things people really wanted to get addressed, but I am ok w/ testing this rather sooner, than later

hack/cmd/bumpso/bumpso.go
pierDipi: Yes, we can iterate, I felt like spending time on patch releases support was not worth at the initial iteration

.github/workflows/bump.yaml
pierDipi: I like explicit versions too

PR #2077, opened 2023-05-16
Install YQ in bump SO automation

pierDipi:
It failed on https://github.com/openshift-knative/serverless-operator/actions/runs/4988589665/jobs/8931480635


.github/workflows/bump.yaml
```diff
@@ -27,6 +27,14 @@ jobs:
           path: ./src/github.com/${{ github.repository }}
           fetch-depth: 0
 
+      - name: Install prerequisites
+        env:
+          YQ_VERSION: 3.4.0
```
matzew, line 32: ZOMG!

PR #2104, opened 2023-05-31
[WIP] Remove skip for more channel-based components

pierDipi:
Good catch @mgencur

PR #2069, opened 2023-05-10
SRVKE-1446: Add eventing-istio-controller for channel-based mTLS

pierDipi:
- Add eventing-istio-controller
- scale down controller when feature flag is disabled and up when enabled

PR #2107, opened 2023-06-01
Add a test for Broker Filter

mgencur:
Fixes https://issues.redhat.com/browse/SRVCOM-1572

The combination of latest REKT tests for Eventing and Eventing Kafka Broker and one more test from this PR will cover all the images except for client-plugin.

## Proposed Changes

-
-
-


maschmid:
Should the `t.Skip("Channel-based tests cannot run in service mesh mode for now")` skip from the test above  be also present in this test? (This test should also be using a channel-based MTbroker?)

pierDipi:
> Should the `t.Skip("Channel-based tests cannot run in service mesh mode for now")` skip from the test above be also present in this test? (This test should also be using a channel-based MTbroker?)

We're removing them https://github.com/openshift-knative/serverless-operator/pull/2104 (fixed by #2069)

PR #2118, opened 2023-06-12
Bump operator CRD to v1.9.0

nak3:
As `olm-catalog/serverless-operator/project.yaml` still sets `v1.6.0` for operator, downstream cannot use some API update of upstream.
This patch bumps the CRD.

PR #2120, opened 2023-06-12
Add `--skip-cleanup-on-fail` to common flag

nak3:
## Proposed Changes

:broom: Add `--skip-cleanup-on-fail` to common flag.
As per title, it is hard to debug without this flag because failed resources are cleaned up.


nak3:
/hold

I will confirm this with https://github.com/openshift-knative/serverless-operator/pull/2118 if this flag leave the failed resources.

PR #2140, opened 2023-07-07
Use APIs from eventing kafka broker in tests

mgencur:
Related to https://issues.redhat.com/browse/SRVKE-1367

## Proposed Changes

- references APIs from eventing-kafka-broker instead of eventing-kafka
- remove dependency on eventing-kafka
-


go.mod
creydr, line 1: Can we remove the replace in line 185 too?

creydr (commented):
Only one question. Otherwise looks good from my side

go.mod
mgencur: Sure. Pushing.

creydr (approved):
/lgtm

PR #2141, opened 2023-07-10
Prevent scaling cluster for single-node openshift

mgencur:
The machineconfigpool is available there but we don't want to scale it up. We want to use the exact cluster that was provided in CI.

Fixes JIRA #

## Proposed Changes

-
-
-

### DIFF
--- a/hack/lib/scaleup.bash
+++ b/hack/lib/scaleup.bash
@@ -68,6 +68,11 @@ function cluster_scalable {
   if ! oc get machineconfigpool &>/dev/null; then
     return 1
   fi
+  # Prevent scaling for single-node OpenShift.
+  if [[ $(oc get machineconfigpool master -ojsonpath='{.status.machineCount}') == 1 && \
+        $(oc get machineconfigpool worker -ojsonpath='{.status.machineCount}') == 0 ]]; then
+    return 1
+  fi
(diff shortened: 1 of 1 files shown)


hack/lib/scaleup.bash
```diff
@@ -68,6 +68,11 @@ function cluster_scalable {
   if ! oc get machineconfigpool &>/dev/null; then
     return 1
   fi
+  # Prevent scaling for single-node OpenShift.
+  if [[ $(oc get machineconfigpool master -ojsonpath='{.status.machineCount}') == 1 && \
+        $(oc get machineconfigpool worker -ojsonpath='{.status.machineCount}') == 0 ]]; then
+    return 1
+  fi
```
cardil, line 75: There is no way of checking if we are on Single-node Openshift, or Microshift? I mean other than checking if there's only one node?

mgencur: OK. According to https://docs.openshift.com/container-platform/4.13/operators/operator_sdk/osdk-ha-sno.html, there is a way. I will update the PR.

cardil: If no, maybe some warn log message be nice, instead of that code comment.

mgencur: Pushed.

cardil (approved):
/lgtm

PR #2143, opened 2023-07-10
Document updated release process

pierDipi:
This is documenting the new more automated release flow


pierDipi:
We will use an issue template for this as checklist

PR #2151, opened 2023-07-12
Add dependabot and its update deps

nak3:
This patch adds dependabot and its update deps.

Currently dependabot bumps lib only when it has a security issue like https://github.com/openshift-knative/serverless-operator/pull/2150.
Also, it does not run update-deps.sh.

`.github/dependabot.yml` adds the dependabot and `.github/workflows/dependabot-deps.yaml` runs `update-deps.sh` when it was triggered.

Note, k8s.io and knative.dev libs need to be consistent with serving/eventing so they are excluded.

### DIFF
--- a/.github/workflows/dependabot-deps.yaml
+++ b/.github/workflows/dependabot-deps.yaml
@@ -19,6 +19,14 @@ jobs:
           path: ./src/github.com/${{ github.repository }}
           fetch-depth: 0
 
+      - name: Install prerequisites
+        env:
+          YQ_VERSION: 3.4.0
+        run: |
+          sudo wget https://github.com/mikefarah/yq/releases/download/${YQ_VERSION}/yq_linux_amd64 -O /usr/bin/yq
+          sudo chmod +x /usr/bin/yq
+          sudo mv /usr/bin/yq /usr/local/bin/yq
+
       - name: Run ./hack/update-deps.sh
         working-directory: ./src/github.com/${{ github.repository }}
         run: ./hack/update-deps.sh
--- a/.github/workflows/dependabot-deps.yaml
+++ b/.github/workflows/dependabot-deps.yaml
@@ -1,0 +1,36 @@
+name: depndabot - update-deps.sh
+
+on:
+  pull_request:
+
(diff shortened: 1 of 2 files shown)


.github/workflows/dependabot-deps.yaml
pierDipi, line 1: Have you tried this action and how it works with existing GH actions checks?

When we were something like this, we ended up with 2 actions (https://github.com/knative-sandbox/eventing-kafka-broker/commit/4886368cd254f6817a776032177446f15957be71) https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/.github/workflows/build-dependabot-pr.yaml and https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/.github/workflows/update-dependabot-pr.yaml but this approach seems simpler so if it works that's even better!

.github/workflows/dependabot-deps.yaml
```diff
@@ -0,0 +1,36 @@
...
+        working-directory: ./src/github.com/${{ github.repository }}
+        run: |
+          if ! git diff --exit-code --quiet
+          then
+            git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
+            git config --local user.name "github-actions[bot]"
+            git add .
+            git commit -m "Run ./hack/update-deps.sh"
+            git push
+          fi
```
pierDipi, line 36: I'm surprised there is no "setup go" here

.github/workflows/dependabot-deps.yaml
nak3: Yes I have added same action in upstream https://github.com/knative-sandbox/net-gateway-api/pull/498 and here is an example https://github.com/knative-sandbox/net-gateway-api/pull/502

nak3: Ah, or you wanted me to verify this on serverless? I will verify it on my repo.

nak3: Thanks. `yq` was necessary. The updated action works fine like https://github.com/nak3/serverless-operator/pull/69

PR #2164, opened 2023-07-14
Drop auto bump for the push event

nak3:
This patch drops auto bump for the `push` event.

Currently the auto-PR is created when `push` (not `pull-request`) event happens. 
We usually check the diff on `Check if everything is consistent` so the auto-PR should not be necessary for push event.

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -80,10 +80,9 @@ jobs:
       #   Since we're generating files based on a floating branch in midstream,
       #   we need to reconcile those files periodically.
       #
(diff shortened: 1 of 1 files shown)


nak3:
@pierDipi As per this PR, can we drop the auto PR for `push` event?
This will solve the current dependabot issue like [this](https://github.com/openshift-knative/serverless-operator/pull/2156) . Also, the auto PR  makes noisy PR on our repos. E.g. your repo has a lot of PRs by this auto PR https://github.com/pierDipi/serverless-operator/pulls ...

pierDipi (approved):
/lgtm
/approve

PR #2172, opened 2023-07-18
Do not pin to some libs

nak3:
There are some libs that are pinned to specific version.
Some of them are not necessary to be pinned so this patch un-pin them.


nak3:
/hold

I will fix CI error.

nak3:
The unit test error is fixed by https://github.com/manifestival/manifestival/pull/94 but maifestival haven't cut the new branch after the fix so we haven't got the fix now yet.

nak3:
/cc @ReToCode @skonto @pierDipi 

This is ready now. Could you please take a look? :pray:

ReToCode (commented):
LGTM in general. Waiting for other reviews.

Just to bring the question up, with dependabot in place, could it be that we get newer dependencies than upstream has? This could in some cases probably result in unexpected behaviour (I think logger and such are no issue, but manifestal or controller-runtime maybe?).

nak3:
> Just to bring the question up, with dependabot in place, could it be that we get newer dependencies than upstream has? This could in some cases probably result in unexpected behaviour (I think logger and such are no issue, but manifestal or controller-runtime maybe?).

Yes, it could be. But the critical dependencies should be [`"k8s.io/*"` and `"knative.dev/*"` which are ignored](https://github.com/openshift-knative/serverless-operator/blob/main/.github/dependabot.yml#L11-L13 ).
So for other deps like `manifestal` or `controller-runtime`, I am thinking that our CI/Build can detect any issues if they have. Otherwise, bumping the dependencies libs are super hard as we need to check upstream's lib version (all upstream's repo - serving, eventing or operstor etc(?)) and then stick the version one by one.

pierDipi (approved):
This LGTM, thanks for doing this!

/lgtm

PR #2174, opened 2023-07-19
Disable internal encryption for Serverless 1.29 and older

mgencur:
The name of the Secret used by internal encryption changed in Serving 1.9 (Serverless 1.30). Upgrades from older versions than 1.30 will fail.

Fixes https://issues.redhat.com/browse/SRVCOM-2605

## Proposed Changes

- Disable internal encryption for Serverless 1.29 and older
- Do not explicitly enable encryption for older releases, it's part of the default config for  tests.

### DIFF
--- a/hack/lib/serverless.bash
+++ b/hack/lib/serverless.bash
@@ -34,16 +34,19 @@ function ensure_serverless_installed {
 
   deploy_serverless_operator "$csv"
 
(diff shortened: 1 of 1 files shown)


mgencur:
/hold
For the kitchensink upgrade tests to pass

PR #2175, opened 2023-07-19
Bump Serving components to release-v1.10

ReToCode:
## Proposed Changes
- Bump Serving components to release-v1.10

/cc @nak3 , @skonto , @pierDipi


ReToCode:
@mgencur you did a fix for the internal-encryption for <1.29. The tests here fail with (I think) the same thing. The secret name changes (also?) from 1.9 -> 1.10. 

I need to dig a bit why this fails, because the new secret is actually created and should be reconciled, but is somehow not:

```
Reconciling Secret: serving-certs at namespace: serving-tests"
type: 'Warning' reason: 'InternalError' public cert in the secret is not ready yet
error":"failed to create Secret serving-certs/serving-tests: secrets \"serving-certs\" already exists
```

Secret name in 1.9 is `ns.Name + "-" + networking.ServingCertName`
And in 1.10 just `networking.ServingCertName`
The label key should be `serving-certs-ctrl` in both versions which is also used as the secretInformer filter.

ReToCode:
Found it, the CA is updated during the process, but the CA field in activators secret does not update. David has fixed this upstream, so back-port [here](https://github.com/openshift-knative/serving/pull/390).

PR #2189, opened 2023-07-25
Pin to OCP dependencies

nak3:
Currently OCP version's deps are not pinned so the version is undetermined.

We should pin to the OCP version like `k8s.io` lib, otherwise we may hit a version incompatible issues. (We are running min/max OCP version so it should be discovered, though.)

Note, I learned that other OCP family is doing this - https://github.com/operator-framework/operator-lifecycle-manager/blob/master/go.mod#L239-L243


ReToCode (approved):
/lgtm
/approve

Should we probably work on getting more recent versions of this in a separate PR? They seem quite dated?

nak3:
> Should we probably work on getting more recent versions of this in a separate PR? They seem quite dated?

The "dated" means `github.com/openshift/client-go` has `20211209144617` (2021 Dec) for example, right?

```
github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20211209144617-7385dd6338e3
```

I aligned it with 4.10 which is a min OCP version so its last commit is in 2021/Dec - https://github.com/openshift/client-go/tree/release-4.10

I think we can use OCP's max version (4.13) if they have backward compatibility but k8s.io libs also aligned with the min k8s (not max) version so maybe it is alright to use 4.10.

ReToCode:
Thanks for the explanation, with knowing that, I definitely agree with:
>  so maybe it is alright to use 4.10.

PR #2190, opened 2023-07-25
:broom: Use Golan'g atomic.Bool

nak3:
Since Go 1.19, atomic.Bool is supported as https://pkg.go.dev/sync/atomic@master#Bool 

Hence this patch replaces the third-party lib with Golang's atomic.


nak3:
Updating Go to 1.19 via https://github.com/openshift/release/pull/41618

PR #2193, opened 2023-07-26
Bump github.com/operator-framework/operator-lifecycle-manager

nak3:
This patch replaces https://github.com/openshift-knative/serverless-operator/pull/2159

I just bumps `github.com/operator-framework/api` then `github.com/operator-framework/operator-lifecycle-manager` instead of bumping `github.com/operator-framework/operator-lifecycle-manager` only.


nak3:
/cc @ReToCode @pierDipi @skonto @matzew 

This should be the last one for current update :tada:

ReToCode:
Thank you!

I don't yet get how operator-framework/olm is now stable. Shouldn't one of these deps be pinned to avoid our "runtime" to upgrade without intention?

nak3:
Umm... It is okay to pin to these deps (or one of them) version, but the `github.com/operator-framework/*` deps are used only by test code in `./test/*`. Should we pin them?

ReToCode:
> Umm... It is okay to pin to these deps (or one of them) version, but the github.com/operator-framework/* deps are used only by test code in ./test/*. Should we pin them?

That is interesting. I thought, we would use OLM also in runtime. If it is just for testing, I think we are fine with not pinning them. So LGTM.

Maybe someone else can also point out his opinion on this? @skonto , @pierDipi , @matzew

pierDipi:
In general, I'm +1 on not pinning dependencies unless there is a real reason

nak3:
Okay, I think we don't need to pin. If there is any issue happens in the feature, we should do that.

nak3:
Could anyone merge this :pray: ?

ReToCode (approved):
/lgtm
/approve

PR #2196, opened 2023-07-26
Make Serving service mesh tests more stable

ReToCode:
Fixes JIRA https://issues.redhat.com/browse/SRVCOM-2356
Partially fixes JIRA https://issues.redhat.com/browse/SRVKS-1099
Partially fixes JIRA https://issues.redhat.com/browse/SRVCOM-2522

## Proposed Changes
We are trying to do https://github.com/openshift-knative/serverless-operator/pull/2171 in parts now. This PR contains the fixes found in https://issues.redhat.com/browse/SRVKS-1099 to make Serving service-mesh tests more stable

Changes:
- Reduce parallelism of mesh tests to 8 because istio-proxies are to slow otherwise
- Use a specific label for the Knative Gateways to omit issues with conflicting gateways (SRVCOM-2356)
- Drop `TERMINATION_DRAIN_DURATION_SECONDS` from metadata-webhook, as this no longer works
- Add `terminationDrainDuration: 35s` to SMCP to make envoy wait longer with termination

/cc @mgencur , @pierDipi

PR #2197, opened 2023-07-26
Increase rekt tests timeouts

pierDipi:
Good catch @mgencur https://github.com/openshift-knative/serverless-operator/pull/2196#issuecomment-1651760074


pierDipi:
This is much better now

/unhold

PR #2209, opened 2023-08-07
Use kube-rbac-proxy with the min supported version

nak3:
As per title, kube-rbac-proxy is still 4.7 on CI. This patch updates it to use kube-rbac-proxy with the min supported version.


ReToCode (approved):
/lgtm
/approve

PR #2210, opened 2023-08-07
Disable cgroups tests on OCP 4.14+

mgencur:
The test is directly accessing files on filesystem and they don't exist for cgroups v2.

Related to https://issues.redhat.com/browse/SRVKS-1111

The test should be re-enabled when it supports cgroups v2

## Proposed Changes

-
-
-


ReToCode (approved):
/lgtm
/approve

PR #2229, opened 2023-08-14
[TEST_ONLY][release-1.30] Fix deprecated field in Kourier bootstrap config

nak3:
As per title, this is test only. We need to fix it on net-kourier.


nak3:
/retest

infra. `ci/prow/4.10-operator-e2e-aws-ocp-410` passed so I think no problem for others...

PR #2249, opened 2023-08-25
Override ingress default certificate in tests

mgencur:
Use a certificate that is defined in the cluster via IngressController custom resources (like on OSD installing in CI).

Fixes https://issues.redhat.com/browse/SRVCOM-2657

## Proposed Changes

-
-
-


mgencur:
> test but the prow here can not validate this change ? 😅

Yeah. This PR will only validate that it doesn't break anything. We will see in the weekly run on OSD whether it helps. But I'm sure it will. We already use this for downstream testing.

PR #2261, opened 2023-09-04
Use v1beta1 DomainMapping CRD

nak3:
As per title, this patch changes to use v1beta1 DomainMapping CRD.
The change is only for some test codes and metadata-webhook which is also used only by test.


skonto:
@nak3 just for clarification, all mid stream branches use v1beta1 or is there conversion happening for v1alpha1?

nak3:
Since long time ago (knative v1.0.0), midstream has v1beta1 DomainMapping CRD.

ReToCode (approved):
/lgtm
/approve

PR #2263, opened 2023-09-05
Update gitops

matzew:
## Proposed Changes

- updating the `channel` for the `gitops` product (according to their documentation)
- adding notes that we may want to retry, if things are "unhealthy"


pierDipi (approved):
/lgtm
/approve

PR #2277, opened 2023-09-21
Fix upgrade sequence for kitchensink tests

mgencur:
* 1.30.0 now replaces 1.29.0 so 1.29.1 is not offered for upgrades
* 1.30.0 was released so it's not in redhat-operators Catalog Sourec

Should fix failures like [this one](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-knative-serverless-operator-main-4.10-upgrade-kitchensink-ocp-410-continuous/1704801507704049664

## Proposed Changes

-
-
-

### DIFF
--- a/olm-catalog/serverless-operator/project.yaml
+++ b/olm-catalog/serverless-operator/project.yaml
@@ -68,9 +68,7 @@ dependencies:
 upgrade_sequence:
     - csv: serverless-operator.v1.29.0
       source: redhat-operators
-    - csv: serverless-operator.v1.29.1
-      source: redhat-operators
     - csv: serverless-operator.v1.30.0
-      source: serverless-operator
+      source: redhat-operators
     - csv: serverless-operator.v1.31.0
       source: serverless-operator
--- a/olm-catalog/serverless-operator/project.yaml
+++ b/olm-catalog/serverless-operator/project.yaml
@@ -67,8 +67,5 @@ dependencies:
         eventing_kafka_broker: knative-v1.9
 upgrade_sequence:
     - csv: serverless-operator.v1.29.0
-      source: redhat-operators
     - csv: serverless-operator.v1.30.0
-      source: redhat-operators
     - csv: serverless-operator.v1.31.0
-      source: serverless-operator
--- a/test/upgrade/installation/serverless.go
+++ b/test/upgrade/installation/serverless.go
@@ -29,11 +29,17 @@ func UpgradeServerlessTo(ctx *test.Context, csv, source string, timeout time.Dur
 		if !strings.Contains(err.Error(), "not found") {
 			return err
 		}
-		// InstallPlan not found in the original catalog source, try the one that was just built.
-		installPlan, err = test.WaitForInstallPlan(ctx,
-			test.OperatorsNamespace, csv, test.ServerlessOperatorPackage, timeout)
-		if err != nil {
-			return err
+		if source != test.ServerlessOperatorPackage {
+			// InstallPlan not found in the original catalog source, try the one that was just built.
+			if _, err := test.UpdateSubscriptionChannelSource(ctx,
+				test.Flags.Subscription, test.Flags.UpgradeChannel, test.ServerlessOperatorPackage); err != nil {
+				return err
+			}
+			installPlan, err = test.WaitForInstallPlan(ctx,
+				test.OperatorsNamespace, csv, test.ServerlessOperatorPackage, timeout)
+			if err != nil {
+				return err
+			}
 		}
 	}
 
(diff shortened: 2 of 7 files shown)


mgencur:
/hold
For kitchensink upgrade tests to pass.

olm-catalog/serverless-operator/project.yaml
```diff
@@ -68,9 +68,7 @@ dependencies:
 upgrade_sequence:
     - csv: serverless-operator.v1.29.0
       source: redhat-operators
-    - csv: serverless-operator.v1.29.1
-      source: redhat-operators
     - csv: serverless-operator.v1.30.0
-      source: serverless-operator
+      source: redhat-operators
```
pierDipi, line 72: So, for previous versions we should always use `redhat-operators` or from 1.31+ we will use `serverless-operator`? I'm wondering if we need to change the bump automation here: https://github.com/openshift-knative/serverless-operator/blob/b2506aa104c5e78995354c5296bc44f6d771f7b2/hack/cmd/bumpso/bumpso.go#L78C2-L86

pierDipi (approved):
/lgtm
/approve

mgencur:
I have fixed the original problem but now the kafka-source-dispatcher is not starting during upgrade due to insufficient CPU:  `"message": "0/6 nodes are available: 3 Insufficient cpu, 3 node(s) had taint {node-role.kubernetes.io/master: }, that the pod didn't tolerate.",`

olm-catalog/serverless-operator/project.yaml
mgencur: `redhat-operator` is used for releases that are already live. `serverless-operator` is used for versions that were not yet released to public (so, they're still in our internal `serverless-operator` catalog source).
Sometimes this sequence will use `serverless-operator` just for the last version in the upgrade sequence. But sometimes even one before. That's for the period when the release branch was already cut and we're starting to develop new version on `main` branch but the previous version was not released yet.

mgencur: Maybe we could get rid of this "source" parameter. We could always try to find the given version in the `redhat-operators` catalogsource. And if it's not there only then switch to `serverless-operator` source.

mgencur: I'm testing a solution locally now.

mgencur: OK. Everything pushed. It works locally.

olm-catalog/serverless-operator/project.yaml
```diff
@@ -67,10 +67,5 @@ dependencies:
         eventing_kafka_broker: knative-v1.9
 upgrade_sequence:
     - csv: serverless-operator.v1.29.0
-      source: redhat-operators
-    - csv: serverless-operator.v1.29.1
-      source: redhat-operators
     - csv: serverless-operator.v1.30.0
-      source: serverless-operator
     - csv: serverless-operator.v1.31.0
-      source: serverless-operator
```
pierDipi, line 76: we can remove this line now
https://github.com/openshift-knative/serverless-operator/blob/b2506aa104c5e78995354c5296bc44f6d771f7b2/hack/cmd/bumpso/bumpso.go#L85

test/upgrade/installation/serverless.go
```diff
@@ -14,14 +15,32 @@ import (
...
 		return err
 	}
 
-	installPlan, err := test.WaitForInstallPlan(ctx, test.OperatorsNamespace, csv, source)
+	installPlan, err := test.WaitForInstallPlan(ctx, test.OperatorsNamespace, csv, source, timeout)
 	if err != nil {
-		return err
+		if !strings.Contains(err.Error(), "not found") {
+			return err
+		}
```
pierDipi, line 31: nit: would the `!apierrors.IsNotFound(err)` work here?

olm-catalog/serverless-operator/project.yaml
mgencur: I think it's already removed in the current PR.

test/upgrade/installation/serverless.go
mgencur: No. It can't be used because the actual error is this, without any additional status:
```
var ErrWaitTimeout = errors.New("timed out waiting for the condition")
```
However, I could change the condition to this:
```
if !errors.Is(err, wait.ErrWaitTimeout) {
	return err
}
```

maschmid:
Instead of attempting to create it and wait for it to fail, we could have perhaps parsed the pacakgemanifests and find the correct catalogsource that does contain the csv, but let's do that later and just make the test work for now.

mgencur:
> Instead of waiting for the installplan to fail, we could have perhaps parsed the pacakgemanifests and find the correct catalogsource that does contain the csv

OK. I will check this approach.

PR #2281, opened 2023-09-25
Determine CatalogSource automatically for upgrade tests

mgencur:
This is a follow-up to https://github.com/openshift-knative/serverless-operator/pull/2277#issuecomment-1733001626

## Proposed Changes

- Determine catalog source which contains the CSV in the given channel


mgencur:
/hold
For kitchensink upgrade tests to pass.

test/subscription.go
```diff
@@ -96,3 +97,24 @@ func CreateOperatorGroup(ctx *Context, name, namespace string) (*operatorsv1.Ope
 			},
 		}, metav1.CreateOptions{})
 }
+
+func GetCatalogSourceWithCSVInChannel(ctx *Context, csv, channel, namespace string) (string, error) {
```
maschmid, line 101: it might still be useful to prefer the defined CatalogSource, in case there are multiple of them containing the CSV (e.g. in case we still build the CI bundle with previous versions as well, we may want to prefer the `serverless-operator` if that one has the CSV, and use any other if the preferred one does not contain it?

maschmid: OK

mgencur:
I'm afraid this whole approach won't work on OCP 4.10. The PackageManifest only contains "currentCSV" on 4.10 while on 4.13 is also includes a list of all CSVs in the channel:
4.10:
https://gist.github.com/mgencur/46d18b486da01242e913da05f2f8812b
4.13:
https://gist.github.com/mgencur/4d395a987ca9ea3aa84e1bf3a824c477

On 4.10, we can't simply determine in which channel/catalogsource the CSV is as long as it's not current CSV but older one.

PR #2287, opened 2023-09-28
Add HPA operator patch

nak3:
The issue is minor and already existed in the released version but please see https://redhat-internal.slack.com/archives/CF5ANN61F/p1695722049622829?thread_ts=1695717539.251089&cid=CF5ANN61F

### DIFF
--- a/hack/patches/013-hpa-operator.patch
+++ b/hack/patches/013-hpa-operator.patch
@@ -1,0 +1,227 @@
+diff --git a/vendor/knative.dev/operator/pkg/reconciler/common/ha.go b/vendor/knative.dev/operator/pkg/reconciler/common/ha.go
+index 3d1004e09..968608658 100644
+--- a/vendor/knative.dev/operator/pkg/reconciler/common/ha.go
++++ b/vendor/knative.dev/operator/pkg/reconciler/common/ha.go
+@@ -18,32 +18,21 @@ package common
+ 
+ import (
+ 	mf "github.com/manifestival/manifestival"
+-	"go.uber.org/zap"
+ 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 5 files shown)


hack/patches/013-hpa-operator.patch
```diff
@@ -0,0 +1,227 @@
...
++
++// When a Podspecable has HPA, the replicas should be controlled by HPAs minReplicas instead of operator.
++// Hence, skip changing the spec.replicas for these Podspecables.
++func hasHorizontalPodAutoscaler(name string) bool {
++	return sets.NewString(
++		"webhook",
++		"activator",
++		"3scale-kourier-gateway",
++		"eventing-webhook",
++		"mt-broker-ingress",
```
maschmid, line 120: I don't think we're having HPA for mt-broker yet in eventing 1.10

(I don't see it in https://github.com/openshift-knative/serverless-operator/blob/main/openshift-knative-operator/cmd/operator/kodata/knative-eventing/1.10/4-mt-channel-broker.yaml 

It seems we are removing them by https://github.com/openshift-knative/eventing/blob/release-v1.10/openshift/release/generate-release.sh , although not sure if we actually still want to be doing that)

maschmid: opening a thread in https://redhat-internal.slack.com/archives/CEXRYS5QC/p1695999818126169

maschmid: opening https://issues.redhat.com/browse/SRVKE-1523  to re-add the HPAs into serverless 1.31

nak3: Okay, we should wait for SRVKE-1523, then.

skonto:
@nak3 let's re-open this.

PR #2288, opened 2023-09-29
Disable tracing test if no Jaeger present

mgencur:
Some scenarios use Zipkin and the tracing is only compatible with Jaeger.

The tracing test then fails with:
```
 tracing_test.go:74: Error verifying traces: timed out waiting for the condition: expecting exactly 1 jaeger pod, got 0
```

Fixes JIRA #

## Proposed Changes

-
-
-

### DIFF
--- a/test/lib.bash
+++ b/test/lib.bash
@@ -139,6 +139,11 @@ function downstream_serving_e2e_tests {
       mv ./test/servinge2e/user_permissions_test.go ./test/servinge2e/user_permissions_notest.go || true
   fi
 
+  # Tracing tests require Jaeger to be installed.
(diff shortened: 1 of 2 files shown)


skonto:
@mgencur should we get this in?

ReToCode (approved):
/lgtm
/approve

mgencur:
> I am wondering how this test was passing previously?
This issue doesn't happen in CI because we clearly separate E2E tests and installing Zipkin. We never run E2E tests after installing Zipkin. 
But these tests are re-used downstream where we might want to install Zipkin for upgrade tests. We also run E2E after doing upgrades in which case it was failing.

mgencur:
/cherry-pick release-1.30

I need this because we also run downgrade tests and then run tests from the previous branch

PR #2294, opened 2023-10-02
Soak tests PoC

maschmid:
Fixes JIRA #SRVCOM-1748

## Proposed Changes
- :gift: Initial add of Soak test (kafkasource so far)

a proof of concept of using reconciler framework for soak tests.

a soak test is some kind of a test that just repeats stuff over and over again until the end of time. Soak tests would run potentially multiple copies of all the defined tests scenarios.

the basic problem we're trying to solve here is how to write features so that they can be reused in different scenarios (e.g. creating a kafkasource once, vs. creating it at each of the iteration)

the proposal here is to: 
* separate naming of the resources from creating them, 
* choosing which features are to be executed only once during setup, and which during each iteration

For any kind of scenario, for which the proposed structure would be too rigid, the test could also just do
```for since.Add(Flags.Duration).After(time.Now()) {``` and just do what they want.

### CODE REVIEW

File: test/soake2erekt/main_test.go [hunk 1]
```diff
@@ -0,0 +1,254 @@
...
+			}
+
+			iteration := 0
+			// Repeat the soak test for the duration specified by --soak-duration flag
+			for since.Add(Flags.Duration).After(time.Now()) {
+				// During each iteration, generate the "iteration" features and run them as Tests
+				// Cleanup all resources for these features at the end of the iteration
+
+				// TODO: can we do this? (use an empty env for the iteration, but keep using the original context?)
+				_, iterationEnv := global.Environment(
```
mgencur, line 174: If I understand it correctly, you create a new environment so as to be able to delete resources from the given iteration via `iterationEnv.References()` ? 
It should work but we're relying on some implementation details that might change in the future. 
Maybe it would be easier to just call `iterationCtx, iterationEnv := soakTestEnvironment(t, namespace)` here, and use the given iteration context and env in this iteration. Then you wouldn't need  to call `iterationCtx = environment.ContextWith(iterationCtx, iterationEnv)` because it's already done when creating the iterationCtx and iterationEnv

test/soake2erekt/main_test.go
```diff
@@ -0,0 +1,254 @@
...
+	return global.Environment(
+		environment.WithNamespace(namespace),
+		knative.WithKnativeNamespace(system.Namespace()),
+		knative.WithLoggingConfig,
+		knative.WithTracingConfig,
+		k8s.WithEventListener,
+		// Enables KnativeService in the scenario.
+		//eventshub.WithKnativeServiceForwarder,
+		environment.WithPollTimings(Flags.PollInterval, Flags.PollDuration),
+		environment.WithTestLogger(t),
```
mgencur, line 70: Maybe I would add a comment to warn about adding "environment.Managed()" which would break all the stuff.

test/soake2erekt/kafkasource_soak_test.go
```diff
@@ -0,0 +1,334 @@
...
+func TestKafkaSourceAddingAndRemovingSoak(t *testing.T) {
+	t.Parallel()
+
+	topicPrefix := feature.MakeRandomK8sName("topic")
+	kafkaSinkPrefix := topicPrefix
+	kafkaSourcePrefix := feature.MakeRandomK8sName("kafkaSource")
+	senderPrefix := feature.MakeRandomK8sName("sender")
+
+	// Number of kafkasources to create and delete at once on each iteration
+	const max = 16
```
mgencur, line 283: Maybe rename to `numKafkaSources` ? Then the comment might not be necessary.

test/soake2erekt/kafkasource_soak_test.go
```diff
@@ -0,0 +1,334 @@
...
+				f := kafkaSourceScenarioInstallKafkaSourceFeature(names)
+				env.Test(ctx, t, f)
+			}
+
+			for j := 0; j < max; j++ {
+				soakEnv := SoakEnvFromContext(ctx)
+				names := namesFn(soakEnv)
+				names.kafkaSource = fmt.Sprintf("%s-%d-%d", kafkaSourcePrefix, soakEnv.Iteration(), j)
+
+				f := kafkaSourceScenarioIsReadyKafkaSourceFeature(names)
```
mgencur, line 318: Sounds like kafkaSourceScenarioIsReadyKafkaSourceFeature and kafkaSourceScenarioInstallKafkaSourceFeature should be merged into one. After all, it's one resource, we're installing it and waiting for it to be ready. I don't see any separate usage of "IsReady" that would justify separating them and having more complex code (e.g. the two loops above)

mgencur (commented):
Hey @maschmid , looks great! I went through the code and suggested some (rather minor) changes. I will take another look tomorrow, with clear mind.

test/soake2erekt/kafkasource_soak_test.go
```diff
@@ -0,0 +1,334 @@
...
+			kafkaTopic:  fmt.Sprintf("%s-%d", topicPrefix, env.CopyID()),
+			kafkaSink:   fmt.Sprintf("%s-%d", kafkaSinkPrefix, env.CopyID()),
+			kafkaSource: kafkaSourceName,
+		}
+	}
+
+	soakTest := SoakTest{
+		NamespacePrefix: "test-kafka-source-stable-",
+		SetupFn: func(ctx context.Context, env environment.Environment, t *testing.T) {
+			// TODO: These are actually equivalent to just env.Test(ctx, t, eventshubReceiverFeature(namesFn(SoakEnvFromContext(ctx)))) , so not sure if we should bother...
```
mgencur, line 205: I would prefer what you have in the comment. For anyone familiar with the REKT framework, it's immediately clear what's happening. Having RunSoakFeatureFnWithMapping is additional layer of abstraction which doesn't bring much, it's just harder to understand the code.

test/soake2erekt/kafkasource_soak_test.go
```diff
@@ -0,0 +1,334 @@
...
+			RunSoakFeatureFnWithMapping(ctx, env, t, kafkaSourceScenarioTopicAndSinkSetupFeature, namesFn)
+			RunSoakFeatureFnWithMapping(ctx, env, t, kafkaSourceScenarioInstallKafkaSourceFeature, namesFn)
+			RunSoakFeatureFnWithMapping(ctx, env, t, kafkaSourceScenarioIsReadyKafkaSourceFeature, namesFn)
+		},
+		IterationFn: func(ctx context.Context, env environment.Environment, t *testing.T) {
+			RunSoakFeatureFnWithMapping(ctx, env, t, kafkaSinkSendFeature, namesFn)
+			RunSoakFeatureFnWithMapping(ctx, env, t, verifySingleEventReceivedFeature, namesFn)
+
+			// we just want to verify the source can send/receive events throughout the soak test, so let it rest here for a while
+			time.Sleep(1 * time.Second)
```
mgencur, line 216: Not sure I understand the meaning of this Sleep. We're running in iterations so each iteration should verify receiving its own event.

mgencur: I mean, the `.AtLeast(atLeast)` function which is used to verify the received event already waits for some time for the event to be present.

test/soake2erekt/main_test.go
maschmid: we need the context to be per soak test, not per iteration, as things like eventshub use the context to store the eventshub store in it... (see `registerEventsHubStore`  ) (so, to be able to use eventshub receiver in soak setup, but use eventshub sender per iteration)

test/soake2erekt/kafkasource_soak_test.go
maschmid: we want to first create all, and then wait for all, so that the actual creation can happen in parallel.

test/soake2erekt/kafkasource_soak_test.go
maschmid: we don't want this iteration cycle to be "as fast as it can", the sleep is there for the thing to not be a stress test.

test/soake2erekt/kafkasource_soak_test.go
```diff
@@ -0,0 +1,334 @@
+package soake2erekt
```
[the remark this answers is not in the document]
maschmid: ok

test/soake2erekt/main_test.go
```diff
@@ -0,0 +1,254 @@
...
+	pkgTest "knative.dev/pkg/test"
+	"knative.dev/reconciler-test/pkg/environment"
+	"knative.dev/reconciler-test/pkg/feature"
+	"knative.dev/reconciler-test/pkg/k8s"
+	"knative.dev/reconciler-test/pkg/knative"
+)
+
+// var global environment.GlobalEnvironment
+
+type SoakFlagsStruct struct {
```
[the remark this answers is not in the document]
maschmid: ok

test/soake2erekt/kafkasource_soak_test.go
```diff
@@ -0,0 +1,334 @@
...
+	"knative.dev/reconciler-test/pkg/feature"
+	"knative.dev/reconciler-test/pkg/manifest"
+	"knative.dev/reconciler-test/pkg/resources/service"
+)
+
+/*
+The scenarios define names of all resources used in the scenario, so they can reference other resources
+in a typesafe-ish kind of way
+*/
+type kafkaSourceScenarioNames struct {
```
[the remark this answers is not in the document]
maschmid: right. Originally, I had tried some other ways, like having interfaces, to then be able to combine the whole thing from some kind of components, but it didn't turn out great...  I'll make this public then for now, maybe we'll see a better way once we have the need to share component between different scenarios.

test/soake2erekt/main_test.go
```diff
@@ -0,0 +1,254 @@
...
+		namespace := test.NamespacePrefix + strconv.Itoa(copyID)
+		t.Run(namespace, func(t *testing.T) {
+			t.Parallel()
+
+			since := time.Now()
+
+			ctx, env := soakTestEnvironment(t, namespace)
+
+			// Execute the setup "features", store the references created during setup for cleanup at the end
+			setupCtx := context.WithValue(ctx, soakKey{}, &soakEnvImpl{
```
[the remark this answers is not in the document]
maschmid: I am not sure, is there a use case for creating the soak context outside of RunSoakTest ?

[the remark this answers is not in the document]
mgencur: OK. Maybe not.

test/soake2erekt/kafkasource_soak_test.go
mgencur: Hmm, using `env.ParallelTest(ctx, t, f)` should have the same effect. There could be just one feature per KafkaSource, which creates and waits for the kafka source to be ready.

test/soake2erekt/kafkasource_soak_test.go
mgencur: Does the 1 second make any functional difference? It seems too short to make any effect. IMO, this is not a stress test because we're always waiting for the event to be received, so the system has enough time to send/receive event before next iteration.

test/soake2erekt/main_test.go
mgencur: Thanks. It was not immediately clear. I think we can do what you did. We just need to be extra careful which `ctx` or `env` is used later in the code.

mgencur: We probably also want to set some env here, maybe at least these three?:
```
knative.WithKnativeNamespace(system.Namespace()),
environment.WithPollTimings(Flags.PollInterval, Flags.PollDuration),
environment.WithTestLogger(t),
```

test/soake2erekt/kafkasource_soak_test.go
mgencur: Since the tests run with high "-parallel" it could create the KafkaSource all in parallel. The behaviour would be similar but probably better to check if it works as expected.

maschmid: in the overall soak test, this will be just one scenario of many, so I am not sure if we want to do too many parallel request in the single scenario...

mgencur: OK. No problem.

test/soak/main_test.go
```diff
@@ -0,0 +1,233 @@
...
+				return
+			}
+
+			iteration := 0
+			// Repeat the soak test for the duration specified by --soak-duration flag
+			for since.Add(Flags.Duration).After(time.Now()) {
+				// During each iteration, generate the "iteration" features and run them as Tests
+				// Cleanup all resources for these features at the end of the iteration
+
+				// TODO: can we do this? (use an empty env for the iteration, but keep using the original context?)
```
mgencur, line 167: Do you want to keep the comment here? We has some discussion around that.

maschmid: well, if we agree we _can_ do it like this... I don't see any alternative besides adding this as an API to reconciler framework

mgencur: I think the tests showed it can be done like this. But if we can formalize the requirements and it would be useful for REKT framework in general, we can contribute this upstream later, and possibly re-use it here. But for now I'd just just what we have here.

mgencur: Maybe we can create an issue upstream in https://github.com/knative-extensions/reconciler-test if the requirements are clear.

PR #2300, opened 2023-10-04
Trying to fix the bump automation for istio authz chart bump

pierDipi:
- set the chart absolute path into env variable
- go to SO path, run the tool with `go run ...` and point the tool to the chart absolute path, using the flag https://github.com/openshift-knative/serverless-operator/blob/2cc5e46b1cdfad082a67a59d4d618b171233747f/hack/cmd/bumpistiochart/bumpistiochart.go#L34

maybe fix https://github.com/openshift-knative/serverless-operator/actions/runs/6403506901/job/17382146920

### DIFF
--- a/.github/workflows/bump.yaml
+++ b/.github/workflows/bump.yaml
@@ -58,8 +58,8 @@ jobs:
           echo "RELEASE_BRANCH=${{ github.ref_name }}" >> $GITHUB_ENV
 
       - name: Bump chart version
-        working-directory: ./src/github.com/openshift-knative/knative-istio-authz-chart
-        run: go run hack/cmd/bumpistiochart/bumpistiochart.go --branch "${RELEASE_BRANCH}"
+        working-directory: ./src/github.com/${{ github.repository }}
+        run: go run hack/cmd/bumpistiochart/bumpistiochart.go --branch "${RELEASE_BRANCH}" --chart-metadata-path "${{ github.workspace }}/src/github.com/openshift-knative/knative-istio-authz-chart/Chart.yaml"
 
       - name: Create branch
         working-directory: ./src/github.com/openshift-knative/knative-istio-authz-chart


.github/workflows/bump.yaml
```diff
@@ -57,9 +57,14 @@ jobs:
...
 
-      - name: Bump chart version
+      - name: Set chart path
         working-directory: ./src/github.com/openshift-knative/knative-istio-authz-chart
-        run: go run hack/cmd/bumpistiochart/bumpistiochart.go --branch "${RELEASE_BRANCH}"
+        run: |
+          echo "ISTIO_AUTHZ_CHART_PATH=$(pwd)" >> $GITHUB_ENV
+
+      - name: Bump chart version
+        working-directory: ./src/github.com/${{ github.repository }}
```
skonto, line 66: Does this point to S-O?

pierDipi: yes

pierDipi: is the repository of the workflow

pierDipi: see line 40

pierDipi: it's not checked out in the Go path, see line 40

pierDipi: that needs to match where SO is checked out and it's not necessarily the same as the Go path

pierDipi: Another approach would be to use ${{ github.workspace }} for the `--chart-metadata-path`

skonto: GOPATH: ${{ github.workspace }}
GOPATH: /home/runner/work/serverless-operator/serverless-operator
So we checkout in the runners home dir it seems and the relative path is based on that.

pierDipi: But it doesn't have to be like that, it is just a case that they match, sorry but I don't get why adding more variables to the line helps in any way

pierDipi: Now, it is consistent with the other usage at line 92-94

skonto: I said ok abov, we are in agreement :) I am just trying to understand the implementation :)

skonto:
@pierDipi should I approve it? We are good?

PR #2301, opened 2023-10-04
Pass QE robot token when checking out helm chart in bump automation

pierDipi:
This should fix the failed to push with 403 https://github.com/openshift-knative/serverless-operator/actions/runs/6405462417/job/17388514221

```
    # Personal access token (PAT) used to fetch the repository. The PAT is configured
    # with the local git config, which enables your scripts to run authenticated git
    # commands. The post-job step removes the PAT.
    #
    # We recommend using a service account with the least permissions necessary. Also
    # when generating a new PAT, select the least scopes necessary.
    #
    # [Learn more about creating and using encrypted secrets](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets)
    #
    # Default: ${{ github.token }}
    token: ''
```

### DIFF
--- a/.github/workflows/bump.yaml
+++ b/.github/workflows/bump.yaml
@@ -45,6 +45,7 @@ jobs:
         with:
           repository: 'openshift-knative/knative-istio-authz-chart'
           path: ./src/github.com/openshift-knative/knative-istio-authz-chart
+          token: ${{ secrets.SERVERLESS_QE_ROBOT }}
           fetch-depth: 0
 
       - name: Set branch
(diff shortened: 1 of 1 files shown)


.github/workflows/bump.yaml
```diff
@@ -66,10 +67,6 @@ jobs:
         run: |
           set -euo pipefail
           
-          git config --global user.email "serverless-support@redhat.com"
-          git config --global user.name "OpenShift Serverless"
-          git config --global user.password "${{ secrets.SERVERLESS_QE_ROBOT }}"
```
skonto, line 71: So this is ignored it seems.

pierDipi:
Merging to try this new solution

PR #2305, opened 2023-10-05
[SRVKS-1154] Do not configure default domain if anything is set

nak3:
Fixes JIRA [SRVKS-1154](https://issues.redhat.com/browse/SRVKS-1154)

## Proposed Changes

Currently default domain is always inserted into `config-domain`. It does not allow users to set the domain like `svc.cluster.local` by default.

Hence, this patch changes to use `ConfigureIfUnsetAny` for `config-domain`. User will be able to change the default domain to `svc.cluster.local`.

### DIFF
--- a/openshift-knative-operator/pkg/common/config.go
+++ b/openshift-knative-operator/pkg/common/config.go
@@ -35,9 +35,9 @@ func ConfigureIfUnset(s *base.CommonSpec, cm, key, value string) {
 	s.Config[cm][key] = value
 }
 
-// ConfigureIfUnsetAny sets a value in the config-domain if any configuration is not already set.
+// ConfigureIfConfigmapUnset sets a value in the given ConifgMap if any configuration is not already set.
 // For example, the config-domain can take an arbitrary domain as a key, so it should be used here.
-func ConfigureIfUnsetAny(s *base.CommonSpec, cm, key, value string) {
+func ConfigureIfConfigmapUnset(s *base.CommonSpec, cm, key, value string) {
 	if s.Config == nil {
 		s.Config = make(map[string]map[string]string, 1)
 	}
--- a/openshift-knative-operator/pkg/common/config.go
+++ b/openshift-knative-operator/pkg/common/config.go
@@ -35,9 +35,10 @@ func ConfigureIfUnset(s *base.CommonSpec, cm, key, value string) {
 	s.Config[cm][key] = value
 }
 
-// ConfigureIfUnsetDefaultDomain sets a value in the given ConfigMap under the given key if it is neither
-// already set the key nor another key becomes the default domain.
-func ConfigureIfUnsetDefaultDomain(s *base.CommonSpec, cm, key, value string) {
+// ConfigureIfUnsetDomain sets a value in the config-domain if it's not already set.
+// config-domain can take a domain as a key so it is different from ConfigureIfUnset.
+func ConfigureIfUnsetDefaultDomain(s *base.CommonSpec, key, value string) {
+	const cm = "domain"
 	if s.Config == nil {
 		s.Config = make(map[string]map[string]string, 1)
 	}
--- a/openshift-knative-operator/pkg/common/config.go
+++ b/openshift-knative-operator/pkg/common/config.go
@@ -34,3 +34,28 @@ func ConfigureIfUnset(s *base.CommonSpec, cm, key, value string) {
 	}
 	s.Config[cm][key] = value
 }
+
+// ConfigureIfUnsetDefaultDomain sets a value in the given ConfigMap under the given key if it is neither
+// already set the key nor another key becomes the default domain.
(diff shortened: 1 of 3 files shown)


nak3:
Hmm... This could be a breaking change for an user who configured config-domain with custom domain without "default" domain. For example -

__KnativeServing CR__ (default domain = `mydefault.domain`)
```
spec:
  config:
    domain:
      some1.domain: |
        selector:
          some1: foo
      some2.domain: |
        selector:
          some2: foo
```

__config-domain CURRENT__

```
spec:
  config:
    domain:
      some1.domain: |
        selector:
          some1: foo
      some2.domain: |
        selector:
          some2: foo
    mydefault.domain: ""
```

__config-domain AFTER THIS PATCH__

```
spec:
  config:
    domain:
      some1.domain: |
        selector:
          some1: foo
      some2.domain: |
        selector:
          some2: foo
```

nak3:
/cc @ReToCode @skonto Could you please take a look?

For https://github.com/openshift-knative/serverless-operator/pull/2305#issuecomment-1748472990, users will not omit default domain and I have written it in release note deraft.

nak3:
upgrade test is not healthy now regardless of this PR https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-knative-serverless-operator-main-4.13-upgrade-tests-aws-ocp-413

openshift-knative-operator/pkg/common/config.go
```diff
@@ -34,3 +34,22 @@ func ConfigureIfUnset(s *base.CommonSpec, cm, key, value string) {
...
+func ConfigureIfUnsetAny(s *base.CommonSpec, cm, key, value string) {
+	if s.Config == nil {
+		s.Config = make(map[string]map[string]string, 1)
+	}
+
+	if s.Config[cm] == nil {
+		s.Config[cm] = make(map[string]string, 1)
+	}
+
+	if len(s.Config[cm]) != 0 {
```
skonto, line 49: Just to be clear in that domain cm there can be only one entry?

ReToCode: I don't thinks so, you can have a list of domains. @nak3 can you maybe add a test case where one key has multiple values in its map? like

domain:
  example.com: xxx
  second.com: xxx

openshift-knative-operator/pkg/common/config.go
```diff
@@ -34,3 +34,22 @@ func ConfigureIfUnset(s *base.CommonSpec, cm, key, value string) {
 	}
 	s.Config[cm][key] = value
 }
+
+// ConfigureIfUnsetAny sets a value in the config-domain if any configuration is not already set.
```
ReToCode, line 38: nit: config-domain, this would work for any config-map.

openshift-knative-operator/pkg/common/config.go
```diff
@@ -34,3 +34,22 @@ func ConfigureIfUnset(s *base.CommonSpec, cm, key, value string) {
...
+// For example, the config-domain can take an arbitrary domain as a key, so it should be used here.
+func ConfigureIfUnsetAny(s *base.CommonSpec, cm, key, value string) {
```
ReToCode, line 40: I had a bit a hard time to distinguish `ConfigureIfUnsetAny` from `ConfigureIfUnset` as one checks for a key and the other one checks for the whole config-map entry. What do you think about `ConfigureIfConfigmapUnset` as a name (or something like this)?

ReToCode:
> For https://github.com/openshift-knative/serverless-operator/pull/2305#issuecomment-1748472990, users will not omit default domain and I have written it in release note deraft.

Hm yes that make sense, but also is the issue that that one customer had, right? So this is more a bug than a feature, but true that it is breaking. @skonto WDYT, can we go with a note in the release-notes?

> upgrade test is not healthy now regardless of this PR

Probably this https://redhat-internal.slack.com/archives/CD87JDUB0/p1697018500414289?

ReToCode:
@skonto gentle ping:

https://github.com/openshift-knative/serverless-operator/pull/2305#issuecomment-1757482586

nak3:
(just keep rebasing)

openshift-knative-operator/pkg/common/config_test.go
```diff
@@ -74,3 +74,83 @@ func TestConfigure(t *testing.T) {
...
+		in: base.ConfigMapData{
+			"foo": map[string]string{},
+		},
+		expected: base.ConfigMapData{
+			"foo": map[string]string{
+				"bar": "baz",
+			},
+		},
+	}, {
+		name: "override",
```
skonto, line 101: Should not this test be named `do not override if set` or something?
If I understand correctly the goal is to configure only if unset right?

skonto:
Sorry for the delay @nak3 I missed this request.

skonto:
This lgtm minus the comment for the test names. @nak3 let me know what you think and I can stamp it.

skonto:
@nak3 could you update it?

PR #2307, opened 2023-10-05
Pull Istio helm chart image only once in generator

dsimansk:
## Proposed Changes

- :broom:  Pull Istio helm chart image only once in generator

This change should address running into Quay API limits in generated files check. 

/cc @ReToCode @pierDipi


dsimansk:
Or maybe we're just completely busted here. :)

dsimansk:
Superseded by no OCI image solution. 

/close

PR #2308, opened 2023-10-05
Get helm chart from Github instead of OCI

ReToCode:
## Proposed Changes
- Get helm chart from Github instead of OCI

### DIFF
--- a/hack/generate/mesh-auth-policies.sh
+++ b/hack/generate/mesh-auth-policies.sh
@@ -20,7 +20,7 @@ source "$(dirname "${BASH_SOURCE[0]}")/../lib/metadata.bash"
 policies_path="$(dirname "${BASH_SOURCE[0]}")/../lib/mesh_resources/authorization-policies/helm"
 chart_version="$(metadata.get project.version | grep -Eo '[0-9]+\.[0-9]+')" # grep removes the patch version in semver
 
-# Pull image template from Github
+# Pull helm chart from Github
 template_cache=$(mktemp -d)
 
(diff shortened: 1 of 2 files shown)


hack/generate/mesh-auth-policies.sh
```diff
@@ -18,10 +18,15 @@ helm > /dev/null || exit 127
 source "$(dirname "${BASH_SOURCE[0]}")/../lib/metadata.bash"
 
 policies_path="$(dirname "${BASH_SOURCE[0]}")/../lib/mesh_resources/authorization-policies/helm"
-if [ -z "${ISTIO_CHART_VERSION:-}" ]; then
-  chart_version="$(metadata.get project.version)"
-else
-  chart_version="${ISTIO_CHART_VERSION}"
+chart_version="$(metadata.get project.version | grep -Eo '[0-9]+\.[0-9]+')" # grep removes the patch version in semver
```
pierDipi, line 21: We still need the logic, this is to have the bump flow work on an unpublished version of the chart

pierDipi: Ah actually, I see now, we don't need that anymore

pierDipi: because we create the branch _before_ generating policies

hack/generate/mesh-auth-policies.sh
```diff
@@ -18,10 +18,15 @@ helm > /dev/null || exit 127
...
+
+# Pull helm chart from Github
+template_cache=$(mktemp -d)
```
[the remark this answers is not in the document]
ReToCode: addeed

PR #2314, opened 2023-10-09
separate execution of kitchensink e2e tests so they don't timeout on 2h CI step limits

maschmid:
CI runs timeout most of the time since the added source tests, https://prow.ci.openshift.org/job-history/gs/origin-ci-test/logs/periodic-ci-openshift-knative-serverless-operator-main-4.13-e2e-kitchensink-ocp-413-continuous

- :broom: Add test-kitchensink-e2e-single-testonly target to be able to run tests separately

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -223,6 +223,7 @@ test-ui-e2e:
 test-kitchensink-e2e-testonly:
 	./test/kitchensink-e2e-tests.sh
 
+# Run only a subset of e2e tests, e.g. "make test-kitchensink-e2e-single-testonly TEST=TestBroker"
 test-kitchensink-e2e-single-testonly:
 	./test/kitchensink-e2e-tests.sh -run $(TEST)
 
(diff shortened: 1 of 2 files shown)


mgencur:
/lgtm
/hold

Hold for kitchensink tests to pass. Just to see if we're hitting the overall job timeout now.

maschmid:
we're hitting the step timeout, I will try to break the kitchensink suite into separate parts, so we can run them as separate steps

Makefile
```diff
@@ -223,10 +223,14 @@ test-ui-e2e:
...
-test-kitchensink-e2e:
+test-kitchensink-e2e-single-testonly:
+	./test/kitchensink-e2e-tests.sh -run $(TEST)
+
+test-kitchensink-e2e-setup:
 	UNINSTALL_STRIMZI="false" ./hack/strimzi.sh
 	SCALE_UP=4 INSTALL_KAFKA="true" ./hack/install.sh
-	./test/kitchensink-e2e-tests.sh
+
+test-kitchensink-e2e: test-kitchensink-e2e-setup test-kitchensink-e2e-testonly
```
mgencur, line 233: Unfortunately, it's not guaranteed that `test-kitchensink-e2e-setup` will run before `test-kitchensink-e2e-testonly`. That burned me in the past. It's better to list the Bash functions under `test-kitchensink-e2e`, one by one.

maschmid: OK

Issue #2315, opened 2023-10-09
test/upgrade_test.kafkaChannelTest data race


maschmid:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift-knative_serverless-operator/2292/pull-ci-openshift-knative-serverless-operator-main-4.13-upgrade-tests-aws-ocp-413/1708842877116747776

Doing these https://github.com/openshift-knative/serverless-operator/blob/main/test/upgrade/upgrade_test.go#L324-L325 inside a parallel test is wrong,

we cannot just change the global env and run channel.ChannelChain  in parallel with both imc and kafkachannel impls...

PR #2319, opened 2023-10-10
Avoid running channel test in parallel for InMemory and Kafka

mgencur:
Fixes https://github.com/openshift-knative/serverless-operator/issues/2315

## Proposed Changes

- Run test for InMemoryChannel first (non-parallel) and for KafkaChannel later in-parallel with other tests
- Make sure GVK is properly set for InMemoryChannel test even if KafkaChannel test runs first, to be on the safe side

### DIFF
--- a/test/upgrade/upgrade_test.go
+++ b/test/upgrade/upgrade_test.go
@@ -256,13 +256,15 @@ func InMemoryChannelPostDowngradeTest() pkgupgrade.Operation {
 }
 
 func inMemoryChannelTest(t *testing.T) {
-	t.Parallel()
-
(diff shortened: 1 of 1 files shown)


pierDipi (commented):
Don't we need to remove it also from line 318 in `kafkaChannelTest` ?

mgencur:
> Don't we need to remove it also from line 318 in kafkaChannelTest ?

No. Because "serial" tests run first, and only after they finish, Golang runs the parallel tests. So, they should not run in parallel with each other even when the kafkaChannelTest is "parallel".

PR #2324, opened 2023-10-10
Create update generated files PR on push that are not for dependabot

pierDipi:
This is an alternative to [1], we need the PR to be created on push events for release branches for the release automation but without touching dependabot PRs.

[1] https://github.com/openshift-knative/serverless-operator/commit/2575eb70145059b53c3276aaffba0aaf6c220cec

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -82,7 +82,7 @@ jobs:
       #
       #   when it's manually triggered or a scheduled run.
       - name: Create Pull Request
-        if: github.event_name == 'workflow_dispatch' || github.event_name == 'schedule'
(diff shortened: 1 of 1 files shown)


.github/workflows/validate.yaml
```diff
@@ -82,7 +82,7 @@ jobs:
       #
       #   when it's manually triggered or a scheduled run.
       - name: Create Pull Request
-        if: github.event_name == 'workflow_dispatch' || github.event_name == 'schedule'
+        if: github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || ( github.event_name == 'push' && !contains(github.ref_name, 'dependabot/') )
```
skonto, line 85: Is the `/` suffix required?

PR #2325, opened 2023-10-11
Start upgrade sequence with 1.30.1

mgencur:
## Proposed Changes

-
-
-

### DIFF
--- a/olm-catalog/serverless-operator/project.yaml
+++ b/olm-catalog/serverless-operator/project.yaml
@@ -67,7 +67,6 @@ dependencies:
         eventing: knative-v1.10
         eventing_kafka_broker: knative-v1.10
 upgrade_sequence:
-    - csv: serverless-operator.v1.30.0
+    - csv: serverless-operator.v1.30.1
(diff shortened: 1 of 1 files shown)


mgencur:
Looks like 1.30.1 is not available on OCP 4.10. I don't get it.

mgencur:
> Looks like 1.30.1 is not available on OCP 4.10. I don't get it.

Ah. It's probably caused by the fact that 4.10 is out of support since September 10

mgencur:
/test 4.13-upgrade-kitchensink-ocp-413
/test 4.13-upgrade-tests-aws-ocp-413
/test 4.13-upstream-e2e-kafka-aws-ocp-413

Running on OCP 4.10 is not required. This version is no longer supported (but CI hasn't been updated yet to have 4.11 as the min. version)

pierDipi:
so we can merge this, right @mgencur ?

/lgtm
/hold

mgencur:
@pierDipi Yes. We can merge this.

PR #2351, opened 2023-10-25
[release-1.30] Add new kube-rbac-proxy with http2-disable option

dsimansk:
Per Slack thread. 

/cc @skonto @ReToCode @matzew


openshift-knative-operator/pkg/monitoring/rbac_proxy_injection.go
```diff
@@ -99,6 +99,7 @@ func InjectRbacProxyContainer(deployments sets.String, cfg base.ConfigMapData) m
 					"--tls-cert-file=" + filepath.Join(mountPath, "tls.crt"),
 					"--tls-private-key-file=" + filepath.Join(mountPath, "tls.key"),
 					"--logtostderr=true",
+					"--http2-disable",
```
matzew, line 102: wouldn't that... default to this behavior anyways?

dsimansk: By this change I don't so

https://github.com/openshift/kube-rbac-proxy/commit/ae32bfa0f11ae2854339ab9af83c44ee1c04fe31#diff-193f95684186b3f36e5beb9ee139212fd891ddbce3f57b79caa5cf328ad4c23aR121

matzew: I think you are right, we need that flag

matzew: Exactly the opposite how we do it in Knative. which is ok...

hack/generate/csv.sh
```diff
@@ -14,7 +14,7 @@ registry_host='registry.ci.openshift.org'
 registry="${registry_host}/openshift"
 client_version="$(metadata.get dependencies.cli)"
 kn_event="${registry_host}/knative/release-${client_version%.*}:client-plugin-event"
-rbac_proxy="registry.ci.openshift.org/origin/4.7:kube-rbac-proxy"
+rbac_proxy="registry.ci.openshift.org/origin/4.12:kube-rbac-proxy"
```
matzew, line 17: that is the same as `quay.io/brancz/kube-rbac-proxy:v0.15.0` ?

asking b/c of this mapping here: https://github.com/openshift/release/blob/master/core-services/image-mirroring/openshift/mapping_origin_4_12#L174C55-L174C100   :sweat_smile:

dsimansk: It has the flag. That's a good news. And I hope it's from OpenShift midstream. 

In a product release that's replaced by `registry.redhat.io` variant. 

https://github.com/openshift/kube-rbac-proxy/tree/release-4.12

ReToCode: Hm, do we maybe (in a followup-PR) want to link the version to https://github.com/openshift-knative/serverless-operator/blob/main/olm-catalog/serverless-operator/project.yaml#L31 instead of having a version in that bash file? @mgencur

ReToCode:
@dsimansk 
Ah just noticed this now, should we not go to the image for 4.10? Release 1.30 also runs on OCP 4.10.

dsimansk:
> @dsimansk Ah just noticed this now, should we not go to the image for 4.10? Release 1.30 also runs on OCP 4.10. And then use 4.11 for `release-1.31` and `main`?

The change is not cherry-picked later than 4.12. I.e. CI registry image `4.11:kube-rbac-proxy` doesn't contain those flags.

matzew:
> registry image `4.11:kube-rbac-proxy` doesn't contain those flags. 

that versioning / mirroring is weird...

ReToCode:
> The change is not cherry-picked later than 4.12. I.e. CI registry image 4.11:kube-rbac-proxy doesn't contain those flags.

Lol. But can we be sure that the 4.12 image also runs "ok" on 4.10?

hack/generate/csv.sh
dsimansk: I've seen on 1.31 it's already derived from ocp min version metadata.

dsimansk: https://github.com/openshift-knative/serverless-operator/blob/main/hack/generate/csv.sh#L17

dsimansk:
@matzew @ReToCode @skonto pls if you get to it in the morning to review & unhold if it looks OK.

ReToCode (approved):
/lgtm
/approve

PR #2352, opened 2023-10-25
[release-1.31] Add new kube-rbac-proxy with http2-disable option

dsimansk:
Similar as #2351 with the twist from product operator bundle. We can test it with OCP max version, that is closer to image used in product.


ReToCode (approved):
/lgtm
/approve

ReToCode:
Are you going to do this to main too?

PR #2355, opened 2023-10-26
[WIP] Move to go 1.20

skonto:
## Proposed Changes

- `go mod edit -go 1.20` , 1.18 is not supported anymore
- updates Golang image builder version
- skips 1.19 moves to the latest supported

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/openshift-knative/serverless-operator
 
-go 1.18
+go 1.20
 
 require (
 	github.com/blang/semver/v4 v4.0.0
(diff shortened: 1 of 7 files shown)


skonto:
@dsimansk 


> We don't have 1.21 available in RHEL yet.

Which part. Could you elaborate? I will change to 1.20 no worries.
I was a bit ambitious and wanted basically to test what breaks ;)

dsimansk:
With Golang toolchain I meant the following version:

- https://github.com/openshift-knative/serverless-operator/blob/main/.github/workflows/validate.yaml#L27

skonto:
@dsimansk do we still suffer from https://redhat-internal.slack.com/archives/CF5ANN61F/p1680250732865129?thread_ts=1680244739.280449&cid=CF5ANN61F? If so @nak3 should we revive https://github.com/knative/hack/pull/179?
cc @ReToCode

PR #2366, opened 2023-11-01
Use s2i GA image and update buildah image to latest released

Kaustubh-pande:
as per title
cc @mgencur @dsimansk


mgencur:
Just a note. We don't have any tests in s-o to verify that this new image actually works. I'm not sure if we should merge it now or after verifying (somewhere else) that this stuff works.

dsimansk:
@Kaustubh-pande perhaps we can try to verify it through func midstream rep. @jrangelramos FYI & PTAL.

PR #2370, opened 2023-11-06
fix pingsource metrics test

maschmid:
attempt to fix pingsource metrics test, e.g. failing in https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift-knative_serverless-operator/2368/pull-ci-openshift-knative-serverless-operator-main-4.13-operator-e2e-aws-ocp-413/1720329077899923456 

It seems to be failing because the metric is gone if pingsources are gone, so we move the metric test to the TestPingSourceToKsvc  (in the same way as we have moved namespaced broker metrics test to  TestSourceNamespacedKafkaBrokerKsvc )

## Proposed Changes
- :broom: Move pingsource metrics test to a test that has PingSource deployed

### DIFF
--- a/test/eventinge2erekt/pingsource_ksvc_test.go
+++ b/test/eventinge2erekt/pingsource_ksvc_test.go
@@ -1,6 +1,9 @@
 package eventinge2erekt
 
 import (
+	"context"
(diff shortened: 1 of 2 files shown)


mgencur:
Nice!

/lgtm

PR #2391, opened 2023-11-15
Set longer progressDeadline for Serving PVC tests

mgencur:
Provisioning a volume took about 6 minutes on OCP 4.15. Setting the deadline to 600s should be enough. The value 600s is same as upstream's default. After running this test, set the value back to default.

Fixes https://issues.redhat.com/browse/SRVCOM-2845

## Proposed Changes

-
-
-


mgencur:
@skonto Please review.

mgencur:
I have sent a similar one to Serving: https://github.com/openshift-knative/serving/pull/532

PR #2410, opened 2023-12-05
disable smcp addons for mesh tests

maschmid:
- :wastebasket: Remove optional smcp addons

It doesn't seem we need those, and it makes the mesh testsuite more stable, as some of the istio deployments (prometheus, primarily) cause OOMs on the nodes during the test runs.


ReToCode (approved):
/lgtm
/approve

Thanks, wanted to do this myself a while ago (If I install manually, I always disable them).

/hold for @pierDipi

maschmid:
The 413-operator-e2e-aws-ocp-413  fail is caused by https://redhat-internal.slack.com/archives/CGYQL9XPG/p1701935641674379  , will fix that in a separate PR

maschmid:
trying to fix the operator-e2e tests in https://github.com/openshift-knative/serverless-operator/pull/2411 first

maschmid:
rebased on top of https://github.com/openshift-knative/serverless-operator/pull/2411 fix

PR #2416, opened 2023-12-13
Fix replacement of CI images - align with new image names

mgencur:
Fixes JIRA #

## Proposed Changes

-
-
-

PR #2417, opened 2023-12-14
[DNM] Test install certmanager for upstream-e2e-kafka test

creydr:
Tests installing cert-manager before running upstream-e2e-kafka tests

/hold


creydr:
Tested in #2416

PR #2427, opened 2024-01-02
Option to disable HA during installation

mgencur:
This is related to periodic CI job on vSphere: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-knative-serverless-operator-main-4.13-vsphere-e2e-vsphere-ocp-413-continuous
The cluster is limited in resources and it often fails with insufficient CPU. 

This PR adds an option to reduce the size of Serverless deployments and free some resources. We might want to use it on other specific platforms in the future as well.

## Proposed Changes

-
-
-


ReToCode (approved):
/lgtm
/approve

mgencur:
@ReToCode Thanks. Unfortunately, there are a few tests which expect the num of replicas == 2 and this value is hardcoded there. This will need a few more changes.

mgencur:
/unhold

I tested this with HA replicas == 1 in https://github.com/openshift/release/pull/47140 and the tests passed.

mgencur:
Reported the kafka failure as https://issues.redhat.com/browse/SRVKE-1558 . It is also visible in https://github.com/openshift-knative/serverless-operator/pull/2429 and it's not related to this PR.

test/e2e/knative_eventing_test.go
```diff
@@ -44,16 +36,29 @@ func TestKnativeEventing(t *testing.T) {
...
+		}
+		if err := v1beta1.UpdateEventingExpectedScale(caCtx,
+			eventingNamespace, eventingNamespace, eventingDeployments, ptr.Int32(eventingHaReplicas)); err != nil {
+			t.Fatalf("Failed to update deployment scale: %v", err)
+		}
 		// Check the desired scale of deployments in the knative eventing namespace
-		for _, deployment := range []string{"eventing-controller", "eventing-webhook", "imc-controller", "imc-dispatcher", "mt-broker-controller"} {
-			if err := test.CheckDeploymentScale(caCtx, eventingNamespace, deployment, eventingHaReplicas); err != nil {
-				t.Fatalf("Failed to verify default HA settings for %q: %v", deployment, err)
+		for _, deployment := range eventingDeployments {
```
ReToCode, line 53: This now also included `mt-broker-filter` and `mt-broker-ingress`. Are they always there?

ReToCode (commented):
Other than that, LGTM.

test/e2e/knative_eventing_test.go
mgencur: Previously it was not there because this part was checking that the deployment had 2 replicas. The broker-filter and broker-ingress used to have just one replica. So, we were only checking that those deployments exist but didn't verify the scale==2. Recently, these were made HA as well. They have two replicas now.
With our KnativeKafka config these two are there.

PR #2429, opened 2024-01-04
Fix local build for Index image

mgencur:
## Proposed Changes

* Unify and reuse "build_image" function in scripts
* Align this function with CI where ci-operator accepts context dir and path to Dockerfile.
* Properly replace reference to registry.ci.openshift.org/ocp/4.14:base with quay variant which is publicly accessible.
* Remove unnecessary checking whether there are any files changed in the build directory via checksums. It is only an optimization for repeated builds against the same cluster but it might sometimes skip the build unnecessarily (e.g. when no files were changed under context dir but some base images were updated).

### DIFF
--- a/hack/lib/catalogsource.bash
+++ b/hack/lib/catalogsource.bash
@@ -44,7 +44,7 @@ function install_catalogsource {
 
     cat "$csv"
 
-    build_image "serverless-bundle" "${rootdir}/olm-catalog/serverless-operator"
+    build_image "serverless-bundle" "${rootdir}" "olm-catalog/serverless-operator/Dockerfile"
 
     logger.debug 'Undo potential changes to the CSV to not pollute the repository.'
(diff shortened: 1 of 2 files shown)


hack/lib/catalogsource.bash
```diff
@@ -103,43 +103,40 @@ EOF
   logger.success "CatalogSource installed successfully"
 }
 
-function build_image {
-  local name build_dir
+# Dockerfiles might include references to images that do not exit but CI operator
```
maschmid, line 106:
```suggestion
# Dockerfiles might include references to images that do not exist but CI operator
```

mgencur:
Reported the kafka failure as https://issues.redhat.com/browse/SRVKE-1558 . It is also visible in https://github.com/openshift-knative/serverless-operator/pull/2427 and it's not related to this PR.

PR #2455, opened 2024-01-22
Fix OLM index image generator

mgencur:
Fixes https://issues.redhat.com/browse/SRVCOM-2898

## Proposed Changes

- Index image offers last three releases for installation. Align this with olm-catalog/serverless-operator/index/Dockerfile which only places CSV from last three versions in the index (through running `/bin/opm`). Any versions older than last three releases were not installable anyway.
- Annotations now properly include the default channel and latest versioned channel, not any previous versioned channels that are not supposed to include the latest release
- Same applies for the bundle Dockerfile for `operators.operatorframework.io.bundle.channels.v1`
- Project yaml includes in `olm.channels.list` only the last three releases. The list was too long and having all possible versions in it didn't make sense as they were not installable. And it was incorrectly used for annotations and dockerfiles (see previous points)

### DIFF
--- a/olm-catalog/serverless-operator/Dockerfile
+++ b/olm-catalog/serverless-operator/Dockerfile
@@ -8,7 +8,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=serverless-operator
 LABEL operators.operatorframework.io.bundle.channel.default.v1="stable"
-LABEL operators.operatorframework.io.bundle.channels.v1="stable,stable-1.29,stable-1.30,stable-1.31,stable-1.32"
+LABEL operators.operatorframework.io.bundle.channels.v1="stable,stable-1.32"
 
 LABEL \
       com.redhat.component="openshift-serverless-1-serverless-operator-bundle-container" \
--- a/hack/cmd/bumpso/bumpso.go
+++ b/hack/cmd/bumpso/bumpso.go
@@ -85,11 +85,14 @@ func run() error {
 		"source": "serverless-operator",
 	})
 
-	channelsList, _, err := unstructured.NestedSlice(project, "olm", "channels", "list")
-	if err != nil {
-		return err
+	defaultChannel, _, _ := unstructured.NestedString(project, "olm", "channels", "default")
+
+	channelsList := []interface{}{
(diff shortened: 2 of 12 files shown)


mgencur:
/hold

I want to give it more testing locally.

mgencur:
Different channels now have different versions available, example:
image

olm-catalog/serverless-operator/Dockerfile
```diff
@@ -8,7 +8,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=serverless-operator
 LABEL operators.operatorframework.io.bundle.channel.default.v1="stable"
-LABEL operators.operatorframework.io.bundle.channels.v1="stable,stable-1.29,stable-1.30,stable-1.31,stable-1.32"
+LABEL operators.operatorframework.io.bundle.channels.v1="stable,stable-1.32"
```
dsimansk, line 11: I believe all channels should be still listed in the annotation. In this case `stable,stable-1.32,stable-1.31,stable-1.30`.

I recall @Kaustubh-pande mentioned that when introducing the current channel format.

mgencur: From [docs](https://olm.operatorframework.io/docs/tasks/creating-operator-bundle/):
```
The label operators.operatorframework.io.bundle.channels.v1 reflects the list of channels the bundle is subscribing to when added into an operator registry
```
Listing all would mean that e.g. serverless 1.32 will be in channel stable-1.30, but the Index won't have it in metadata so it won't be installable anyway. 
I'd like to know the reason why to keep all versions here.

mgencur: When I look at productization files there are just two channels listed: https://pkgs.devel.redhat.com/cgit/containers/openshift-serverless-operator-bundle/tree/Dockerfile?h=openshift-serverless-1.31-rhel-8#n11

Kaustubh-pande: @dsimansk we don't need all the channels there, it is something introducing a channel. The 1.32 release is part of stable and stable-1.32.

dsimansk: Agreed, the description makes more sense then. I might have a bit blurred memory of the channels purpose. All good!

PR #2473, opened 2024-02-01
Use ubi-minimal from CI registry

dsimansk:
There're several error in other projects that try to rebuild S-O on spot: 

https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift-knative_eventing-kafka-broker/967/pull-ci-openshift-knative-eventing-kafka-broker-release-next-414-test-e2e-aws-ocp-414/1752855085005672448#1:build-log.txt%3A1138

It seems the fails are intermittent either on ubi image pull or image stream creation. I have seen the same error on client midstream as well. Not happening on every run though. 

- Trying to verify if we can work with CI image of ubi to fix other project building from S-O base

### DIFF
--- a/knative-operator/Dockerfile
+++ b/knative-operator/Dockerfile
@@ -8,7 +8,7 @@ COPY . .
 ENV GOFLAGS="-mod=vendor"
 RUN go build -o /tmp/operator ${BASE}/knative-operator/cmd/manager
 
-FROM registry.access.redhat.com/ubi8/ubi-minimal
+FROM registry.ci.openshift.org/ocp/ubi-minimal:8
 USER 65532
 
 COPY --from=builder /tmp/operator /ko-app/operator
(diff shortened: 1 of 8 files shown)


mgencur:
I think we'll need to put back a modified version of [replace_images](https://github.com/openshift-knative/serverless-operator/commit/5d688143df9e207e507a89a8bd17f175b2932bdf#diff-d6ee23a8f1c82d66144020b6ba63537a521631d860104bf14f1a2a6e717eb1d3L110) function and use it for local builds (when running make images with DOCKER_REPO_OVERRIDE set; pulling from "ocp" namespace in CI registry only works in CI or with authfile).

mgencur:
@dsimansk Sent https://github.com/openshift-knative/serverless-operator/pull/2487 to fix local builds.

PR #2483, opened 2024-02-08
Fix cert-manager deployment availability check

creydr:
Currently we wait for the deployment `ready` condition of the cert-manager deployments. This fails, was deployments use the `available` condition (instead of `ready` - see [deployment conditions](https://github.com/kubernetes/kubernetes/blob/f07b47c3d1e596f53640285afb0d53a81fe6f259/pkg/apis/apps/types.go#L547-L560)):

```
error: timed out waiting for the condition on deployments/cert-manager-webhook
error: timed out waiting for the condition on deployments/cert-manager
```

from https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift-knative_eventing-kafka-broker/967/pull-ci-openshift-knative-eventing-kafka-broker-release-next-411-test-reconciler-aws-411/1755029368633036800

This PR addresses it and uses `--for condition=available` to check for availability of the deployment.


pierDipi (approved):
:man_facepalming: 

Thanks!

/lgtm
/approve

PR #2484, opened 2024-02-08
Remove toolchain directive

dsimansk:
Per title, ditching `toolchain` to avoid any havoc downstream. We can't guarantee `1.21.6` will be in prod env. 

/cc @ReToCode @pierDipi @skonto

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,8 +2,6 @@ module github.com/openshift-knative/serverless-operator
 
 go 1.21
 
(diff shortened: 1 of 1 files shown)


ReToCode (approved):
With "prod env" you mean build env, right? I'm fine with removing this for now, but we need to work on making sure this will eventually work - as this will be injected again and again locally.

/lgtm
/approve

PR #2486, opened 2024-02-12
Scale up cluster when running Kafka tests

pierDipi:
We don't have enough nodes to run Kafka tests, see https://github.com/openshift-knative/eventing-kafka-broker/pull/957#issuecomment-1938378652


pierDipi:
@mgencur @dsimansk

PR #2492, opened 2024-02-14
Provide spec.registry like other operator CRs

matzew:
## Proposed Changes

- Like upstream adding `registry` field from its `base` for providing image overrides

### DIFF
--- a/knative-operator/pkg/apis/operator/v1alpha1/knativekafka_types.go
+++ b/knative-operator/pkg/apis/operator/v1alpha1/knativekafka_types.go
@@ -38,6 +38,11 @@ type KnativeKafkaSpec struct {
 	// +optional
 	Logging *Logging `json:"logging,omitempty"`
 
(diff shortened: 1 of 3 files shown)


matzew:
/hold

let's see if we really want this

dsimansk:
Isn't that an override we don't want to change downstream in the product?

matzew:
> Isn't that an override we don't want to change downstream in the product?

That said, overriding the images on `KnativeServing` and `KnativeEventing` types is a supported feature.  However missing on the `KnativeKafka` CR. FWIW, I guess we are _not_ actually documenting that. I guess for a reason :sweat_smile: 

FWIW, the PR is on hold for now.

matzew:
/close

we are actually removing that filed from upstream:
https://github.com/openshift-knative/serverless-operator/blob/main/olm-catalog/serverless-operator/hack/002-serving-drop-unsupported-fields.patch

PR #2470, opened 2024-01-30
[SRVKS-1189] Generate istio net policies if we have to

skonto:
Fixes JIRA #SRVKS-1189

## Proposed Changes

-  By default when OSSM is installed we get (absence implies true):
```
  profiles:
  - default
  security:
    manageNetworkPolicy: true
```
In this case two default net OSSM policies are created [nets.txt](https://github.com/openshift-knative/serverless-operator/files/14097615/nets.txt), one for allowing access from openshift routes and another that allows access to pods with the following label set:
```
                "podSelector": {
                    "matchLabels": {
                        "maistra.io/expose-route": "true"
                    }
                },
````

In order to integrate with those default net policies we generate three extra Serverless network policies when istio is enabled:

   a) webhook
   b) net-istio-webhook
   c) allow-from-openshift-monitoring-ns

The first two are meant to allow access from the  Api server to the webhooks and the last one for Prometheus to be able to scrape the Serving pods.

- We this patch we provide the option to disable the generation of our net policies when user sets: `manageNetworkPolicy: false` . With no net policies traffic works as expected. Also enabling monitoring by setting in Serving CR (injects kube-rbac-proxy)
```
  config:
    observability:
      metrics.backend-destination: "prometheus"
```
and by setting in smcp:
```
  proxy:
    networking:
      trafficControl:
        inbound:
          excludedPorts:
          - 8444 # metrics
          - 8022 # serving: wait-for-drain k8s pre-stop hook
```
allows prometheus to scrape targets without any net policies.

image
In other words having no net policies relaxes restrictions.
- TODO: test midstream serving against removing the net policies (I think we don't have tests for istio there).

PR #2493, opened 2024-02-14
[Eventing] Generate istio net policies if we have to

pierDipi:
This is similar to https://github.com/openshift-knative/serverless-operator/pull/2470 but for Eventing


ReToCode (approved):
/lgtm
/approve

PR #2499, opened 2024-02-19
[RELEASE-1.32] Fix secret informer filtering transfomer

skonto:
Fixes JIRA #SRVKS-1179

## Proposed Changes

- Fixes secret informer filtering transformer for kourier 
- See discussion in https://issues.redhat.com/browse/SRVKS-1179

TODO: cherry-pick into main


mgencur:
/lgtm

The kitchensink upgrade tests are expeted to fail. They need a fix in project.yaml because a new version went live (1.31.1) so the upgrade sequence is not correct.

PR #2506, opened 2024-02-21
Use default teardown timeout for wathola which is now 60s

mgencur:
This aims to backport https://github.com/openshift-knative/eventing/commit/e1bb5051deccd8a8b3f23e9d72fed061f55650a2 and use it.

## Proposed Changes

- The receiver is actively waiting for the number of received events to be equal to the number delivered via "Finished" event. Only then it shuts down. Previously, there was a hard coded timeout which was much shorter and it could happen that all events were delivered much sooner or the timeout was not sufficient. 
- Update deps, especially `k8s.io/apimachinery` to align it with midstream repos
- Use generic sets as the old ones are deprecated (as a result of updating k8s.io/apimachinery)


mgencur:
The "Validate" check fails because it depends on another PR: https://github.com/openshift-knative/serverless-operator/pull/2570

PR #2513, opened 2024-02-22
[main] adding a little note for CURRENT_VERSION_IMAGES overrides

matzew:
## Proposed Changes

- :broom: adding a little note, when wondering about the `generated-files` target going nuts


pierDipi:
No need to run tests for this.

PR #2514, opened 2024-02-26
[SRVCOM-2167] Add Eventing dashboards to Developer View

pierDipi:
As per title.

Notes:
- I had to rename `eventNamespace` variable to just `namespace` so that ODC picks it up from the project (context variable)

Dashboard list:
image

One rendered dashboard:
image


knative-operator/deploy/resources/dashboards/eventing/grafana-dash-knative-eventing-broker.yaml
```diff
@@ -118,7 +119,7 @@ data:
           "tableColumn": "",
           "targets": [
             {
-              "expr": "sum(rate(mt_broker_ingress_event_count{namespace_name=~\"$eventNamespace\"}[1m]))",
+              "expr": "sum(rate(mt_broker_ingress_event_count{namespace_name=~\"$namespace\"}[1m]))",
```
skonto, line 122: Interesting could be that this is a convention, I think I hit that before but it is so long ago.

pierDipi: Yes, when I applied them initially without this change, it was showing another select option as `eventNamespace` and when renamed to `namespace` it remove that option as it is taking it from the top level project select option

skonto:
In the past odc folks reviewed the dashboards for the developer view. Should we ask them if it is ok to add them?

pierDipi:
We met with the ODC team yesterday and there is no concern to proceed with this change.

/cc @skonto @matzew

pierDipi:
@matzew @skonto can we merge this?

creydr (commented):
/lgtm

PR #2520, opened 2024-02-27
[main] Bump min K8s version to 1.25

ReToCode:
As per title.

### DIFF
--- a/olm-catalog/serverless-operator/project.yaml
+++ b/olm-catalog/serverless-operator/project.yaml
@@ -23,13 +23,13 @@ requirements:
         # The min version validation in `vendor/knative.dev/pkg/version/version.go`
         # is ignored as it is overridden by fake version via KUBERNETES_MIN_VERSION.
         # This value is used for CSV's min version validation.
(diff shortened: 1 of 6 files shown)


olm-catalog/serverless-operator/project.yaml
```diff
@@ -23,13 +23,13 @@ requirements:
...
+        minVersion: 1.25.0
     golang: '1.21'
     nodejs: 16.x
     ocpVersion:
-        min: '4.11'
-        max: '4.14'
-        label: 'v4.11'
+        min: '4.12'
+        max: '4.15'
+        label: 'v4.12'
```
matzew, line 32: these are the only changes, rest is from the `make generated-files`, right?

(asking b/c it is just one commit)

matzew (approved):
/lgtm
/approve

/hold
(you can never be sure w/ CI ... :trollface: )

olm-catalog/serverless-operator/project.yaml
ReToCode: Yepp

PR #2521, opened 2024-02-27
Pass --config param to prowcopy command in github action

mgencur:
This is for proper generation of CI jobs. The command needs to have access to the config so that it can see which tests are on-demand.

Follow-up to https://github.com/openshift-knative/hack/pull/124

## Proposed Changes

-
-
-


pierDipi (approved):
/lgtm
/approve

mgencur:
Merged as the changes are not actually tested in this CI.

PR #2523, opened 2024-02-27
eventing-core bump to 1.13

creydr:
As per title


olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
```diff
@@ -893,29 +893,29 @@ spec:
                       - name: "IMAGE_net-istio-webhook__webhook"
                         value: "registry.ci.openshift.org/openshift/net-istio-webhook:knative-v1.11"
                       - name: "IMAGE_eventing-controller__eventing-controller"
-                        value: "registry.ci.openshift.org/openshift/knative-eventing-controller:knative-v1.11"
+                        value: "registry.ci.openshift.org/openshift/knative-eventing-controller:knative-v1.13"
```
ReToCode, line 896: that should be 1.12, right?

creydr: We are planning to go directly to 1.13 to have OIDC and eventtype discovery in

ReToCode: 🤷  this could cause interesting problems. So operator will be 1.12, go deps (in S-O) will be 1.12 and actual running images 1.13? 

/cc @skonto @dsimansk

skonto: @creydr I think in the past you did skip version but only at the content level not real version is that true?

creydr: > I think in the past you did skip version but only at the content level not real version is that true?

Let me bring in @pierDipi and @matzew as I don't know the details how this was done the last time

dsimansk: > @creydr I think in the past you did skip version but only at the content level not real version is that true? Would not this confuse the customers as they would expect 1.12 and the usual Knative version upgrade policy?

Yep, that's what Matthias already mentioned. IMO the content was 1.13, but appeared as 1.12 on the surface level, i.e. image versions.

matzew: For the manifests, we have done that before. Shipping a newer version as `-1`

matzew: This is a test how this really goes. the operand is in theory indepdent from the operator (and APIs are stable)

pierDipi:
It's probably a real error due to EKB tests being still on 1.11, I believe we need to bump every component at once
```
request: validation failed: Channel.Spec.Subscribers changed by user system:admin which was not the system:serviceaccount:knative-eventing:eventing-controller service account: spec.subscribers
        {[]v1.SubscriberSpec}:
```

matzew:
@creydr can you rebase this?

creydr:
> @creydr can you rebase this?

I am currently working on #2527, so not sure if we need this anymore

PR #2529, opened 2024-02-29
Remove version check in KnativeEventing

creydr:
Currently the knative operator checks between upgrades/downgrades the already installed version and prevents skipping minor versions:

`Version migration is not eligible with message: not supported to upgrade or downgrade across multiple MINOR versions`

When the operator is managed via OLM this check is not needed (and can lead to problems, on intentional minor version skips).
Following our discussions in https://redhat-internal.slack.com/archives/CEXRYS5QC/p1708945005582059?thread_ts=1708599613.277839&cid=CEXRYS5QC and https://redhat-internal.slack.com/archives/CEXRYS5QC/p1709116187511639?thread_ts=1709044877.485049&cid=CEXRYS5QC, we remove the check for knative-eventing.

@pierDipi: we probably need to backport this to 1.32 too, to make downgrading multiple minor versions possible too.


pierDipi (approved):
/lgtm
/approve

PR #2542, opened 2024-03-05
[RELEASE-1.32] Add required label for trusted-ca ConfigMap

ReToCode:
Fixes JIRA SRVKS-1218

## Proposed Changes
- Adds a label to our injected CA cluster bundle to avoid operator reconciliation war on OCP 4.15
- Values have to line up with `cluster-network-operator` as in https://github.com/openshift/cluster-network-operator/pull/2111

/assign @skonto 
/assign @maschmid

### DIFF
--- a/knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
+++ b/knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
@@ -55,7 +55,7 @@ const (
 	// https://github.com/openshift/cluster-network-operator/pull/2111
 	// Otherwise, both operators will revert each other's changes.
 	trustedCAOwningAnnotationKey   = "openshift.io/owning-component"
-	trustedCAOwningAnnotationValue = "Networking / cluster-network-operator"
+	trustedCAOwningAnnotationValue = "Serverless Operator"
 
 	// certVersionKey is an annotation key used by the Serverless operator to annotate the Knative Serving
 	// controller's PodTemplate to make it redeploy on certificate changes.
--- a/knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
+++ b/knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
@@ -51,6 +51,12 @@ const (
 	// Docs: https://docs.openshift.com/container-platform/4.3/networking/configuring-a-custom-pki.html#certificate-injection-using-operators_configuring-a-custom-pki
 	trustedCAKey = "config.openshift.io/inject-trusted-cabundle"
 
+	// cluster-network-operator enforces the following annotation to be there
+	// https://github.com/openshift/cluster-network-operator/pull/2111
+	// Otherwise, both operators will revert each other's changes.
+	trustedCAOwningAnnotationKey   = "openshift.io/owning-component"
+	trustedCAOwningAnnotationValue = "Networking / cluster-network-operator"
+
 	// certVersionKey is an annotation key used by the Serverless operator to annotate the Knative Serving
 	// controller's PodTemplate to make it redeploy on certificate changes.
 	certVersionKey = socommon.ServingDownstreamDomain + "/mounted-cert-version"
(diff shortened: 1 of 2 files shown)


ReToCode:
>  error: build error: creating build container: copying system image from manifest list: parsing image configuration: fetching blob: received unexpected HTTP status: 502 Bad Gateway 

/retest

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
```diff
@@ -51,6 +51,12 @@ const (
 	// Docs: https://docs.openshift.com/container-platform/4.3/networking/configuring-a-custom-pki.html#certificate-injection-using-operators_configuring-a-custom-pki
 	trustedCAKey = "config.openshift.io/inject-trusted-cabundle"
 
+	// cluster-network-operator enforces the following annotation to be there
+	// https://github.com/openshift/cluster-network-operator/pull/2111
+	// Otherwise, both operators will revert each other's changes.
+	trustedCAOwningAnnotationKey   = "openshift.io/owning-component"
+	trustedCAOwningAnnotationValue = "Serverless Operator"
```
skonto, line 58: Should we use some other format like: "serverless-operator" see [here](https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/bindata/bootkube/manifests/secret-service-network-serving-signer.yaml#L10) for example.

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
```diff
@@ -271,7 +277,9 @@ func (r *ReconcileKnativeServing) ensureCustomCertsConfigMap(instance *operatorv
 	if err != nil {
 		return fmt.Errorf("error reconciling serviceCACM: %w", err)
 	}
-	trustedCACM, err := r.reconcileConfigMap(instance, certs.Name+"-trusted-ca", nil, map[string]string{trustedCAKey: "true"}, nil)
+	trustedCACM, err := r.reconcileConfigMap(instance, certs.Name+"-trusted-ca",
```
skonto, line 280: I guess that is the only cm we should take care right?

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
ReToCode: Is this defined somewhere? The one from cluster-network-operator has blanks and `-` in it. So not sure if this is a must or not. Eventing already did their PR with the same value.

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
ReToCode: I think so yes, as this is not a global thing.

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
maschmid: not sure where it's defined, but in https://github.com/openshift/cluster-network-operator/pull/2111/files#diff-1400085807145384b35f8566e22539835342354c8a55689cc4126ea2c66b122aR134 the comment says

"the jira component name for the cluster-network-operator"

We don't use OCPBUGS jira, so we can't really follow the same pattern anyway.

skonto:
Upgrade job failed with:
>
=== Failed
=== FAIL: test/upgrade TestServerlessUpgradeContinual/Run/KafkaBrokerContinualTest (997.96s)

ReToCode:
Running again is probably not going to help: https://redhat-internal.slack.com/archives/CD87JDUB0/p1709733070634629?thread_ts=1709711546.714569&cid=CD87JDUB0

knative-operator/pkg/controller/knativeserving/knativeserving_controller.go
ReToCode: As Eventing already merged like this, I think we can just keep it. It does not really matter anyway (nothing will display it).

dsimansk:
In case we want to proceed, adding lgtm. 

/lgtm

ReToCode:
Merging as everything (except the Kafka thing passed).

PR #2543, opened 2024-03-05
[release-1.32] Replaces 1.31.1 on 1.32

matzew:
## Proposed Changes

- :broom: Update the replace for latest 1.31.x release

PR #2546, opened 2024-03-06
[TEST ONLY] [release-1.32] Replaces 1.31.1 on 1.32 - only upgrades

mgencur:
Based on https://github.com/openshift-knative/serverless-operator/pull/2543, but only run upgrades, not downgrades.

## Proposed Changes

-
-
-

### DIFF
--- a/olm-catalog/serverless-operator/project.yaml
+++ b/olm-catalog/serverless-operator/project.yaml
@@ -6,7 +6,7 @@ project:
     # Also make sure to update values under `olm.previous` by copying from `olm.replaces` and `olm.skipRange`.
     version: 1.32.0
 olm:
-    replaces: 1.31.0
+    replaces: 1.31.1
     skipRange: '>=1.31.0 <1.32.0'
     channels:
         default: stable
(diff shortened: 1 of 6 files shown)


olm-catalog/serverless-operator/project.yaml
```diff
@@ -6,7 +6,7 @@ project:
     # Also make sure to update values under `olm.previous` by copying from `olm.replaces` and `olm.skipRange`.
     version: 1.32.0
 olm:
-    replaces: 1.31.0
+    replaces: 1.31.1
```
matzew, line 9: I am not really sure we need this.  as we have a range?

Usually it would be ok if something _like_ 1.31.1 comes out **_AFTER_** the 1.32 (think LTS)

mgencur: Right.

matzew: /close 
as discussed in the comments above
this should _NOT_ be needed, at all

PR #2547, opened 2024-03-06
[release-1.32] Include 1.31.1 in upgrade sequence for kitchensink tests

matzew:
The version must be included because after 1.31.0, OLM offers 1.31.1 and not 1.32.0


matzew:
/close

well this is just for kitchen sink. and that is not executed here.

PR #2553, opened 2024-03-07
Add kafka contract cleanup script

Cali0707:
This is a possible fix to the serverless upgrade test failures by calling a bash script to clean things up during the downgrade stage of the tests

## Proposed Changes

-
-
-


test/upgrade/installation/serverless.go
```diff
@@ -251,3 +260,16 @@ func setStorageToAlpha(ctx *test.Context, name string) error {
 		return err
 	})
 }
+
+func downgradeKafkaContractsWithScript() error {
+	c := exec.Command("bash", "-s", "-", "kafka-broker-brokers-triggers", "kafka-channel-channels-subscriptions", "kafka-sink-sinks", "kafka-source-dispatcher-0", "kafka-source-dispatcher-1")
```
creydr, line 265: Maybe only one thing for the script: Should we include the names of the main configmaps already in the script, so support would not need to know the names of the CMs?
And maybe we can pass the namespace as a parameter to the script, so we could also use this (in the rare case) of using the namespaced-broker.

Cali0707: Yeah we could change that!

creydr: Sry, I could have responded earlier: #2555

Cali0707:
Closing since we have #2554 merged (thanks @creydr !!!)

PR #2554, opened 2024-03-08
[release-1.32] Add bash script to downgrade kafka contracts and run as part of downgrade tests

creydr:
Same as #2553, but includes the linter fixes.

Credits go to @Cali0707!

/cc @Cali0707  @matzew


mgencur:
Hey, this looks great! It will make testing much easier. 

One question: The code modifies a config map in knative-eventing NS. Will this be later overwritten/reconciled by the operator to the original value? Or not. Thanks

matzew:
That is a GOOD question

creydr:
> One question: The code modifies a config map in knative-eventing NS. Will this be later overwritten/reconciled by the operator to the original value? Or not. Thanks

Not sure if I understood you correctly, but the downgraded kafka-controller knows only the left fields in the contract. Or which component do you mean to reconcile the configmaps?

I am only a bit wondering about the order we execute this. Currently we do this, before we know that the new/old (1.10) operator is fully running: 
https://github.com/openshift-knative/serverless-operator/blob/ca3472c69e61a924113addedd1e7b80317b54c9b/test/upgrade/installation/serverless.go#L196-L206

So IIUC we could have the case, that we update the configmaps while still the 1.11 kafka-controller running and reconciling the configmaps back to include the `trustBundles` fields :thinking: (but I saw @Cali0707 changing this in 8136b0df3597875b5062deb4d2c619e2a9c195c0 :shrug: )

matzew:
I think the question was, if the Serverless Operator (owing those CMs, b/c applying them), is not reverting the change of the script.

matzew:
But, didn't we also notice that even on the stuck situation, on a broker cluster, things start to work just fine again, if the reconcilation is triggered? e.g. as we provided manually a new `Broker` or `KafkaSink` on our servers?

When we did that the CM was completely untouched. So, I **_guess_**, that's fine?  But not sure

But we are not in a rush here to merge this

mgencur:
> But, didn't we also notice that even on the stuck situation, on a broker cluster, things start to work just fine again, if the reconcilation is triggered?

That's true. Anyway, the upgrade tests should show if this works. I'd probably run them at least twice on this PR.

matzew (approved):
/lgtm
/approve

PR #2557, opened 2024-03-08
:broom: Switch to Node 20 LTS

cardil:
## Proposed Changes

- :broom: Switch to Node 20 LTS


matzew:
@matejvasek  where do we generally stand w/ Node 20 ? 
I think in upstream we also have older nodes?

cardil:
> @matejvasek where do we generally stand w/ Node 20 ? I think in upstream we also have older nodes?

At serverless-operator we just use node for testing. I think it makes 100% sense to follow the current supported Node version from func.

cardil:
It turned out that the setup here, after we switched to RHEL8 as base for build image, was invalid: `v10.24.0`, see the [build #1767238324688261120 @ L1886](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift-knative_serverless-operator/2557/pull-ci-openshift-knative-serverless-operator-main-414-ui-e2e-aws-414/1767238324688261120#1:build-log.txt%3A1886)

Maybe we should switch to RHEL9, like what @mgencur is doing with https://github.com/openshift-knative/serving/pull/650. Also, the OCP is using the RHEL9 for a while now. Then, we can do `dnf install @node:20`, and install the RHEL provided node, instead of relying on nodesource.com' distribution. WDYT?

Something else isn't clear to me. I'm changing the Dockerfiles in this PR. But it seems the originals are being used for the CI, not the ones in this PR, see the [build-log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-knative_serverless-operator/2557/pull-ci-openshift-knative-serverless-operator-main-414-ui-e2e-aws-414/1767238324688261120/artifacts/build-logs/root-amd64.log). But, [our `openshift/release` configuration](https://github.com/openshift/release/blob/23c22818edd3a0c9dd492c246cde3d665e73ab10/ci-operator/config/openshift-knative/serverless-operator/openshift-knative-serverless-operator-main__414.yaml#L34-L36) says the project Dockerfile should be used, so I presume the current, pre-merged files, not after the merge. No?!

mgencur:
> Makes sense. But, that means there is no way to test the changes before the merge?

You can possibly create your own fork, create CI setup for it, and then send a PR for openshift/release targeting your fork. Some "formal" checks on the PR will fail but it will be possible to run "rehearsal" jobs.

cardil:
> @mgencur: We use RHEL8 because it's closer to what what is productized in the end.

Yes, but this is my point. To switch the base to RHEL 9 everywhere: in CI, testing, and productization. As I said. The OCP is being based on RHEL 9 for a couple of releases now, see: https://docs.openshift.com/container-platform/4.13/release_notes/ocp-4-13-release-notes.html#ocp-4-13-about-this-release

cardil:
> @mgencur: You can possibly create your own fork, create CI setup for it, and then send a PR for openshift/release targeting your fork. Some "formal" checks on the PR will fail but it will be possible to run "rehearsal" jobs.

That can be done in one PR? The _"create CI setup for it, and then send a PR for openshift/release targeting your fork"_?

If so, this is the same security thread, as running Dockerfiles from a PR. Most CI's allow to run a modified CI configuration, if it was posted by an approved contributor. Here it is even stranger, as the OpenShift CI/Prow supports [the `/ok-to-test` command](https://prow.k8s.io/command-help) for newcommers.

cardil:
@mgencur I did the exact test you proposed. Here you can see the successful job: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/49784/rehearse-49784-periodic-ci-cardil-serverless-operator-node20-main-414-ui-e2e-aws-414-c/1767657276690141184

cardil:
/test 414-test-upgrade-aws-414

Looks like infra to me. Either way. This isn't connected.

PR #2577, opened 2024-03-25
SRVCOM-2642 Include must-gather image in CSV relatedImages

mgencur:
Fixes https://issues.redhat.com/browse/SRVCOM-2642

## Proposed Changes

-
-
-


Kaustubh-pande:
I think we can cherry pick in 1.32 as well, with 1.32.1 it should get resolved.
/lgtm

mgencur:
Let's minimize the payload for 1.32.1. We can include this in 1.33. The fix is only required for OCP 4.16 and 1.32.1 is not supported there yet.

PR #2578, opened 2024-03-25
Bump code dependencies to 1.13

ReToCode:
⚠️ See the separated commits for easier review (and because the GH UI dies if you open without filter 🤷)
 
# Basic changes
- Bump version of dependants to 1.12 (pointing to midstream 1.12 branches which are 1.13). So this is aligned with the yamls.
- Bump operator to version 1.13 (as this is from upstream)
- Re-order the dependencies to update upstream deps first. Otherwise the script was failing, as eventing deps depend on newer upstream deps
- Removed K8s deps replacement, see conversation [here](https://redhat-internal.slack.com/archives/CD87JDUB0/p1710855618298849?thread_ts=1710845630.434839&cid=CD87JDUB0) and [here](https://redhat-internal.slack.com/archives/CEXRYS5QC/p1711362464033969)
- Updated the `006-version.patch` to new controller-runtime version
- Replaces the [other approach](https://github.com/openshift-knative/serverless-operator/pull/2573)

PR #2580, opened 2024-03-26
add my-restricted-sasl-user for EKB tests

maschmid:
- :broom: Add my-restricted-sasl-user for EKB tests


maschmid:
I see we're adding it already in https://github.com/openshift-knative/serverless-operator/pull/2578

PR #2581, opened 2024-03-26
Use promotion by name for operator images

mgencur:
## Proposed Changes

-
-
-

PR #2575, opened 2024-03-22
[release-1.32] Bump to 1.32.1

pierDipi:
Update metadata for patch release

Waiting for https://github.com/openshift/release/pull/50173

PR #2582, opened 2024-03-27
Use promotion by name for operator images

mgencur:
This brings some changes from the following two PRs into main:
* https://github.com/openshift-knative/serverless-operator/pull/2581
* https://github.com/openshift-knative/serverless-operator/pull/2575

## Proposed Changes

- Use promotion by name where the format is `registry.ci.openshift.org/knative/main:serverless-knative-operator` instead of  `registry.ci.openshift.org/knative/serverless-openshift-knative-operator:main`. The reason is: old images that are promoted by "tag" in CI are cleaned up if no CI config uses them whereas images promoted by name remain there.
- Use the new promotion names (release-X.Y.Z instead of openshift-serverless.vX.Y.Z)
- Make sure the catalogsource script fails if the sed command is not able to replace the bundle image. Replacing the image is crucial for correct functioning.

### DIFF
--- a/hack/lib/catalogsource.bash
+++ b/hack/lib/catalogsource.bash
@@ -64,8 +64,13 @@ function install_catalogsource {
     logger.debug "Create a backup of the index Dockerfile."
     cp "${rootdir}/${index_dorkerfile_path}" "${rootdir}/_output/bkp.Dockerfile"
 
-    # Replace the nightly bundle reference with the previously built bundle
-    sed -i "s_\(.*\)\(registry.ci.openshift.org/knative/${CURRENT_VERSION_IMAGES}:serverless-bundle\)\(.*\)_\1image-registry.openshift-image-registry.svc:5000/$OLM_NAMESPACE/serverless-bundle:latest\3_" "${rootdir}/${index_dorkerfile_path}"
+    # Replace bundle reference with previously built bundle
+    bundle="registry.ci.openshift.org/knative/release-${CURRENT_VERSION}:serverless-bundle"
+    if ! grep "${bundle}" "${rootdir}/${index_dorkerfile_path}"; then
+      logger.error "Bundle ${bundle} not found in Dockerfile."
+      return 1
+    fi
+    sed -i "s_\(.*\)\(${bundle}\)\(.*\)_\1image-registry.openshift-image-registry.svc:5000/$OLM_NAMESPACE/serverless-bundle:latest\3_" "${rootdir}/${index_dorkerfile_path}"
 
     build_image "serverless-index" "${rootdir}" "${index_dorkerfile_path}"
 
--- a/hack/lib/catalogsource.bash
+++ b/hack/lib/catalogsource.bash
@@ -65,7 +65,7 @@ function install_catalogsource {
     cp "${rootdir}/${index_dorkerfile_path}" "${rootdir}/_output/bkp.Dockerfile"
 
     # Replace the nightly bundle reference with the previously built bundle
-    sed -i "s_\(.*\)\(registry.ci.openshift.org/knative/openshift-serverless-v${CURRENT_VERSION}:serverless-bundle\)\(.*\)_\1image-registry.openshift-image-registry.svc:5000/$OLM_NAMESPACE/serverless-bundle:latest\3_" "${rootdir}/${index_dorkerfile_path}"
+    sed -i "s_\(.*\)\(registry.ci.openshift.org/knative/${CURRENT_VERSION_IMAGES}:serverless-bundle\)\(.*\)_\1image-registry.openshift-image-registry.svc:5000/$OLM_NAMESPACE/serverless-bundle:latest\3_" "${rootdir}/${index_dorkerfile_path}"
 
     build_image "serverless-index" "${rootdir}" "${index_dorkerfile_path}"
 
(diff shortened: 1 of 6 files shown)


hack/lib/catalogsource.bash
```diff
@@ -64,8 +64,13 @@ function install_catalogsource {
     logger.debug "Create a backup of the index Dockerfile."
     cp "${rootdir}/${index_dorkerfile_path}" "${rootdir}/_output/bkp.Dockerfile"
 
-    # Replace the nightly bundle reference with the previously built bundle
-    sed -i "s_\(.*\)\(registry.ci.openshift.org/knative/openshift-serverless-v${CURRENT_VERSION}:serverless-bundle\)\(.*\)_\1image-registry.openshift-image-registry.svc:5000/$OLM_NAMESPACE/serverless-bundle:latest\3_" "${rootdir}/${index_dorkerfile_path}"
+    # Replace bundle reference with previously built bundle
+    bundle="registry.ci.openshift.org/knative/release-${CURRENT_VERSION}:serverless-bundle"
```
pierDipi, line 68:
```suggestion
    bundle="registry.ci.openshift.org/knative/${CURRENT_VERSION_IMAGES}:serverless-bundle"
```

pierDipi: So that we don't need to hardcode `release-` prefix

mgencur: There's a problem with using CURRENT_VERSION_IMAGES. It will have a different value on "main" and on older branches. So, in the end it will replace the last line of the following snippet on "main" and the 3rd line when running on older branches
```
RUN /bin/opm render --skip-tls-verify -o yaml registry.ci.openshift.org/knative/openshift-serverless-v1.31.0:serverless-bundle \
      registry.ci.openshift.org/knative/openshift-serverless-v1.32.0:serverless-bundle \
      registry.ci.openshift.org/knative/release-1.32.1:serverless-bundle >> /configs/index.yaml || \
    /bin/opm render --skip-tls-verify -o yaml registry.ci.openshift.org/knative/openshift-serverless-v1.31.0:serverless-bundle \
      registry.ci.openshift.org/knative/openshift-serverless-v1.32.0:serverless-bundle \
      registry.ci.openshift.org/knative/serverless-bundle:main >> /configs/index.yaml
```
This will be problemmatic on "main" because it will replace the image but at the same time the first command before `||` might succeed and then we get unwanted result.

mgencur: Actually, with the code above on main branch it would not replace anything with CURRENT_VERSION_IMAGES

pierDipi: I don't fully get it

```
export CURRENT_VERSION_IMAGES=${CURRENT_VERSION_IMAGES:-"main"}
```

if `CURRENT_VERSION_IMAGES = main`

that becomes 

```
registry.ci.openshift.org/knative/main:serverless-bundle
```

or on a branch

```
registry.ci.openshift.org/knative/release-1.33.0:serverless-bundle
```

isn't that what we want?

pierDipi: But I'm fine with your approach anyway

mgencur:
Interesting, I'm getting this when installing the previous (1.32.0) version [link](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-knative_serverless-operator/2582/pull-ci-openshift-knative-serverless-operator-main-415-test-upgrade-aws-415/1772908033509167104/artifacts/test-upgrade-aws-415/openshift-gather-extra/artifacts/pods/openshift-serverless_knative-operator-webhook-5fb8f8c4db-thdt5_knative-operator.log)
```
{"level":"error","ts":"2024-03-27T09:41:10.330Z","logger":"cmd","msg":"","error":"failed to load KafkaChannel manifest: stat : no such file or directory","stacktrace":"main.main\n\t/go/src/github.com/openshift-knative/serverless-operator/knative-operator/cmd/manager/main.go:104\nruntime.main\n\t/usr/lib/golang/src/runtime/proc.go:267"}
```

hack/lib/catalogsource.bash
mgencur: Yes. The above is true. But it will replace different parts of the `/bin/opm render` command. When it does it on main branch and 1.32.1 already exists then the first `opm render` command will succeed and the resulting Index image won't have changes from the current PR.

dsimansk: I wonder, whether it's required to have exact patch version in the image name. Hence also having floating promotion image config.

Would a floating branch name worked the same? I guess it will brake patch version upgrade tests `x.y.0 -> x.y.1`. But once Serverless `x.y.0` is published into `registry.redhat.io`. In fact is used for upgrade test purposes.

My idea is if we can simplify patch release setup on the CI side with floating tags based on branch name.
 
```
registry.ci.openshift.org/knative/release-1.33:serverless-bundle
```

mgencur: I think @pierDipi was suggesting something similar: https://redhat-internal.slack.com/archives/CD87JDUB0/p1711535777573639?thread_ts=1711440425.458889&cid=CD87JDUB0
Worth exploring this option. It's true that we wouldn't be able to test upgrades across micro versions in CI, but we're anyway not testing it correctly now. We only bump the operators but the operands are same (i.e. 1.11 being used in 1.32.0 and 1.32.1).
I'd probably try it in a separate PR, with more testing. 
We need this PR to unblock other PRs.

mgencur: I did a few attempts but I haven't figured out how to omit the base z-stream (e.g. 1.31.0, or 1.30.0) and work only with latest micro release.
My attempts can be seen here: https://github.com/mgencur/serverless-operator/commits/attemts_upgrades_across_channels/
Last three commits have some changes and every commit includes an error message in the commit message.

PR #2583, opened 2024-03-27
Generate Index yaml entries for older channels correctly

mgencur:
For older channels, we ignore micro-releases and only use the initial version from those streams because we don't know how many micro release will be in that stream in advance.

Related to https://github.com/openshift-knative/serverless-operator/pull/2575/files#r1537601825

## Proposed Changes

-
-
-


pierDipi (approved):
/lgtm
/approve

/hold for tests

PR #2594, opened 2024-04-03
Restart OLM pods to prevent OCPBUGS-3139

ReToCode:
# Changes
- Temporarily delete OLM catalog pods if they are failing
- This can be removed once https://issues.redhat.com/browse/OCPBUGS-31391 is fixed

### DIFF
--- a/hack/lib/catalogsource.bash
+++ b/hack/lib/catalogsource.bash
@@ -10,6 +10,25 @@ function ensure_catalogsource_installed {
 }
 
 function install_catalogsource {
+  # Ensure the catalog source pods are running correctly
+  # to avoid https://issues.redhat.com/browse/OCPBUGS-31391.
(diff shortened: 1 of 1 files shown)


ReToCode:
Seems to look good. @pierDipi @mgencur a bit hacky, but I think this should help for now. PTAL.

mgencur:
@ReToCode Hmm. I think the workaround is applied too late. We might hit it earlier when installing Cert Manager:
```
 08:55:08.218 DEBUG:   [[ $(oc get deploy -n cert-manager cert-manager --no-headers | wc -l) != 1 ]] : Waiting until non-zero (max 600 sec.)
.........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
09:06:12.613 ERROR:   Time out of 600 exceeded 
```

PR #2604, opened 2024-04-05
Remove inactive people from reviewers

pierDipi:
As per title


pierDipi:
Thanks, merging manually as there is no need for tests!

PR #2616, opened 2024-04-18
switch strimzi from zookeeper to kraft

maschmid:
- :gift: Use kraft for strimzi by default

to get rid of flakiness in zookeeper startup

### DIFF
--- a/hack/lib/strimzi.bash
+++ b/hack/lib/strimzi.bash
@@ -20,89 +20,92 @@ function install_strimzi_operator {
 function install_strimzi_cluster {
   logger.info "Applying Strimzi Cluster file"
   cat <<-EOF | oc apply -f -
-    apiVersion: kafka.strimzi.io/v1beta2
-    kind: Kafka
-    metadata:
-      name: my-cluster
-      namespace: kafka
-    spec:
-      kafka:
-        version: 3.7.0
-        replicas: 3
-        listeners:
-          # PLAINTEXT
-          - name: plain
-            port: 9092
-            type: internal
-            tls: false
-          # SSL
-          - name: tls
-            port: 9093
-            type: internal
-            tls: true
-            authentication:
-              type: tls
-          # protocol=SASL_SSL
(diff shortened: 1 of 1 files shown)


hack/lib/strimzi.bash
```diff
@@ -20,89 +20,92 @@ function install_strimzi_operator {
...
+metadata:
+  name: kafka
+  namespace: kafka
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  replicas: 3
+  roles:
+    - controller
+    - broker
```
pierDipi, line 94: With this configuration, the 3 replicas will be both controllers and brokers, right?

hack/lib/strimzi.bash
```diff
@@ -20,89 +20,92 @@ function install_strimzi_operator {
...
+  storage:
+    type: jbod
+    volumes:
+    - id: 0
+      type: persistent-claim
+      size: 100Gi
+      deleteClaim: false
```
pierDipi, line 101: I see we use `false` just like the previous one, but what's the reason ?

maschmid: We've inherited that from https://github.com/strimzi/strimzi-kafka-operator/blob/main/examples/kafka/kafka-persistent.yaml , it defaults to `false` , so I assume it's just there in the example to show that it exists.

It doesn't really matter, we don't delete Kafka in CI, and the PVC (and its PV) would be deleted with the cluster anyway

creydr:
@maschmid @pierDipi do you want to have this in, before we cut 1.33?

`error=failed to load configuration: got unexpected http 404 status code from configresolver: failed to get config: could not find any config for branch main on repo openshift-knative/serverless-operator`

/retest

matzew:
I'd _not_ block the release for this - but its good to move towards this setup, in the next weeks. IMO

maschmid:
It doesn't block 1.33 , but feel free to merge  it right away.

pierDipi (approved):
/lgtm
/approve

PR #2643, opened 2024-05-09
Remove FIPS annotation

pierDipi:
This is the follow up to the temporary removal of the FIPS support for 1.33+ until we know when we can support it back

### DIFF
--- a/templates/csv.yaml
+++ b/templates/csv.yaml
@@ -3,7 +3,7 @@ kind: ClusterServiceVersion
 metadata:
   annotations:
     operatorframework.io/suggested-namespace: "openshift-serverless"
-    operators.openshift.io/infrastructure-features: '["disconnected", "fips", "proxy-aware"]'
+    operators.openshift.io/infrastructure-features: '["disconnected", "proxy-aware"]'
     operators.openshift.io/valid-subscription: '["OpenShift Container Platform", "OpenShift Platform Plus"]'
     alm-examples: |-
       [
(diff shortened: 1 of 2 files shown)


templates/csv.yaml
```diff
@@ -3,7 +3,7 @@ kind: ClusterServiceVersion
 metadata:
   annotations:
     operatorframework.io/suggested-namespace: "openshift-serverless"
-    operators.openshift.io/infrastructure-features: '["disconnected", "fips", "proxy-aware"]'
+    operators.openshift.io/infrastructure-features: '["disconnected", "proxy-aware"]'
     operators.openshift.io/valid-subscription: '["OpenShift Container Platform", "OpenShift Platform Plus"]'
```
Kaustubh-pande, line 7: Hi @pierDipi can you please also add below annotations,
  
```
features.operators.openshift.io/disconnected: "true"
features.operators.openshift.io/fips-compliant: "false"
features.operators.openshift.io/proxy-aware: "true"
features.operators.openshift.io/tls-profiles: "false"
features.operators.openshift.io/token-auth-aws: "false"
features.operators.openshift.io/token-auth-azure: "false"
features.operators.openshift.io/token-auth-gcp: "false"
```

Kaustubh-pande (changes requested):

templates/csv.yaml
pierDipi: I think those can be added as a separate PR

Kaustubh-pande: Okay, I'll open a PR with the above changes.

PR #2663, opened 2024-05-15
Incorporate new upstream Serving encryption

ReToCode:
Fixes JIRA https://issues.redhat.com/browse/SRVKS-1198

## Proposed Changes
- Install cert-manager always (as Downstream Serving tests now need it as well)
- Deploy Serving cert-manager issuers & certificates (aligned with upstream, we can't use the Eventing ones)
- Remove `internal-encryption` in favour of `system-internal-tls` and `cluster-local-domain-tls`
- Add a new test to make sure Serving encryption works
- Refactor test helpers to be shared

Please consider my comments for simpler review.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,6 @@ install-operator:
 
 install-all:
 	UNINSTALL_STRIMZI="false" ./hack/strimzi.sh
-	UNINSTALL_CERTMANAGER="false" ./hack/certmanager.sh
 	./hack/tracing.sh
 	SCALE_UP=4 INSTALL_KAFKA="true" ENABLE_TRACING=true ./hack/install.sh
 
@@ -33,23 +32,19 @@ install-serving-with-mesh:
 	MESH=true SCALE_UP=4 INSTALL_SERVING=true INSTALL_EVENTING="false" ./hack/install.sh
 
 install-eventing:
-	UNINSTALL_CERTMANAGER="false" ./hack/certmanager.sh
 	INSTALL_SERVING="false" ./hack/install.sh
 
 install-kafka:
-	UNINSTALL_CERTMANAGER="false" ./hack/certmanager.sh
 	SCALE_UP=4 INSTALL_SERVING="false" INSTALL_KAFKA="true" ./hack/install.sh
 
 install-kafka-with-mesh:
 	UNINSTALL_MESH="false" ./hack/mesh.sh
-	UNINSTALL_CERTMANAGER="false" ./hack/certmanager.sh
 	TRACING_BACKEND=zipkin TRACING_NAMESPACE=knative-eventing ./hack/tracing.sh
 	UNINSTALL_STRIMZI="false" ./hack/strimzi.sh
 	MESH=true SCALE_UP=5 INSTALL_SERVING=false INSTALL_EVENTING=true INSTALL_KAFKA=true TRACING_BACKEND=zipkin TRACING_NAMESPACE=knative-eventing ENABLE_TRACING=true ./hack/install.sh
 
 install-kafka-with-keda:
 	UNINSTALL_KEDA="false" ./hack/keda.sh
-	UNINSTALL_CERTMANAGER="false" ./hack/certmanager.sh
 	SCALE_UP=4 INSTALL_SERVING="false" INSTALL_KAFKA="true" ENABLE_KEDA="true" ./hack/install.sh
 
 install-strimzi:
@@ -75,12 +70,6 @@ install-previous:
 
 install-previous-with-kafka:
 	INSTALL_PREVIOUS_VERSION="true" INSTALL_KAFKA="true" ./hack/install.sh
-
-install-mesh:
-	UNINSTALL_MESH="false" ./hack/mesh.sh
-
-uninstall-mesh:
-	UNINSTALL_MESH="true" ./hack/mesh.sh
 
 install-mesh:
 	UNINSTALL_MESH="false" ./hack/mesh.sh
@@ -157,7 +146,6 @@ test-upstream-e2e-mesh-testonly:
 	MESH=true TEST_KNATIVE_KAFKA=false TEST_KNATIVE_SERVING=true TEST_KNATIVE_EVENTING=true TEST_KNATIVE_KAFKA_BROKER=true TEST_KNATIVE_UPGRADE=false ./test/upstream-e2e-tests.sh
 
 install-for-mesh-e2e:
-	UNINSTALL_CERTMANAGER="false" ./hack/certmanager.sh # This avoids early failures for the skipped Eventing TLS tests
 	UNINSTALL_MESH="false" ./hack/mesh.sh
 	TRACING_BACKEND=zipkin TRACING_NAMESPACE=knative-eventing ./hack/tracing.sh
 	UNINSTALL_STRIMZI="false" ./hack/strimzi.sh
@@ -186,7 +174,6 @@ test-upstream-e2e-no-upgrade: upstream-e2e
 
 upstream-e2e-kafka:
 	UNINSTALL_STRIMZI="false" ./hack/strimzi.sh
-	UNINSTALL_CERTMANAGER="false" ./hack/certmanager.sh
 	TRACING_BACKEND=zipkin ./hack/tracing.sh
 	SCALE_UP=5 INSTALL_KAFKA="true" TRACING_BACKEND=zipkin ENABLE_TRACING=true ./hack/install.sh
 	TEST_KNATIVE_KAFKA_BROKER=true TEST_KNATIVE_E2E=true TEST_KNATIVE_UPGRADE=false ./test/upstream-e2e-tests.sh
--- a/hack/install.sh
+++ b/hack/install.sh
@@ -19,5 +19,6 @@ dump_state.setup
 scale_up_workers
 create_namespaces "${SYSTEM_NAMESPACES[@]}"
 
+install_certmanager
 ensure_catalogsource_installed
 ensure_serverless_installed
--- a/hack/lib/certmanager.bash
+++ b/hack/lib/certmanager.bash
@@ -82,6 +82,7 @@ function sync_trust_bundle {
    for ns in "${namespaces[@]}"; do
      echo "Syncing trust-bundle for namespace: ${ns}"
      oc create namespace "${ns}" --dry-run=client -o yaml | oc apply -f -
+     oc label namespace "${ns}" knative.openshift.io/part-of="openshift-serverless" --overwrite
      oc create configmap -n "${ns}" knative-ca-bundle --from-file=tls.crt --from-file=ca.crt \
         --dry-run=client -o yaml | kubectl apply -n "${ns}" -f - || return $?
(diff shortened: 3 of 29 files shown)


Makefile
```diff
@@ -82,12 +77,6 @@ install-mesh:
 uninstall-mesh:
 	UNINSTALL_MESH="true" ./hack/mesh.sh
 
-install-mesh:
```
ReToCode, line 85: duplicate

hack/install.sh
```diff
@@ -19,5 +19,6 @@ dump_state.setup
 scale_up_workers
 create_namespaces "${SYSTEM_NAMESPACES[@]}"
 
+install_certmanager
```
ReToCode, line 22: we always need cert-manager now for Serving tests, so we can install it in any make target

hack/lib/certmanager.bash
```diff
@@ -39,22 +39,21 @@ function deploy_certmanager_operator {
   oc wait deployments -n ${deployment_namespace} cert-manager-webhook --for condition=available --timeout=600s
   oc wait deployments -n ${deployment_namespace} cert-manager --for condition=available --timeout=600s
 
+  # serving resources
```
ReToCode, line 42: generalized to be used by serving + eventing

serving/ingress/pkg/reconciler/ingress/resources/route_test.go
```diff
@@ -258,7 +259,7 @@ func TestMakeRoute(t *testing.T) {
 			name: "valid, passthrough by BYO cert",
 			ingress: ingress(
 				withTLS(networkingv1alpha1.IngressTLS{Hosts: []string{"custom.example.com"}, SecretName: "someSecretName"}),
-				withRules(rule(withHosts([]string{localDomain, externalDomain}))),
+				withRules(rule(withHosts([]string{"custom.example.com"}))),
```
ReToCode, line 262: Serving now expects the TLS hosts to be the same as the route rule hosts to figure out what is public/cluster-local, so aligned these.

test/e2e-tests.sh
```diff
@@ -15,12 +15,12 @@ dump_state.setup # test
 if [[ $MESH == "true" ]]; then
   # net-istio does not use knative-serving-ingress namespace.
   export INGRESS_NAMESPACE="knative-serving"
-else
-  trust_router_ca
 fi
 
 logger.success '🚀 Cluster prepared for testing.'
 
+trust_router_ca
```
ReToCode, line 22: we always need to trust the CA, also for mesh with TLS (not yet available downstream).

test/servinge2e/helpers.go
```diff
@@ -6,13 +6,15 @@ import (
...
+	corev1 "k8s.io/api/core/v1"
 	pkgTest "knative.dev/pkg/test"
 	"knative.dev/pkg/test/spoof"
+	servingv1 "knative.dev/serving/pkg/apis/serving/v1"
 	servingTest "knative.dev/serving/test"
 )
 
 const (
-	helloworldText = "Hello World!"
+	HelloworldText = "Hello World!"
```
ReToCode, line 17: was defined in multiple places, so making it public here

test/servinge2e/helpers.go
```diff
@@ -40,3 +42,26 @@ func MakeSpoofingClient(ctx *test.Context, url *url.URL) (*spoof.SpoofingClient,
 		true,
 		servingTest.AddRootCAtoTransport(context.Background(), ctx.T.Logf, &servingTest.Clients{KubeClient: ctx.Clients.Kube}, true))
 }
+
+// HTTPProxyService returns a knative service acting as "http proxy", redirects requests towards a given "host". Used to test cluster-local services
```
ReToCode, line 46: moved as its not just used in "servicemesh" tests.

test/servinge2e/servicemesh/helpers.go
```diff
@@ -25,20 +21,3 @@ type testCase struct {
 	usePrivateService bool                  // Whether http proxy should call target's service private service
 	checkResponseFunc spoof.ResponseChecker // Function to be used to check response
 }
-
-// HTTPProxyService returns a knative service acting as "http proxy", redirects requests towards a given "host". Used to test cluster-local services
-func HTTPProxyService(name, namespace, gateway, target string, serviceAnnotations, templateAnnotations map[string]string) *servingv1.Service {
```
ReToCode, line 30: moved to helper in a package one level up.

serving/ingress/pkg/reconciler/ingress/resources/route.go
```diff
@@ -144,9 +144,12 @@ func makeRoute(ci *networkingv1alpha1.Ingress, host string, rule networkingv1alp
 
 	// Target the HTTPS port and configure passthrough when:
 	// * the passthrough annotation is set.
-	// * the ingress.spec.tls is set. (DomainMapping with BYP cert.)
-	// * the internal-encryption is enabled.
-	if _, ok := annotations[EnablePassthroughRouteAnnotation]; ok || len(ci.Spec.TLS) > 0 || isInternalEncryptionEnabled(rule) {
+	// * the ingress.spec.tls is set for an external domain (e.g. DomainMapping with BYP cert.)
+	// * the destination service uses a https port.
+	if _, ok := annotations[EnablePassthroughRouteAnnotation]; ok ||
+		len(ci.GetIngressTLSForVisibility(networkingv1alpha1.IngressVisibilityExternalIP)) > 0 ||
```
skonto, line 150: Was not `len(ci.Spec.TLS) > 0` enough in the past?

ReToCode: Yes, but now we also have TLS for cluster-local-domains (visibility = cluster-local), then we should not do passthrough as the "external-gateway" part then still uses http. This basically filters out the new thing to be the same as before.

test/servinge2e/encryption/encryption_test.go
```diff
@@ -0,0 +1,69 @@
...
+	ksvc = test.WithServiceReadyOrFail(ctx, ksvc)
+
+	// Get cert-manager CA cert that signed the cluster-local-domain certs
+	ca, err := getCertManagerCA(ctx.Clients)
+	if err != nil {
+		t.Fatalf("Could not get cert-manager CA: %v", err)
+	}
+
+	// Check if the service is reachable with https on cluster-local
+	httpProxy := test.WithServiceReadyOrFail(ctx, servinge2e.HTTPProxyService(ksvc.Name+"-proxy", test.Namespace,
```
skonto, line 65: Is there http direct on cluster local? :thinking:

test/servinge2e/encryption/helper.go
```diff
@@ -0,0 +1,41 @@
...
+	"time"
+
+	"github.com/openshift-knative/serverless-operator/test"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"knative.dev/networking/pkg/certificates"
+)
+
+// this needs to match the ClusterIssuer in hack/lib/certmanager_resources/serving-ca-issuer.yaml.
```
skonto, line 15: s/this/This

test/servinge2e/helpers.go
skonto: If we returned the same text as upstream we could avoid this.

serving/ingress/pkg/reconciler/ingress/resources/route.go
skonto: @ReToCode what is needed to access the service if someone wants to use ca verification at the client side?
Does he need to get the certificate (self-signed in this case) from the corresponding ns? What happens if someone replaces the ingress cert as in here: https://docs.openshift.com/container-platform/4.15/security/certificates/replacing-default-ingress-certificate.html?

ReToCode: Depends on from where:
- Caller is outside of the cluster, it needs to trust the OCP ingress CA (the one you linked). Out of scope of Knative.
- Caller is inside the cluster, it needs to trust the CA that they put in cert-manager. As described in the docs.

ReToCode: I think the confusion comes from this:
- The pass-through here is only relevant for `system-internal-tls`, there we have some code in S-O that copies the OCP ingress cert (the same as OCP routers use) to kourier to make it still work. Not really sure why we did that in the first place, but it works for now.

test/servinge2e/encryption/encryption_test.go
ReToCode: http direct?

serving/ingress/pkg/reconciler/ingress/resources/route.go
skonto: > Caller is outside of the cluster, it needs to trust the OCP ingress CA (the one you linked). Out of scope of Knative.

If it is passthrough from what I read the same OCP cert should be presented at the Kourier ingress side for a TLS connection. no?

mgencur: > If it is passthrough from what I read the same OCP cert should be presented at the Kourier ingress side for a TLS connection, no? Thus the copy. @mgencur any background info on this one?

What backgroung info do you mean?

The certificate for OpenShift ingress is copied to the net-kourier-controller [here](https://github.com/openshift-knative/serverless-operator/pull/2663/files#diff-f2a6e099ca204caa0f720b4b7ac02b8f2e3f0d47029f9aa26ca95e24cc74fbadR104) so the end user can trust the OCP ingress certificate without change and still use the internal-encryption.

I think this is to simplify the process for the user as the user doesn't have to setup trust to any new certificates presented by the Kourier gateway. And Kourier gateway doesn't have to generate new certificates.

I guess this could be also achieved by generating new certificates for Kourier, then using a re-encrypt Route for Knative services and the certificate from Kourier could be automically setup for the Route.

mgencur: Using a re-encrypt route would involve one aditional decryption+encryption on the data path.

ReToCode: + we discussed yesterday, not sure if this helps with the http2 case. With passthrough, haproxy does not touch the http connection. I assume with reencrypt it would be touched, which could break it (I haven't tested that though).

test/servinge2e/encryption/helper.go
ReToCode: @skonto as this is just cosmetics, can we skip this to get this merged?

hack/install.sh
ReToCode: For Serving we now always have the one test making sure TLS stuff works, so we always need it. Also it has to be there first, so we can populate CRDs and the trust bundle upfront.

test/servinge2e/helpers.go
ReToCode: True, but separate topic.

test/servinge2e/encryption/encryption_test.go
skonto: redirect I mean :)

PR #2665, opened 2024-05-15
Apply OpenShift Ingress certificate to Kourier system-internal-tls is…

mgencur:
… used

Key "internal-encryption" is deprecated as part of https://github.com/knative/networking/pull/858

## Proposed Changes

- Apply OpenShift Ingress certificate when system-internal-tls key is defined in KnativeServing:
```
spec:
  config:
    network:
      system-internal-tls: Enabled
```

### DIFF
--- a/openshift-knative-operator/pkg/serving/kourier.go
+++ b/openshift-knative-operator/pkg/serving/kourier.go
@@ -9,6 +9,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/client-go/kubernetes/scheme"
+	networking "knative.dev/networking/pkg/config"
 	"knative.dev/operator/pkg/apis/operator/base"
 	"knative.dev/pkg/network"
 )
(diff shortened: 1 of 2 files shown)


mgencur:
OK. Seems like this will be done as part of https://github.com/openshift-knative/serverless-operator/pull/2663/files
Closing.

openshift-knative-operator/pkg/serving/kourier.go
```diff
@@ -101,7 +98,8 @@ func addKourierEnvValues(ks base.KComponent) mf.Transformer {
 	}
 
 	networkCM := ks.GetSpec().GetConfig()[networkCMName]
-	if encrypt := networkCM[InternalEncryptionKey]; strings.ToLower(encrypt) == "true" {
+	if strings.ToLower(networkCM[networking.SystemInternalTLSKey]) == "true" ||
```
ReToCode, line 101: I don't think we need to support the old key longer, we deprecated it in the docs. Also the new value is `enabled` and has a type that can be used. Check out https://github.com/openshift-knative/serverless-operator/pull/2663.

PR #2676, opened 2024-05-27
Bump Operator to 1.14

creydr:
Bump operator to 1.14.1 and update patches

### DIFF
--- a/olm-catalog/serverless-operator/manifests/operator_v1beta1_knativeeventing_crd.yaml
+++ b/olm-catalog/serverless-operator/manifests/operator_v1beta1_knativeeventing_crd.yaml
@@ -2192,6 +2192,12 @@ spec:
                         - type: string
                       description: An eviction is allowed if at least "minAvailable" pods selected by "selector" will still be available after the eviction, i.e. even in the absence of the evicted pod.  So for example you can prevent all voluntary evictions by specifying "100%".
                       x-kubernetes-int-or-string: true
+                    maxUnavailable:
+                      anyOf:
+                        - type: integer
+                        - type: string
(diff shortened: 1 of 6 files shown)


creydr:
@skonto @pierDipi could you PTAL?

olm-catalog/serverless-operator/manifests/operator_v1beta1_knativeserving_crd.yaml
```diff
@@ -2352,5 +2369,5 @@ spec:
       clientConfig:
         service:
           name: operator-webhook
-          namespace: default
+          namespace: knative-operator
```
pierDipi, line 2372: Is this change correct for us?

olm-catalog/serverless-operator/manifests/operator_v1beta1_knativeeventing_crd.yaml
```diff
@@ -2272,5 +2278,5 @@ spec:
       clientConfig:
         service:
           name: operator-webhook
-          namespace: default
+          namespace: knative-operator
```
pierDipi, line 2281: same question

olm-catalog/serverless-operator/manifests/operator_v1beta1_knativeserving_crd.yaml
ReToCode: Hm needs to be tested. IMHO we rewrite all namespaces to our own ones anyway. So probably yes?

pierDipi: it's managed by OLM ? https://github.com/openshift-knative/serverless-operator/blob/473a9c576ba29bea28c653791e9541798e9d9dfc/olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml#L1283-L1304

PR #2680, opened 2024-05-31
Only run Serving encryption tests for kourier

ReToCode:
## Proposed Changes
- Only run Serving encryption tests for Kourier.
- As we run istio tests with mesh enabled and encryption is not supported/needed in that mode.
- And as, encryption for istio downstream is currently not supported in general.
- More context [here](https://redhat-internal.slack.com/archives/CF5ANN61F/p1716965246865489)

### DIFF
--- a/test/lib.bash
+++ b/test/lib.bash
@@ -149,38 +149,38 @@ function downstream_serving_e2e_tests {
       --kubeconfigs "${kubeconfigs_str}" \
       --imagetemplate "${IMAGE_TEMPLATE}" \
       "$@"
+
+    # Enable Serving encryption (only supported on Kourier - at least for now)
+    configure_cm network system-internal-tls:enabled
+    configure_cm network cluster-local-domain-tls:enabled
+
+    logger.info "Restart controller to enable cert-manager integration"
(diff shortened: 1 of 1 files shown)


ReToCode:
`system-internal-tls` definitely makes no sense with istio. I'm currently testing if `cluster-local-domain-tls` would work (just to know), even though we do not plan on supporting that with istio either.

ReToCode:
Lint is unrelated, separate PR so

/override "validate/lint"

skonto:
> I'm currently testing if cluster-local-domain-tls would work (just to know), even though we do not plan on supporting that with istio either.

Is this tested by creating a separate gateway and a workload accessing with https the ingress gaetway, with Istio in mesh mode? Is not this a feature RHOAI wanted to solve the tls issue?

ReToCode:
@skonto @mgencur  PTAL

mgencur:
LGTM

skonto:
Maybe get the lint one in first? Then stamp this one?

ReToCode:
Should not matter, but we can 👍

PR #2706, opened 2024-06-12
Reduce duplication in Konflux setup

mgencur:
[Slack discussion](https://redhat-internal.slack.com/archives/CD87JDUB0/p1718193215608399)

## Proposed Changes

- Use pipelineRef instead of pipelineSpec and reference common Pipeline in for both "push" and "pull request" variants.


mgencur:
Cool. The changes are used in the Konflux run:
image
Removing the induced error now.

PR #2733, opened 2024-06-26
Workaround for OCPBUGS-35731 Repeat deleting service accounts

mgencur:
This is a workaround for https://issues.redhat.com/browse/OCPBUGS-35731
The bug description has references to CI runs that suffered this bug. It's pretty frequent.

The scope is as narrow as possible:
* Do this only downstream because the bug is on OCP
* Repeating deleting only service accounts, not other resources

## Proposed Changes

- Delete ServiceAccount if the first attempt failed.


mgencur:
/hold

For 416-operator-e2e-aws-416  to pass

PR #2749, opened 2024-07-03
Change kube rbac proxy's default log level to zero and make it configurable

### LINKED ISSUE #2740: Adjustable Logging Level for kube-rbac-proxy Container
Currently, the log level of the `kube-rbac-proxy` container is set to 10 (_--v=10_), which logs sensitive information such as tokens to the logs.

```
  kube-rbac-proxy:
    Image:         registry.redhat.io/openshift4/ose-kube-rbac-proxy@sha256:a6ea11d4df5b439e6ceaf8e9f8799a56aee1cfbfad774af968e01c1e175514c0
    Image ID:      registry.redhat.io/openshift4/ose-kube-rbac-proxy@sha256:a6ea11d4df5b439e6ceaf8e9f8799a56aee1cfbfad774af968e01c1e175514c0
    Port:          <none>
    Host Port:     <none>
    Args:
      --secure-listen-address=0.0.0.0:8444
      --upstream=http://127.0.0.1:9090/
      --tls-cert-file=/etc/tls/private/tls.crt
      --tls-private-key-file=/etc/tls/private/tls.key
      --logtostderr=true
      --http2-disable
      --v=10
```

I couldn't find a method to update the args information via Custom Resource (CR). It would be greatly beneficial if support could be added to adjust the log level of the `kube-rbac-proxy` container. Please let me know if you need any additional information.

skonto:
Fixes JIRA #

## Proposed Changes

- Fixes #2740, similar issue to https://github.com/kubernetes-sigs/kubebuilder/issues/2434.
- User can inject the desired log level at the corresponding operator CR as follows:
```
spec:
  config:
    logging:
      loglevel.kube-rbac-proxy: "10"
```

### DIFF
--- a/openshift-knative-operator/pkg/monitoring/rbac_proxy_injection.go
+++ b/openshift-knative-operator/pkg/monitoring/rbac_proxy_injection.go
@@ -18,9 +18,9 @@ import (
 )
 
 const (
-	RBACContainerName     = "kube-rbac-proxy"
-	rbacProxyImageEnvVar  = "IMAGE_KUBE_RBAC_PROXY"
-	KubeRbacProxyLogLevel = 0
+	RBACContainerName            = "kube-rbac-proxy"
+	rbacProxyImageEnvVar         = "IMAGE_KUBE_RBAC_PROXY"
+	DefaultKubeRbacProxyLogLevel = 0
 )
 
 var defaultKubeRBACProxyRequests = corev1.ResourceList{
@@ -33,7 +33,7 @@ func InjectRbacProxyContainer(deployments sets.Set[string], cfg base.ConfigMapDa
 		Requests: defaultKubeRBACProxyRequests,
 		Limits:   corev1.ResourceList{},
 	}
-	logLevel := KubeRbacProxyLogLevel
+	logLevel := DefaultKubeRbacProxyLogLevel
 	if cfg != nil {
 		if cfg["deployment"] != nil {
 			if cpuRequest, ok := cfg["deployment"]["kube-rbac-proxy-cpu-request"]; ok {
@@ -48,10 +48,10 @@ func InjectRbacProxyContainer(deployments sets.Set[string], cfg base.ConfigMapDa
 			if memLimit, ok := cfg["deployment"]["kube-rbac-proxy-memory-limit"]; ok {
 				resources.Limits["memory"] = resource.MustParse(memLimit)
 			}
-			if cfg["logging"] != nil {
-				if logLevelStr, ok := cfg["logging"]["loglevel.kube-rbac-proxy"]; ok {
-					logLevel, _ = strconv.Atoi(logLevelStr)
-				}
+		}
+		if cfg["logging"] != nil {
+			if logLevelStr, ok := cfg["logging"]["loglevel.kube-rbac-proxy"]; ok {
+				logLevel, _ = strconv.Atoi(logLevelStr)
 			}
 		}
 	}
--- a/openshift-knative-operator/pkg/monitoring/rbac_proxy_injection.go
+++ b/openshift-knative-operator/pkg/monitoring/rbac_proxy_injection.go
@@ -35,22 +35,22 @@ func InjectRbacProxyContainer(deployments sets.Set[string], cfg base.ConfigMapDa
 	}
 	logLevel := DefaultKubeRbacProxyLogLevel
 	if cfg != nil {
-		if cfg["deployment"] != nil {
-			if cpuRequest, ok := cfg["deployment"]["kube-rbac-proxy-cpu-request"]; ok {
+		if deploymentData := getCmDataforName(cfg, "config-deployment"); deploymentData != nil {
+			if cpuRequest, ok := deploymentData["kube-rbac-proxy-cpu-request"]; ok {
 				resources.Requests["cpu"] = resource.MustParse(cpuRequest)
 			}
-			if memRequest, ok := cfg["deployment"]["kube-rbac-proxy-memory-request"]; ok {
+			if memRequest, ok := deploymentData["kube-rbac-proxy-memory-request"]; ok {
 				resources.Requests["memory"] = resource.MustParse(memRequest)
 			}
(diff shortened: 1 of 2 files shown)


openshift-knative-operator/pkg/monitoring/rbac_proxy_injection.go
```diff
@@ -31,18 +33,26 @@ func InjectRbacProxyContainer(deployments sets.Set[string], cfg base.ConfigMapDa
...
+			}
+			if cpuLimit, ok := cfg["deployment"]["kube-rbac-proxy-cpu-limit"]; ok {
+				resources.Limits["cpu"] = resource.MustParse(cpuLimit)
+			}
+			if memLimit, ok := cfg["deployment"]["kube-rbac-proxy-memory-limit"]; ok {
+				resources.Limits["memory"] = resource.MustParse(memLimit)
+			}
+			if cfg["logging"] != nil {
+				if logLevelStr, ok := cfg["logging"]["loglevel.kube-rbac-proxy"]; ok {
+					logLevel, _ = strconv.Atoi(logLevelStr)
```
skonto, line 53: In the deployment override checking above we panic when input is bad, not sure if we should do this here for consistency :thinking:

openshift-knative-operator/pkg/monitoring/rbac_proxy_injection.go
```diff
@@ -31,18 +33,26 @@ func InjectRbacProxyContainer(deployments sets.Set[string], cfg base.ConfigMapDa
...
+			if memRequest, ok := cfg["deployment"]["kube-rbac-proxy-memory-request"]; ok {
+				resources.Requests["memory"] = resource.MustParse(memRequest)
+			}
+			if cpuLimit, ok := cfg["deployment"]["kube-rbac-proxy-cpu-limit"]; ok {
+				resources.Limits["cpu"] = resource.MustParse(cpuLimit)
+			}
+			if memLimit, ok := cfg["deployment"]["kube-rbac-proxy-memory-limit"]; ok {
+				resources.Limits["memory"] = resource.MustParse(memLimit)
+			}
+			if cfg["logging"] != nil {
```
ReToCode, line 51: This is not under the `deployment` right? Then should the if not be one level to the left?

openshift-knative-operator/pkg/monitoring/rbac_proxy_injection.go
ReToCode: No real opinion on this. Both ways are fine I think (as this is just logging).

openshift-knative-operator/pkg/monitoring/rbac_proxy_injection.go
```diff
@@ -17,8 +18,9 @@ import (
 )
 
 const (
-	RBACContainerName    = "kube-rbac-proxy"
-	rbacProxyImageEnvVar = "IMAGE_KUBE_RBAC_PROXY"
+	RBACContainerName     = "kube-rbac-proxy"
+	rbacProxyImageEnvVar  = "IMAGE_KUBE_RBAC_PROXY"
+	KubeRbacProxyLogLevel = 0
```
ReToCode, line 23: Maybe name this `DefaultKubeRbacProxyLogLevel`?

matzew:
@skonto  this is for all three CRs? `KnativeServing`, `KnativeEventing` and our custom `KnativeKafka` ?

openshift-knative-operator/pkg/monitoring/rbac_proxy_injection.go
pierDipi, line 51: There is also `config-logging` that can be used to change the `config-logging` ConfigMap

skonto: The convention in the CR is that we omit the prefix afaik. We don't use the `config-`.

openshift-knative-operator/pkg/monitoring/rbac_proxy_injection.go
```diff
@@ -31,18 +33,26 @@ func InjectRbacProxyContainer(deployments sets.Set[string], cfg base.ConfigMapDa
...
+				resources.Limits["memory"] = resource.MustParse(memLimit)
+			}
 		}
-		if memLimit, ok := cfg["deployment"]["kube-rbac-proxy-memory-limit"]; ok {
-			resources.Limits["memory"] = resource.MustParse(memLimit)
+		if cfg["logging"] != nil {
+			if logLevelStr, ok := cfg["logging"]["loglevel.kube-rbac-proxy"]; ok {
+				logLevel, _ = strconv.Atoi(logLevelStr)
+			}
 		}
```
pierDipi, line 56: There is also `config-logging`, in addition to just `logging` that can be used to change the `config-logging` ConfigMap

skonto:
>  this is for all three CRs? 

It should be as `InjectRbacProxyContainer` is called from these components.

openshift-knative-operator/pkg/monitoring/rbac_proxy_injection.go
pierDipi: But it's still possible https://github.com/knative/operator/blob/df34e6e004acb96e8f0d867f0a2920bd3829cf56/pkg/reconciler/common/config_maps.go#L27-L41

openshift-knative-operator/pkg/monitoring/rbac_proxy_injection.go
```diff
@@ -131,3 +141,14 @@ func ExtensionDeploymentOverrides(overrides []base.WorkloadOverride, deployments
 	}
 	return operator.OverridesTransform(ovs, nil)
 }
+
+func getCmDataforName(cfg base.ConfigMapData, name string) map[string]string {
+	if cfg[name] != nil {
```
skonto, line 146: Priority is the same as in the knative operator, see reference from @pierDipi

openshift-knative-operator/pkg/monitoring/rbac_proxy_injection.go
skonto: ok fixed it although was not sure what we support downstream.

PR #2751, opened 2024-07-03
tempo tracing backend

maschmid:
Fixes JIRA [SRVCOM-3168](https://issues.redhat.com/browse/SRVCOM-3168)

## Proposed Changes

- :gift: Add tempo tracing backend and make it default

### DIFF
--- a/hack/lib/tracing.bash
+++ b/hack/lib/tracing.bash
@@ -8,8 +8,13 @@ function install_tracing {
       dedicate_node_to_zipkin
     fi
     install_zipkin_tracing
-  else
+  elif [[ "${TRACING_BACKEND}" == "tempo" ]]; then
+    install_tempo_tracing
+  elif [[ "${TRACING_BACKEND}" == "otel" ]]; then
     install_opentelemetry_tracing
+  else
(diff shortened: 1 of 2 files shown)


maschmid:
It's weird becase there is apparently no TraceQueryParameters parameter to make the `limit=0` be anything else than 0 ... 

We might need to swith to using the http query interface instead of the grpc one...

maschmid:
OK, so apparently the `SearchDepth` field is used as the `limit` in the query, so it might actually work after all... 

```
level=info ts=2024-07-15T09:28:02.978075258Z caller=searchsharding.go:254 msg="sharded search query request stats and SearchMetrics" tenant=single-tenant query="end=1721035682&limit=1000&maxDuration=1m0s&minDuration=0s&start=1721035082&tags=service.name=testtracestartedatactivator-00001-deployment-9c895bd4b-cq94r+name=/" duration_seconds=6.729543ms request_throughput=2.93541478225193e+06 total_requests=1 started_requests=1 cancelled_requests=0 finished_requests=1 totalBlocks=0 inspectedBytes=19754 inspectedTraces=110 totalBlockBytes=0
```

PR #2766, opened 2024-07-09
Enable pprof support in ctrl runtime

skonto:
## Proposed Changes

- We recently had an issue that could have needed pprof support see [discussion](https://stackoverflow.com/questions/64057727/is-it-ok-to-use-golang-pprof-on-production-without-effecting-performance).
- Right now pprof is not configurable in knative-openshift deployment
- ctrl runtime is not configurable via a cm, there was a [design](https://github.com/kubernetes-sigs/controller-runtime/blob/main/designs/component-config.md) in the past but it seems not done. Ideally we would like to set up things the Knative way at some point: install a handler that depending on a cm value uses the pprof handler or some dummy that returns 403. The latter allows to dynamically switch on/off pprof support and get the data at the time of the issue without pod restart.
- Here we re-use the pprof support ctrl manager which is static thus we need a restart. That is ok for now given the low number of issues we have faced in practice and that pod restart would work for most cases (but would require some time for the issue to re-appear once a pod is restarted).
- [Probably it would be ok](https://stackoverflow.com/questions/64057727/is-it-ok-to-use-golang-pprof-on-production-without-effecting-performance) even to have it on by default but let's not be that brave for now.
- User can enable this temporarily by editing the csv in case of debugging a case.
- Pprof is exposed only locally for security reasons, available via port-forward (but could be open on ":8008" as Knative if we want to be less restrictive).
- See sample heap profile [here](https://github.com/openshift-knative/serverless-operator/assets/7945591/9910e314-91d5-4cbd-ae60-6ab990dd70d5). It shows that ctrl runtime `dynamicclient` consumes most part of memory during cache initialization.
-  Tested it locally by editing the csv (pod is restarted automatically with the right env var value set).

### DIFF
--- a/knative-operator/cmd/manager/main.go
+++ b/knative-operator/cmd/manager/main.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"strconv"
 	"time"
 
 	"github.com/openshift-knative/serverless-operator/knative-operator/pkg/apis"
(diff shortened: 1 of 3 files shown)


ReToCode (approved):
/lgtm
/approve

/hold for @matzew

skonto:
@matzew gentle ping

matzew (approved):
/lgtm
/approve

PR #2774, opened 2024-07-17
Use spot instances for periodic runs in CI/AWS

mgencur:
Fixes https://issues.redhat.com/browse/SRVCOM-2911

## Proposed Changes

- Following the tutorial for setting up spot instances from https://docs.google.com/document/d/1qEQEggDJhJY9IzymwmbmekZZpbLtT2fK34L986QphRk/edit
- The goal is to use "spot" instances in CI only for periodic runs for now. If the stability is good enough we can think of using them for pre-submit phase as well.

> They cost 70-90% less than On-Demand VMs. (NB: Spot instances do not receive discounts from Savings Plans)
AWS may reclaim the capacity at any time (though in practice 95% of instances last at least 2 hours, and 90% last longer than 8 hours). AWS gives a 2-minute warning before an interruption happens, so that the instance may cleanly terminate/hibernate.

### DIFF
--- a/hack/dependencies.sh
+++ b/hack/dependencies.sh
@@ -16,6 +16,5 @@ set -Eeuo pipefail
 
 debugging.setup
 
-use_spot_instances
 scale_up_workers
 create_namespaces "${SYSTEM_NAMESPACES[@]}"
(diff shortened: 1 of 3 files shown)


ReToCode (approved):
/lgtm
/approve

PR #2329, opened 2023-10-13
SRVCOM-2745: bump to next patch release 1.30.2

dsimansk:
Fixes JIRA #SRVCOM-2745

## Proposed Changes

- SRVCOM-2745: bump to next patch release 1.30.2

/cc @Kaustubh-pande @mgencur @ReToCode @matzew

PR #2786, opened 2024-07-23
[release-1.32] Bump to 1.32.2

pierDipi:
As per title

After https://github.com/openshift/release/pull/54689

### DIFF
--- a/olm-catalog/serverless-operator/index/configs/index.yaml
+++ b/olm-catalog/serverless-operator/index/configs/index.yaml
@@ -4,24 +4,24 @@ name: stable
 package: serverless-operator
 entries:
   - name: "serverless-operator.v1.31.0"
-  - name: "serverless-operator.v1.32.1"
+  - name: "serverless-operator.v1.32.0"
     replaces: "serverless-operator.v1.31.0"
-    skipRange: ">=1.31.0 <1.32.1"
+    skipRange: ">=1.31.0 <1.32.0"
   - name: "serverless-operator.v1.32.2"
-    replaces: "serverless-operator.v1.32.1"
-    skipRange: ">=1.32.1 <1.32.2"
+    replaces: "serverless-operator.v1.32.0"
+    skipRange: ">=1.32.0 <1.32.2"
 ---
 schema: olm.channel
 name: stable-1.32
 package: serverless-operator
 entries:
   - name: "serverless-operator.v1.31.0"
-  - name: "serverless-operator.v1.32.1"
+  - name: "serverless-operator.v1.32.0"
     replaces: "serverless-operator.v1.31.0"
-    skipRange: ">=1.31.0 <1.32.1"
+    skipRange: ">=1.31.0 <1.32.0"
   - name: "serverless-operator.v1.32.2"
-    replaces: "serverless-operator.v1.32.1"
-    skipRange: ">=1.32.1 <1.32.2"
+    replaces: "serverless-operator.v1.32.0"
+    skipRange: ">=1.32.0 <1.32.2"
 ---
 schema: olm.channel
 name: stable-1.31
--- a/olm-catalog/serverless-operator/index/configs/index.yaml
+++ b/olm-catalog/serverless-operator/index/configs/index.yaml
@@ -4,24 +4,24 @@ name: stable
 package: serverless-operator
 entries:
   - name: "serverless-operator.v1.31.0"
-  - name: "serverless-operator.v1.32.0"
-    replaces: "serverless-operator.v1.31.0"
-    skipRange: ">=1.31.0 <1.32.0"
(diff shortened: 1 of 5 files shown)


olm-catalog/serverless-operator/index/configs/index.yaml
```diff
@@ -4,24 +4,24 @@ name: stable
...
-    replaces: "serverless-operator.v1.31.0"
-    skipRange: ">=1.31.0 <1.32.0"
   - name: "serverless-operator.v1.32.1"
-    replaces: "serverless-operator.v1.32.0"
-    skipRange: ">=1.32.0 <1.32.1"
+    replaces: "serverless-operator.v1.31.0"
+    skipRange: ">=1.31.0 <1.32.1"
+  - name: "serverless-operator.v1.32.2"
+    replaces: "serverless-operator.v1.32.1"
+    skipRange: ">=1.32.1 <1.32.2"
```
mgencur, line 12: TBH, I'm not sure which version should be replaced here.  think this needs to be `skipRange: ">=1.32.0 <1.32.2"`. For 1.30.0 we did something different. See [here](https://github.com/openshift-knative/serverless-operator/commit/db0039bef8b2ea4a55cce305e66366c1f5d02765#diff-55554988a6ad1a56cfb485f9de15784449897484786cdbfcd8c3ffbb97be92e1R12)

olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
```diff
@@ -1394,5 +1394,5 @@ spec:
       image: "registry.ci.openshift.org/openshift/knative-eventing-kafka-broker-post-install:knative-v1.11"
     - name: "KAFKA_IMAGE_knative-kafka-storage-version-migrator__migrate"
       image: "registry.ci.openshift.org/openshift/knative-eventing-migrate:knative-v1.11"
-  replaces: serverless-operator.v1.32.0
-  version: 1.32.1
+  replaces: serverless-operator.v1.32.1
```
mgencur, line 1397: Not sure if the "replaces" thing should stay at 1.32.0 or should be bumped. We did a different thing with 1.30.2: https://github.com/openshift-knative/serverless-operator/commit/db0039bef8b2ea4a55cce305e66366c1f5d02765#diff-55554988a6ad1a56cfb485f9de15784449897484786cdbfcd8c3ffbb97be92e1R12

olm-catalog/serverless-operator/index/configs/index.yaml
mgencur: From [this doc](https://olm.operatorframework.io/docs/concepts/olm-architecture/operator-catalog/creating-an-update-graph/):

> If you want to have direct (one hop) updates to this version from multiple previous versions but keep those previous versions available to users for install, always use skipRange in conjunction with replaces pointing to the immediate previous version of the operator version in question.

I think this line should be `skipRange: ">=1.32.0 <1.32.2"` but the replaces should be what it is now: `serverless-operator.v1.32.1`

mgencur: Actually, I would do the same as we did for 1.30.2 in the [PR](https://github.com/openshift-knative/serverless-operator/pull/2329) for 1.30.2: Keep `olm.replaces` in project.yaml at 1.32.0 and regenerate all files.
This part should then look like this:
```
schema: olm.channel
name: stable
package: serverless-operator
entries:
   - name: "serverless-operator.v1.31.0"
   - name: "serverless-operator.v1.32.0"
     replaces: "serverless-operator.v1.31.0"
     skipRange: ">=1.31.0 <1.32.0"
   - name: "serverless-operator.v1.32.2"
     replaces: "serverless-operator.v1.32.0"
     skipRange: ">=1.32.0 <1.32.2"
```

mgencur: This will also align this with the skipRange in CSV: https://github.com/openshift-knative/serverless-operator/pull/2786/files#diff-5a9371d1cdfd376365aba4ae0d4297d8d9d4e0fd85cc6d8886a4b79059b66f73R71

mgencur: Historically, we have used the strategy that first version of each stream is installable (e.g. 1.30.0) and then the last micro-release (i.e. 1.30.2). Versions between these two are not directly installable:
* 1.30.0 - installable
* 1.30.1 - not installable
* 1.30.2 - installable
This is achieved by `skipRange: ">=1.30.0 <1.30.2"` and `replaces: "serverless-operator.v1.30.0"`. 
If we used `replaces: "serverless-operator.v1.30.1"` then 1.30.1 would be installable as well.

PR #2788, opened 2024-07-23
[release-1.33] Bump to 1.33.2

pierDipi:
As per title

After https://github.com/openshift/release/pull/54681


mgencur:
Let's do the same as in https://github.com/openshift-knative/serverless-operator/pull/2786. Use 1.33.0 for "replaces"

pierDipi:
Done

PR #2792, opened 2024-07-24
[SRVCOM-3200] Bring up the hookserver

skonto:
Fixes JIRA #https://issues.redhat.com/browse/SRVCOM-3200

## Proposed Changes

- See discussion [here](https://redhat-internal.slack.com/archives/CD87JDUB0/p1721825624524249?thread_ts=1721816918.734599&cid=CD87JDUB0)
- Tested and now when trying to create a new CR it fails:

> $ oc apply ...
namespace/test4 created
Error from server (Forbidden): error when creating "wrong.yaml": admission webhook "validating.knativeservings.operator.serverless.openshift.io" denied the request: KnativeServing may only be created in knative-serving namespace

### DIFF
--- a/knative-operator/cmd/manager/main.go
+++ b/knative-operator/cmd/manager/main.go
@@ -135,6 +135,7 @@ func main() {
 
 	decoder := admission.NewDecoder(mgr.GetScheme())
 
+	// This call adds the server to the manager as a Runnable
 	_ = mgr.GetWebhookServer()
 	// Serving Webhooks
 	hookServer.Register("/mutate-knativeservings", &webhook.Admission{Handler: knativeserving.NewConfigurator(decoder)})
--- a/knative-operator/cmd/manager/main.go
+++ b/knative-operator/cmd/manager/main.go
@@ -135,6 +135,7 @@ func main() {
 
 	decoder := admission.NewDecoder(mgr.GetScheme())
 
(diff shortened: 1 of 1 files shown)


knative-operator/cmd/manager/main.go
```diff
@@ -135,6 +135,8 @@ func main() {
 
 	decoder := admission.NewDecoder(mgr.GetScheme())
 
+	// This call adds the server to the manager as a Runnable
+	_ = mgr.GetWebhookServer()
```
skonto, line 139:
```
func (cm *controllerManager) GetWebhookServer() webhook.Server {
	cm.webhookServerOnce.Do(func() {
		if cm.webhookServer == nil {
			panic("webhook should not be nil")
		}
		if err := cm.Add(cm.webhookServer); err != nil {
			panic(fmt.Sprintf("unable to add webhook server to the controller manager: %s", err))
		}
	})
	return cm.webhookServer
}
```

Without this webhook server is not available.

skonto: We used to have that: https://github.com/openshift-knative/serverless-operator/blob/release-1.26/knative-operator/cmd/manager/main.go#L105.

ReToCode (approved):
/lgtm
/approve

aliok:
/lgtm

Thanks @skonto

PR #2796, opened 2024-07-30
Bump Golang version to 1.22 in CI image

dsimansk:
Fixes JIRA #

## Proposed Changes

- Bump Golang version to 1.22 in CI image

/cc @ReToCode @pierDipi @mgencur

### DIFF
--- a/.github/workflows/bump.yaml
+++ b/.github/workflows/bump.yaml
@@ -20,9 +20,7 @@ jobs:
       GOPATH: ${{ github.workspace }}
     steps:
       - name: Setup Golang
-        uses: actions/setup-go@v5
-        with:
-          go-version: 1.21.x
+        uses: openshift-knative/hack/actions/setup-go@main
 
       - name: Checkout openshift-knative/serverless-operator
         uses: actions/checkout@v4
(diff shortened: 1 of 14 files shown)


ReToCode:
Probably also need to bump all the GH actions?

dsimansk:
> Probably also need to bump all the GH actions?

Let's try to share that version across repos: https://github.com/openshift-knative/hack/pull/189

ReToCode (approved):
/lgtm
/approve

ReToCode:
> Successfully set up Go version 1.21.x

There is one more place it seems.

ReToCode:
As per @pierDipi comment, maybe also bump it in the go.mod file?

dsimansk:
> As per @pierDipi comment, maybe also bump it in the go.mod file?

Ahh, to prevent `toolchain` directive addition through Dependabot PRs, right?

ReToCode:
> Ahh, to prevent toolchain directive addition through Dependabot PRs, right?

Yes but also because we read it in hack and use it as the go version (as far as I understood)

ReToCode:
I meant https://github.com/openshift-knative/hack/blob/f2602e4241ac18b3da338788e15701c0610bd413/cmd/generate/generate.go#L160, but probably not relevant for S-O as we generate the Dockerfiles here from the template. But in Serving and Co we should align the go version in go.mod as well.

/lgtm
/approve

PR #2803, opened 2024-08-05
Index metadata generation for Kitchensink upgrade tests

mgencur:
Install all versions for upgrade tests from the currently built CatalogSource named "serverless-operator" and not from redhat-operators CatalogSource.
This fixes a number of recurring problems in kitchensink upgrade tests:
* When Serverless z-stream is not published to OCP's that are EOL ([example failure here](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-knative-serverless-operator-main-412-kitchensink-upgrade-aws-412-c/1820274369956417536)) - on OCP 4.12 the Serverless 1.33.1 was not published while on OCP 4.15 it was. This has the effect that different `upgrade_sequence`s are required for different OCP versions, which is hard to maintain.
* When z-stream is actually published then the upgrade_sequence is immediately out of date and tests fail. [example here](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-knative-serverless-operator-main-415-kitchensink-upgrade-aws-415-c/1817757401651089408). This requires a PR every time we publish a micro version.

## Proposed Changes

-  Do not use redhat-operator catalog source anymore which is out of our control. Use the test catalog source that we build in CI.
- Remove unused elements from the project.yaml (`upgrade_sequence`, `olm.previous`)
- Put same entries in the index image yaml for all channels. The purpose is to be able to switch channels. The older channels are otherwise unused. It is enough to have `stable`, and `stable-X.Y`.
- The generators are now able to produce index image yaml and index dockerfile with any number of bundles/csvs included. This is determined by INDEX_IMAGE_NUM_CSVS variable.


mgencur:
/hold

For kitchensink tests to successfully install all versions.

mgencur:
/hold
This needs a fix

mgencur:
/unhold

The Kitchensink upgrade tests fail with a common problem that we have been seeing in other runs. The Broker-related tests can't cleanup the namespace because the Trigger is hanging there. It's unrelated to current PR and will be investigated separately.

hack/generate/index.sh
```diff
@@ -2,29 +2,85 @@
...
+
+  echo "---" >> "${target}"
+  cat "${channel_entry_yaml}" >> "${target}"
+}
+
+# Clear the file.
+rm -f "${target}"
+
+add_channel_entries "stable" "${target}"
+add_channel_entries "stable-1.34" "${target}"
```
pierDipi, line 86: Should this values be taken from `project.yaml` ?

mgencur: Right. Good idea.

olm-catalog/serverless-operator/project.yaml
```diff
@@ -13,11 +13,6 @@ olm:
         list:
             - stable
             - stable-1.34
```
pierDipi, line 15: If the channel list will only contain the latest 2 releases, we will need to change `bumpso` tool to only have 

```
	channelsList := []interface{}{
		defaultChannel,
		fmt.Sprintf("stable-%d.%d", newVersion.Major, newVersion.Minor),
  }
```

PR #2822, opened 2024-08-26
[WIP] Use latest for artifacts directory instead of specific versions

pierDipi:
This helps with testing release-next


maschmid:
It looks fine by me, I'd prefer if anybody from the serving team would give an "aye" as well. 
/cc @skonto

vendor/knative.dev/operator/pkg/apis/operator/v1beta1/knativeeventing_lifecycle.go
```diff
@@ -132,7 +133,7 @@ func (es *KnativeEventingStatus) GetVersion() string {
 
 // SetVersion sets the currently installed version of the component.
 func (es *KnativeEventingStatus) SetVersion(version string) {
-	es.Version = version
+	es.Version = os.Getenv("KNATIVE_EVENTING_VERSION")
 }
```
pierDipi, line 137: This and the serving one, allows decoupling the directory name from the version in the status and since we only support one version per CSV

hack/patches/set_upstream_version.patch
```diff
@@ -0,0 +1,42 @@
...
++	"os"
+ 	"strings"
+ 
+ 	"k8s.io/apimachinery/pkg/runtime/schema"
+@@ -132,7 +133,7 @@ func (es *KnativeEventingStatus) GetVersion() string {
+ 
+ // SetVersion sets the currently installed version of the component.
+ func (es *KnativeEventingStatus) SetVersion(version string) {
+-	es.Version = version
++	es.Version = os.Getenv("KNATIVE_EVENTING_VERSION")
```
skonto, line 18: Could we push the change upstream in the future?

PR #2831, opened 2024-09-03
Only parse `transport-encryption` feature flag in KnativeKafka

pierDipi:
This helps with installing newer Eventing versions with "unknown" feature flags


creydr (commented):
/lgtm

PR #2833, opened 2024-09-03
:bug: Prevent slash from appearing in random IDs

cardil:
Follows up on https://github.com/knative/actions/pull/224

## Proposed Changes

- :bug: Prevent slash from appearing in random IDs

/kind bug


pierDipi (approved):
/lgtm
/approve

PR #2837, opened 2024-09-04
[SRVKS-1264] Remove deprecated TLS secret

dsimansk:
Fixes JIRA #SRVKS-1264

## Proposed Changes

- [SRVKS-1264] Remove deprecated TLS secret

Meanwhile, I'm waiting for OCP cluster to spin up. I'm trying very naive approach to remove the dangling secret.

/cc @skonto

### DIFF
--- a/openshift-knative-operator/pkg/serving/extension.go
+++ b/openshift-knative-operator/pkg/serving/extension.go
@@ -37,8 +37,7 @@ const (
 	controlProtocolCertificatesReconcilerLease = "controller.knative.dev.control-protocol.pkg.certificates.reconciler.reconciler"
 )
 
-var isDomainMappingRemoved atomic.Bool
-var isSecretRemoved atomic.Bool
+var oldResourceRemoved atomic.Bool
 
 // NewExtension creates a new extension for a Knative Serving controller.
 func NewExtension(ctx context.Context, impl *controller.Impl) operator.Extension {
@@ -170,18 +169,11 @@ func (e *extension) Reconcile(ctx context.Context, comp base.KComponent) error {
 		common.ConfigureIfUnset(&ks.Spec.CommonSpec, monitoring.ObservabilityCMName, monitoring.ObservabilityBackendKey, "none")
 	}
 
-	if !isDomainMappingRemoved.Load() {
+	if !oldResourceRemoved.Load() {
 		if err := e.cleanupOldResources(ctx, ks.GetNamespace()); err != nil {
 			return err
 		}
-		isDomainMappingRemoved.Store(true)
-	}
-
-	if !isSecretRemoved.Load() {
-		if err := e.cleanupOldSecrets(ctx, ks.GetNamespace()); err != nil {
-			return err
-		}
-		isSecretRemoved.Store(true)
+		oldResourceRemoved.Store(true)
 	}
 
 	return monitoring.ReconcileMonitoringForServing(ctx, e.kubeclient, ks)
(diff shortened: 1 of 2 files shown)


test/servinge2e/deprecated_resources_test.go
```diff
@@ -52,3 +52,15 @@ func TestDomainMappingResource(t *testing.T) {
 		}
 	})
 }
+
+// SRVKS-1264 - cleanup of old internal TLS secrets
+func TestOldTLSSecret(t *testing.T) {
```
skonto, line 57: Given that 1.11 (1.32) is the last time the secret was deployed I am wondering if the test is useful here. I think this matters for the 1.33 branch during upgrades.

openshift-knative-operator/pkg/serving/extension.go
```diff
@@ -251,3 +259,12 @@ func (e *extension) cleanupOldResources(ctx context.Context, ns string) error {
 	}
 	return nil
 }
+
+// SRVKS-1264 - cleanup of old internal TLS secrets
+func (e *extension) cleanupOldSecrets(ctx context.Context, ns string) error {
```
skonto, line 264: Maybe move this to `cleanupOldResources` ? If it gets too long we split that method later.

test/servinge2e/deprecated_resources_test.go
dsimansk: I was just hoping it's part of upgrade test, but haven't checked that.

dsimansk:
Codegen verification will fail due to patch error. It'll be fixed by https://github.com/openshift-knative/serverless-operator/pull/2838.

openshift-knative-operator/pkg/serving/extension.go
dsimansk: Updated per suggestion, thanks!

openshift-knative-operator/pkg/serving/extension.go
```diff
@@ -214,8 +214,10 @@ func (e *extension) fetchLoggingHost(ctx context.Context) string {
 	return route.Status.Ingress[0].Host
 }
 
+// cleanupOldResources util function to clean up old, deprecated or dangling resources from Serving features.
 func (e *extension) cleanupOldResources(ctx context.Context, ns string) error {
 	client := e.kubeclient
+	// DomainMapping related resources
```
skonto, line 220: s/DomainMapping/old resources

test/servinge2e/deprecated_resources_test.go
skonto: Are we planning to backport this?

openshift-knative-operator/pkg/serving/extension.go
dsimansk: I've actually added closing DomainMapping section at L254 in case we will need to split the functions to have a bit of reference.

test/servinge2e/deprecated_resources_test.go
dsimansk: It's still worth to have in 1.33.z if we happen to release another z-stream patch.

openshift-knative-operator/pkg/serving/extension.go
dsimansk: I see, there are leases related to net-certmanager. Added another comment and removed "closing" DomainMapping. It should be a bit more clearer now.

test/servinge2e/deprecated_resources_test.go
skonto: Ok let's add it.

skonto:
It seems generation fails again.

openshift-knative-operator/pkg/serving/extension.go
```diff
@@ -231,6 +233,7 @@ func (e *extension) cleanupOldResources(ctx context.Context, ns string) error {
 	if err != nil {
 		return err
 	}
+	// Leases - DomainMapping and CertManager related
```
skonto, line 236: Nit: Maybe we could remove the comments as it is a straightforward and grouping is done per resource type not based on dm, certmanager etc.

PR #2838, opened 2024-09-05
Fix Dockerfiles and patches

mgencur:
Fix this after merging https://github.com/openshift-knative/hack/pull/243

## Proposed Changes

-
-
-


pierDipi (approved):
/lgtm

PR #2841, opened 2024-09-09
Fix TLS feature flag parsing for KnativeKafka

creydr:
Switch to use `feature.NewFlagsConfigFromMap()` to parse feature flags, as this also ignores the values case (`strings.EqualFold()`)


pierDipi (commented):
This doesn't solve the issue with "unknown flags" (future flags)

https://github.com/openshift-knative/serverless-operator/pull/2831

creydr:
> This doesn't solve the issue with "unknown flags" (future flags)
> 
> #2831

As we set up the configmap data "manually" it does not fail on unknown fields (the data contains only the transport-encryption FF then)

pierDipi (approved):
/lgtm
/approve

PR #2843, opened 2024-09-09
Update client midstream to 1.14

dsimansk:
- Update client version 
- Dockerfiles are re-generated per latest tooling changes. Also fixed by #2842.


knative-operator/Dockerfile
```diff
@@ -15,7 +15,6 @@ FROM $GO_RUNTIME
 
 ARG VERSION=
 
-RUN microdnf install tzdata
```
skonto, line 18: What was the goal of this before?

dsimansk: I think, it's added for some other component not S-O needs.

dsimansk: https://github.com/openshift-knative/hack/pull/250 ahh so it's part of ubi8-minimal already.

PR #2870, opened 2024-09-13
Force builder images to Golang 1.22

dsimansk:
- Force builder images to Golang 1.22

/cc @pierDipi @mgencur

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -321,7 +321,7 @@ generate-dockerfiles:
 	rm -rf /tmp/serverless-operator-generator
 	$(shell go env GOPATH)/bin/generate \
 		--generators dockerfile \
-		--dockerfile-image-builder-fmt "registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.22-openshift-4.17%.0s"  \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -321,6 +321,7 @@ generate-dockerfiles:
 	rm -rf /tmp/serverless-operator-generator
 	$(shell go env GOPATH)/bin/generate \
 		--generators dockerfile \
+		--dockerfile-image-builder-fmt "registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.22-openshift-4.17%.0s"  \
```
dsimansk, line 324: `%s.0` is truncating string formatting to zero length, otherwise there's an extra param error in the result.

mgencur: We could probably fix it somehow in openshift-knative/hack and allow passing just string in which case it would be used as is.

dsimansk:
Update with a latest commit to `generate`.

pierDipi (approved):
/lgtm

skonto:
/override "Red Hat Konflux / serverless-index-main-on-pull-request"

PR #2878, opened 2024-09-18
Adding image overrides for base image for Konflux

mgencur:
Fixes https://issues.redhat.com/browse/SRVCOM-3350

Konflux will only read these overrides when the corresponding s-o branch exists (e.g. for midstream release-v1.15 the release-1.35 needs to exist in s-o). Maybe we'll want a fix for this in openshift-knative/hack.

Adding this according to instructions in https://github.com/openshift-knative/hack/pull/228

## Proposed Changes

- Override both images GO_RUNTIME and GO_BUILDER. This will allow pinning the images with specific SHA


mgencur:
Closing in favor of https://github.com/openshift-knative/serverless-operator/pull/2878

PR #2879, opened 2024-09-18
Fix source path for upstream tests

pierDipi:
The path doesn't really depend on GOPATH but it's rather static as configured by `test-source-image.Dockerfile` to be `/go/src/...`

### DIFF
--- a/hack/lib/vars.bash
+++ b/hack/lib/vars.bash
@@ -29,10 +29,11 @@ PREVIOUS_CSV="$(metadata.get project.name).v$PREVIOUS_VERSION"
 export CURRENT_VERSION PREVIOUS_VERSION CURRENT_CSV PREVIOUS_CSV
 
 # Directories below are filled with source code by ci-operator
-export KNATIVE_SERVING_HOME="${GOPATH}/src/knative.dev/serving"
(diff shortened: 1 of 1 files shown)


hack/lib/vars.bash
```diff
@@ -29,10 +29,11 @@ PREVIOUS_CSV="$(metadata.get project.name).v$PREVIOUS_VERSION"
 export CURRENT_VERSION PREVIOUS_VERSION CURRENT_CSV PREVIOUS_CSV
 
 # Directories below are filled with source code by ci-operator
-export KNATIVE_SERVING_HOME="${GOPATH}/src/knative.dev/serving"
-export KNATIVE_EVENTING_HOME="${GOPATH}/src/knative.dev/eventing"
-export KNATIVE_EVENTING_KAFKA_BROKER_HOME="${GOPATH}/src/knative.dev/eventing-kafka-broker"
-export KNATIVE_EVENTING_ISTIO_HOME="${GOPATH}/src/knative.dev/eventing-istio"
+export TEST_SOURCE_BASE_PATH=${TEST_SOURCE_BASE_PATH:-"/go/src/knative.dev"}
```
pierDipi, line 32: Giving the option to override the base path just in case we're relying on that in downstream testing

ReToCode (approved):
/lgtm
/approve

PR #2887, opened 2024-09-24
Use istio.sidecar.inject label instead of annotation

ReToCode:
Fixes JIRA https://issues.redhat.com/browse/SRVKS-1273

## Proposed Changes
- Use istio.sidecar.inject label instead of annotation

/assign @skonto


docs/full-mesh.md
```diff
@@ -77,17 +77,19 @@ spec:
...
+    labels:
       "sidecar.istio.io/inject": "true"
+    annotations:
       "sidecar.istio.io/rewriteAppHTTPProbers": "true"
 EOF
 ```
 
-Then, create KnativeService with `sidecar.istio.io/inject: "true"`, `sidecar.istio.io/rewriteAppHTTPProbers: "true"` and `serving.knative.openshift.io/enablePassthrough: "true"` annotations in your namespace,
+Then, create KnativeService with the `sidecar.istio.io/inject: "true"` label and `sidecar.istio.io/rewriteAppHTTPProbers: "true"`, `serving.knative.openshift.io/enablePassthrough: "true"` annotations in your namespace,
 which is one of the namespaces in the `ServiceMeshMemberRoll`.
```
dsimansk, line 93: These doc files are not source of truth for any public content right?

ReToCode: Yes, not sure why they are there in the first place. Just updated every reference I could find.

skonto:
The only issue I had was answered here https://github.com/knative-extensions/net-istio/pull/1369#issuecomment-2373108380

PR #2827, opened 2024-09-02
SRVCOM-3319: Generate CSV with Konflux images (serving, eventing, operator)

pierDipi:

PR #2897, opened 2024-09-26
Define project.tag in project.yaml

mgencur:
* Allows to run `make install` on release branches and it will us the right index image
* This change will also populate the VERSION arg properly in Dockerfiles as the version is read from `project.tag` which exists in other repositories but didn't exist in s-o so far ([example](https://github.com/openshift-knative/serverless-operator/blob/main/knative-operator/Dockerfile#L16))

[Slack thread](https://redhat-internal.slack.com/archives/CLUCMK7R6/p1727352333044709?thread_ts=1727347265.990459&cid=CLUCMK7R6)

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -55,7 +55,7 @@ jobs:
         if: (github.event_name == 'pull_request' && github.base_ref != 'main') || ((github.event_name == 'push' || github.event_name == 'workflow_dispatch') && github.ref_name != 'main')
         run: |
           tag="release-$(yq read ./olm-catalog/serverless-operator/project.yaml 'project.version')"
-          echo "CURRENT_VERSION_IMAGES=${tag}" >> $GITHUB_ENV
+          yq write ./olm-catalog/serverless-operator/project.yaml 'project.tag' ${tag}
         shell: bash
 
       - name: Regenerate all generated files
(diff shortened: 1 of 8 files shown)


.github/workflows/validate.yaml
```diff
@@ -55,7 +55,7 @@ jobs:
         if: (github.event_name == 'pull_request' && github.base_ref != 'main') || ((github.event_name == 'push' || github.event_name == 'workflow_dispatch') && github.ref_name != 'main')
         run: |
           tag="release-$(yq read ./olm-catalog/serverless-operator/project.yaml 'project.version')"
-          echo "CURRENT_VERSION_IMAGES=${tag}" >> $GITHUB_ENV
+          yq write ./olm-catalog/serverless-operator/project.yaml 'project.tag' ${tag}
```
pierDipi, line 58: Should we do this write in `bumpso` ? https://github.com/openshift-knative/serverless-operator/blob/main/hack/cmd/bumpso/bumpso.go

mgencur: I don't think so. For main branch we want this to be always `project.tag: main`. But for release branches we want it to be specific, e.g. `release-1.34.0`.
AFAIK, we run the `bumpso` github action against main: https://github.com/openshift-knative/serverless-operator/blob/main/.github/workflows/bump.yaml#L29
Not sure if I'm missing something?

mgencur: OK. Looks like this is completely replaced by https://github.com/openshift-knative/serverless-operator/pull/2827 .
The version of index image will be determined from project version, e.g. for 1.35.0 it will be `docker://quay.io/redhat-user-workloads/ocp-serverless-tenant/serverless-operator-135/serverless-index:latest` 
Closing.

PR #2918, opened 2024-10-04
Use Konflux built func-utils image in CSV

creydr:
As per title


hack/generate/csv.sh
```diff
@@ -27,12 +27,14 @@ if [[ ${USE_RELEASE_NEXT:-} == "true" ]]; then
...
   knative_backstage_plugins_images_release
   knative_serving_images_release
+  knative_kn_plugin_func_images_release
 else
   default_knative_eventing_images
   default_knative_eventing_istio_images
   default_knative_eventing_kafka_broker_images
   default_knative_backstage_plugins_images
   default_knative_serving_images
+  default_knative_kn_plugin_func_images
```
creydr, line 37: Not sure on this one, if we simply should call `knative_kn_plugin_func_images` instead (and update knative_kn_plugin_func_images to not set any tags) or if keeping it this way, makes cleanup with the other functions more streamline as they are then all similar :shrug:

pierDipi (approved):
/lgtm

pierDipi:
Since there is no tests for this I'll merge it without wasting CI cycles

hack/generate/csv.sh
pierDipi: I'm not sure I get it? PR looks good though

PR #2919, opened 2024-10-04
[WIP] Use production ready Envoy image and get latest external func  images

pierDipi:
- Use latest images that are used by func and kourier
- Use production ready Envoy image


pierDipi:
https://github.com/openshift-knative/serverless-operator/pull/2959

PR #2920, opened 2024-10-07
Use registry.redhat.io reference in CSV

mgencur:
Related [slack discussion](https://redhat-internal.slack.com/archives/CKR568L8G/p1728045413397479)

## Proposed Changes

- When building Bundle image in CI, replace the references back to Konflux quay.io in the CSV being used. This is for testing purposes.
- When using an external Index image that was built earlier, generate an ImageContentSourcePolicy as we can't directly manipulate the CSV. This ICSP will redirect OpenShift to pull image from different locations (Konflux workspace in quay instead of registry.redhat.io) 
- Do not use registry.redhat.io for bundle image itself in the catalog otherwise the Konflux build for Index image fails as it can't pull the image from registry.redhat.io. The catalog needs to be reworked as part of https://issues.redhat.com/browse/SRVCOM-3369 anyway.


hack/lib/catalogsource.bash
```diff
@@ -161,6 +209,7 @@ function build_image() {
 
 function delete_catalog_source {
   logger.info "Deleting CatalogSource $OPERATOR"
+  oc delete imagecontentsourcepolicy --ignore-not-found=true serverless-image-content-source-policy
```
pierDipi, line 212: nit: Technically, should this be done after deleting the `catalogsource` ?

pierDipi: and perhaps even after deleting the `serverless-index` deployment ?

mgencur: We could move it. But it's not necessary. The ImageContentSourcePolicy is required when the pods try to pull the images. They were already pulled and the pods started. It would be necessary only if the catalog source pod restarted and needed to pull the images again. I can move this in a follow up (unless this PR needs some other changes too).

pierDipi:
Merged

PR #2933, opened 2024-10-11
Remove multi-arch builds in GH actions

pierDipi:
We have Konflux builds now that are multi-arch and production ready


ReToCode (approved):
/lgtm
/approve

PR #2936, opened 2024-10-14
Support patch releases for Konflux builds

pierDipi:
This will support pushing images to a different tag as soon as we bump the patch version

### DIFF
--- a/hack/generate/update-pipelines.sh
+++ b/hack/generate/update-pipelines.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+readonly version=$(yq read olm-catalog/serverless-operator/project.yaml project.version)
+# shellcheck disable=SC2038
+find .tekton/ \( -name "*-push.yaml" \) | xargs -I{} yq write --inplace "{}" "spec.params(name==additional-tags).value[0]" "${version}"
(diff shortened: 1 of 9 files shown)


hack/generate/update-pipelines.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+readonly version=$(yq read olm-catalog/serverless-operator/project.yaml project.version)
+# shellcheck disable=SC2038
+find .tekton/ \( -name "*-push.yaml" \) | xargs -I{} yq write --inplace "{}" "spec.params(name==additional-tags).value[0]" "${version}"
```
creydr, line 7: Not really sure, but could we add a comment [here](https://github.com/openshift-knative/hack/blob/d63bd441112550420fc587cd4548252b215670df/pkg/prowgen/prowgen_konflux.go#L316), that the version should be on `Tags[0]`?
Just to reduce the probability that we break this in future

pierDipi: I will do

pierDipi: https://github.com/openshift-knative/hack/pull/339

pierDipi:
Konflux pipelines still running after 1 day :fire:

pierDipi:
https://github.com/openshift-knative/serverless-operator/pull/2936/checks?check_run_id=31492343304

PR #2953, opened 2024-10-17
Use correct FBC image when installing catalog source

mgencur:
## Proposed Changes

- Include specific bundle in the catalog only if the respective entry was added in the channel. If the entry is already there it means the bundle was already added.
- The regenerated catalog now includes version 1.34 that comes from registry.redhat.io as the release is already out and baked in the production catalog. This replaces the previous entry that was coming from CI registry.


hack/lib/images.bash
```diff
@@ -224,6 +247,11 @@ function latest_registry_redhat_io_image_sha() {
   image_without_tag=${image_without_tag%@*} # Remove sha, if any
 
   image=$(image_with_sha "${image_without_tag}:latest")
+
+  if [ "${image}" = "" ]; then
+    exit 1
+  fi
```
creydr, line 253: can this ever be, as we append `:latest` anyhow?

creydr (commented):
/lgtm

PR #2954, opened 2024-10-17
Allow forcing konflux index

mgencur:
Related to https://issues.redhat.com/browse/SRVCOM-3382

This will be followed by a PR in openshift-knative/hack that will generate periodic jobs with this parameter: `FORCE_KONFLUX_INDEX=true`

For example, instead of:
```
GOPATH=/tmp/go PATH=$PATH:/tmp/go/bin SKIP_MESH_AUTH_POLICY_GENERATION=true
        make upstream-e2e-kafka
```
we will use:
```
GOPATH=/tmp/go PATH=$PATH:/tmp/go/bin SKIP_MESH_AUTH_POLICY_GENERATION=true FORCE_KONFLUX_INDEX=true
        make upstream-e2e-kafka
```

## Proposed Changes

-
-
-


mgencur:
@pierDipi Using the Index image is fixed in https://github.com/openshift-knative/serverless-operator/pull/2953 , there it uses the right OCP-specific version.

PR #2969, opened 2024-10-23
Do not fail if FORCE_KONFLUX_INDEX is undefined

mgencur:
An error spotted in https://github.com/openshift-knative/serving/pull/929#issuecomment-2430201540

## Proposed Changes

-
-
-


pierDipi (approved):
/lgtm

PR #2979, opened 2024-10-30
optionally skip certmanager install on "make install"

maschmid:
- :broom: optionally skip certmanager install on "make install" with INSTALL_CERTMANAGER=fals


creydr (approved):
/lgtm

PR #2986, opened 2024-11-04
Wait for imageStreamTag to be imported before build

maschmid:
- :broom: Explicitly import the BuildConfig FROM imageStreamTag before starting the build

Sometimes we're getting errors like
> The ImageStreamTag "ubi-minimal:latest" is invalid: from: Error resolving ImageStreamTag ubi-minimal:latest in namespace openshift-marketplace: unable to find latest tagged image

let's see if explicitly importing the imagestream before running the "oc start-build" helps.

### DIFF
--- a/hack/lib/catalogsource.bash
+++ b/hack/lib/catalogsource.bash
@@ -201,7 +201,7 @@ function replace_images() {
 }
 
 function build_image() {
-  local name from_dir dockerfile_path tmp_dockerfile
+  local name from_dir dockerfile_path tmp_dockerfile image_stream_tag
   name=${1:?Pass a name of image to be built as arg[1]}
   from_dir=${2:?Pass context dir}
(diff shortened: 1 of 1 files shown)


maschmid:
/hold 

it seems the additional import start a build that will fail, as the builds only expect a binary build. we should also disable triggers (esp. the image change trigger)

maschmid:
/unhold

calling "oc import-image" is unecessary. The build creation already imports the image, so instead, just wait for the imagestreamtag to be imported (wait for its `.image.dockerImageReference` to not be empty)

maschmid:
> ! oc get imagestreamtag -n "openshift-marketplace" "scratch" -o json | jq -re .image.dockerImageReference : Waiting until non-zero (max 60 sec.)

we'll need a special case for sratch

PR #3004, opened 2024-11-07
Bump 1.34.1

Kaustubh-pande:
as per title

cc @dsimansk @pierDipi

### DIFF
--- a/olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
+++ b/olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
@@ -68,14 +68,14 @@ metadata:
       Deploy and manage event-driven serverless applications and functions using Knative.
     repository: https://github.com/openshift-knative/serverless-operator
     support: Red Hat
-    olm.skipRange: '>=1.33.0 <1.34.0'
+    olm.skipRange: '>=1.34.0 <1.34.1'
     operators.openshift.io/must-gather-image: quay.io/openshift-knative/must-gather
   labels:
     operatorframework.io/arch.amd64: supported
     operatorframework.io/arch.ppc64le: supported
     operatorframework.io/arch.s390x: supported
     operatorframework.io/arch.arm64: supported
-  name: serverless-operator.v1.34.0
+  name: serverless-operator.v1.34.1
   namespace: placeholder
 spec:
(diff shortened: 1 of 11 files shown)


olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml
```diff
@@ -1316,11 +1316,11 @@ spec:
         - knativeeventings.operator.knative.dev
   relatedImages:
     - name: "knative-operator"
-      image: "registry.ci.openshift.org/knative/release-1.34.0:serverless-openshift-knative-operator"
+      image: "registry.ci.openshift.org/knative/main:serverless-openshift-knative-operator"
```
Kaustubh-pande, line 1319: I think this should be release-1.34.1 but not sure why it take main by default @pierDipi @dsimansk is something i missed?

dsimansk: There's a env var to set during generator cmd. Just a sec... :)

dsimansk: https://github.com/openshift-knative/serverless-operator/blob/main/Makefile#L350-L353

The following should do the trick. 
```
CURRENT_VERSION_IMAGES=release-1.34.1 make generated-files
```

pierDipi: ^^ @Kaustubh-pande

dsimansk:
@mgencur could you please take a look, if there's a way to fix upgrade tests? Or we need to have images tagged first as `release-1.34.1`?

mgencur:
Hey. For 1.33.2 we were just removing the test from execution: https://github.com/openshift-knative/serverless-operator/commit/b4282fe789b001e0d71a4305d3037a4983a4d4f9

skonto:
Yeah that test should be removed unless images make a difference from one S-O version to the other (not using a fixed tag).

dsimansk:
Lint error is fine to proceed with, it's unhappy with unused functions.

dsimansk:
@mgencur FYI the tests "stabilized" without the triggerv2 downgrade part.

PR #3023, opened 2024-11-14
[SRVKS-1288] Fix Kourier ns cleanup

skonto:
Fixes JIRA # SRVKS-1288

## Proposed Changes

- Removed the manual delete of the knative-serving-ingress ns
- Makes sure that the manifest used during deletion points to the right ns for the kourier resources.
So far we used a hack where we just deleted the ns.
- With this PR the behavior is made similar to Eventing, Serving namespaces. For the latter we keep the ns but delete the resources under it. The goal here is the same. This should fit better with gitops where the namespace is managed by the external tool, thus we should not delete it.

### DIFF
--- a/vendor/knative.dev/operator/pkg/reconciler/knativeserving/knativeserving.go
+++ b/vendor/knative.dev/operator/pkg/reconciler/knativeserving/knativeserving.go
@@ -19,14 +19,15 @@ package knativeserving
 import (
 	"context"
 	"fmt"
+	"os"
 
-	mf "github.com/manifestival/manifestival"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 
 	"knative.dev/pkg/logging"
 	pkgreconciler "knative.dev/pkg/reconciler"
-
+	mf "github.com/manifestival/manifestival"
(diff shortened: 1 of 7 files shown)


skonto:
Accidently re-run the job. It has passed [here](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift-knative_serverless-operator/3017/pull-ci-openshift-knative-serverless-operator-main-417-test-upgrade-aws-417/1856995652349726720)

skonto:
> upgrade.go:15: Serverless upgrade failed: knative kafka upgrade failed: knativekafka knative-kafka is not in desired state, got: &{TypeMeta:{Kind:KnativeKafka

vendor/knative.dev/operator/pkg/reconciler/knativeserving/knativeserving.go
```diff
@@ -97,6 +103,20 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, original *v1beta1.Knative
 	return nil
 }
 
+func overrideKourierNamespace(ks base.KComponent) mf.Transformer {
```
matzew, line 106: would it make sense to move this to upstream, eventually?

matzew:
LGTM, but @ReToCode or @dsimansk  mind taking a look too?

PR #3035, opened 2024-11-18
lower CPU requests for dispatchers in kitchensink test

maschmid:
- :broom: lower CPU requests for dispatchers in kitchensink test


test/lib.bash
```diff
@@ -372,7 +372,7 @@ function downstream_kitchensink_e2e_tests {
 
   logger.info "Running Knative kitchensink tests"
 
-  local images_file
+  local images_file, workloads_patch
```
mgencur, line 375: Using comma won't work.
The variables should be separated by space

PR #3054, opened 2024-11-26
Separate installing cert manager operator and certificates

mgencur:
Allows for testing proper CA setup downstream instead of self-signed certificates and still re-using the installation function for certmanager operator.

Fixes JIRA #

## Proposed Changes

-
-
-


mgencur:
Our [Dockerfile](https://github.com/openshift-knative/serverless-operator/blob/main/olm-catalog/serverless-operator-index/Dockerfile) for Index image is now pulling registry.ci.openshift.org/knative/release-1.35.0:serverless-bundle on line 14 as the bundle now exists. It doesn't pull the the previously built "image-registry.openshift-image-registry.svc:5000/openshift-marketplace/serverless-bundle:latest" and the release-1.35.0 bundle references images at registry.redhat.io which are not pullable yet.
We'll need to fix this.

Issue #2989, opened 2024-11-05
Release 1.35


matzew:
# Release Checklist

## Resources
- https://github.com/openshift-knative/serverless-operator/blob/main/README.md#create-a-new-version
- [Video](https://drive.google.com/drive/u/0/folders/1j1KF_1dUl6x0oiN2aVKlapAYXo1dGUDG) from pierdipi showing the release process

## Pre-checks
- [x] Check if the OCP versions in the [release-matrix](https://docs.google.com/spreadsheets/d/1HTxR37_MM03_JQImW-KmAFk1oEQ83fXzfGxdfWQSX6o/edit#gid=0) match [our current CI config](https://github.com/openshift-knative/hack/tree/main/config). If not, fix the versions first
- [x] Check if there are open PRs for Serverless-Operator that need to be merged
- [x] Check with the teams if they have all the necessary patches in the dependent repositories
- [x] [Run the `validate` action in `serverless-operator` and wait for it to complete](https://github.com/openshift-knative/serverless-operator/actions/workflows/validate.yaml)
- [x] [Merge PRs created by GitHub actions](https://github.com/openshift-knative/serverless-operator/pulls/app%2Fgithub-actions)
- [x] Make sure that Serverless-Operator CI and periodics are passing. Check for example the periodic jobs in [#serverless-ci](https://redhat.enterprise.slack.com/archives/CHTTRCUBC) 

## Cutting the new release branch
- [x] Create a new branch from `main` with the pattern `release-x.xx`
- [x] [Approve CI setup for `release-1.X` branch in `openshift/release`](https://github.com/openshift/release/pulls/serverless-qe) and wait for the PR to be merged.
- [ ] [Approve PRs created by GitHub actions bot in `serverless-operator`](https://github.com/openshift-knative/serverless-operator/pulls/app%2Fgithub-actions)
- [ ] [Only for release-1.35+] [Revoke and recreate the token for the Konflux auto apply workflow](https://github.com/openshift-knative/hack/tree/main/cmd/konflux-apply#revoke-and-recreate-token)
- [ ] [Only for release-1.35+] [Approve PRs created by `serverless-qe` bot in `hack`](https://github.com/openshift-knative/hack/pulls/serverless-qe)
  - [x] wait for the PR to be merged
  - [ ] wait for the [Generate CI config workflow to succeed](https://github.com/openshift-knative/hack/actions/workflows/release-generate-ci.yaml?query=event%3Apush++)
[...]

PR #3056, opened 2024-11-27
Fix SO index

pierDipi:
Our Dockerfile for Index image is now pulling
`registry.ci.openshift.org/knative/release-1.35.0:serverless-bundle` on line 14 as the bundle now exists.

It doesn't pull the the previously built
`image-registry.openshift-image-registry.svc:5000/openshift-marketplace/serverless-bundle:latest` and the release-1.35.0 bundle references images at `registry.redhat.io` which are not pullable yet.

Originally reported here: https://github.com/openshift-knative/serverless-operator/pull/3054#issuecomment-2503220449

### DIFF
--- a/olm-catalog/serverless-operator-index/Dockerfile
+++ b/olm-catalog/serverless-operator-index/Dockerfile
@@ -11,10 +11,6 @@ RUN /bin/opm init serverless-operator --default-channel=stable --output yaml >> 
 RUN /bin/opm render --skip-tls-verify -o yaml \
 registry.ci.openshift.org/knative/release-1.33.0:serverless-bundle \
 registry.ci.openshift.org/knative/release-1.34.0:serverless-bundle \
-      registry.ci.openshift.org/knative/release-1.35.0:serverless-bundle >> /configs/index.yaml || \
-    /bin/opm render --skip-tls-verify -o yaml \
-registry.ci.openshift.org/knative/release-1.33.0:serverless-bundle \
-registry.ci.openshift.org/knative/release-1.34.0:serverless-bundle \
       quay.io/redhat-user-workloads/ocp-serverless-tenant/serverless-operator-135/serverless-bundle@sha256:0d5feb3c9f59246d16db1c687e3c1085fcc8a360cd51974197a878944df06b1d >> /configs/index.yaml
 
 # The base image is expected to contain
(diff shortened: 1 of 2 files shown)


creydr (commented):
Thanks for fixing

/lgtm

skonto:
We need this after each cut?

olm-catalog/serverless-operator-index/Dockerfile
```diff
@@ -10,10 +10,6 @@ COPY olm-catalog/serverless-operator-index/configs /configs
 RUN /bin/opm init serverless-operator --default-channel=stable --output yaml >> /configs/index.yaml
 RUN /bin/opm render --skip-tls-verify -o yaml \
 registry.ci.openshift.org/knative/release-1.33.0:serverless-bundle \
-registry.ci.openshift.org/knative/release-1.34.0:serverless-bundle \
-      registry.ci.openshift.org/knative/release-1.35.0:serverless-bundle >> /configs/index.yaml || \
```
skonto, line 14: or make it point to 1.36?

mgencur:
> We need this after each cut? Do we need to document it in the template as in here
No. That's a one-time change. We had that before because we wanted to be able to install Serverless operator from "main" and from release branch without rebuilding the Bundle (i.e. faster setup - for periodic jobs or local runs). But now we need to rebuild the Bundle every time because we replace registry.redhat.io references with quay.io. So, the bundle will be rebuilt correctly for each branch.

olm-catalog/serverless-operator-index/Dockerfile
mgencur: Not needed. See my comment below.

skonto:
Is there going to be a 1.36 entry in there?

mgencur:
> Is there going to be a 1.36 entry in there with registry.redhat.io references?

No.

pierDipi:
> We need this after each cut? Do we need to document it in the template as in here: #2989?

no, I don't think so

pierDipi:
```
error getting cluster operators: Get "https://api.serverless-ocp-4-17-amd64-aws-us-east-1-6544x.serverless.devcluster.openshift.com:6443/apis/config.openshift.io/v1/clusteroperators": dial tcp 3.227.130.244:6443: connect: connection refused
ClusterOperators:
	clusteroperators are missing
```

Can we merge anyway?

PR #3117, opened 2024-12-03
:bug: Check for empty pull-secrets

cardil:
## Proposed Changes

- :bug: Check for empty pull-secrets on catalog-source pod

/kind bug

Ref.: https://redhat-internal.slack.com/archives/CLUCMK7R6/p1733249078088419


dsimansk:
Closing to stop further CI runs etc. If we still need such check, it definitely needs a rebase. 
/close

PR #3133, opened 2024-12-05
[release-1.35] Regenerate CSV

pierDipi:


Makefile
```diff
@@ -286,6 +286,8 @@ generate-ci-config:
 
 generate-catalog:
 	./hack/generate/catalog.sh
+	./hack/generate/override-snapshot.sh \
+  	.konflux-release/
```
creydr, line 290: Is this intended?

pierDipi: it would be failing otherwise

creydr: In #3132 I moved them to a separate target, because the override snapshot generation is anyhow "async" to the catalog update. 
But not blocking on this
/lgtm

pierDipi: I see, we can do that too

PR #3142, opened 2024-12-06
[release-1.35] Add Konflux configurations

serverless-qe:
Add Konflux components and pipelines


pierDipi:
One FBC pipeline succeeded

PR #3151, opened 2024-12-10
Link to Build Status Dashboard in release issue template

pierDipi:
Draft so that we can skip running CI and waste resources


creydr (approved):
/lgtm

PR #3153, opened 2024-12-11
Move override-snapshot generation into separate Makefile target

creydr:
Moving the override-snapshot generation into a separate Makefile target, so we can trigger it directly e.g. via https://github.com/openshift-knative/hack/actions/workflows/run-command-repository.yaml

Also adding it to the validate GH workflow to get an early signal when we have problems on the override-snapshot generation (e.g. `Error: entity not found in registry`)


creydr:
No need to wait for prow with this

PR #3154, opened 2024-12-11
[release-v1.35] Update catalog with latest bundle image

creydr:
Getting the latest bundle into the catalog.

Needed manual run of `./hack/generate/catalog.sh`, as `make generate-catalog` currently also updates the override-snapshot which fails ATM due to `entity not found in registry`


creydr:
Closing in favor of #3156

PR #3220, opened 2024-12-18
Print CVEs when override snapshot is generated

creydr:
Printing the CVEs (if any) of the images from the override-snapshot.


creydr:
```
jq: error: pick/1 is not defined at <top-level>, line 1:
```
GH runner `ubuntu-latest` seems to be still on [jq 1.6](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#tools), which does not have `pick`, which came in [1.7](https://github.com/jqlang/jq/releases/tag/jq-1.7)

/hold

PR #3231, opened 2024-12-19
[WIP] Add override snapshot revision check

creydr:
Adds a check for the component override-snapshot to check if the revisions for components from the same repo align


hack/generate/override-snapshot.sh
```diff
@@ -90,6 +90,41 @@ EOF
   rm -rf "${tmp_catalog_dir}"
 }
```
pierDipi, line 92:
```suggestion
# verify that the revisions (git commit) for components from the same repo match
```

creydr:
/hold
need to rework as we don't want to block the verify action run. Will check if we can move this test to a prow check or an optional GH action

creydr:
Might be superseded by https://github.com/openshift-knative/serverless-operator/pull/3338
/hold

creydr:
Closing in favor of #3338

PR #3298, opened 2025-01-09
[WIP] Add makefile target to check for CVEs in components of override-snapshot

creydr:
Adding a script & makefile target to check that the components in the override-snapshot don't have any high/critical CVEs


creydr:
Doing it on the CSV level already: #3343

PR #3344, opened 2025-01-17
Make the name of the override-snapshot predictable

creydr:
As discussed moving away from `generatedName` of the override-snapshot


pierDipi (approved):
/lgtm
/approve

PR #3345, opened 2025-01-17
Add target to apply override-snapshot

creydr:
Makes later the workflow to apply the snapshots independent from location of override-snapshots and how they are getting applied


pierDipi (approved):
/lgtm

PR #3355, opened 2025-01-21
Exclude override-snapshot generateName from diff check

creydr:
Excluding the override-snapshot name from the diff check


mgencur:
The `validate-lint` check fails with:
```
func ServerlessDowngradeOperations(ctx *test.Context, glob environment.GlobalEnvironment) []pkgupgrade.Operation {
                                                        ^
```
But that's a left-over from previous PRs.

PR #3360, opened 2025-01-27
Fix SO version in override snapshot generation

creydr:
Using the correct SO version in the image references to prevent `error resolving name for image ref quay.io/redhat-user-workloads/ocp-serverless-tenant/serverless-operator-136-fbc-418/serverless-index-135-fbc-418:latest` like [here](https://github.com/openshift-knative/serverless-operator/actions/runs/12985285072/job/36209834038)


pierDipi (approved):
/lgtm
/approve

PR #3372, opened 2025-02-03
[main] Update Konflux configurations

serverless-qe:
Update Konflux components and pipelines

### DIFF
--- a/.tekton/docker-build.yaml
+++ b/.tekton/docker-build.yaml
@@ -468,9 +468,9 @@ spec:
     taskRef:
       params:
       - name: name
-        value: sast-shell-check-oci-ta
+        value: sast-unicode-check-oci-ta
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-sast-shell-check-oci-ta:0.1@sha256:b00fa0294e6318ff8130ca11358ae0b3764c4e6c9a12ec7a953e1a813ca2e231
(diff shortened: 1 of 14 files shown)


pierDipi:
```
invalid result reference in pipeline task "fbc-target-index-pruning-check": "RENDERED_CATALOG_DIGEST" is not a named result returned by pipeline task "validate-fbc"
```

creydr:
> ```
> invalid result reference in pipeline task "fbc-target-index-pruning-check": "RENDERED_CATALOG_DIGEST" is not a named result returned by pipeline task "validate-fbc"
> ```

Seems we didn't get for ~2 week the "Update Konflux references" PRs (last one was #3346). The current referenced validate-fbc task uses sha `27e492e513c4cf5758ad87b1b75d59f98acd05c938e00fbc183446970e96128a` (from 2025-01-21) : https://github.com/openshift-knative/serverless-operator/blob/ab8cf342a99f99162a976bfa92333f9f1b78bd5f/.tekton/fbc-builder.yaml#L315,
while the latest is actually `bda6f60adb0b9f4785ae3e3746206dffd9fcff21c19f97a7ca11dddc2a586902` (from 2025-02-04): https://github.com/openshift-knative/hack/blob/a12678327b21ca21b8e0dcb2a5b567912b99d5bf/pkg/konfluxgen/kustomize/fbc-builder.yaml#L306

So I'd wait if we get a new "Update Konflux references" over the weekend (they usually get created on Saturday/Sunday) and then recheck. If they don't get created, this is something which we probably should check too.

@rudyredhat1 fyi

PR #3379, opened 2025-02-12
Add 'Update support end dates' to release issue template

pierDipi:
Update release doc

Opened as draft to avoid running CI jobs


creydr (approved):
/lgtm

PR #3461, opened 2025-03-04
Create ImageDigestMirrorSet for fbc-fips-check-oci-ta task

pierDipi:
See:
- https://github.com/openshift-knative/serverless-operator/pull/3399
- https://github.com/konflux-ci/build-definitions/tree/main/task/fbc-fips-check-oci-ta/0.1


dsimansk:
Do we have a way to verify it in the PR? Any of the check include it?

PR #3479, opened 2025-03-05
[wip] [test] [SRVKS-1301][RELEASE-1.35] Wait for webhook services

skonto:
## Proposed Changes

- waits for any webhook service that exists in the current manifest before applying the new wh configurations. Eliminates the error during the reconciliation phase. We could re-enqueue ideally instead of waiting (better practice) but an error would be printed if we just fail the reconciliation in case some service is not ready (similar to the object has changed one). I think better is to wait for deployments before applying webhook configs. I will check.
- I will push this upstream
- I will create a PR on main too


skonto:
closing

PR #3480, opened 2025-03-05
[main] Update Konflux configurations

serverless-qe:
Update Konflux components and pipelines


.tekton/images-mirror-set.yaml
creydr, line 1: This file should not be deleted
/hold
@pierDipi fyi

creydr: resolved
/unhold

PR #3563, opened 2025-03-20
Increase tempo stack memory limits further

maschmid:
Still getting OOMKilled in tempo on tracing tests


creydr:
"Duplicate" of #3595

PR #3568, opened 2025-03-25
[release-1.35] Run make generated-files

serverless-qe:
Generated by running `make generated-files`


creydr:
This is to update the catalog with the latest bundle which contains cve fixes of client and serving components 

/lgtm
/approve

PR #3573, opened 2025-03-27
serverless-operator is designed for FIPS, update annotation

pierDipi:
Also remove deprecated annotation `operators.openshift.io/infrastructure-features` https://docs.redhat.com/en/documentation/openshift_dedicated/4/html/operators/developing-operators#osdk-csv-manual-annotations-deprecated_osdk-generating-csvs


rudyredhat1:
The prow jobs are failing for most of the PRs , so can we override them and skip the prow jobs. WDYT?

creydr:
Updated branch to get latest CI fixes from main branch

PR #3583, opened 2025-04-02
WIP: add ConfigMap to webhook dependant resources

dsimansk:
Fixes JIRA #

## Proposed Changes

-
-
-


maschmid:
alternatively, we may just drop the config.webhook.serving.knative.dev validatingwebhookconfiguration altogether, as we manage the CMs in the operator anyway?  (and if we really need to control that, the SO operator should be the webhook that guards them then... )

openshift-knative-operator/cmd/openshift-knative-operator/kodata/knative-serving/latest/2-serving-core.yaml
```diff
@@ -9164,7 +9164,7 @@ webhooks:
     - key: app.kubernetes.io/component
       operator: In
       values: ["autoscaler", "controller", "logging", "networking", "observability", "tracing", "net-certmanager"]
-  timeoutSeconds: 10
+  timeoutSeconds: 30
```
maschmid, line 9167: I thought we wanted to decrease it  (so that the delay is not that long if the issue occurs? )

dsimansk:
@maschmid per slack thread. I've changed the `failurePolicy` and `timeout` on the validating webhook.

openshift-knative-operator/cmd/openshift-knative-operator/kodata/knative-serving/latest/2-serving-core.yaml
dsimansk: Ops my bad then, I have misinterpreted the meaning of this timeout.

skonto:
@maschmid @dsimansk In the upstream operator we can deploy the webhook deployments first and then wait for them. Then apply webhook cgfs and only after that the rest of the resources. This way even during upgrades there will be no resources that can cause the webhook not found error. Ideally you don't want to create the cms without validation.

dsimansk:
@skonto ins't it already done in `CheckDeployments`? At least that was my impression from https://github.com/knative/operator/blob/main/pkg/reconciler/common/deployments.go#L48

skonto:
@dsimansk Befire we call checkDeployments here https://github.com/knative/operator/blob/main/pkg/reconciler/knativeserving/knativeserving.go#L126 we call manifest.Install https://github.com/knative/operator/blob/main/pkg/reconciler/common/install.go#L43 which installs all things except mutation, validation webhook cfgs and the certificate.

skonto:
So far we are facing the following issues:
a) fresh install emits a warning due to the certificate being deployed too early. I fixed this upstream.
b) user does a non clean uninstall and webhook cgs are left around. This would fail with a re-install due to the certificate being deployed too early. Fixed with the PR for a)
c) user does a non clean uninstall and webhook cgs are left around. This would fail with a re-install due to cms being deployed too early.

I suggest we fix c) as commented above.

PR #3584, opened 2025-04-07
Update catalog

creydr:
Manually updating the catalog (via `make generate-catalog`) to get the latest images from the bundle into the catalog, as we have some broken catalog ATM for 1.36, which causes the validate action to fail (https://github.com/openshift-knative/serverless-operator/actions/runs/14302395079)


creydr:
@rudyredhat1 @Kaustubh-pande could you PTAL? This should help on https://github.com/openshift-knative/serverless-operator/actions/runs/14302395079

creydr:
/override "Validate / Generated files are committed (pull_request)"

creydr:
Merging, as eventing components need the yaml update via the validate job, which should be fixed after the FBCs of this pr are built

PR #3590, opened 2025-04-09
Update to Golang 1.23

creydr:
Updating to Golang 1.23, as some "downstream" projects require go 1.23 (`go: go.mod requires go >= 1.23.0 (running go 1.22.9; GOTOOLCHAIN=local)`, e.g. [here](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift-knative_serverless-operator/3589/pull-ci-openshift-knative-serverless-operator-main-417-upstream-e2e/1909861411991851008)

PR #3591, opened 2025-04-10
Update build Dockerfile to Golang 1.23

creydr:
Update the build dockerfile to Golang 1.23 too, as this was missed in #3590.
I'll update the automation for this as a follow up, as this changes currently a lot and I want to test this separately, so we get unblocked on main first


matzew (approved):
/lgtm
/approve

the validate / `make generated-files` was failing, though :man_shrugging:

PR #3641, opened 2025-04-28
use floats for "le"

maschmid:
In OpenShift 4.19 

```
With Prometheus v3, the classic histogram's "le" and summary's "quantile" labels values will be floats.
```

per https://issues.redhat.com/browse/MON-4129

- :broom: Use floats in "le" label prometheus query


creydr:
Merging, as all of them passed already

PR #3671, opened 2025-05-12
Update catalogs after bundle update

creydr:
Update catalog after latest bundle update (https://github.com/openshift-knative/serverless-operator/pull/3668) as validate action seems to be broken ATM


creydr:
@Kaustubh-pande @rudyredhat1 PTAL

PR #3695, opened 2025-05-20
Update dependencies to 1.17

creydr:
Updating the dependencies to 1.17 versions.

Only updating the update-deps.sh file didn't work and had some dependency issues. So I bumped the versions in the [`replaces`](https://github.com/creydr/serverless-operator/blob/2df326a4a0c5539d018bfa19d12a16e4883edebc/go.mod#L178-L204) in the go.mod from the Knative, OpenShift and Kubernetes components each manually and ran `go mod tidy` afterwards in 2df326a4a0c5539d018bfa19d12a16e4883edebc. Afterwards I run the `update-deps.sh --upgrade` and `make generated-files` (each a separate commit)

Part of #3575


creydr:
Why did this trigger a retest of all prow jobs :thinking: :facepalm: :money_with_wings:

dsimansk:
> Why did this trigger a retest of all prow jobs 🤔 🤦 💸

That's a prow's funny way of rebasing...

PR #3736, opened 2025-06-18
[SRVCOM-3428] remove workaround installing cert-manager from upstream instead of OperatorHub on OCP 4.18

simkam:
This removes the workaround added in 4dba7ea308e77b12a48a10295fe882869103ebe9.

Fixes https://issues.redhat.com/browse/SRVCOM-3428

## Proposed Changes

- :broom: remove workaround installing cert-manager from upstream instead of OperatorHub on OCP 4.18

### DIFF
--- a/hack/lib/certmanager.bash
+++ b/hack/lib/certmanager.bash
@@ -17,14 +17,7 @@ function deploy_certmanager_operator {
 
   deployment_namespace="cert-manager"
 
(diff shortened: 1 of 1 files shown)


simkam:
@maschmid I wonder whether we won't get to the same situation with 4.20 (it seems available in OperatorHub for 4.19). Shouldn't we rather keep it and update the version to 4.20?

PR #3757, opened 2025-08-01
[release1.35] update to 1.35.2

maschmid:
- :broom: Update to 1.35.2

Also,

1. Fixing the `replaces` and `skipRange` versions

AFAIU, the correct versions for replaces and skipRange are,

for a patch release
```
replaces: x.y.(z-1)
skipRange: >=x.y.0 <x.y.z
```

for a .0 release
```
replaces: x.(y-1).0
skipRange: >=x.(y-1).0 <x.y.0
```

(the previous version of the script didn't properly generate the replaces and skipRanges for z>1)

2. Fixing the "default channel"

For older releases, we don't want to release into the "stable" channel, just to the "stable-x.y" channel, but the operator "default channel" (the `operators.operatorframework.io.bundle.channel.default.v1` annotation) needs to be kept "stable".

(this patch introduces a new `olm.channels.operator_default` to make this be set separately, so the "default" channel to use for testing can be set to the "stable-x.y", while keeping the bundle annotation stay "stable".)

3. Use the actually released bundles for the catalog and index image use in CI

We don't have a mechanism in CI nor in Konflux to store working bundles for previous patch releases. The only way we can properly build an index with .0,.1,.2,... bundles is to use actually released bundles from registry.redhat.io. 

This patch changes how the olm-catalog/serverless-operator-index/Dockerfile is generated. Instead of adding the "old" CI bundles in build-time via "opm redner" as before, we do run "opm render" in "make generate-catalog", ( which is the only time we have credentials to pull images from registry.redhat.io )

This patch also fixes the "previous" operator versions for patch releases greater than .1  (e.g. for "1.35.3", the previous bundles would be "1.35.2", "1.35.1", "1.35.0", "1.34.0", "1.33.0", ... )

4. Fix kitchensink-upgrade test to actually use the test catalog

Probably for the above reasons of not actually properly generating previous versions, the upgrade tests used a hack to use redhat-operators if the actual bundle was not in the test catalog. This is now fixed.

This patch also fixes the expected upgrade patch. Given the skipRanges, the path from 1.35.0 goes directly to 1.35.2 , so this patch skips the >1.35.0 <1.35.x versions in the kitchensink-upgrade csvs.

5. There are some other minor hacks added for the kitchensink test, due to some resources not available in previous releases. (we could drop those if we scale back the upgrade path to just .e.g 1.35.0 -> 1.35.2 )


maschmid:
we need to push at least one 1.35.2 build to CI before we build the catalog, so dropping the catalog changes for now

PR #3778, opened 2025-08-25
Bump patch version for 1.36 to 1.36.1

Kaustubh-pande:
as per title
cc @maschmid @creydr


maschmid:
The linter should be fixed by updating the linter version in .github/workflows/validate.yaml  

(see this change in the commit from `main`)
https://github.com/openshift-knative/serverless-operator/commit/211f4da632cae2424b2d8850b420c0dd14905a0f#diff-9237aa56078dfbd8fcac3198563f306ada599aabaa43aca884905129779c5846L186-R186

maschmid:
I _think_ the `opm render` in the serverless-operator-index Dockerfile will fail unless we add the policy.json (not sure if with this version as well, but if it does fail, it needs

https://github.com/openshift-knative/serverless-operator/commit/ee2cb4a494b58cec9348b2bc17e3559ea446cd8a#diff-180daa8d7473215db71d63194c7ab9ccd92af88c4501b87c6cfd58037cf561d8R10-R12

```
COPY olm-catalog/serverless-operator-index/policy.json /etc/containers/policy.json
```

and the `olm-catalog/serverless-operator-index/policy.json` file https://github.com/openshift-knative/serverless-operator/blob/main/olm-catalog/serverless-operator-index/policy.json 

(but let's see if the serverless-operator-index build does actually fail in the 419-images jobs or the 419-operator-e2e )

PR #3781, opened 2025-08-27
[main] micro version bundle generation fixes

maschmid:
Changes originally made for 1.35 in https://github.com/openshift-knative/serverless-operator/pull/3757 , moving these to main.


maschmid:
kitchensink-upgrade tests are failing because we don't have KafkaSource v1 in 1.35.0 , so this is unrelated to these changes. We'll need to fix them later.

PR #3797, opened 2025-09-16
fix: make images install-all works on macos

Cali0707:
This fixes some build issues I ran into when trying to run `make images install-all` from a macbook on `release-1.35`.

## Proposed Changes

- Add `TARGET_ARCH` and `TARGET_OS` flags to the makefile and scripts


creydr (approved):
Thanks
/lgtm

PR #3798, opened 2025-09-19
Run on-cluster-builds in separate namespace to OLM

maschmid:
On OCP 4.20, the on-cluster builds of the Operator bundle and index images fail, 

see https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/69430/rehearse-69430-periodic-ci-openshift-knative-serverless-operator-main-ocp4.20-lp-interop-operator-e2e-interop-aws-ocp420/1968994633430077440 

The problem is that OCP 4.20 introduces NetworkPolicies in openshift namespaces, so the builds can't pull the UBI image. 

A clean solution is to move the builds from OLM_NAMESPACE somewhere else, hence introducing a `ON_CLUSTER_BUILDS_NAMESPACE` env, defaulting to `openshift-serverless-builds` for this purpose.

## Proposed Changes

- :bug: Run on-cluster-builds of bundle and index images in a separate namespace (openshift-serverless-builds)


maschmid:
420-operator-e2e will still fail because the Tempo operator is not yet available in OCP 4.20 catalog.

cardil (approved):
/lgtm

PR #3871, opened 2025-10-20
[release-1.37] use replaces 1.36.1

maschmid:
- :broom: Use replaces: 1.36.1, update generating scripts to use replaces and skipRange from metadata


maschmid:
> @maschmid shouldn't the change result in generated changes in actual OLM manifest?

It should with `make generate-catalog`  ("generated-files" does not generate catalog)

maschmid:
ok, let me try again :)

PR #3892, opened 2025-10-31
SRVCOM-3730: Adjust leader election parameters to match OpenShift recommendations

Fedosin:
Fixes JIRA #SRVCOM-3730

https://issues.redhat.com/browse/SRVCOM-3730


maschmid:
/override 421-kitchensink-upgrade

kitchensink-upgrade is known to be broken on 4.21

PR #4087, opened 2026-04-23
Use only currently available metrics in the scaling-debugging dash

maschmid:
- :bug: Fix broken scaling-debugging  dashboard

Fixes the missing "_seconds" name in the metrics.
Drops the revision name label from `kn_autoscaler_scrape_duration_seconds_bucket`, as it currently doesn't have one. (so, unfortunately, we will be only showing the histogram per-configuration, not per-revision)


creydr (approved):
/lgtm

PR #4083, opened 2026-04-21
[release-1.37] Run make generated-files

serverless-qe:
Generated by running `make generated-files`

PR #4094, opened 2026-04-28
[release-1.37] make generated-files

maschmid:


maschmid:
I thought I don't need this as https://github.com/openshift-knative/serverless-operator/pull/4083 exists, but I am not sure about that konflux fail building the ocp 4.22 index. 

This one doesn't generate catalogs, so perhaps it would be safer for now.

PR #4098, opened 2026-05-05
Added CRDs to must gather collection for easier use for analyze tools

davidtluong:
Adding CRDs to must gather so when support get a must gather archive we can use omc with it instead of manually looking at the files

## Proposed Changes
Add CRDs to must gather

- :gift: Add new feature


davidtluong:
@dsimansk Do I need to do anything to progress the status of this? I don't believe my change should've led to any of these tests failing considering I'm just adding one line to the must gather.

PR #4107, opened 2026-05-13
fix grafana-dash-knative-eventing-channel query for kafkachannels

maschmid:
kn_eventing_dispatch_latency_ms_count has `kn_kafkachannel_namespace` , 
kn_eventing_dispatch_duration_seconds_count has `kn_channel_namespace`

So when using just `kn_channel_namespace` labels, we don't see the kafkachannel namespaces.

Use label_replace to rename the `kn_kafkachannel_namespace` labels to `kn_channel_namespace` for the namespace query.

- :broom: fix grafana-dash-knative-eventing-channel query for kafkachannels


creydr (approved):
Good catch

/lgtm

PR #4167, opened 2026-09-15
[release-1.37] update go to 1.26 in build-image

maschmid:
- :broom: Update go to 1.26 in build-image


maschmid:
hm.. running `golangci-lint.1.64.8 run --path-prefix=. --timeout=10m0s --verbose` locally oomkills too.

